PR #41, opened 2019-06-01
deprecate block format v0

jkowalski:


jkowalski:
Note this is a breaking change for very old repository formats.

PR #55, opened 2019-06-28
Sftp storage

jbrodriguez:
Implements an sftp backed storage provider.

Uses go/ssh and pkg/sftp as main supporting modules and conceptually shares logic with filesystem and webdav providers.

A couple of notes:
- The provider assumes the user has a working public/private key connection to the ssh server. No other authentication method is supported
- The repository path must exist on the server
- (testing related) The pkg/sftp server doesn't offer a way to set a filesystem root, so it runs from the local directory which is basically repo/blob/sftp during testing. So the tests leave some debris behind. Additionally, that's the reason why id_rsa and known_hosts are there at all.
- Encrypted keyfiles are currently not supported (but it could be done)


repo/blob/sftp/sftp_storage_test.go
```diff
@@ -0,0 +1,214 @@
...
+	// the public key of a received connection
+	// with the entries in the authorized_keys file.
+	authorizedKeysBytes, err := ioutil.ReadFile(filepath.Join(cwd, "known_hosts"))
+	if err != nil {
+		t.Errorf("failed to load authorized_keys, err: %v", err)
+	}
+
+	authorizedKeysMap := map[string]bool{}
+	for len(authorizedKeysBytes) > 0 {
+		pubKey, _, _, rest, err := ssh.ParseAuthorizedKey(authorizedKeysBytes)
```
golangcibot: shadow: declaration of "err" shadows declaration at line 90 (from `govet`)

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -0,0 +1,214 @@
...
+	}); err != nil {
+		t.Fatalf("unable to clear GCS bucket: %v", err)
+	}
+
+	if err := st.Close(ctx); err != nil {
+		t.Fatalf("err: %v", err)
+	}
+}
+
+func server(t *testing.T) {
```
golangcibot: cyclomatic complexity 19 of func `server` is high (> 15) (from `gocyclo`)

repo/blob/sftp/sftp_storage.go
```diff
@@ -0,0 +1,338 @@
...
+func (s *sftpStorage) ListBlobs(ctx context.Context, prefix blob.ID, callback func(blob.Metadata) error) error {
+	var walkDir func(string, string) error
+
+	walkDir = func(directory string, currentPrefix string) error {
+		entries, err := s.cli.ReadDir(directory)
+		if err != nil {
+			return err
+		}
+
+		for _, e := range entries {
```
golangcibot: 99-126 lines are duplicate of `repo/blob/filesystem/filesystem_storage.go:89-116` (from `dupl`)

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -0,0 +1,214 @@
...
+	assertNoError(t, st.PutBlob(ctx, t3, []byte{1}))
+
+	if err := st.ListBlobs(ctx, "", func(bm blob.Metadata) error {
+		return st.DeleteBlob(ctx, bm.BlobID)
+	}); err != nil {
+		t.Fatalf("unable to clear sftp storage: %v", err)
+	}
+
+	blobtesting.VerifyStorage(ctx, t, st)
+	// blobtesting.AssertConnectionInfoRoundTrips(ctx, t, st)
```
golangcibot: commentedOutCode: may want to remove commented-out code (from `gocritic`)

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	private, err := ssh.ParsePrivateKey(privateBytes)
+	if err != nil {
+		log.Fatal("Failed to parse private key: ", err)
+	}
+
+	c.AddHostKey(private)
+
+	listener, _ := net.Listen("tcp", "0.0.0.0:2222")
```
golangcibot: G102: Binds to all network interfaces (from `gosec`)

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -0,0 +1,219 @@
...
+		t.Fatalf("err: %v", err)
+	}
+}
+
+func server(t *testing.T) {
+	debugStream := os.Stderr
+
+	c := createConfig(t)
+
+	listener, _ := net.Listen("tcp", ":2222")
```
golangcibot: G102: Binds to all network interfaces (from `gosec`)

repo/blob/sftp/sftp_storage.go
jbrodriguez: This one, I'm not sure how to fix.

I could surely create a common function that adds a `readDir` function argument that is provider specific.

But it would only serve filesystem and sftp, webdav's logic is slightly different.

jkowalski:
thanks for the PR, can you rebase and squish your changes into a single commit? this will be easier to review

jkowalski:
@jbrodriguez I've just sent out #56 that should help with the code duplication issue. Let me know what you think.

repo/blob/sftp/sftp_storage.go
jkowalski: See #56

repo/blob/sftp/sftp_storage.go
```diff
@@ -0,0 +1,338 @@
...
+	return walkDir(s.Path, "")
+}
+
+func (s *sftpStorage) PutBlob(ctx context.Context, blobID blob.ID, data []byte) error {
+	_, filePath := s.getDirPathAndFilePath(blobID)
+	tmpPath := fmt.Sprintf("%v-%v", filePath, rand.Int63())
+
+	f, err := s.createTempFileAndDir(tmpPath)
+	if err != nil {
+		fmt.Printf("putblob-createtempfileanddir(%s)\n", err)
```
jkowalski: please remove logging statements or use `kopialogging` package

repo/blob/sftp/sftp_storage.go
```diff
@@ -0,0 +1,338 @@
...
+		return nil, err
+	}
+	defer file.Close()
+
+	var hostKey ssh.PublicKey
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		fields := strings.Split(scanner.Text(), " ")
+
+		if len(fields) != 3 {
```
jkowalski: nit: a comment here explaining file structure would be nice

repo/blob/sftp/sftp_storage.go
```diff
@@ -0,0 +1,338 @@
...
+			continue
+		}
+
+		if strings.Contains(fields[0], host) {
```
jkowalski: is that substring match intentional? I think it will match all hosts if "" is provided.

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -0,0 +1,219 @@
...
+		Keyfile:    keyfile,
+		KnownHosts: filepath.Join(cwd, "known_hosts"),
+	})
+
+	if err != nil {
+		t.Fatalf("unable to connect to SSH: %v", err)
+	}
+
+	assertNoError(t, st.PutBlob(ctx, t1, []byte{1}))
+	time.Sleep(1 * time.Second) // sleep a bit to accommodate Apple filesystems with low timestamp resolution
```
jkowalski: is that needed?

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -0,0 +1,219 @@
...
+		t.Fatalf("unable to clear sftp storage: %v", err)
+	}
+
+	blobtesting.VerifyStorage(ctx, t, st)
+
+	// delete everything again
+	if err := st.ListBlobs(ctx, "", func(bm blob.Metadata) error {
+		return st.DeleteBlob(ctx, bm.BlobID)
+	}); err != nil {
+		t.Fatalf("unable to clear GCS bucket: %v", err)
```
jkowalski: perhaps extract a helper function, also remove reference to "GCS"

jkowalski:
Overall looks very good. Please let me know if #56 would make sense to use here.

Also would it make sense to also have a real SFTP host for testing (OpenSSH or something like that on some linux box?) We could store credentials in Travis config and only run this test when some environment variable is set (like for GCS)

jkowalski:
Just submitted #56. Can you see if it makes sense to rebase on top of that? The only non-trivial thing is a replacement for ReadDir() - I think it should be relatively straightforward to return some custom os.FileInfo implementation - does not need to be complete, just name, size and mod time.

jbrodriguez:
Hi, I've been busy, so couldn't take a look at your changes. I'll get to it tomorrow.

repo/blob/sftp/sftp_storage.go
```diff
@@ -0,0 +1,272 @@
...
+	if int(length) > len(data) {
+		return nil, errors.New("invalid length")
+	}
+
+	return data[0:length], nil
+}
+
+func (s *sftpImpl) PutBlobInPath(ctx context.Context, dirPath, path string, data []byte) error {
+	randSuffix := make([]byte, 8)
+	if _, err := rand.Read(randSuffix); err != nil {
```
golangcibot: G404: Use of weak random number generator (math/rand instead of crypto/rand) (from `gosec`)

repo/blob/sftp/sftp_storage.go
jbrodriguez: Added some reference info about the known_hosts file

repo/blob/sftp/sftp_storage.go
jbrodriguez: host is a required argument, can't be empty.
The check is there to extract the public key for the specific host, if it's there.
If it's not there, that means the user doesn't have a working public/private key connection to the ssh server.

repo/blob/sftp/sftp_storage_test.go
jbrodriguez: just copy/pasted filesystem tests 😁

jbrodriguez:
> Also would it make sense to also have a real SFTP host for testing (OpenSSH or something like that on some linux box?) We could store credentials in Travis config and only run this test when some environment variable is set (like for GCS)

That would be great. I don't have access to such a linux box, but I'd add to the tests if you can set it up.

jbrodriguez:
Also, I think I covered all the comments you had.

I don't quite get the weak random generator issue, since it's the same implementation as filesystem.

Let me know if you prefer it to be the same as webdav's.

repo/blob/sftp/sftp_storage.go
jkowalski: wouldn't this have side effects if hou had hosts "foo", "foo-a" and "foo-b". Now strings.Contains("foo-a", "foo") is true, but that's not indended, I think.

repo/blob/sftp/sftp_storage.go
```diff
@@ -0,0 +1,274 @@
+package sftp
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
```
jkowalski: just replace this with `crypto/rand` to resolve the issue?

jkowalski:
> Also, I think I covered all the comments you had.
> 
> I don't quite get the weak random generator issue, since it's the same implementation as filesystem.
> 
> Let me know if you prefer it to be the same as webdav's.

I resolved all but 2 issues. The rand fix is simple. I have one minor concern about the host matching, otherwise looks good.

jkowalski:
BTW I tried your changes against some test sftp account and worked on first try. Great stuff!
I'll think about a way to test this e2e.

repo/blob/sftp/sftp_storage.go
jbrodriguez: That's right. vscode autocomplete found math/rand instead of crypto/rand.

repo/blob/sftp/sftp_storage.go
jbrodriguez: I implement more robust logic, leveraging ssh.ParseKnownHosts go function.

This requires the hostname to be exactly equal to the host entry found in known_hosts.

There maybe some edge cases that this code isn't taking into account (hashed entries, wildcards, etc), but that should be on the users to solve, since they have full control over the --host, --keyfile params, as well as the known_hosts file.

jkowalski:
> BTW I tried your changes against some test sftp account and worked on first try. Great stuff!
> I'll think about a way to test this e2e.

Can we do something similar to how GCS is tested (separate PR is fine), basically, assume that the caller will pass the configuration through environment variables, something like:

`KOPIA_SFTP_TEST_SERVER`
`KOPIA_SFTP_TEST_PATH`
`KOPIA_SFTP_IDENTITY_FILE`

When those are not provided, the test will be skipped. I have a SFTP host I could use for this testing on travis, where all the credentials will be stored encrypted.

jkowalski:
Looks great, thanks for your contribution!

jbrodriguez:
> Can we do something similar to how GCS is tested (separate PR is fine), basically, assume that the caller will pass the configuration through environment variables, something like:
> 
> `KOPIA_SFTP_TEST_SERVER`
> `KOPIA_SFTP_TEST_PATH`
> `KOPIA_SFTP_IDENTITY_FILE`
> 
> When those are not provided, the test will be skipped. I have a SFTP host I could use for this testing on travis, where all the credentials will be stored encrypted.

Sure, I'll work on it (most likely next week)

PR #56, opened 2019-06-30
refactored common bits of blob providers that use sharding

jkowalski:
(filesystem, webdav and upcoming sftp) into shared package


jkowalski:
@jbrodriguez This should help with your PR, curious what you think.

PR #62, opened 2019-07-11
Add e2e tests for sftp storage provider

jbrodriguez:
Uses environment variables from the CI server, to connect to a test
sftp server


jbrodriguez:
Ok, I see it failed because of the original "localhost" test.

Not sure why it fails now, while it worked before.

Should  I remove that test ?

jkowalski:
Yes. Can you disable the localhost test to get it to pass in the meantime ?

jkowalski:
LGTM, thanks! I'll work on pointing this at some linux box in travis.

PR #68, opened 2019-07-16
Improve sftp provider logic

jbrodriguez:
Tests are failing because pkg/sftp won't overwrite an existing file
(Rename function) and the test is actually doing that with
blobtesting.VerifyStorage.

The solution is to use pkg/sftp's PosixRename function:
"PosixRename renames a file using the posix-rename@openssh.com
extension which will replace newname if it already exists."

Additionally, the provider now creates the path on the server, if it
doesn't exist.


jkowalski:
Nice, thanks for fixing the issue!

Issue #93, opened 2019-08-24
High memory usage seen with a large number of small files


ntolia:
I created a 5 test directories (on an NVME device) with 2M files each. 100 bytes of random data in each one of those files.

```
$ kopia repository create s3 --bucket=<test-bucket> --prefix=kopia/
$ time kopia snapshot create /mnt/ssd/smallfiles
...

real	13m36.700s
user	72m46.181s
sys	5m23.254s
```

I found the memory usage to be quite high. See `pidstat` below but I saw it grow to ~21GB on one of these runs.
```
$ pidstat -r -p 5049 1
Linux 4.15.0-1044-aws (ip-172-31-10-90) 	08/24/19 	_x86_64_	(16 CPU)

04:13:01      UID       PID  minflt/s  majflt/s     VSZ     RSS   %MEM  Command
04:13:02     1000      5049     36.00      0.00 16696284 15266768  47.87  kopia
...
```

Dataset Generation Script (not my own but found elsewhere):
```python
!/usr/bin/env python3
import os
import sys
import argparse
from multiprocessing import Pool
from functools import partial


def create_files(args, i):
    path = os.path.join(args.path, 'my-dir-name-%d' % i)
    os.mkdir(path)
    for j in range(args.filenum):
        with open(os.path.join(path, 'my-random-file-%d.data' % j), 'wb') as f:
            f.write(os.urandom(args.filesize))


def main(argv):
    parser = argparse.ArgumentParser(description='Create directories and files for backup test')
    parser.add_argument('-p', '--path', type=str, required=True, help='Path to the directory')
    parser.add_argument('-d', '--dirnum', type=int, required=True, help='Number of subdirectories')
    parser.add_argument('-f', '--filenum', type=int, required=True, help='Number of files per directory')
    parser.add_argument('-s', '--filesize', type=int, required=True, help='File size')
    args = parser.parse_args()

    if os.path.exists(args.path):
        print('ERROR: destination directory %s already exists' % args.path, file=sys.stderr)
        return 1
    elif not os.path.exists(os.path.dirname(args.path)):
        print('ERROR: parent directory for destination %s does not exist' % args.path, file=sys.stderr)
        return 1

    os.mkdir(args.path)
    with Pool(min(args.dirnum, 8)) as p:
        p.map(partial(create_files, args), range(args.dirnum))


if __name__ == '__main__':
    sys.exit(main(sys.argv))
```

and I then ran
```
[...]

PR #95, opened 2019-08-24
localfs: improved performance and memory usage when scanning large directories

jkowalski:
- parallelized os.Lstat() x 16 (dramatically improves speed)
- discarded unused portions of os.FileInfo (uses 60% less RAM on macOS)

BEFORE:
10:47:03.670 [kopia/localfs] listed 200000 entries in 43.871211686s using 79126528 bytes of heap

After:
10:49:12.439 [kopia/localfs] listed 200000 entries in 1.953018184s using 30515200 bytes of heap

### DIFF
--- a/fs/localfs/local_fs.go
+++ b/fs/localfs/local_fs.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"sync"
+	"time"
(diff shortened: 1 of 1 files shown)


jkowalski:
This should improve #93 a bit, but is not the complete answer.

PR #101, opened 2019-08-25
Switch file log level to info

ntolia:
Reduces multi-GB log files. Main culprits were:

- [upload.go:385] could not find cache entry for...
- [content_manager.go:790] WriteContent("...") - new
- [content_manager.go:396] adding ... length=... deleted=false


jkowalski:
LGTM, thanks!

PR #105, opened 2019-08-28
Reduced locking in content manager to improve throughput with parallel uploads

jkowalski:


ntolia:
Definitely seems to have help with a dramatic increase in bandwidth usage and reduction in completion time. For 10 1 GB files, the old code had upload times of:

```
real	5m53.927s
user	4m54.245s
sys	0m13.971s
```

The new code has upload times of:
```
real    1m28.624s
user    4m59.487s
sys     0m28.693s
```

If I reduced parallelism to 1 on my test VM (16 cores) but with the new code, I got:
```
real    8m25.286s
user    4m33.292s
sys     0m13.549s
```

ntolia:
With this code and some reasonable data (80 1 GB files, `--parallel` not modified/default), I can get sustained bandwidth of 1.2+ Gbps (as reported by `iftop`) to an AWS S3 bucket in the same region as my VM. Memory usage is high (10+ GB) but mostly steady.

ntolia:
Also, final E2E numbers for the 80 GB upload is shown below. Works out to ~0.94 Gbps.

```
real    11m18.138s
user    48m59.034s
sys     3m53.217s
```

jkowalski:
those numbers look great, I figured out why I was getting random failures in stress test - due to reduced locking the test was getting hit by two forgotten mutexes in `Refresh()` and `CompactIndexes()` - this is now fixed.

PR #113, opened 2019-09-11
linter: upgraded to golangci-lint 1.18.0, fixed or suppressed warnings

jkowalski:


jkowalski:
looks like golangci.com is having a bad-hair day. submitting anyway since the lint also runs on travis

PR #120, opened 2019-10-31
Documentation about the encryption of the format blob

julio-lopez:
Notes about the encryption of the format blob.

### DIFF
--- a/site/content/docs/Encryption/_index.md
+++ b/site/content/docs/Encryption/_index.md
@@ -4,11 +4,11 @@ linkTitle: "Encryption"
 weight: 4
 ---
 
+### Format Blob Encryption
+
 Kopia uses a standard envelope encryption technique to de-couple the repository passphrase from the keys used for encrypting and authenticating the contents of the repository.
 
 Each repository has a format blob containing configuration parameters for the repository. The format blob itself is serialized and persistent as JSON, such as the one below.
-
-not a good random initial value
 
 ```json
 {
--- a/site/content/docs/Encryption/_index.md
+++ b/site/content/docs/Encryption/_index.md
@@ -1,0 +1,104 @@
+---
+title: "Encryption"
+linkTitle: "Encryption"
+weight: 4
+---
+
+Kopia uses a standard envelope encryption technique to de-couple the repository passphrase from the keys used for encrypting and authenticating the contents of the repository.
+
+Each repository has a format blob containing configuration parameters for the repository. The format blob itself is serialized and persistent as JSON, such as the one below.
+
+not a good random initial value
+
+```json
+{
+  "tool": "https://github.com/kopia/kopia",
+  "buildVersion": "v0.3.0",
+  "buildInfo": "unknown",
+  "uniqueID": "bm90IGEgZ29vZCByYW5kb20gaW5pdGlhbCB2YWx1ZQ==",
+  "keyAlgo": "scrypt-65536-8-1",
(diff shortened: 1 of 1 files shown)


julio-lopez:
Only pushed initial content for the format blob, probably the wording needs to be clarified, the formatting improved. Not sure whether or not the new page needs to be linked from anywhere.

site/content/docs/Encryption/_index.md
```diff
@@ -0,0 +1,104 @@
...
+title: "Encryption"
+linkTitle: "Encryption"
+weight: 4
+---
+
+Kopia uses a standard envelope encryption technique to de-couple the repository passphrase from the keys used for encrypting and authenticating the contents of the repository.
+
+Each repository has a format blob containing configuration parameters for the repository. The format blob itself is serialized and persistent as JSON, such as the one below.
+
+not a good random initial value
```
jkowalski: not sure what `not a good random inital value` is?

site/content/docs/Encryption/_index.md
```diff
@@ -0,0 +1,104 @@
+---
+title: "Encryption"
+linkTitle: "Encryption"
+weight: 4
+---
+
```
jkowalski: nit: perhaps add a section here for format block encryption.

site/content/docs/Encryption/_index.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+* `Version` identifies the version of the format blob.
+* The `tool` and `buildInfo` fields are informational
+* `buildVersion` is the version of Kopia, which is also indirectly used as the repository format version.
+* `UniqueID` is a randomly generated identifier for the repository. This is also used as the input for various encryption operations.
+* `keyAlgo` identifies the password-based key derivation function (PBKDF). Only _scrypt_ with the currently recommended cost parameters (N=65536, r=8, p=1) is supported at the moment. The main purpose of this field is to be able to change and extend the key derivation algorithm in the future. For example, by increasing the cost parameters for _scrypt_ or using a different algorithm altogether.
+* `encryption` identifies the encryption algorithm that was used to encrypt the encryptedBlockFormat field.
+* `encryptedBlockFormat` is a ciphertext containing among others, the encryption secrets and parameters used for encrypting the repository content. Below is additional information about its plaintext content and how it is encrypted.
```
jkowalski: might be worth mentioning that `blockFormat` can alternatively be specified

site/content/docs/Encryption/_index.md
julio-lopez: uh? me neither.
```suggestion
```

site/content/docs/Encryption/_index.md
julio-lopez: Makes sense. Any suggestions for linking, maybe to command line options?

site/content/docs/Encryption/_index.md
julio-lopez: Oh, now I remember => `bm90IGEgZ29vZCByYW5kb20gaW5pdGlhbCB2YWx1ZQ==`
And it was only an hour ago 😞

site/content/docs/Encryption/_index.md
julio-lopez: OK, I guess I understand your comment better now, the unencrypted contents can be specified / inlined in the format blob. Got it.

Issue #114, opened 2019-09-15
implement extract command


purplesrl:
I have tested this and its a great tool, especially because of the sftp and webdav, very fast and modern, but unfortunately there is one major flaw: there is no extract command, which is present in all other backup programs.

I was thinking:

kopia extract [snapshot_id] [PATH] [EXTRACT_PATH] <--test> <--verify> <-a>

Problem with FUSE: I am backing up on a VPS, as many people I am sure do, in the case of OpenVZ, there is no fuse by default and many providers are unwilling to enable it because it uses too much resources, I tried also to mount remotely but it will not work, so I am unable to use your program because I cannot restore the files.

jkowalski:
basic support for restore is now submitted, we will file additional issues for performance and feature improvements.

PR #126, opened 2019-11-06
Allow passing an fs.Entry to TreeWalker.Callback

julio-lopez:
This allows leveraging `TreeWalker` for other uses. Initially, it will be used to address #114.

This is only re-factoring, no functional changes.
The PR is split into self-contained commits to facilitate reviewing.
* 8c4754d is code movement only (to a new package)
* b1b29da allows specifying a function to a `TreeWalker` to extract an identifier from an `fs.Entry`
* Lastly, edf2cce changes the signature of `TreeWalker.Callback` to pass an `fs.Entry` instead of an  `object.ID`.

`TreeWalker` could be moved to the `kopia/fs` package, but it's not done in this PR.

### DIFF
--- a/snapshot/gc/gc.go
+++ b/snapshot/gc/gc.go
@@ -1,0 +1,124 @@
+package gc
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/internal/kopialogging"
+	"github.com/kopia/kopia/internal/units"
+	"github.com/kopia/kopia/repo"
+	"github.com/kopia/kopia/repo/content"
+	"github.com/kopia/kopia/repo/manifest"
+	"github.com/kopia/kopia/repo/object"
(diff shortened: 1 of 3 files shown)


snapshot/gc/gc.go
```diff
@@ -0,0 +1,130 @@
+package gc
```
golangcibot: /usr/local/go/src/strings/strings.go:980:20: DecodeRuneInString not declared by package utf8 (from `typecheck`)

julio-lopez: ?

julio-lopez: `make lint` works locally

@jkowalski Any ideas?

jkowalski: yeah, this bot is weird, just ignore it for now

jkowalski:
LGTM, thanks for doing this!

snapshot/gc/gc.go
julio-lopez: Thanks!

PR #130, opened 2019-11-07
Ensure Mutex.Unlock() in cachefs.Cache.getEntries()

julio-lopez:
Motivation: An unlock is missing in the error path

### DIFF
--- a/fs/cachefs/cache.go
+++ b/fs/cachefs/cache.go
@@ -118,6 +118,7 @@ func (c *Cache) getEntries(ctx context.Context, id string, expirationTime time.D
 	}
 
 	c.mu.Lock()
+	defer c.mu.Unlock()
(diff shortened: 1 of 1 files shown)


fs/cachefs/cache.go
```diff
@@ -118,6 +118,7 @@ func (c *Cache) getEntries(ctx context.Context, id string, expirationTime time.D
 	}
 
 	c.mu.Lock()
+	defer c.mu.Unlock()
```
julio-lopez: `c.mu.Unlock()` was missing at line 131 below.

jkowalski: good catch, wondering why I did it like that.

jkowalski:
LGTM, since this was clearly untested, perhaps we can add unit tests here...

Issue #124, opened 2019-11-01
Allow specifying the hostname during snapshot creation


julio-lopez:
It should be possible to specify/override the OS hostname during snapshot creation.

Scenarios where this is useful or even required include:
* Cloud-native (containerized) platforms, such as Kubernetes, where the hostname is not stable, and may change across different executions of a Pod / container associated with the same "logical volume" (PVC/PV in k8s terms)
* Mobile clients, such as laptops, that may get their hostname assigned via DHCP and thus the hostname may change depending on configuration, network connection and so on.

This could be implemented via an optional flag to the `snapshot create` command.

jkowalski:
Take a look at `getHostName()` in `cli/command_snapshot_create.go` - should be pretty trivial to plumb through a flag.

jkowalski:
While at it we may want to support overriding username.

PR #139, opened 2019-11-22
[#124] Add optional flags to override hostname and username

pavannd1:
## Change Overview

- Add optional flags to override hostname and username
- Add the flags to `snapshot`, `policy` and `server start` commands
- Add a test case in the E2E tests to create and list snapshots with overridden hostname and username

## Issues

- #124 

## Test Plan

Added a case to the existing E2E test
```
=== RUN   TestEndToEnd
PASS
ok      github.com/kopia/kopia/tests/end_to_end_test    21.270s
```


jkowalski:
Thank you!

PR #140, opened 2019-11-22
Minor cleanups

julio-lopez:
Flagged by `golangci` while testing version 1.21.0

There were other (noisy?) failures not addressed in this PR. Punting for now.


jkowalski:
Thank you!

PR #142, opened 2019-11-26
lint: switched to 1.21 and fixed a ton of whitespace issues discovered

jkowalski:


internal/ospath/ospath_xdg.go
```diff
@@ -13,6 +13,7 @@ func init() {
 	} else {
 		userSettingsDir = filepath.Join(os.Getenv("HOME"), ".config")
 	}
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

jkowalski: done

PR #144, opened 2019-12-06
[114] Restore command

julio-lopez:
Behavior:

- Creates top directory on restore
- Fails when target directories or files already exist
- Allows restoring to local root directory
- Restores of file attributes => mod time, mode, owner info
- Only a directory can be specified as a parameter

Not implemented:

- Restoring attributes of the top folder
- Restoring symlinks
- Restoring a single file

End-to-end test included

#114

### DIFF
--- a/cli/command_restore.go
+++ b/cli/command_restore.go
@@ -1,0 +1,43 @@
+package cli
+
+import (
+	"context"
+
+	"github.com/kopia/kopia/repo"
+	"github.com/kopia/kopia/snapshot/snapshotfs"
+)
+
+const restoreCommandHelp = `Restore a directory from a snapshot into the specified target path.
+
+The target path must not exist and it is created by the restore command. The case when the target is the root directory in the local machine is an exception, in this case no attempt is made to create the root directory and the contents of the source directory are copied to the local root directory.
+The restore command conservatively refuses to overwrite previously existing files or directories.
+
+The source to be restored is specified in the form of a directory ID and optionally a sub-directory path.
+
+For example, the following source and target arguments will restore the contents of the 'kffbb7c28ea6c34d6cbe555d1cf80faa9' directory into a new, local directory named 'd1'
+
+'restore kffbb7c28ea6c34d6cbe555d1cf80faa9 d1'
+
+Similarly, the following command will restore the contents of a subdirectory 'subdir/subdir2' under 'kffbb7c28ea6c34d6cbe555d1cf80faa9'  into a new, local directory named 'sd2'
+
+'restore kffbb7c28ea6c34d6cbe555d1cf80faa9/subdir1/subdir2 sd2'
+`
+
+var (
+	restoreCommand           = app.Command("restore", restoreCommandHelp)
+	restoreCommandSourcePath = restoreCommand.Arg("source-path", "Source directory ID/path in the form of a directory ID and optionally a sub-directory path. For example, ' kffbb7c28ea6c34d6cbe555d1cf80faa9' or 'kffbb7c28ea6c34d6cbe555d1cf80faa9/subdir1/subdir2'").Required().String()
+	restoreCommandTargetPath = restoreCommand.Arg("target-path", "Path of the directory for the contents to be restored").Required().String()
+)
+
+func runRestoreCommand(ctx context.Context, rep *repo.Repository) error {
+	oid, err := parseObjectID(ctx, rep, *restoreCommandSourcePath)
+	if err != nil {
+		return err
+	}
+
+	return snapshotfs.Restore(ctx, rep, *restoreCommandTargetPath, oid)
+}
+
+func init() {
+	restoreCommand.Action(repositoryAction(runRestoreCommand))
+}
--- a/cli/command_restore.go
+++ b/cli/command_restore.go
@@ -7,25 +7,41 @@ import (
 	"github.com/kopia/kopia/snapshot/snapshotfs"
 )
 
-const restoreCommandHelp = `Restore a directory from a snapshot into the specified target path.
+const (
+	restoreCommandHelp = `Restore a directory from a snapshot into the specified target path.
 
-The target path must not exist and it is created by the restore command. The case when the target is the root directory in the local machine is an exception, in this case no attempt is made to create the root directory and the contents of the source directory are copied to the local root directory.
-The restore command conservatively refuses to overwrite previously existing files or directories.
+The target path must not exist and it is created by the restore command. The
+case when the target is the root directory in the local machine is an exception,
+in this case no attempt is made to create the root directory and the contents of
+the source directory are copied to the local root directory.
+The restore command conservatively refuses to overwrite previously existing
+files or directories.
 
-The source to be restored is specified in the form of a directory ID and optionally a sub-directory path.
+The source to be restored is specified in the form of a directory ID and
+optionally a sub-directory path.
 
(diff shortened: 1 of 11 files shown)


julio-lopez:
Opening this PR on behalf of @pavannd1 who has been doing the heavy lifting here and currently on a break.

cli/command_restore.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+
+	"github.com/kopia/kopia/repo"
+	"github.com/kopia/kopia/snapshot/snapshotfs"
+)
+
+const restoreCommandHelp = `Restore a directory from a snapshot into the specified target path.
+
+The target path must not exist and it is created by the restore command. The case when the target is the root directory in the local machine is an exception, in this case no attempt is made to create the root directory and the contents of the source directory are copied to the local root directory.
```
golangcibot: line is 299 characters (from `lll`)

cli/command_restore.go
```diff
@@ -0,0 +1,43 @@
...
+'restore kffbb7c28ea6c34d6cbe555d1cf80faa9 d1'
+
+Similarly, the following command will restore the contents of a subdirectory 'subdir/subdir2' under 'kffbb7c28ea6c34d6cbe555d1cf80faa9'  into a new, local directory named 'sd2'
+
+'restore kffbb7c28ea6c34d6cbe555d1cf80faa9/subdir1/subdir2 sd2'
+`
+
+var (
+	restoreCommand           = app.Command("restore", restoreCommandHelp)
+	restoreCommandSourcePath = restoreCommand.Arg("source-path", "Source directory ID/path in the form of a directory ID and optionally a sub-directory path. For example, ' kffbb7c28ea6c34d6cbe555d1cf80faa9' or 'kffbb7c28ea6c34d6cbe555d1cf80faa9/subdir1/subdir2'").Required().String()
```
golangcibot: line is 281 characters (from `lll`)

fs/localfs/copy.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, e := range entries {
+		if err := copyEntry(ctx, e, filepath.Join(targetPath, e.Name())); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func createDirectory(path string) error {
```
jkowalski: why do we want to fail if directory exists? 

I think `os.MkdirAll()` should be sufficient - this will make the restore retriable/idempotent, otherwise recovering from partial restore is not ideal.

fs/localfs/copy.go
```diff
@@ -0,0 +1,164 @@
...
+		return errors.Wrap(err, "failed to stat "+targetPath)
+	}
+
+	r, err := f.Open(ctx)
+	if err != nil {
+		return errors.Wrap(err, "unable to open snapshot file for "+targetPath)
+	}
+	defer r.Close() //nolint:errcheck
+
+	if _, err = io.Copy(out, r); err != nil {
```
jkowalski: it would be good to make this atomic - can we use https://github.com/natefinch/atomic or something ?

fs/localfs/copy.go
```diff
@@ -0,0 +1,164 @@
...
+func copyFileContent(ctx context.Context, targetPath string, f fs.File) error {
+	var out *os.File
+
+	switch _, err := os.Stat(targetPath); {
+	case os.IsNotExist(err):
+		if out, err = os.Create(targetPath); err != nil {
+			return errors.Wrap(err, "failed to create new file "+targetPath)
+		}
+		defer out.Close() //nolint:errcheck
+	case err == nil:
```
jkowalski: this should be success, I think, again to be idempotent

fs/localfs/copy.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return errors.Wrap(err, "unable to open snapshot file for "+targetPath)
+	}
+	defer r.Close() //nolint:errcheck
+
+	if _, err = io.Copy(out, r); err != nil {
+		return errors.Wrapf(err, "unable to restore %q content", targetPath)
+	}
+
+	return out.Sync()
```
jkowalski: do we need to sync here? That might kill performance of restores.

fs/localfs/copy.go
```diff
@@ -0,0 +1,164 @@
...
+	return copyDirectoryContent(ctx, d, targetPath)
+}
+
+func copyDirectoryContent(ctx context.Context, d fs.Directory, targetPath string) error {
+	entries, err := d.Readdir(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, e := range entries {
```
jkowalski: for high-latency repositories this will be pretty slow unless we parallelize. Fortunately there's `parallelwork.Queue` which can be used to pretty efficiently schedule work whose size is unknown when you begin. Basically - instead of just doing work in the for loop just schedule each entry in the queue and let the queue do the magic.

fs/localfs/copy.go
```diff
@@ -0,0 +1,164 @@
...
+
+// Copy copies e into targetPath in the local file system. If e is an
+// fs.Directory, then the contents are recursively copied.
+// The targetPath must not exist, except when the target path is the root
+// directory. In that case, e must be a fs.Directory and its contents are copied
+// to the root directory.
+// Copy does not overwrite files or directories and returns an error in that
+// case. It also returns an error when the the contents cannot be restored,
+// for example due to an I/O error.
+func Copy(ctx context.Context, targetPath string, e fs.Entry) error {
```
jkowalski: possible future improvement: introduce abstraction for writing to an abstract filesystem, so we can restore to a zip/tar file too in addition to a local directory.

jkowalski:
This looks great, few usability suggestions:

- there should be some log/progress output - would be easy if you use `parallelwork.Queue` as suggested in one of the comments.
- restore should be easy to restart after a crash/connection failure, so we should accept pre-existing directories and just copy there

For the future, restoring to a ZIP/TAR file would be great.

fs/localfs/copy.go
julio-lopez: Will do in a sec.

fs/localfs/copy.go
julio-lopez: Mmmm, we may be able to remove it. 

* Yes, it may kill performance. OTOH, `r.Close()` may (or may not) trigger a sync, and whether or not that ends up being synchronous is OS/FS dependent. This is not explicitly specified in `close()` for POSIX, and most implementations end up being asynchronous, and thus the application has no way of knowing whether or not the data was persisted.
* The intent here was to capture write errors and surface those as soon as possible. So, this would be the _safest_ option. At the same time, it seems reasonable that the default behavior could be to *NOT sync*, and have a flag to overwrite that.

I'll go ahead and remove the `Sync()` and at least capture the error from `out.Close()` when there have been no other errors.

Thoughts?

julio-lopez: Removed the `Sync()` when leveraging `atomic.WriteFile(...)`

fs/localfs/copy.go
julio-lopez: I'll submit another patch to avoid failing in this case and the one below for files.

jkowalski:
thanks, LGTM, we can work on future improvements as separate PRs.

julio-lopez:
Jarek, thanks for the review.
I'll submit follow up PRs to implement your suggestions. I'll add flags, and we'll figure out what makes sense as default values depending on feedback / usage.

PR #148, opened 2019-12-11
Minor cleanups

julio-lopez:


jkowalski:
is there a linter that flagged these?

julio-lopez:
IIRC, there was a linter included in `gometalinter` that would flag the undocumented or not consistently documented export symbols. These were flagged by whatever flavor-of-the-week linter used by VS code, they popped up as I was going through the code, but it is not comprehensive. Let me see if I can dig something.

PR #151, opened 2019-12-13
Snapshot restore command

redgoat650:
Snapshot restore will take a snapshot ID and restore the
associated snapshot to the target path.
- Looks up the manifest with the snapshot ID
- Gets the snapshot root entry
- Copies the snapshot from the root entry to the target path

Because it uses the parent manifest with the copied permissions,
the restored directory will have the permissions of the original
source directory.


jkowalski:
@redgoat650 can you rebase and re-push?

PR #152, opened 2019-12-13
Added UI to browse directories and download files in `kopia server --ui`

jkowalski:
To try:

```
make install && kopia server --ui
```

Navigate to http://localhost:51515/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,11 @@ include tools/tools.mk
 
 -include ./Makefile.local.mk
 
-build:
-	go build github.com/kopia/kopia/...
+install: html-ui-bindata
+	go install -tags embedhtml
+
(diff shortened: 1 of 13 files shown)


julio-lopez:
Hi @jkowalski 

I tried the UI. Overall, looks good. There are a couple of things I noticed. 

* Would it make sense to group / filter by host alone, besides `user@host`
* When there are many pages, the page list overflows outside the view to the right. See screenshot below.
* Changing the sort order affects the table layout in a "surprising" way, it's somehow not smooth. See before and after screenshots
* In the snapshots view, Sorting by path does not seem to sort correctly.
 
Long list of pages

Default before sorting

Sorting by path ascending

Sorting by path descending

PR #156, opened 2019-12-17
htmlui: addressed feedback from PR #152

jkowalski:
See #152


julio-lopez:
Tried it. LGTM.

Nit: There is a slightly jerky transition for table in the top snapshots page  when the width of the columns changes due to the with of the elements in them changing. For example, if you have paths that are short and show in the first page, and then the second page has paths that are long. Anyway, it's not a big deal.

PR #163, opened 2020-01-02
Change default behavior for localfs.Copy

julio-lopez:
Addresses some of the feedback from #144

Allow copying to an existing directory
Remove special handling for the root directory, no longer needed.
Add copy options to control behavior
Allow overwriting an existing file (via options)
Expose copy options as flags


jkowalski:
LGTM

PR #164, opened 2020-01-03
lint: upgraded to 1.22.2 and make lint issues a build failure

jkowalski:
fixed or silenced linter warnings, mostly due to magic numeric constants

### DIFF
--- a/internal/fshasher/fshasher.go
+++ b/internal/fshasher/fshasher.go
@@ -25,7 +25,7 @@ func Hash(ctx context.Context, e fs.Entry) ([]byte, error) {
 	}
 
 	tw := tar.NewWriter(h)
-	defer tw.Close()
+	defer tw.Close() //nolint:errcheck
 
 	if err := write(ctx, tw, "", e); err != nil {
 		return nil, err
@@ -114,9 +114,10 @@ func writeFile(ctx context.Context, w io.Writer, f fs.File) error {
 	if err != nil {
 		return err
 	}
-	defer r.Close()
 
-	_, err = io.Copy(w, r)
+	if _, err = io.Copy(w, r); err != nil {
+		return err
+	}
 
-	return err
+	return r.Close()
 }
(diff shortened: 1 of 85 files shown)


tools/tools.mk
```diff
@@ -35,11 +35,12 @@ $(BINDATA_TOOL):
 	go build -o $(BINDATA_TOOL) github.com/go-bindata/go-bindata/go-bindata
 
 # linter
-LINTER_TOOL=$(TOOLS_DIR)/bin/golangci-lint
+LINTER_TOOL=$(TOOLS_DIR)/bin/golangci-lint-$(GOLANGCI_LINT_VERSION)
```
julio-lopez: nice!

repo/content/index.go
```diff
@@ -164,10 +164,6 @@ func (b *index) GetInfo(contentID ID) (*Info, error) {
 }
 
 func (b *index) entryToInfo(contentID ID, entryData []byte) (Info, error) {
-	if len(entryData) < 20 {
```
julio-lopez: Is this check performed inside `e.parse` ?

internal/fshasher/fshasher.go
```diff
@@ -114,9 +114,10 @@ func writeFile(ctx context.Context, w io.Writer, f fs.File) error {
...
 	}
-	defer r.Close()
 
-	_, err = io.Copy(w, r)
+	if _, err = io.Copy(w, r); err != nil {
+		return err
+	}
 
-	return err
+	return r.Close()
```
julio-lopez: This actually changes the behavior and would leave `r` open when `io.Copy` fails.

repo/content/index.go
jkowalski: Yes

internal/fshasher/fshasher.go
jkowalski: fixed

jkowalski: one sec, let me fix it again :)

jkowalski: done, I initially misread the code and thought that it's the `w` that needs to be closed, simply adding `//nolint:errcheck` is the right thing to do for read-only streams

PR #171, opened 2020-01-16
testing: split end_to_end_test into separate files.

jkowalski:
refactored test helpers to separate package
made all tests parallel and improved the code structure


tests/end_to_end_test/main_test.go
```diff
@@ -0,0 +1,81 @@
...
+	sharedTestDataDirBase string
+	sharedTestDataDir1    string
+	sharedTestDataDir2    string
+	sharedTestDataDir3    string
+)
+
+func oneTimeSetup() error {
+	var err error
+
+	sharedTestDataDirBase, err = ioutil.TempDir("", "kopia-test")
```
julio-lopez: Just skimmed through the tests, and AFAICT, these directories are only read from the individual tests and not modified otherwise. And for tests that need specific input setups or so, then `makeScratchDir` below is used.

Is that correct?

Also, `dir1`, `dir2`, `dir3` have approximately similar structures. What's the rationale for generating 3 separate ones? is it an increase in coverage?

Makefile
```diff
@@ -117,7 +118,7 @@ dist-binary:
 	go build -o dist/integration/kopia github.com/kopia/kopia
 
 integration-tests: dist-binary
-	KOPIA_EXE=$(CURDIR)/dist/integration/kopia $(GO_TEST) -v -count=1 -timeout 90s github.com/kopia/kopia/tests/end_to_end_test
+	KOPIA_EXE=$(CURDIR)/dist/integration/kopia $(GO_TEST) -count=1 -parallel $(PARALLEL) -timeout 300s github.com/kopia/kopia/tests/end_to_end_test
```
julio-lopez: Any time reductions or increases due to running the tests in parallel?

jkowalski: It's about 50% reduction in wall clock time:

Before:

```
$ time make integration-tests
ok  	github.com/kopia/kopia/tests/end_to_end_test	67.782s
make integration-tests  55.67s user 15.15s system 102% cpu 1:09.38 total
```

After:

```
$ time make integration-tests
ok  	github.com/kopia/kopia/tests/end_to_end_test	35.230s
make integration-tests  93.23s user 19.32s system 305% cpu 36.885 total
```

tests/end_to_end_test/main_test.go
jkowalski: added more interesting shapes and sizes of the tree along with test API to create more if needed.

PR #175, opened 2020-01-24
cli: improvements for 'kopia server' and client

jkowalski:
Those will make it possible to securely host 'kopia server' embedded
in a desktop app that runs in the background and can access UI.

- added support for using and generating TLS certificates
- added /api/v1/shutdown API to remotely trigger server shutdown
- added support for automatically shutting down server if no requests
  arrive in certain amount of time
- added support for generating and printing random password to STDERR

TLS supports 3 modes:

1. serve TLS using externally-provided cert/key PEM files
2. generate & write PEM files, then serve TLS using them
3. generate and use emphemeral cert/key (prints SHA256 fingerprint)

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -3,7 +3,6 @@ package cli
 
 import (
 	"context"
-	"net/http"
 	"os"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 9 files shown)


julio-lopez:
Skimmed through it. LG overall.

Issue #179, opened 2020-01-29
Robustness testing toolset implementation


redgoat650:
Umbrella issue to track progress on implementation of robustness and data verification test framework and tests.

Plan for initial implementation:
- Wrap fio to in a helper library for modifying and manipulating files and file contents
- Wrap fswalker in a helper library for capturing file system state and checking integrity
- Build framework to track verification metadata and issue actions against the repository and the data set
- Implement a test suite making use of the test framework to validate data after kopia restore

redgoat650:
### Goal

Ensure and verify end-to-end data consistency and long-term operational correctness in Kopia.

### Motivation and Requirements

Continuously ensuring data correctness for Kopia’s operations at increasingly larger scales over an extended period of time is highly valuable. It is worthwhile to implement a robustness framework and collection of jobs for verifying data integrity over long time scales, such as weeks, months or years. While continuous integration jobs ensure the correctness of changes at a particular point in time, executing the robustness framework tasks in the context of the CI tools, such as Travis, is impractical primarily due to resource constraints. Instead, robustness jobs should run periodically (semi-continuously) independent of the PR verification development workflow.

### Requirements

The robustness framework must have the ability to:

- Verify the consistency of the data stored in and restored from snapshots using an external end-to-end approach based on an implementation that is separate from the internal code paths used in Kopia. For example, by computing and verifying whole file / directory checksums.

- Automatically detect and highlight encountered data inconsistencies.

- Automatically evolve and grow a Kopia repository by performing operations such as new data generation, new snapshot creation, snapshot restore, data deletion and so on.

- Verify the consistency of a long-lived repository as it evolves over time. After an arbitrarily large number of operations have been executed through the framework, it should be possible to verify the data in any snapshot created by the framework in  the past.

- Handle increasingly larger repositories and data sources as time passes.

### Design

The Robustness Framework provides support for executing sequences of actions on a Kopia repository over an extended period of time and verifying the consistency of the snapshots in the repository. At the core of the framework is a flexible execution engine that (1) coordinates the execution of actions; and (2) manages persistent long-term metadata about the actions that have been executed on the repo. The framework allows the execution of arbitrary workloads composed by sequences of actions. The engine supports executing discrete actions such as:

- writing files to a local directory used as a data source for a snapshot;

- computing data verification checksums and creating a snapshot of a data source;

- restoring and verifying a snapshot;

- deleting a snapshot;

- executing GC;

- deleting files/directories from the local data directory and updating verification metadata.

[...]

PR #180, opened 2020-01-29
[robustness testing] Adding library for file generation using 'fio'

redgoat650:
Adding a helper library that wraps fio execution. This is the basic initial check-in that implements the runners, configs, and a single WriteFiles helper. It should be enough to unblock subsequent tasks that will use fio to generate data sets for kopia snapshot verification. More helper workloads can be added as needed.

In this implementation the tests will all skip from test main if the `FIO_EXE` env variable is not set. Adding fio to the CI environment will be addressed as a separate PR.

Tracking progress in issue https://github.com/kopia/kopia/issues/179

### DIFF
--- a/tests/tools/fio/config.go
+++ b/tests/tools/fio/config.go
@@ -1,0 +1,17 @@
+package fio
+
+import "strings"
+
+// Config structures the fields of a FIO job run configuration
+type Config []Job
+
+// String implements the stringer interface, formats the Config
+// as it would appear in a well-formed fio config file.
+func (cfg Config) String() string {
+	ret := []string{}
+	for _, job := range cfg {
+		ret = append(ret, job.String())
+	}
+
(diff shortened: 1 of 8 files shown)


redgoat650:
Travis CI is failing for an unrelated issue with the MakeBucket region, connecting to play.minio.io. Assumption in the test (and in minio documentation and everywhere else) is that the sever region is `us-east-1`. However currently the server appears to be in `us-east-2`. The MakeBucket command is failing silently (error is ignored in case the bucket already exists) and subsequent API calls are failing. If I change the MakeBucket region to `us-east-2` the test passes. I've made a branch that tries both regions before returning an error. I may open a PR for that if the problem persists.

redgoat650:
PR #181 is a workaround for the (unrelated) CI test failure.

PR #181, opened 2020-01-30
Parse MakeBucket error and try multiple regions for play.minio.io

redgoat650:
Attempt to connect to play.minio.io on either us-east-1 or us-east-2. Fail if the client can't create a bucket for either region.

Error will be parsed for the MakeBucket command in the TestS3Storage test. Current behavior is to ignore all errors. New behavior is to ignore nil error or already-exists error, try a new region if it returns wrong-region error, and fail for all other errors.

This PR acts as a workaround for the CI errors in the TestS3Storage test, e.g. those seen in #180

### DIFF
--- a/repo/blob/s3/s3_storage_test.go
+++ b/repo/blob/s3/s3_storage_test.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


julio-lopez:
The issue with play.min.io seems to have gone away. Closing this for now. We can revisit if needed. Does that work @redgoat650 ?
And thanks for looking into it.

PR #198, opened 2020-02-06
CoC and Governance Proposal

ntolia:
This is a proposal for introducing a lightweight governance proposal and a
Code of Conduct for Kopia.

Original Sources of CoC and Governance:
- https://www.contributor-covenant.org/version/2/0/code_of_conduct.html
- http://oss-watch.ac.uk/resources/meritocraticgovernancemodel

No changes to the CoC have been made.

Modulo the source credit, changes specific to the Governance model have been made as a separate patch.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,6 +1,8 @@
+# Kopia Governance
+
 # Overview
 
-This is a meritocratic, consensus-based community project. Anyone with an interest in the project can join the community, contribute to the project design and participate in the decision making process. This document describes how that participation takes place and how to set about earning merit within the project community.
+Kopia is a meritocratic, consensus-based community project. Anyone with an interest in the project can join the community, contribute to the project design and participate in the decision making process. This document describes how that participation takes place and how to set about earning merit within the project community.
 
 # Roles And Responsibilities
 
@@ -14,7 +16,8 @@ The project asks its users to participate in the project and community as much a
 * informing developers of strengths and weaknesses from a new user perspective
 * providing moral support (a ‘thank you’ goes a long way)
 * providing financial support (the software is open source, but its developers need to eat)
-* Users who continue to engage with the project and its community will often become more and more involved. Such users may find themselves becoming contributors, as described in the next section.
+
+Users who continue to engage with the project and its community will often become more and more involved. Such users may find themselves becoming contributors, as described in the next section.
 
 ## Contributors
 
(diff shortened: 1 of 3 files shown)


GOVERNANCE.md
```diff
@@ -14,7 +16,8 @@ The project asks its users to participate in the project and community as much a
 * informing developers of strengths and weaknesses from a new user perspective
 * providing moral support (a ‘thank you’ goes a long way)
 * providing financial support (the software is open source, but its developers need to eat)
-* Users who continue to engage with the project and its community will often become more and more involved. Such users may find themselves becoming contributors, as described in the next section.
```
ntolia: This was a copy-paste error from the webpage. Fixed in this patch.

GOVERNANCE.md
```diff
@@ -0,0 +1,120 @@
...
+* reporting bugs
+* identifying requirements
+* providing graphics and web design
+* programming
+* assisting with project infrastructure
+* writing documentation
+* fixing bugs
+* adding features
+
+Contributors engage with the project through the issue tracker, mailing list, the Slack channel, or by writing or editing documentation. They submit changes to the project itself via patches, which will be considered for inclusion in the project by existing committers (see next section). The Slack channel is the most appropriate place to ask for help when making that first contribution.
```
jkowalski: nit: perhaps link to https://slack.kopia.io on the first reference to Slack.

GOVERNANCE.md
```diff
@@ -0,0 +1,120 @@
...
+It is important to recognize that committership is a privilege, not a right. That privilege must be earned and once earned it can be removed by the PMC for conduct inconsistent with the [Code of Conduct](https://github.com/kopia/kopia/blob/master/CODE_OF_CONDUCT.md) or if they drop below a level of commitment and engagement required to be a Committer, as determined by the PMC. The PMC also reserves the right to remove a person for any other reason inconsistent with the goals of Kopia.
+
+A committer who shows an above-average level of contribution to the project, particularly with respect to its strategic direction and long-term health, may be nominated to become a member of the PMC. This role is described below.
+
+## Project Management Committee
+
+The current list of PMC members and their GitHub handles are:
+
+* Jarek Kowalski (jkowalski)
+* Julio Lopez (julio-lopez)
```
jkowalski: perhaps mention that it's desired for PMC to have 3 members, and third member can be added by unanimous decision of existing PMC members

GOVERNANCE.md
ntolia: Done.

GOVERNANCE.md
ntolia: I agree that an odd number is better. However, I am not a lawyer but am concerned if mentioning we want 3 but not having 3 might invalidate any part of this governance document. Currently, this works as-is and I believe that as long as you are supportive of getting to three, we should be good to go.

Moving forward, we might want to add a line like Vitess that also caps the PMC size.

jkowalski: makes sense. LGTM

PR #200, opened 2020-02-06
Update website

ntolia:
- New image of armor inspired by kopia's Polish meaning and the
  protection it provides
- Tweak text for greater clarity

Image from https://unsplash.com/photos/JOzv_pAkcMk under a permissive
license that doesn't require attribution -
https://unsplash.com/license


site/content/_index.html
```diff
@@ -13,7 +13,7 @@
 	<a class="btn btn-lg btn-secondary mr-3 mb-4" href="https://github.com/kopia/kopia/releases/">
 		Releases <i class="fab fa-github ml-2 "></i>
 	</a>
-	<p class="lead mt-5">Manage secure backups of your PCs, Macs and servers using your own cloud storage.</p>
+	<p class="h3 mt-5">Compressed, Deduplicated, and Encrypted Data Backups Using Your Own Cloud Storage.</p>
```
jkowalski: Not sure if Compressed is important here, but definitely not the first attribute. I would say Encrypted, Deduplicated and Compressed perhaps in this order.

jkowalski: also on a desktop the white text lands on a portion of image that is relatively bright, making it somewhat hard to see. ok on iPhone, though. Can we use different color perhaps?

ntolia: Will tweak text. Let me look at what the styles have to offer and will push another commit when I get to it.

ntolia: Also, the reason I put in compressed is that it is a competitive differentiator for kopia compared to some of the other tools out there.

ntolia: Tweak image and text. Please take a look now. If it still doesn't work, we might have to change the image.

jkowalski: that looks great

PR #204, opened 2020-02-07
Use Kopia favicons instead of docsy ones

ntolia:
Generated based on docs from https://www.docsy.dev/docs/adding-content/iconsimages/


jkowalski:
This is cool, I was thinking of ditching the shaded icon and going with flat everywhere (and b/w for the tray icon). WDYT?

jkowalski:
BTW, I've noticed that those hires icons are not full-size, but small and centered? Is that intentional? Where do they get used?

Also not sure how you generated all those sizes, I normally use Inkscape to export from SVG master to the size I need, which has the best quality.

ntolia:
> This is cool, I was thinking of ditching the shaded icon and going with flat everywhere (and b/w for the tray icon). WDYT?

I have learned (the hard way), that icons without too many "kinks" in the curves lines are better for stickers, etc. I agree with the flat idea but it depends on how it looks. If you would like a new icon, I have a designer I could talk to to get something done for us.

ntolia:
> BTW, I've noticed that those hires icons are not full-size, but small and centered? Is that intentional? Where do they get used?

There were auto-generated from icongen based of the helpful theme docs - https://www.docsy.dev/docs/adding-content/iconsimages/#add-your-favicons. They were generated from the SVG found in the repo.

> Also not sure how you generated all those sizes, I normally use Inkscape to export from SVG master to the size I need, which has the best quality.

I am assuming the theme does magic depending on the platform being used.

ntolia:
> BTW, I've noticed that those hires icons are not full-size, but small and centered? Is that intentional? Where do they get used?

Actually, not sure why they are small and centered :(. I can just add one favicon or I can look into this and we can iterate. If you have Inkscape, please feel free to upload images for the given dimensions.

ntolia:
@jkowalski Fixed using https://realfavicongenerator.net/

PR #208, opened 2020-02-07
Add fio install and robustness-tool-tests makefile target

redgoat650:
Fio install target is added, a robustness tools target invokes it and sets the FIO_EXE environment variable, launches tests from the robustness tools directory (for now).

Currently fio is only installed if the `apt-get` binary is found, otherwise install skips to avoid error. The `FIO_EXE` env is populated with the `fio` executable if it is found (independent of whether it was explicitly installed), empty otherwise. The tests will skip if that env is empty.

@jkowalski - I welcome any suggestions you have on how to do this better. There doesn't seem to be a common fio binary that can be downloaded the same way the other tools are in `tools.mk`. The documentation lists different binary download links for every os type and distro: https://fio.readthedocs.io/en/latest/fio_doc.html#binary-packages

### DIFF
--- a/tools/tools.mk
+++ b/tools/tools.mk
@@ -55,6 +55,15 @@ else
 	curl -LsS https://github.com/gohugoio/hugo/releases/download/v$(HUGO_VERSION)/hugo_extended_$(HUGO_VERSION)_macOS-64bit.tar.gz | tar zxv -C $(TOOLS_DIR)/hugo
 endif
 
+# fio 
+# - this install is currently conditional, only supported on systems with apt-get capability. Tests relying on fio should be skipped if it is not found.
+# - fio is installed using the package manager, and is thus not installed into TOOLS_DIR. It is not purged with the clean-tools target.
+fio:
+ifneq (, $(strip $(shell which apt-get)))
+	sudo apt-get -qq update
+	sudo apt-get -y install fio
+endif
+
 # linter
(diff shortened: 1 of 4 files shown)


tools/tools.mk
```diff
@@ -55,6 +55,15 @@ else
 	curl -LsS https://github.com/gohugoio/hugo/releases/download/v$(HUGO_VERSION)/hugo_extended_$(HUGO_VERSION)_macOS-64bit.tar.gz | tar zxv -C $(TOOLS_DIR)/hugo
 endif
 
+# fio 
+# - this install is currently conditional, only supported on systems with apt-get capability. Tests relying on fio should be skipped if it is not found.
+# - fio is installed using the package manager, and is thus not installed into TOOLS_DIR. It is not purged with the clean-tools target.
+fio:
+ifneq (, $(strip $(shell which apt-get)))
```
julio-lopez: @jkowalski thoughts?

jkowalski: can we avoid making the build Linux only and at least have macOS functionality? Also, installing any tool globally would be a first for Kopia - currently all tools are installed under "tools/.tools" so they are hermetic and independent of what the user has in the system.

jkowalski: one thought would be to use docker ?

redgoat650:
Closing in favor of #292 - using a docker image to wrap fio, thanks to @jkowalski for the suggestion

PR #210, opened 2020-02-08
revamped progress output and cleaned up logging

jkowalski:
See https://asciinema.org/a/J2QUzQg4zsr9oSqIVlBYgZukC

### DIFF
--- a/cli/cli_progress.go
+++ b/cli/cli_progress.go
@@ -3,127 +3,142 @@ package cli
 import (
 	"fmt"
 	"strings"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/kopia/kopia/internal/units"
-	"github.com/kopia/kopia/snapshot"
+	"github.com/kopia/kopia/snapshot/snapshotfs"
 )
 
-const maxUnshortenedPath = 60
+const spinner = `|/-\`
 
-type singleProgress struct {
-	desc      string
-	startTime time.Time
-	progress  int64
-	total     int64
+type cliProgress struct {
+	snapshotfs.NullUploadProgress
+
+	// all int64 must precede all int32 due to alignment requirements on ARM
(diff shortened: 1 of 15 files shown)


cli/cli_progress.go
```diff
@@ -3,127 +3,142 @@ package cli
...
+	p.maybeOutput()
 }
 
-func (p *singleProgress) toString(details bool) string {
-	if p.total == 0 {
-		return fmt.Sprintf("empty %v", p.desc)
-	}
+func (p *cliProgress) UploadedBytes(numBytes int64) {
+	atomic.AddInt64(&p.uploadedBytes, numBytes)
+	atomic.AddInt32(&p.uploadedFiles, 1)
```
julio-lopez: Does a call here really correspond to a (source) file being uploaded? or does it (strictly) refer to uploaded blobs?
In case it is the latter, it makes sense to rename the field, maybe to `uploadedBlobs` ?
In either case, it'd be good to expand the documentation for `UploadBytes` in the interface for clarity.

cli/command_snapshot_create.go
```diff
@@ -117,9 +120,22 @@ func snapshotSingleSource(ctx context.Context, rep *repo.Repository, u *snapshot
...
+
+	progress.Finish()
+
+	var maybePartial string
+	if manifest.IncompleteReason != "" {
+		maybePartial = " partial"
+	}
 
-	_, err = policy.ApplyRetentionPolicy(ctx, rep, sourceInfo, true)
+	printStderr("\nCreated%v snapshot with root %v and ID %v in %v\n", maybePartial, manifest.RootObjectID(), snapID, time.Since(t0).Truncate(time.Second))
```
julio-lopez: It turns out that we were parsing this. This modification changed the format and the order.

Filed #272 to track defining the output format for various commands

PR #212, opened 2020-02-10
Place fswalker packages behind build flag

redgoat650:
Temporary workaround for compile issues on MacOS and Windows due
to upstream fswalker bug. Only build the reporter and walker
packages as GOOS=linux for now.


julio-lopez:
Upstream fix for macOS google/fswalker#25

PR #207, opened 2020-02-07
Ignore read errors based on policy settings

redgoat650:
Added an error handling policy section. Can independently control error handling for directory and file read errors, toggle-able from the `policy set` command to either "true", "false", or "inherit". If any read error is hit, the error handling will check the effective policy on whether to ignore it or not. Currently there is no differentiation between read error types, though in the future we may want to add the `errors.Is(err, os.ErrPermission)` conditional.

Fix was implemented such that the policy ignores read errors ONLY on child entries of the source. So a snapshot will still fail if the source root directory itself can't be read, but you can ignore the error if a file or a subdirectory in the snapshot source root can't be read. I did this to address some otherwise strange behavior where you would successfully snapshot (because you ignored the error), but couldn't restore that snapshot because nothing really happened during the operation.

Fixes #177

PR #221, opened 2020-02-11
[minor] Fix error handling policy print

redgoat650:
Fixes print for error handling policy. Previously was printing
pointer address, now dereference pointer to print boolean itself.
Nil check in case the caller passes a policy with nil. In that
case print "false" which is the behavior in uploader for nil
pointer.

Issue was introduced when changing bool pair implementation
to bool pointer from PR suggestions in #207 and uncaught
at the time the swap was made.

before:
```
Error handling policy:
  Ignore file read errors:       0xc005f4dbfc       (default)
  Ignore directory read errors:  0xc005f4dbfd       (default)
```

after:
```
Error handling policy:
  Ignore file read errors:        true       (defined for this target)
  Ignore directory read errors:  false       (default)
```

### DIFF
--- a/cli/command_policy_show.go
+++ b/cli/command_policy_show.go
@@ -160,14 +160,24 @@ func printFilesPolicy(p *policy.Policy, parents []*policy.Policy) {
 func printErrorHandlingPolicy(p *policy.Policy, parents []*policy.Policy) {
 	printStdout("Error handling policy:\n")
 
+	ignoreFile := false
+	if p.ErrorHandlingPolicy.IgnoreFileErrors != nil {
+		ignoreFile = *p.ErrorHandlingPolicy.IgnoreFileErrors
+	}
+
 	printStdout("  Ignore file read errors:       %5v       %v\n",
-		p.ErrorHandlingPolicy.IgnoreFileErrors,
+		ignoreFile,
 		getDefinitionPoint(parents, func(pol *policy.Policy) bool {
 			return pol.ErrorHandlingPolicy.IgnoreFileErrors != nil
 		}))
 
+	ignoreDir := false
+	if p.ErrorHandlingPolicy.IgnoreDirectoryErrors != nil {
+		ignoreFile = *p.ErrorHandlingPolicy.IgnoreDirectoryErrors
+	}
+
 	printStdout("  Ignore directory read errors:  %5v       %v\n",
-		p.ErrorHandlingPolicy.IgnoreDirectoryErrors,
+		ignoreDir,
 		getDefinitionPoint(parents, func(pol *policy.Policy) bool {
 			return pol.ErrorHandlingPolicy.IgnoreDirectoryErrors != nil
 		}))
--- a/cli/command_policy_show.go
+++ b/cli/command_policy_show.go
@@ -161,13 +161,13 @@ func printErrorHandlingPolicy(p *policy.Policy, parents []*policy.Policy) {
 	printStdout("Error handling policy:\n")
 
 	printStdout("  Ignore file read errors:       %5v       %v\n",
-		p.ErrorHandlingPolicy.IgnoreDirectoryErrorsOrDefault(false),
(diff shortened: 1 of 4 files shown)


cli/command_policy_show.go
```diff
@@ -160,14 +160,24 @@ func printFilesPolicy(p *policy.Policy, parents []*policy.Policy) {
 func printErrorHandlingPolicy(p *policy.Policy, parents []*policy.Policy) {
 	printStdout("Error handling policy:\n")
 
+	ignoreFile := false
+	if p.ErrorHandlingPolicy.IgnoreFileErrors != nil {
```
jkowalski: nit: you can bind to local variable to avoid repeating long variable names.

ignoreFile := false
if ig := p.ErrorHandlingPolicy.IgnoreFileErrors; ig != nil {
  ignoreFile = *ig
}

cli/command_policy_show.go
```diff
@@ -160,14 +160,24 @@ func printFilesPolicy(p *policy.Policy, parents []*policy.Policy) {
...
 	printStdout("  Ignore file read errors:       %5v       %v\n",
-		p.ErrorHandlingPolicy.IgnoreFileErrors,
+		ignoreFile,
 		getDefinitionPoint(parents, func(pol *policy.Policy) bool {
 			return pol.ErrorHandlingPolicy.IgnoreFileErrors != nil
 		}))
 
+	ignoreDir := false
+	if p.ErrorHandlingPolicy.IgnoreDirectoryErrors != nil {
+		ignoreFile = *p.ErrorHandlingPolicy.IgnoreDirectoryErrors
```
golangcibot: ineffectual assignment to `ignoreFile` (from `ineffassign`)

cli/command_policy_show.go
```diff
@@ -160,14 +160,24 @@ func printFilesPolicy(p *policy.Policy, parents []*policy.Policy) {
...
+
 	printStdout("  Ignore file read errors:       %5v       %v\n",
-		p.ErrorHandlingPolicy.IgnoreFileErrors,
+		ignoreFile,
 		getDefinitionPoint(parents, func(pol *policy.Policy) bool {
 			return pol.ErrorHandlingPolicy.IgnoreFileErrors != nil
 		}))
 
+	ignoreDir := false
+	if p.ErrorHandlingPolicy.IgnoreDirectoryErrors != nil {
```
golangcibot: if statements should only be cuddled with assignments used in the if statement itself (from `wsl`)

cli/command_policy_show.go
jkowalski: btw, perhaps it's worth exposing helpers on ErrorHandlingPolicy itself, since this code is repeated already?

p.ErrorHandlingPolicy.IgnoreDirectoryErrorsOrDefault(defaultValue)
p.ErrorHandlingPolicy.IgnoreFileErrorsOrDefault(defaultValue)

WDYT ?

redgoat650: yeah good idea, I'll give it a shot

redgoat650: ✔️

cli/command_policy_show.go
```diff
@@ -160,24 +160,14 @@ func printFilesPolicy(p *policy.Policy, parents []*policy.Policy) {
...
 		}))
 
-	ignoreDir := false
-	if p.ErrorHandlingPolicy.IgnoreDirectoryErrors != nil {
-		ignoreFile = *p.ErrorHandlingPolicy.IgnoreDirectoryErrors
-	}
-
 	printStdout("  Ignore directory read errors:  %5v       %v\n",
-		ignoreDir,
+		p.ErrorHandlingPolicy.IgnoreFileErrorsOrDefault(false),
```
jkowalski: you have swapped File/Directory output

redgoat650: good catch. copy pasted wrong

jkowalski:
squish-merged, thanks for the PR!

PR #226, opened 2020-02-13
Add comparer interface and fswalker implementation

redgoat650:
Add comparer interface which gathers data on a path and
compares that data to a new path, returning error if the path
differs in any way from the input data. The details of what
constitutes a difference is left to the implementation.

FSWalker implementation uses Walk and Report to do the data
gathering and comparison. Filters are applied to sort out any
modifications that might be expected (e.g. ctime, atime, mtime,
rename of root directory after restore) from the report.

### DIFF
--- a/tests/tools/fswalker/fswalker.go
+++ b/tests/tools/fswalker/fswalker.go
@@ -7,15 +7,15 @@ package fswalker
 import (
 	"context"
 	"encoding/json"
-	"errors"
-	"fmt"
 	"io"
+	"log"
 	"path/filepath"
 	"strings"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/google/fswalker"
 	fspb "github.com/google/fswalker/proto/fswalker"
+	"github.com/pkg/errors"
 
 	"github.com/kopia/kopia/tests/robustness/checker"
 	"github.com/kopia/kopia/tests/tools/fswalker/reporter"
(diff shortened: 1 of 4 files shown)


tests/tools/fswalker/fswalker.go
```diff
@@ -0,0 +1,204 @@
...
+
+	err = validateReport(report)
+	if err != nil {
+		printReportSummary(report, reportOut)
+
+		b, marshalErr := json.MarshalIndent(report, "", "   ")
+		if marshalErr != nil {
+			_, reportErr := reportOut.Write([]byte(marshalErr.Error()))
+			if reportErr != nil {
+				return fmt.Errorf("error while writing marshal error (%v): %v", marshalErr.Error(), reportErr.Error())
```
jkowalski: use errors.Wrap() or Wrapf()

tests/tools/fswalker/fswalker.go
```diff
@@ -0,0 +1,204 @@
...
+// on the provided file path, unmarshals the comparison data as a fswalker Walk,
+// and generates a fswalker report comparing the two Walks. If there are any differences
+// an error is returned, and the full report will be written to the provided writer
+// as JSON.
+func (chk *WalkCompare) Compare(ctx context.Context, path string, data []byte, reportOut io.Writer) error {
+	beforeWalk := &fspb.Walk{}
+
+	err := proto.Unmarshal(data, beforeWalk)
+	if err != nil {
+		return err
```
jkowalski: can we use use errors.Wrap() instead of bare error propagation, same thing below?

PR #230, opened 2020-02-14
Add WalkCrossDevice option to walk policy

redgoat650:
Add the walk policy flag WalkCrossDevice to the fswalker Walk calls. This will avoid potential issues when running in a docker container where a FS tree is made of many overlays. Without the flag set, the Walk operation skips over files on devices that do not match the device at the base path.


julio-lopez:
@redgoat650 Can you provide an example of failure messages?

redgoat650:
Ex:
```
2020/02/14 02:37:04 skipping "/tmp/164698556/test-file": file is on different device
```
Worth noting: this does not propagate as an error, but just logged as a warning

PR #247, opened 2020-02-18
Add Asciicinema recording to the website

ntolia:


jkowalski:
is there a way to move the preview to be fully visible without scrolling and make it smaller? perhaps slightly to the right, like https://agones.dev/site/?

ntolia:
The Agones site has a bunch of custom CSS for this and in-built asciicinema support. Might be hard here without a heavy lift. That said, we can always iterate.

PR #261, opened 2020-02-21
Makefile: publish Kopia UI builds from PRs in separate repository

jkowalski:


jkowalski:
turns out GH token is not available

PR #268, opened 2020-02-23
Kopia UI & CLI: support for progress indicator (percentage based on last-known snapshot size)

jkowalski:


cli/cli_progress.go
```diff
@@ -99,6 +102,15 @@ func (p *cliProgress) output() {
 		units.BytesStringBase10(uploadedBytes),
 	)
 
+	if p.previousTotalSize > 0 {
+		percent := (float64(hashedBytes+cachedBytes) * 100.0 / float64(p.previousTotalSize))
+		if percent > 100 {
```
golangcibot: mnd: Magic number: 100, in <condition> detected (from `gomnd`)

snapshot/snapshotfs/upload.go
```diff
@@ -729,7 +729,20 @@ func (u *Uploader) Upload(
...
+	for _, m := range previousManifests {
+		if s := m.Stats.TotalFileSize; s > maxPreviousTotalFileSize {
+			maxPreviousTotalFileSize = s
+		}
+
+		if s := m.Stats.TotalFileCount; s > maxPreviousFileCount {
+			maxPreviousFileCount = s
+		}
+	}
+
```
julio-lopez: Would it make sense to make these values larger than the previous max by some factor?
The question is what that factor should be. At some point, for very large snapshots, a fraction of the existing snapshot size becomes much larger than a reasonable ingest / source growth rate.
Anyway, just a random idea, not even sure it makes sense.

PR #269, opened 2020-02-24
BREAKING: changed how hostname/username are handled

jkowalski:
The hostname/username are now persisted when connecting to repository in a local config file.

This prevents weird behavior changes when hostname is suddenly changed, such as when moving between networks.

repo.Repository will now expose Hostname/Username properties which are always guarnateed to be set, and are used throughout.

Removed --hostname/--username overrides when taking snapshot et.al.


julio-lopez:
> This breaks the current integration with other tools. A transition period is needed where the existing flags are supported, in particular for `snapshot create` and `restore`.

Actually, make that `snapshot create` only.

jkowalski:
> > This breaks the current integration with other tools. A transition period is needed where the existing flags are supported, in particular for `snapshot create` and `restore`.
> 
> Actually, make that `snapshot create` only.

Done

PR #277, opened 2020-02-28
crypto: deprecated crypto algorithms and replaced with better alternatives

jkowalski:
New ciphers are using authenticated encryption with associated data
(AEAD) and per-content key derived using HMAC-SHA256:

* `AES256-GCM-HMAC-SHA256`
* `CHACHA20-POLY1305-HMAC-SHA256`

They support content IDs of arbitrary length and are quite fast:

On my 2019 MBP:

```
- BLAKE2B-256 + AES256-GCM-HMAC-SHA256 - 648.7 MiB / second
- BLAKE2B-256 + CHACHA20-POLY1305-HMAC-SHA256 - 597.1 MiB / second
- HMAC-SHA256 + AES256-GCM-HMAC-SHA256 351 MiB / second
- HMAC-SHA256 + CHACHA20-POLY1305-HMAC-SHA256 316.2 MiB / second
```
Previous ciphers had several subtle issues:

* `SALSA20` encryption, used weak nonce (64 bit prefix of content ID),
  which means that for any two contents, whose IDs that have the same
  64-bit prefix, their plaintext can be decoded from the ciphertext
  alone.

* `AES-{128,192,256}-CTR` were not authenticated, so we were
  required to hash plaintext after decryption to validate. This is not
  recommended due to possibility of subtle timing attacks if an attacker
  controls the ciphertext.

* `SALSA20-HMAC` was only validating checksum and not that the ciphertext
  was for the correct content ID.

New repositories cannot be created using deprecated ciphers, but they
will still be supported for existing repositories, until at least 0.6.0.

The users are encouraged to migrate to one of new ciphers when 0.5.0 is
out.

### DIFF
--- a/repo/encryption/aes256_gcm_hmac_sha256_encryptor.go
+++ b/repo/encryption/aes256_gcm_hmac_sha256_encryptor.go
@@ -1,0 +1,66 @@
+package encryption
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/hmac"
+	"crypto/sha256"
+
+	"github.com/pkg/errors"
+)
+
+var zeroAES256GCMNonce = make([]byte, 12)
+
+type aes256GCMHmacSha256 struct {
+	keyDerivationSecret []byte
+}
+
+// aeadForContent returns cipher.AEAD using key derived from a given contentID.
+func (e aes256GCMHmacSha256) aeadForContent(contentID []byte) (cipher.AEAD, error) {
+	h := hmac.New(sha256.New, e.keyDerivationSecret)
+	if _, err := h.Write(contentID); err != nil {
+		return nil, errors.Wrap(err, "unable to derive encryption key")
+	}
+
+	key := h.Sum(nil)
+
+	c, err := aes.NewCipher(key)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to create AES-256 cipher")
+	}
+
+	return cipher.NewGCM(c)
+}
+
+func (e aes256GCMHmacSha256) Decrypt(input, contentID []byte) ([]byte, error) {
+	a, err := e.aeadForContent(contentID)
+	if err != nil {
+		return nil, err
+	}
+
+	return a.Open(nil, zeroAES256GCMNonce, input, contentID)
+}
+
+func (e aes256GCMHmacSha256) Encrypt(input, contentID []byte) ([]byte, error) {
+	a, err := e.aeadForContent(contentID)
+	if err != nil {
+		return nil, err
+	}
+
+	return a.Seal(nil, zeroAES256GCMNonce, input, contentID), nil
+}
+
+func (e aes256GCMHmacSha256) IsAuthenticated() bool {
+	return true
+}
+
+func init() {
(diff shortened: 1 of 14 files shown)


repo/encryption/encryption_test.go
```diff
@@ -73,14 +75,119 @@ func TestRoundTrip(t *testing.T) {
...
+			v, err := enc.Encrypt(payload, contentID)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if !t.Failed() {
+				t.Errorf("missing ciphertext sample")
+			}
+
+			t.Logf("%q: %q,", encryptionAlgo, hex.EncodeToString(v))
```
julio-lopez: Since going down this block always fails the test, might as well collapse it to be explicit:

```suggestion
			t.Errorf("missing ciphertext sample for %q: %q,", encryptionAlgo, hex.EncodeToString(v))
```

repo/encryption/aes256_gcm_hmac_sha256_encryptor.go
```diff
@@ -0,0 +1,63 @@
...
+	h := hmac.New(sha256.New, e.keyDerivationSecret)
+	if _, err := h.Write(contentID); err != nil {
+		return nil, errors.Wrap(err, "unable to derive encryption key")
+	}
+
+	key := h.Sum(nil)
+
+	c, err := aes.NewCipher(key)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to crete AES-256 cipher")
```
julio-lopez:
```suggestion
		return nil, errors.Wrap(err, "unable to create AES-256 cipher")
```

repo/encryption/aes256_gcm_hmac_sha256_encryptor.go
```diff
@@ -0,0 +1,63 @@
...
+
+var zeroAES256GCMNonce = make([]byte, 12)
+
+type aes256GCMHmacSha256 struct {
+	keyDerivationSecret []byte
+}
+
+// aeadForContent returns cipher.AEAD using key derived from a given contentID.
+func (e aes256GCMHmacSha256) aeadForContent(contentID []byte) (cipher.AEAD, error) {
+	h := hmac.New(sha256.New, e.keyDerivationSecret)
```
julio-lopez: Not sure where the right place would be, but it may make sense to enforce a minimum length for `keyDerivationSecret` to prevent incorrect use.
Notice that:
* since this is not the key directly passed to the cipher below, the cipher does not validate that `keyDerivationSecret` has a minimum length
* the `init()` below generated as 32-byte key, but there are no checks to prevent accidentally creating a short `keyDerivationSecret`

jkowalski: it just so happens that deriveKey() always takes 32 bytes today (that's the reason for // nolint:unparam). I added error condition to make sure it fails for attempts to derive shorter keys.

repo/encryption/aes256_gcm_hmac_sha256_encryptor.go
jkowalski: done

repo/encryption/encryption_test.go
jkowalski: done

julio-lopez:
LG +1 👍

PR #289, opened 2020-03-01
testing: better fix for ignoring ErrClosed

jkowalski:


jkowalski:
it turns our that os.ErrClosed is wrapped in *os.PathError

PR #291, opened 2020-03-02
testing: added lease mechanism in GCS to protect tests from stepping on each other's toes

jkowalski:
This will be only configured on Travis.

### DIFF
--- a/repo/blob/gcs/gcs_storage_test.go
+++ b/repo/blob/gcs/gcs_storage_test.go
@@ -4,12 +4,19 @@ import (
 	"io/ioutil"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/kopia/kopia/internal/blobtesting"
 	"github.com/kopia/kopia/internal/testlogging"
+	"github.com/kopia/kopia/tests/testlease"
 
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/blob/gcs"
+)
+
+const (
+	testLeaseID       = "gcs-test"
+	testLeaseDuration = 10 * time.Minute
 )
 
 func TestGCSStorage(t *testing.T) {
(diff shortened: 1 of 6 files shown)


tests/testlease/testlease.go
```diff
@@ -0,0 +1,235 @@
...
+	w := oh.If(cond).NewWriter(ctx)
+
+	if err := json.NewEncoder(w).Encode(ld); err != nil {
+		return false, err
+	}
+
+	if err := w.Close(); err != nil {
+		if ae, ok := err.(*googleapi.Error); ok {
+			if ae.Code == http.StatusPreconditionFailed {
+				// already exists
```
julio-lopez: The comment may be misleading, the condition could also be `GenerationMatch`

tests/testlease/testlease.go
```diff
@@ -0,0 +1,235 @@
...
+	ctx := context.Background()
+
+	myOwner := uuid.New().String()
+
+	if err := tryAcquireLease(ctx, t, oh, myOwner, leaseDuration); err != nil {
+		t.Fatalf("unable to acquire lease: %v", err)
+	}
+
+	defer releaseLeaseIfStillOwned(ctx, t, oh, myOwner)
+	cb()
```
julio-lopez: Not requesting a change, just an observation.
There isn't anything attempting to guarantee that `cb` does not run past the lease duration. It seems that'd it'd be to the caller to choose a lease duration that matches the test timeout.
If `cb` took a `Context`, then maybe a deadline could be set on the context and pass it down. Nevertheless it'd be up to `cb` to honor the context cancellation.

repo/blob/gcs/gcs_storage_test.go
```diff
@@ -23,35 +30,37 @@ func TestGCSStorage(t *testing.T) {
...
 
-	ctx := testlogging.Context(t)
-	st, err := gcs.New(ctx, &gcs.Options{
-		BucketName:                   bucket,
-		ServiceAccountCredentialJSON: credData,
-	})
+	testlease.RunWithLease(t, testLeaseID, testLeaseDuration, func() {
+		ctx := testlogging.Context(t)
+		st, err := gcs.New(ctx, &gcs.Options{
+			BucketName:                   bucket,
```
julio-lopez: If a prefix is specified, then the tests could share the bucket and not step into each other toes. Would that work?
In fact, we'd want to test when multiple repos share the same bucket.
Same for the S3 tests.

repo/blob/s3/s3_storage_test.go
```diff
@@ -187,8 +197,10 @@ func TestS3StorageMinioSTS(t *testing.T) {
 		t.Skip("endpoint not reachable")
 	}
 
-	createBucket(t, options)
-	testStorage(t, options)
+	testlease.RunWithLease(t, testLeaseID, testLeaseDuration, func() {
```
julio-lopez: It seems that it is desirable to run this one with a lease. The observation has been that either the created user or generated token disappear under load.

jkowalski:
abandoning this, the issue with not being able to run on PRs is not solvable this way.

PR #292, opened 2020-03-02
Use docker to execute fio in robustness tool

redgoat650:
Update the fio runner to use a docker image if the appropriate environment variable is set. Docker image is built via a makefile target and used in the robustness tool tests. The image uses `fio` as an entrypoint, so from the standpoint of the fio runner package implementation, it behaves similarly to a local executable fio.

This is a replacement implementation for #208, which I will close shortly.

### DIFF
--- a/tests/tools/fio/fio.go
+++ b/tests/tools/fio/fio.go
@@ -36,9 +36,7 @@ const (
 
 	// FioDockerImageEnvKey specifies the docker image tag to use. If
 	// FioExeEnvKey is set, the local executable will be used instead of
-	// docker, even if this variable is also set. The exception is if
-	// FioUseDockerEnvKey is not an empty string, which will force
-	// use of the fio docker image independent of FioExeEnvKey.
+	// docker, even if this variable is also set.
 	FioDockerImageEnvKey = "FIO_DOCKER_IMAGE"
 
 	// LocalFioDataPathEnvKey is the local path where fio data will be
@@ -49,10 +47,6 @@ const (
 	// relative to the docker host. If left blank, defaults to local fio data path
 	// (works unless running via docker from within a container, e.g. for development)
 	HostFioDataPathEnvKey = "HOST_FIO_DATA_PATH"
-
-	// FioUseDockerEnvKey forces the fio runner to use the docker image, even if
-	// an executable path is provided.
-	FioUseDockerEnvKey = "FIO_USE_DOCKER"
 )
 
 // Known error messages
@@ -75,7 +69,6 @@ func NewRunner() (fr *Runner, err error) {
 	imgStr := os.Getenv(FioDockerImageEnvKey)
 	localFioDataPathStr := os.Getenv(LocalFioDataPathEnvKey)
 	hostFioDataPathStr := os.Getenv(HostFioDataPathEnvKey)
-	forceDocker := os.Getenv(FioUseDockerEnvKey) != ""
 
 	var exeArgs []string
 
@@ -89,7 +82,7 @@ func NewRunner() (fr *Runner, err error) {
 	}
 
 	switch {
-	case exeStr != "" && !forceDocker:
+	case exeStr != "":
 		// Provided a local FIO executable to run
 		Exe = exeStr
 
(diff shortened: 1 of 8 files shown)


tests/tools/fio/workload.go
```diff
@@ -4,29 +4,46 @@ import (
...
+
 	_, _, err = fr.RunConfigs(Config{
 		{
-			Name: fmt.Sprintf("write-%vB-%v", sizeB, numFiles),
-			Options: opt.Merge(Options{
-				"size":      strconv.Itoa(int(sizeB)),
-				"nrfiles":   strconv.Itoa(numFiles),
-				"directory": fullPath,
-			}),
+			Name: fmt.Sprintf("writeFiles"),
```
julio-lopez:
```suggestion
			Name: "writeFiles",
```
Or do you want to preserve some identifying info in the name?

tests/tools/fio/fio.go
```diff
@@ -5,45 +5,130 @@
...
+	// LocalFioDataPathEnvKey is the local path where fio data will be
+	// accessible. If not specified, defaults to the default temp directory (os.TempDir)
+	LocalFioDataPathEnvKey = "LOCAL_FIO_DATA_PATH"
+
+	// HostFioDataPathEnvKey specifies the path where fio data will be written,
+	// relative to the docker host. If left blank, defaults to local fio data path
+	// (works unless running via docker from within a container, e.g. for development)
+	HostFioDataPathEnvKey = "HOST_FIO_DATA_PATH"
+
+	// FioUseDockerEnvKey forces the fio runner to use the docker image, even if
```
julio-lopez: Is the "force" env var really needed? aren't `FIO_EXE` and `FIO_DOCKER_IMAGE` enough to control the desired behavior? or is there a particular case not handled with those?

tests/tools/fio/fio.go
```diff
@@ -5,45 +5,130 @@
...
+		// data directory path and we are not running from within a container already
+		if hostFioDataPathStr == "" {
+			hostFioDataPathStr = dataDirParent
+		}
+
+		exeArgs = []string{
+			"run",
+			"--rm",
+			"-v",
+			fmt.Sprintf("%s:%s", hostFioDataPathStr, fioDataContainerPath), // /c/Users/usr/fio-data:/fio-data
```
julio-lopez: Is the comment at the end needed?
```suggestion
			fmt.Sprintf("%s:%s", hostFioDataPathStr, fioDataContainerPath),
```

tests/tools/fio/fio.go
```diff
@@ -54,27 +139,82 @@ func NewRunner() (*Runner, error) {
...
+	nrFiles := rand.Intn(maxTestFiles) + 1
+
+	opt := Options{}.WithNumFiles(nrFiles).WithFileSize(fileSizeB)
+
+	err := fr.WriteFiles(subDirPath, opt)
+	if err != nil {
+		return errors.Wrap(err, "unable to perform writes")
+	}
+
+	defer fr.DeleteRelDir("test") //nolint:errcheck
```
julio-lopez: Would this work if `defer fr.DeleteRelDir("test")` were prior to `fr.WriteFiles`? What happens when `WriteFiles` returns an error after having written some files, but not completing successfully?

julio-lopez:
LG +1 👍
A few minor comments

tests/tools/fio/fio.go
redgoat650: good call, will move it

tests/tools/fio/fio.go
redgoat650: yeah it's not strictly needed. Was added as a convenience since setting and clearing this dumb switch was slightly easier than setting and clearing a path to the fio executable, plus it reduces ambiguity on what will be used if both EXE and DOCKER_IMAGE are set. I'll remove it.

julio-lopez: The precedence for which one is used should be documented. The ambiguity was present anyway when the "force" env var was not specified, or had a falsey value

Issue #284, opened 2020-02-29
TestSnapshotFail is still flaky and pretty slow


jkowalski:
TestSnapshotFail often flakes out

Example, in which it took 3 minutes:
https://api.travis-ci.org/v3/job/656585634/log.txt

```
    snapshot_fail_test.go:195: 'kopia snapshot create /var/folders/17/5mc7816d3mndxjqgplq6057w0000gn/T/kopia-test833197020/TestSnapshotFail/a7cfc43a/dir0' succeeded, but expected failure
```

Can we have a goal for this test to complete in <30s?

Ideas for improving:

- add retries as much as possible
- parallelize all the cases (perhaps create N sources instead of trying to mutate one in place)?
- perhaps test the bulk of those conditions directly Uploader API and either mockfs or localfs to avoid the overhead of running kopia executable all the time?

jkowalski:
btw, I've added retries to travis, so this should be less annoying. In the log retry can be identified by searching for "attempt 2"

jkowalski:
Stil getting flakes despite 3 retries:

snapshot_delete_test.go:239: error reading out of the stderr pipe: read |0: file already closed
3978

this one actually failed 3 times in a row in https://api.travis-ci.org/v3/job/656770936/log.txt

redgoat650:
Hmm not sure why that's happening. Feel free to revert the prev fix and I can look into it when I get the chance

redgoat650:
This could be a separate issue/quirk that was masked by the clobbered error, since most of the time the Output() returned after ReadAll (i.e. errors from the latter would be often masked unintentionally, which is why the snap fail test looked more like a flake than a regular problem).

redgoat650:
Update: 
- #293 should fix the test flakes. All tests were susceptible to the bug, but `TestSnapshotFail` issues many more `RunAndExpectFailure` commands, making it much more likely to be the one that exhibited the flaky behavior (the most common symptom was error=<nil> when expecting non-nil, because the error variable was clobbered by the second goroutine).
- I have a branch that parallelizes the TestSnapshotFail test, should be good to go, and I will open a PR soon.

PR #298, opened 2020-03-03
Parallelizing snapshot fail test to decrease test time

redgoat650:
Make each test case combination a sub-test, and running independently it in parallel. This gives a 3-4x speedup vs master in local testing.


redgoat650:
Addresses #284

PR #302, opened 2020-03-06
kopia-ui: misc

jkowalski:
- launch server on a randomized port in KopiaUI
- added support for connecting to remote server
- embedded version info in the app and htmlui

### DIFF
--- a/tools/tools.mk
+++ b/tools/tools.mk
@@ -140,6 +140,27 @@ else
 	curl -LsS https://github.com/goreleaser/goreleaser/releases/download/$(GORELEASER_VERSION)/goreleaser_$$(uname -s)_$$(uname -m).tar.gz | tar zx -C $(TOOLS_DIR)/goreleaser-$(GORELEASER_VERSION)
 endif
 
+ifeq ($(TRAVIS_PULL_REQUEST),false)
+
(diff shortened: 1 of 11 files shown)


tools/tools.mk
```diff
@@ -140,6 +140,27 @@ else
...
+KOPIA_VERSION:=$(TRAVIS_TAG:v%=%)
+else
+# travis, non-tagged release
+KOPIA_VERSION:=$(shell date +%Y%m%d).0.$(TRAVIS_BUILD_NUMBER)
+endif
+
+else
+
+# non-travis, or travis PR
+KOPIA_VERSION:=$(shell date +%Y%m%d).0.0-$(shell git rev-parse --short HEAD)
```
julio-lopez: non-essential, but would this work?
```suggestion
KOPIA_VERSION:=$(shell date +%Y%m%d)-$(shell git describe --candidates=1 HEAD)
```

The version would be something like
`20200305-0.5.0-test5-<count>-g<short-commit-id>`

jkowalski: .0.0- is to make this semver (which KopiaUI actually cares about for update checks)

julio-lopez: `$(shell date +%Y.%m.%d)-<whatever>`

PR #304, opened 2020-03-07
cli: 'snapshot migrate' improvements to help with data migration

jkowalski:
- cleaned up migration progress output
- fixed migration idempotency
- added migration of policies
- renamed --parallelism to --parallel
- improved e2e test
- do not prompt for password to source repository if persisted

### DIFF
--- a/cli/command_snapshot_migrate.go
+++ b/cli/command_snapshot_migrate.go
@@ -15,22 +15,19 @@ import (
 )
 
 var (
-	migrateCommand      = snapshotCommands.Command("migrate", "Migrate snapshots from another repository")
-	migrateSourceConfig = migrateCommand.Flag("source-config", "Configuration file for the source repository").Required().ExistingFile()
-	migrateSources      = migrateCommand.Flag("sources", "List of sources to migrate").Strings()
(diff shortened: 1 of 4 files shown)


cli/command_snapshot_migrate.go
```diff
@@ -96,18 +107,89 @@ func runMigrateCommand(ctx context.Context, destRepo *repo.Repository) error {
...
+
+func migrateAllPolicies(ctx context.Context, sourceRepo, destRepo *repo.Repository) error {
+	policies, err := policy.ListPolicies(ctx, sourceRepo)
+	if err != nil {
+		return errors.Wrap(err, "unable to list source policies")
+	}
+
+	for _, pol := range policies {
+		if err := migrateSinglePolicy(ctx, sourceRepo, destRepo, pol.Target()); err != nil {
+			log(ctx).Warningf("unable to migrate policy for %v: %v", pol.Target(), err)
```
julio-lopez: It is surprising that the function signature has an error return value, but then it swallows the errors in some cases. it'd may be desirable to not swallow the error. Thoughts?

PR #314, opened 2020-03-09
Inject time in Kopia components

julio-lopez:
Motivation: Allow time injection for (unit) tests, to more easily test and verify time-dependent invariants.

Add time injection support for:
- `repo.Manager`
- `manifest.Manager`
- `snapshot.Uploader`

Then, wire up to these components. The `content.Manager` already had support for time injection, but was not wired up from the time function passed to repo creation.

Add an `internal/faketime` package for testing. Mainly code movement from testing code in the `repo/content` package. Motivation: make it available to other packages outside content
Also, add simple tests for `faketime` functions.

### DIFF
--- a/internal/faketime/faketime.go
+++ b/internal/faketime/faketime.go
@@ -30,3 +30,36 @@ func AutoAdvance(t time.Time, dt time.Duration) func() time.Time {
 		return ret
 	}
 }
+
+// TimeAdvance allows controlling the passage of time. Intended to be used in
+// tests.
+type TimeAdvance struct {
+	mu sync.Mutex
+	t  time.Time
+}
+
+// NewTimeAdvance creates a TimeAdvance with the given start time
+func NewTimeAdvance(start time.Time) *TimeAdvance {
+	return &TimeAdvance{t: start}
+}
+
+// NowFunc returns a time provider function for t
+func (t *TimeAdvance) NowFunc() func() time.Time {
+	return func() time.Time {
+		t.mu.Lock()
+		defer t.mu.Unlock()
+
+		return t.t
+	}
+}
+
+// Advance advances t by dt, such that the next call to t.NowFunc()() returns
+// current t + dt
+func (t *TimeAdvance) Advance(dt time.Duration) time.Time {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	t.t = t.t.Add(dt)
+
+	return t.t
+}
--- a/internal/faketime/faketime.go
+++ b/internal/faketime/faketime.go
@@ -1,0 +1,32 @@
+// Package faketime fakes time for tests
+package faketime
+
+import (
+	"sync"
+	"time"
+)
+
+// Frozen returns a function that always returns t
+func Frozen(t time.Time) func() time.Time {
+	return func() time.Time {
+		return t
+	}
+}
+
+// AutoAdvance returns a time source function that returns a time equal to
+// 't + ((n - 1) * dt)' wheren n is the number of serialized invocations of
+// the returned function. The returned function will generate a time series of
+// the form [t, t+dt, t+2dt, t+3dt, ...]
+func AutoAdvance(t time.Time, dt time.Duration) func() time.Time {
+	var mu sync.Mutex
+
+	return func() time.Time {
+		mu.Lock()
+		defer mu.Unlock()
+
+		ret := t
+		t = t.Add(dt)
+
+		return ret
+	}
+}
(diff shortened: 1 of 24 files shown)


julio-lopez:
Broken down into self-contained commits to facilitate reviewing. Just follow the commit train.

internal/faketime/faketime.go
```diff
@@ -30,3 +30,36 @@ func AutoAdvance(t time.Time, dt time.Duration) func() time.Time {
 		return ret
 	}
 }
+
+// TimeAdvance allows controlling the passage of time. Intended to be used in
+// tests.
+type TimeAdvance struct {
+	mu sync.Mutex
+	t  time.Time
```
jkowalski: nit (optional): this could be int64 and use atomic.AddInt64() to advance without a mutex.

jkowalski:
this is awesome, I have one request: can we add a test that scans all *.go files and asserts that time.Now() is not used except in whitelisted cases? 

This could even be a shell script/makefile target (which we can only run on mac/linux, not windows)

```find . -name '*.go' | grep -v _test | xargs grep -e time.Now -e time.Since -e time.Until | ... to-be-continued```

File: internal/faketime/faketime.go (hunk 1 above)
Comment (julio-lopez) [reply internal/faketime/faketime.go]: Then the type could not be `time.Time`.

We could have something like what's shown below, is that what you had in mind.

Not sure it makes much of a difference in tests though.

File: internal/faketime/faketime.go [hunk 2]
```diff
@@ -30,3 +30,36 @@ func AutoAdvance(t time.Time, dt time.Duration) func() time.Time {
...
+}
```
julio-lopez:
```suggestion
type TimeAdvance struct {
	base  time.Time
	int64 advance
}
```

Then
```go
func (t *TimeAdvance) NowFunc() func() time.Time {
	return func() time.Time {
		dt := time.Duration(atomic.LoadInt64((*int64)(&t.delta)))

		return t.base.Add(dt)
	}
}

func (t *TimeAdvance) Advance(dt time.Duration) time.Time {
	advance := atomic.AddInt64((*int64)(&t.delta), int64(t.delta))

	return t.base.Add(time.Duration(advance))
}
```

internal/faketime/faketime.go
```diff
@@ -30,3 +30,36 @@ func AutoAdvance(t time.Time, dt time.Duration) func() time.Time {
 		return ret
 	}
 }
+
+// TimeAdvance allows controlling the passage of time. Intended to be used in
+// tests.
+type TimeAdvance struct {
```
julio-lopez: Actually, this name stutters, suggestions?

julio-lopez:
> this is awesome, I have one request: can we add a test that scans all *.go files and asserts that time.Now() is not used except in whitelisted cases?
> 
> This could even be a shell script/makefile target (which we can only run on mac/linux, not windows)
> 
> `find repo/ -name '*.go' | grep -v _test | xargs grep -e time.Now -e time.Since -e time.Until | ... to-be-continued`

Probably the easiest way to implement the whitelist would be a "no-lint-style" comment such as `// allow:no-inject-time`

Thoughts?

internal/faketime/faketime.go
jkowalski: oh, I meant time.Unix(0, atomic.AddInt64(...))

jkowalski: but that's fine either way.

snapshot/snapshotfs/source_snapshots.go
```diff
@@ -40,7 +40,7 @@ func (s *sourceSnapshots) Sys() interface{} {
 }
 
 func (s *sourceSnapshots) ModTime() time.Time {
-	return time.Now()
+	return s.rep.Time()
```
jkowalski: BTW this is completely arbitrary, perhaps (separate PR) we could use last repo sync time?

internal/faketime/faketime.go
```diff
@@ -34,32 +35,28 @@ func AutoAdvance(t time.Time, dt time.Duration) func() time.Time {
...
+		return t.base.Add(dt)
 	}
 }
 
 // Advance advances t by dt, such that the next call to t.NowFunc()() returns
 // current t + dt
 func (t *TimeAdvance) Advance(dt time.Duration) time.Time {
-	t.mu.Lock()
-	defer t.mu.Unlock()
+	advance := atomic.AddInt64((*int64)(&t.delta), int64(dt))
```
jkowalski: i'm actually surprised this works (casting to *int64).

<optional>
in my comment I was thinking about keeping two numbers, which would be int64.

Something like:

```
type TimeAdvance struct {
  timestampNanoseconds int64
  advanceNanoseconds int64
}
```

then use `time.Unix(0, atomic.LoadInt64(&t.timestampNanoseconds))`
</optional>

julio-lopez: This can be simplified so some of the casts can be removed.

PR #315, opened 2020-03-09
Add snapshotter interface and kopia implementation

redgoat650:
Snapshotter interface describes an entity that can create,
restore, and delete snapshots, as well as manage a repository.

Add kopia implementation of the snapshotter interface.

### DIFF
--- a/tests/robustness/snap/snap.go
+++ b/tests/robustness/snap/snap.go
@@ -1,0 +1,22 @@
+// Package snap describes entities that are capable of performing
+// common snapshot operations
+package snap
+
+// Snapshotter is an interface that describes methods
+// for taking, restoring, deleting snapshots, and
+// tracking them by a string snapshot ID.
+type Snapshotter interface {
+	RepoManager
+	CreateSnapshot(sourceDir string) (snapID string, err error)
+	RestoreSnapshot(snapID string, restoreDir string) error
+	DeleteSnapshot(snapID string) error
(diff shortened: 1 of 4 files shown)


tests/robustness/snap/snap.go
```diff
@@ -0,0 +1,22 @@
+// Package snap describes entities that are capable of performing
+// common snapshot operations
+package snap
+
+// Snapshotter is an interface that describes methods
+// for taking, restoring, deleting snapshots, and
+// tracking them by a string snapshot ID.
+type Snapshotter interface {
```
jkowalski: is there a reason why we have all these interfaces? this feels very Java-ish...

redgoat650: ultimately the framework can be applied to any backup tool that can implement this interface. I wanted to maintain the flexibility of running experiments and workloads against any arbitrary tool for apples-to-apples comparison against kopia.

jkowalski: ok, could it be perhaps separate repo under kopia (longer term, not now)?

redgoat650: For sure. This was discussed with @julio-lopez when scoping out the framework. We decided to add it under the rough umbrella of kopia test tools for now, but keeping it independent enough that it can be eventually transitioned to a separate repo (ideally without much overhead to do so).

PR #316, opened 2020-03-09
Trivial cleanup for internal diff

julio-lopez:


internal/diff/diff.go
```diff
@@ -241,7 +241,7 @@ func (c *Comparer) compareFiles(ctx context.Context, f1, f2 fs.File, fname strin
 	}
 
 	if f2 != nil {
-		newName = filepath.Clean("new/" + fname)
+		newName = filepath.Join("new", fname)
```
jkowalski: i think there was a reason for doing it this way, but I can't figure out what it was...

julio-lopez: AFAICT, `Join` cleans the path after joining.

PR #318, opened 2020-03-10
repo: refactored public API

jkowalski:
This is 99% mechanical:

Extracted repo.Repository interface that only exposes high-level object and manifest management methods, but not blob nor content management.

Renamed old *repo.Repository to *repo.DirectRepository

Reviewed codebase to only depend on repo.Repository as much as possible, but added way for low-level CLI commands to use DirectRepository.

### DIFF
--- a/repo/manifest/manifest_manager.go
+++ b/repo/manifest/manifest_manager.go
@@ -122,25 +122,11 @@ func (m *Manager) GetMetadata(ctx context.Context, id ID) (*EntryMetadata, error
 
 // Get retrieves the contents of the provided manifest item by deserializing it as JSON to provided object.
 // If the manifest is not found, returns ErrNotFound.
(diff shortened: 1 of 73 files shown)


repo/manifest/manifest_manager.go
```diff
@@ -122,25 +122,11 @@ func (m *Manager) GetMetadata(ctx context.Context, id ID) (*EntryMetadata, error
...
-	}
-
-	b, err := m.GetRaw(ctx, id)
-	if err != nil {
-		return err
-	}
-
-	if err := json.Unmarshal(b, data); err != nil {
-		return errors.Wrapf(err, "unable to unmashal %q", id)
+		return nil, err
```
julio-lopez: Is this block and the one below duplicated?

PR #320, opened 2020-03-10
Fix Travis CI Windows job

julio-lopez:
Add `--force` to curl installation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
 - |-
     case $TRAVIS_OS_NAME in
       windows)
-        choco install --no-progress -y make unzip curl
+        choco install --no-progress --force -y make unzip curl
         ;;
       osx):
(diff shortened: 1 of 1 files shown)


jkowalski:
I don't think this will fix it, the root cause appears to be:

```
unzip (exited 1) - unzip not installed. An error occurred during installation:
129 The operation has timed out
```

julio-lopez:
> I don't think this will fix it, the root cause appears to be:
> 
> ```
> unzip (exited 1) - unzip not installed. An error occurred during installation:
> 129 The operation has timed out
> ```

You are right. On the failure in another PR, I stopped reading past:

```
Warning:
 - curl - curl v7.68.0 already installed.
 Use --force to reinstall, specify a version to install, or try upgrade.
```

It was actually a warning and not an error. The root cause was right below as you pointed out.

PR #331, opened 2020-03-11
Improved upload performance and memory usage

jkowalski:
This is first in series of refactorings to dramatically reduce memory usage of large uploads.

Improves #93

### DIFF
--- a/snapshot/snapshotfs/upload.go
+++ b/snapshot/snapshotfs/upload.go
@@ -9,10 +9,10 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"sync"
 	"sync/atomic"
 
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/kopia/kopia/fs"
 	"github.com/kopia/kopia/fs/ignorefs"
@@ -70,13 +70,13 @@ func (u *Uploader) cancelReason() string {
 	return ""
 }
 
-func (u *Uploader) uploadFileInternal(ctx context.Context, relativePath string, f fs.File, pol *policy.Policy) entryResult {
+func (u *Uploader) uploadFileInternal(ctx context.Context, relativePath string, f fs.File, pol *policy.Policy) (*snapshot.DirEntry, error) {
 	u.Progress.HashingFile(relativePath)
 	defer u.Progress.FinishedHashingFile(relativePath, f.Size())
 
 	file, err := f.Open(ctx)
 	if err != nil {
-		return entryResult{err: errors.Wrap(err, "unable to open file")}
+		return nil, errors.Wrap(err, "unable to open file")
 	}
 	defer file.Close() //nolint:errcheck
 
@@ -88,36 +88,36 @@ func (u *Uploader) uploadFileInternal(ctx context.Context, relativePath string, 
 
 	written, err := u.copyWithProgress(writer, file, 0, f.Size())
 	if err != nil {
-		return entryResult{err: err}
+		return nil, err
 	}
 
 	fi2, err := file.Entry()
 	if err != nil {
-		return entryResult{err: err}
+		return nil, err
 	}
 
 	r, err := writer.Result()
 	if err != nil {
-		return entryResult{err: err}
+		return nil, err
 	}
 
 	de, err := newDirEntry(fi2, r)
 	if err != nil {
-		return entryResult{err: errors.Wrap(err, "unable to create dir entry")}
+		return nil, errors.Wrap(err, "unable to create dir entry")
 	}
 
 	de.FileSize = written
 
-	return entryResult{de: de}
+	return de, nil
 }
 
-func (u *Uploader) uploadSymlinkInternal(ctx context.Context, relativePath string, f fs.Symlink) entryResult {
+func (u *Uploader) uploadSymlinkInternal(ctx context.Context, relativePath string, f fs.Symlink) (*snapshot.DirEntry, error) {
 	u.Progress.HashingFile(relativePath)
 	defer u.Progress.FinishedHashingFile(relativePath, f.Size())
 
 	target, err := f.Readlink(ctx)
 	if err != nil {
-		return entryResult{err: errors.Wrap(err, "unable to read symlink")}
+		return nil, errors.Wrap(err, "unable to read symlink")
 	}
 
 	writer := u.repo.Objects.NewWriter(ctx, object.WriterOptions{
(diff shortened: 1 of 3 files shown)


snapshot/snapshotfs/upload.go
```diff
@@ -423,117 +435,65 @@ func (u *Uploader) prepareWorkItems(ctx context.Context, dirRelativePath string,
...
-		wg.Add(1)
+	for _, ent := range results {
+		if ent == nil {
+			continue
+		}
 
-		go func() {
-			defer wg.Done()
+		if ent.Type == snapshot.EntryTypeFile {
+			u.stats.TotalFileCount++
```
julio-lopez: What guarantees that the uploader stats aren't accessed concurrently?

snapshot/snapshotfs/upload.go
```diff
@@ -423,117 +435,65 @@ func (u *Uploader) prepareWorkItems(ctx context.Context, dirRelativePath string,
...
-	ch := toChannel(workItems)
+		return nil
+	})
 
-	for i := 0; i < workerCount; i++ {
-		wg.Add(1)
+	for _, ent := range results {
+		if ent == nil {
+			continue
+		}
```
julio-lopez:
```suggestion
	results = notNilEntries(results)
	for _, ent := range results {
```

snapshot/snapshotfs/upload.go
```diff
@@ -423,117 +435,65 @@ func (u *Uploader) prepareWorkItems(ctx context.Context, dirRelativePath string,
...
-			return errCancelled
 		}
+	}
 
-		if result.err != nil {
-			if ignoreFileErrs {
-				u.stats.ReadErrors++
-
-				log(ctx).Warningf("unable to hash file %q: %s, ignoring", it.entryRelativePath, result.err)
+	return notNilEntries(results), resultErr
```
julio-lopez:
```suggestion
	return results, resultErr
```

snapshot/snapshotfs/upload.go
```diff
@@ -255,29 +255,52 @@ func (u *Uploader) uploadDir(ctx context.Context, rootDir fs.Directory, policyTr
...
+		return nil
+	})
 
-	return nil
-}
+	// launch N workers in parallel
+	for i := 0; i < parallel; i++ {
+		eg.Go(func() error {
+			for ewi := range ch {
+				if u.IsCancelled() {
```
julio-lopez: Given that this function may run for a long time, that is, process many entries, it seems that it should also check `ctx.Err()`

Separate PR: at some point it'd be good to use `context.Context` for upload cancellation as well.

snapshot/snapshotfs/upload.go
```diff
@@ -255,29 +255,52 @@ func (u *Uploader) uploadDir(ctx context.Context, rootDir fs.Directory, policyTr
...
+					return errCancelled
+				}
+
+				entryRelativePath := relativePath + "/" + ewi.entry.Name()
```
julio-lopez: Does it make sense to use `path.Join(relativePath, ewi.entry.Name())` here?

snapshot/snapshotfs/upload.go
```diff
@@ -255,29 +255,52 @@ func (u *Uploader) uploadDir(ctx context.Context, rootDir fs.Directory, policyTr
...
+				return nil
+			}
 		}
-	}
+		return nil
+	})
 
-	return nil
-}
+	// launch N workers in parallel
```
julio-lopez: Should `parallel` be capped to the number of entries as well for directories with just a couple of entries?

snapshot/snapshotfs/upload.go
```diff
@@ -255,29 +255,52 @@ func (u *Uploader) uploadDir(ctx context.Context, rootDir fs.Directory, policyTr
 	return de, err
 }
 
-func (u *Uploader) foreachEntryUnlessCancelled(relativePath string, entries fs.Entries, cb func(entry fs.Entry, entryRelativePath string) error) error {
-	for _, entry := range entries {
-		if u.IsCancelled() {
-			return errCancelled
-		}
+func (u *Uploader) foreachEntryUnlessCancelled(ctx context.Context, parallel int, relativePath string, entries fs.Entries, cb func(ctx context.Context, index int, entry fs.Entry, entryRelativePath string) error) error {
+	type entryWithIndex struct {
```
julio-lopez: Given that the entries will be sorted to create the corresponding snapshot directory, would it make sense to send the results back through a channel instead of passing an index for the entry? This is assuming the result type is always the same.

snapshot/snapshotfs/upload.go
```diff
@@ -255,29 +255,52 @@ func (u *Uploader) uploadDir(ctx context.Context, rootDir fs.Directory, policyTr
...
+			return nil
+		})
+	}
+
+	return eg.Wait()
 }
 
 func (u *Uploader) processSubdirectories(ctx context.Context, relativePath string, entries fs.Entries, policyTree *policy.Tree, previousEntries []fs.Entries, dirManifest *snapshot.DirManifest, summ *fs.DirectorySummary) error {
-	return u.foreachEntryUnlessCancelled(relativePath, entries, func(entry fs.Entry, entryRelativePath string) error {
+	return u.foreachEntryUnlessCancelled(ctx, 1, relativePath, entries, func(ctx context.Context, index int, entry fs.Entry, entryRelativePath string) error {
```
julio-lopez: This is brittle. The update in line [349](https://github.com/kopia/kopia/pull/331/commits/a7ed40fb05e7b729259a0a8b8f62da584d946143#diff-d711584d735b389e3d0f01af1055bc6bR349) works because `parallel=1`, but would blow otherwise.

snapshot/snapshotfs/upload.go
jkowalski: cancellation can't be driven off "ctx", because "ctx" is used to actually write to storage, and if we cancelled the parent context it would prevent any writes

snapshot/snapshotfs/upload.go
jkowalski: fixed

snapshot/snapshotfs/upload.go
jkowalski: rewrote using channels

snapshot/snapshotfs/upload.go
jkowalski: rewrote using channels, which is much cleaner

snapshot/snapshotfs/upload.go
jkowalski: done

snapshot/snapshotfs/upload.go
jkowalski: implemented this suggestion, much cleaner!

snapshot/snapshotfs/upload.go
jkowalski: done

snapshot/snapshotfs/upload.go
jkowalski: done

snapshot/snapshotfs/upload.go
jkowalski: added constant to make this more obvious, we can work on making this actually parallelizable later

snapshot/snapshotfs/upload.go
julio-lopez: Understood. 
Thanks for the explanation.

snapshot/snapshotfs/upload.go
```diff
@@ -590,18 +601,20 @@ func uploadDirInternal(
...
+	}
 
 	defer func() {
-		summ.IncompleteReason = u.cancelReason()
+		dirManifest.Summary.IncompleteReason = u.cancelReason()
 	}()
 
-	log(ctx).Debugf("reading directory %v", dirRelativePath)
-
+	t0 := u.repo.Time()
```
julio-lopez: if `t0` is used for reporting elapsed time, then it is OK and maybe preferable to use `time.Now()`

PR #340, opened 2020-03-12
Refactor content stats

julio-lopez:
Ensures that the stats are read and reset using atomic primitives as well


jkowalski:
LGTM!

PR #345, opened 2020-03-13
performance: introduced buf.Pool which helps reuse memory buffers

### LINKED ISSUE #93: High memory usage seen with a large number of small files

### LINKED ISSUE #125: Investigate memory usage during snapshot of 100GB+ tree
Follow up to #93 

The setup is:
* 100K files (single directory)
* Each file is 1MB.

Then create a snapshot of this directory.

* Multi-GB RSS (1-8GB) was observed in a few initial runs.
* Running with memory profiling enabled results in RSS ~ 0.5GB.

jkowalski:
content manager:

Previously we would store special field Payload for contents
that were added but never flushed to the store and it was not
encrypted. This required special handling different for pending
vs flushed contents.

This change maintains pending pack buffer ready to be flushed
and appends encrypted contents to it, which avoids data copying.
The buffers are pooled to avoid allocations.

This change also introduces 'buf.Pool', which allows efficient management of ephemeral in-memory buffer using multiple buffers.

## Impact of this PR:

```
10 GB: 10 dirs x 10 files x 100 mb each:

before: time=50.152      cpu=372.200      memory=824.699     vmem=5841.207
after:  time=45.133      cpu=200.400      memory=725.645     vmem=5659.527

10 GB: 10 dirs x 10000 files x 100 kb each:

before: time=43.130      cpu=545.600      memory=656.676     vmem=5724.605
after:  time=43.108      cpu=603.400      memory=407.422     vmem=5392.953

100 MB: 100 directories x 10000 files x 100 bytes each:

before: time=108.431     cpu=149.100      memory=834.781     vmem=5825.094
after:  time=54.108      cpu=100.000      memory=696.168    vmem= 5747.988
```

Fixes #125
Fixes #93

### DIFF
--- a/internal/buf/pool.go
+++ b/internal/buf/pool.go
@@ -2,19 +2,20 @@
 package buf
 
 import (
+	"context"
 	"sync"
 	"sync/atomic"
 	"time"
-)
 
-const (
-	maxAllocAttempts       = 3
-	allocationRetryTimeout = 500 * time.Millisecond
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
 )
 
 type segment struct {
-	nextUnallocated   int32  // high water mark
-	allocatedBufCount int32  // how many outstanding users of the segment there are
+	mu sync.RWMutex
+
+	nextUnallocated   int    // high water mark
+	allocatedBufCount int    // how many outstanding users of the segment there are
 	data              []byte // the underlying buffer from which we're allocating
 	pool              *Pool
 }
@@ -22,12 +23,11 @@ type segment struct {
 // Buf represents allocated slice of memory pool. At the end of using the buffer, Release() must be called to
 // reclaim memory.
 type Buf struct {
-	nextUnallocated         int32
-	previousNextUnallocated int32
-
 	Data []byte
 
-	segment *segment // segment from which the data was allocated
+	nextUnallocated         int
+	previousNextUnallocated int
+	segment                 *segment // segment from which the data was allocated
 }
 
 // IsPooled determines whether data slice is part of a pool.
@@ -39,51 +39,53 @@ func (b *Buf) Release() {
 		return
 	}
 
-	notify := false
+	atomic.AddInt64(&b.segment.pool.totalReleasedBytes, int64(len(b.Data)))
+	atomic.AddInt64(&b.segment.pool.totalReleasedBuffers, 1)
+
+	b.segment.mu.Lock()
+	defer b.segment.mu.Unlock()
 
 	// best effort compare-and-swap, which will pop the buffer off the stack in its appropriate segment
-	if atomic.CompareAndSwapInt32(&b.segment.nextUnallocated, b.nextUnallocated, b.previousNextUnallocated) {
-		// popped from the stack in the segment
-		notify = true
+	if b.segment.nextUnallocated == b.nextUnallocated {
+		b.segment.nextUnallocated = b.previousNextUnallocated
 	}
 
-	if atomic.AddInt32(&b.segment.allocatedBufCount, -1) == 0 {
+	b.segment.allocatedBufCount--
+	if b.segment.allocatedBufCount == 0 {
 		// last allocated Buf, we can reset 'next' to zero
-		atomic.StoreInt32(&b.segment.nextUnallocated, 0)
-
-		notify = true
-	}
-
-	if notify {
-		// this segment just became free, notify other goroutines doing Allocate()
-		select {
-		case b.segment.pool.segmentReleased <- struct{}{}: // notified
-		default: // nobody is waiting
-		}
+		b.segment.nextUnallocated = 0
 	}
 
 	b.Data = nil
-
 	b.segment = nil
 }
 
-func (s *segment) allocate(count int) (Buf, bool) {
-	n := int32(count)
+func (s *segment) allocate(n int) (Buf, bool) {
+	// quick check using shared lock
+	s.mu.RLock()
+	haveRoom := s.nextUnallocated+n <= len(s.data)
+	s.mu.RUnlock()
 
-	for {
-		// see if we have capacity in this segment
-		v := atomic.LoadInt32(&s.nextUnallocated)
-		if v+n > int32(len(s.data)) {
-			// out of space in this segment
-			return Buf{}, false
-		}
-
-		if atomic.CompareAndSwapInt32(&s.nextUnallocated, v, v+n) {
-			atomic.AddInt32(&s.allocatedBufCount, 1)
-
-			return Buf{v + n, v, s.data[v : v+n : v+n], s}, true
-		}
+	if !haveRoom {
+		return Buf{}, false
 	}
+
+	// we likely have space, allocate under exclusive lock
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// see if we have capacity in this segment
+	v := s.nextUnallocated
+
+	if v+n > len(s.data) {
+		// out of space in this segment
+		return Buf{}, false
+	}
+
+	s.allocatedBufCount++
+	s.nextUnallocated += n
+
+	return Buf{s.data[v : v+n : v+n], v + n, v, s}, true
 }
 
 // Pool manages allocations of short-term data buffers from a pool.
@@ -91,7 +93,7 @@ func (s *segment) allocate(count int) (Buf, bool) {
 // Note that buffers managed by the pool are meant to be extremely short lived and are suitable
 // for in-memory operations, such as encryption, compression, etc, but not for I/O buffers of any kind.
 // It is EXTREMELY important to always release memory allocated from the Pool. Failure to do so will
-// result in
+// result in memory leaks.
 //
 // The pool uses N segments, with each segment tracking its high water mark usage.
 //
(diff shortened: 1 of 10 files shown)


internal/buf/pool.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+
+		// add one more segment up to specified limit
+		if p.InitializeSegments(1) {
+			continue
+		}
+
+		// wait until some segment becomes free or some time passes, whichever comes first
+		select {
+		case <-p.segmentReleased:
```
julio-lopez: If this is a frequent code path and `<-p.segmentReleased` is a more frequent case than the one below, this ends up allocating many timers that are not released until the timer fires. Now, this may not make a difference given the `allocationRetryTimeout` and GC frequency.

internal/buf/pool.go
```diff
@@ -0,0 +1,205 @@
...
+	notify := false
+
+	// best effort compare-and-swap, which will pop the buffer off the stack in its appropriate segment
+	if atomic.CompareAndSwapInt32(&b.segment.nextUnallocated, b.nextUnallocated, b.previousNextUnallocated) {
+		// popped from the stack in the segment
+		notify = true
+	}
+
+	if atomic.AddInt32(&b.segment.allocatedBufCount, -1) == 0 {
+		// last allocated Buf, we can reset 'next' to zero
```
julio-lopez: I think there's a race here.
- Suppose the case where this is the last `Buf` in this segment.
- The last `Release` executes up to the line above, but not line 52.

In the mean time an allocate on the same same segment executes up to line 81, but not 82.
At this point `s.nextUnallocated` is after the `Buf.data` for this allocation

When 52 executes, it would reset `s.nextUnallocated` to 0 and it would corrupt the segment and subsequent allocations on the same segment.

julio-lopez:
@jkowalski I think that we should be able to merge all the patches that are not related to `buf.Pool` now, and then merge `buf.Pool` related patches separately. WDYT?

internal/buf/pool.go
```diff
@@ -0,0 +1,205 @@
...
+		n = remaining
+	}
+
+	if n == 0 {
+		return false
+	}
+
+	var newSegments []*segment
+
+	newSegments = append(newSegments, p.segments...)
```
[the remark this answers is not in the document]
jkowalski: i removed maxSegments in the latest version. Unless there's a `Buf` leak, this should not be needed.

internal/buf/pool.go
```diff
@@ -0,0 +1,205 @@
...
+			pool: p,
+		})
+	}
+
+	p.segments = newSegments
+
+	return true
+}
+
+// Allocate allocates a slice of the buffer
```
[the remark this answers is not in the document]
jkowalski: done

internal/buf/pool.go
```diff
@@ -0,0 +1,205 @@
...
+func (p *Pool) Allocate(n int) Buf {
+	// requested more than the pool can cache, allocate throw-away buffer.
+	if p == nil || n > p.segmentSize {
+		return Buf{0, 0, make([]byte, n), nil}
```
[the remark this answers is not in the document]
jkowalski: added instrumentation

internal/buf/pool.go
jkowalski: removed this whole channel business, we can simply add more segments

internal/buf/pool.go
jkowalski: fixed by adding mutex and removing atomic

internal/buf/pool.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}
+}
+
+// Pool manages allocations of short-term data buffers from a pool.
+//
+// Note that buffers managed by the pool are meant to be extremely short lived and are suitable
+// for in-memory operations, such as encryption, compression, etc, but not for I/O buffers of any kind.
+// It is EXTREMELY important to always release memory allocated from the Pool. Failure to do so will
+// result in
```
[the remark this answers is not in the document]
jkowalski: done

internal/buf/pool.go
```diff
@@ -39,59 +39,61 @@ func (b *Buf) Release() {
...
-		if atomic.CompareAndSwapInt32(&s.nextUnallocated, v, v+n) {
-			atomic.AddInt32(&s.allocatedBufCount, 1)
+	// we likely have space, allocate under exclusive lock
+	s.mu.Lock()
+	defer s.mu.Unlock()
 
-			return Buf{v + n, v, s.data[v : v+n : v+n], s}, true
-		}
+	// see if we have capacity in this segment
+	v := s.nextUnallocated
```
julio-lopez: nit: maybe rename `v` to `nu` or `next` or even `prev`? Not sure what would make it clearer.

internal/buf/pool.go
```diff
@@ -39,59 +39,61 @@ func (b *Buf) Release() {
...
+
+	if v+n > len(s.data) {
+		// out of space in this segment
+		return Buf{}, false
 	}
+
+	s.allocatedBufCount++
+	s.nextUnallocated += n
+
+	return Buf{s.data[v : v+n : v+n], v + n, v, s}, true
```
julio-lopez: use named initializers for `Buf` ?

internal/buf/pool.go
```diff
@@ -107,26 +109,71 @@ func (s *segment) allocate(count int) (Buf, bool) {
...
+
 	return p
 }
 
+// Close closes the pool
+func (p *Pool) Close() {
+	close(p.closed)
+}
+
+func (p *Pool) reportMetrics() {
```
julio-lopez: nit: pass the `Context` from the caller (go routine above)?

internal/buf/pool.go
```diff
@@ -188,18 +227,7 @@ func (p *Pool) Allocate(n int) Buf {
...
-			continue
-		}
-
-		// wait until some segment becomes free or some time passes, whichever comes first
-		select {
-		case <-p.segmentReleased:
-		case <-time.After(allocationRetryTimeout):
-		}
+		// add one more segment
+		p.AddSegments(1)
```
julio-lopez: As discussed earlier, a concern here is that there should be a way of freeing up _excess_ segments as they become empty. Perhaps keep a minimum number of segments in the pool, but releasing extra unused segments. Or perhaps having a maximum number of unused segments that can be kept around. Thoughts?

Related to this, it'd be good to track the number of unused / empty segments. That is, on `Release` increment a counter of unused segments when the allocated buffer count reaches 0, or when a segment is initially created. Decrease the counter on allocate when allocating from a currently empty segment.

jkowalski: good ideas, will do in separate PR

internal/buf/pool.go
jkowalski: done

internal/buf/pool.go
jkowalski: done

internal/buf/pool.go
jkowalski: done

PR #358, opened 2020-03-16
cli: allow override of snapshot start time and end time

### LINKED ISSUE #354: Set the timestamp of a snapshot
I was wondering if there's a flag for `kopia snapshot create` similar to `--hostname` that will override the timestamp of a snapshot?

We're looking at migrating from an older backup system and it would be nice to insert past backups into the kopia repository along with new ones.

Novex:
This PR introduces two new arguments `--start-time` and `--end-time` to the `kopia snapshot create' command` to override the start time and end the end time that the snapshot is created.

If the start time is specified but no end time is, then the end time will be set to the start time so it doesn't look like a backup ran for (potentially) years and vice-versa for end time specified without a start time.

Fixes #354

### DIFF
--- a/cli/command_snapshot_create.go
+++ b/cli/command_snapshot_create.go
@@ -16,6 +16,7 @@ import (
 
 const (
 	maxSnapshotDescriptionLength = 1024
+	timeFormat                   = "2006-01-02 15:04:05 MST"
 )
 
 var (
@@ -29,6 +30,8 @@ var (
 	snapshotCreateParallelUploads         = snapshotCreateCommand.Flag("parallel", "Upload N files in parallel").PlaceHolder("N").Default("0").Int()
 	snapshotCreateHostname                = snapshotCreateCommand.Flag("hostname", "Override local hostname.").String()
 	snapshotCreateUsername                = snapshotCreateCommand.Flag("username", "Override local username.").String()
+	snapshotCreateStartTime               = snapshotCreateCommand.Flag("start-time", "Override snapshot start timestamp.").String()
+	snapshotCreateEndTime                 = snapshotCreateCommand.Flag("end-time", "Override snapshot end timestamp.").String()
 )
 
 func runBackupCommand(ctx context.Context, rep *repo.Repository) error {
@@ -54,6 +57,20 @@ func runBackupCommand(ctx context.Context, rep *repo.Repository) error {
 	onCtrlC(u.Cancel)
 
 	u.Progress = progress
+
+	startTime, err := parseTimestamp(*snapshotCreateStartTime)
+	if err != nil {
+		return errors.Wrap(err, "could not parse start-time")
+	}
+
+	endTime, err := parseTimestamp(*snapshotCreateEndTime)
+	if err != nil {
+		return errors.Wrap(err, "could not parse end-time")
+	}
+
+	if startTimeAfterEndTime(startTime, endTime) {
+		return errors.New("start time override cannot be after the end time override")
+	}
 
 	if len(*snapshotCreateDescription) > maxSnapshotDescriptionLength {
 		return errors.New("description too long")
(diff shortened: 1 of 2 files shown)


cli/command_snapshot_create.go
```diff
@@ -129,6 +133,32 @@ func snapshotSingleSource(ctx context.Context, rep *repo.Repository, u *snapshot
 
 	manifest.Description = *snapshotCreateDescription
 
+	if *snapshotCreateStartTime != "" {
```
jkowalski: please move parsing to be done sooner. otherwise folks may spend hours uploading only to find out they had a typo in the flag and the snapshot will fail.

cli/command_snapshot_create.go
```diff
@@ -129,6 +133,32 @@ func snapshotSingleSource(ctx context.Context, rep *repo.Repository, u *snapshot
...
+		startTime, err := time.Parse(timeFormat, *snapshotCreateStartTime)
+
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("Could not parse start-time, should be like the format '%s'", timeFormat))
```
jkowalski: errors.Wrapf()

cli/command_snapshot_create.go
```diff
@@ -129,6 +133,32 @@ func snapshotSingleSource(ctx context.Context, rep *repo.Repository, u *snapshot
...
+		if *snapshotCreateEndTime == "" {
+			manifest.EndTime = startTime
+		}
+	}
+
+	if *snapshotCreateEndTime != "" {
+		endTime, err := time.Parse(timeFormat, *snapshotCreateEndTime)
+
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("Could not parse end-time, should be like the format '%s'", timeFormat))
```
jkowalski: errors.Wrapf()

cli/command_snapshot_create.go
```diff
@@ -129,6 +133,32 @@ func snapshotSingleSource(ctx context.Context, rep *repo.Repository, u *snapshot
...
+		}
+
+		if *snapshotCreateStartTime == "" {
+			manifest.StartTime = endTime
```
jkowalski: i'm thinking: if somebody specifies one flag but not both, we can take actual snapshot duration and compute the other. would that be useful?

Also can you add one or two test cases for this logic? we already have CLI integration tests for a whole bunch of cases.

cli/command_snapshot_create.go
Novex: Thanks for picking up on this - I didn't even think about that!

cli/command_snapshot_create.go
Novex: It wouldn't be useful for my use case, but I'm happy to work out how to do it :)

I'll add tests as well, the existing ones serve as a good example

jkowalski: try `tests/end_to_end_test/snapshot_create_test.go`

jkowalski:
BTW, you have some linter errors, before re-submitting the PR please do `make lint; make test; make integration test`

jkowalski:
also, if you have questions, best to ask them on `slack.kopia.io`

snapshot/snapshotfs/upload.go
```diff
@@ -779,8 +784,36 @@ func (u *Uploader) Upload(
...
 
+	s.StartTime = startTime
+	s.EndTime = endTime
+
+	u.overrideTimestamps(s, startTime, endTime)
+
 	return s, nil
 }
+
+func (u *Uploader) overrideTimestamps(
```
jkowalski: can this logic be outside of the uploader? original idea was for uploader to potentially be shared between multiple snapshots and and having shared overrides does not work.

you don't really need "u" in this function - just the two overrides that can be stored in the "cli" package.

jkowalski: btw, other than this the PR looks very good.

Novex: Yep that makes sense - I was trying to avoid creating new objects or passing values down the chain but I think I worked out a good way to do it

jkowalski:
LGTM! Let's wait for Travis and we can merge it.

Novex:
Awesome! Thank you for the great review process!

It looks like Travis doing some maintenance but it should run eventually - just ping me if anything else needs to be done

jkowalski:
Merged after linux and windows succeeded, macOS is having some problems on Travis right now.

PR #361, opened 2020-03-17
content: wait for cache sweep goroutine to close

jkowalski:
- also fixed couple cases where we were not closing content manager


repo/content/content_cache.go
```diff
@@ -261,6 +265,7 @@ func newContentCacheWithCacheStorage(ctx context.Context, st, cacheStorage blob.
 		return nil, err
 	}
 
+	c.asyncWG.Add(1)
```
julio-lopez:
```suggestion
	c.asyncWG.Add(1)

```

PR #365, opened 2020-03-18
perf: added performance harness tool

jkowalski:
that runs kopia, collects basic stats and uploads to a provided server for long-term storage


jkowalski:
i'm going to close it for now, had couple more ideas

PR #369, opened 2020-03-19
object: added AsyncWrites to ObjectWriter, which improves performance…

### LINKED ISSUE #351: Implement asynchronous WriteContent inside ObjectWriter
Right now ObjectWriter waits for WriteContent to complete. This could be improved with double or triple buffering, where ObjectWriter has 2 or 3 buffers and switches to the next one after asynchronously sending one to content manager. 

It could improve performance of pathological case of a single very large file, which today can't be paralellized.

jkowalski:
… of uploading of a single file

Fixes #351

### DIFF
--- a/repo/object/object_manager_test.go
+++ b/repo/object/object_manager_test.go
@@ -301,79 +301,111 @@ func TestReaderStoredBlockNotFound(t *testing.T) {
 }
 
 func TestEndToEndReadAndSeek(t *testing.T) {
-	ctx := testlogging.Context(t)
-	_, om := setupTest(t)
+	for _, asyncWrites := range []int{0, 4, 8} {
+		asyncWrites := asyncWrites
 
-	for _, size := range []int{1, 199, 200, 201, 9999, 512434, 5012434} {
-		// Create some random data sample of the specified size.
-		randomData := make([]byte, size)
-		cryptorand.Read(randomData) //nolint:errcheck
+		t.Run(fmt.Sprintf("async-%v", asyncWrites), func(t *testing.T) {
+			t.Parallel()
 
-		writer := om.NewWriter(ctx, WriterOptions{})
-		if _, err := writer.Write(randomData); err != nil {
-			t.Errorf("write error: %v", err)
-		}
+			ctx := testlogging.Context(t)
+			_, om := setupTest(t)
 
-		objectID, err := writer.Result()
-		t.Logf("oid: %v", objectID)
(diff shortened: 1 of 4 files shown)


repo/object/object_manager_test.go
```diff
@@ -301,79 +301,111 @@ func TestReaderStoredBlockNotFound(t *testing.T) {
...
 
-		objectID, err := writer.Result()
-		t.Logf("oid: %v", objectID)
+			for _, size := range []int{1, 199, 200, 201, 9999, 512434, 5012434} {
+				// Create some random data sample of the specified size.
+				randomData := make([]byte, size)
+				cryptorand.Read(randomData) //nolint:errcheck
 
-		writer.Close()
+				writer := om.NewWriter(ctx, WriterOptions{})
```
julio-lopez: Should `WriterOptions{}` include `asyncWrites` here?

repo/object/object_writer.go
```diff
@@ -155,12 +222,21 @@ func maybeCompressedContentBytes(comp compression.Compressor, output *bytes.Buff
 }
 
 func (w *objectWriter) Result() (ID, error) {
+	// no need to hold a lock on w.indirectIndexGrowMutex, since growing index only happens synchronously
+	// and never in parallel with calling Result()
```
julio-lopez: This is making an assumption about how `Result` is called w.r.t `Write(..) => `flushBuffer(...)`. It is essentially saying that index growth, and thus modifications to `w.indirectIndex`, only happen in this go routine. This works, but seems fragile. A change in the implementation outside this function, could easily result in a race here due to this assumption. Not sure what would be a good way to prevent that, and whether that type of change is already covered by unit tests and would be caught by those.

jkowalski: yeah, ObjectWriter is fundamentally not goroutine-safe to the caller. tests did catch it, though, which is the reason I added the growth mutex.

snapshot/snapshotfs/upload.go
```diff
@@ -494,10 +504,31 @@ func (u *Uploader) maybeIgnoreCachedEntry(ctx context.Context, ent fs.Entry) fs.
...
-		workerCount = runtime.NumCPU()
+	par := u.effectiveParallelUploads()
+
+	var workerCount, asyncWritesPerFile int
+
+	if len(entries) < par {
+		workerCount = len(entries)
+
+		if len(entries) > 0 {
+			asyncWritesPerFile = par / len(entries)
```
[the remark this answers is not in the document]
jkowalski: good point, we can do something more fancy here if needed.

PR #392, opened 2020-03-24
Minor cleanup for snapshot time override

julio-lopez:


jkowalski:
looks like e2e test is failing

cli/command_snapshot_create.go
```diff
@@ -178,6 +168,7 @@ func snapshotSingleSource(ctx context.Context, rep *repo.Repository, u *snapshot
 
 		if endTimeOverride.IsZero() {
 			// Calculate the correct end time based on current duration if they're not specified
+			duration := manifest.EndTime.Sub(manifest.StartTime)
```
Novex: The e2e tests are failing here - you can't calculate the duration after you've modified the StartTime with the override (same as below with the EndTime as well)

julio-lopez: duh! Let me revert this. I completely missed it.

PR #393, opened 2020-03-24
policy: deduplicate multiple policies for the same source

### LINKED ISSUE #391: Deduplicate multiple policies for the same source
#385 uncovered that we sometimes land in a situation where two live policies are set for the same source. This race is not something that can be completely avoided, we should automatically pick one of the policies as the winner and auto-delete the other one in manifest manager. 

We can either use time or a guid to pick the winner, both are generated client-side.

jkowalski:
in policy manager, fixes #391

### DIFF
--- a/snapshot/policy/policy_manager.go
+++ b/snapshot/policy/policy_manager.go
@@ -98,29 +98,32 @@ func GetDefinedPolicy(ctx context.Context, rep *repo.Repository, si snapshot.Sou
 		return nil, ErrPolicyNotFound
 	}
 
-	if len(md) == 1 {
-		p := &Policy{}
+	// arbitrality pick first pick ID to return in case there's more than one
+	// this is possible when two repository clients independently create manifests at approximately the same time
+	// so it should not really matter which one we pick.
+	// see https://github.com/kopia/kopia/issues/391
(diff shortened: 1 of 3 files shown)


snapshot/policy/policy_manager.go
```diff
@@ -98,29 +98,32 @@ func GetDefinedPolicy(ctx context.Context, rep *repo.Repository, si snapshot.Sou
 		return nil, ErrPolicyNotFound
 	}
 
-	if len(md) == 1 {
-		p := &Policy{}
+	// arbitrality pick first pick ID to return in case there's more than one
+	// this is possible when two repository clients independently create manifests at approximately the same time
+	// so it should not really matter which one we pick.
+	// see https://github.com/kopia/kopia/issues/391
```
julio-lopez: If they are not expected to be sorted by a particular criteria this may end up picking different policies across invocations. Which one gets chosen should be deterministic to have consistent behavior across executions / clients.

jkowalski: there's really no guarantee that repo that's open at this time but not flushed at this time hasn't added yet another conflicting manifest, so this won't guarantee much, I think.

PR #395, opened 2020-03-25
repo: deprecated NONE algorithm, will not be available for new repositories

### LINKED ISSUE #388: Remove support for unencrypted repos
Support for the "no encryption" cipher should be disabled in released kopia binaries. It should only be used for testing purposes.

This avoids unintentional and incorrect misuses.

jkowalski:
Fixes #388

### DIFF
--- a/internal/repotesting/repotesting.go
+++ b/internal/repotesting/repotesting.go
@@ -48,7 +48,7 @@ func (e *Environment) Setup(t *testing.T, opts ...func(*repo.NewRepositoryOption
 		BlockFormat: content.FormattingOptions{
 			HMACSecret: []byte{},
 			Hash:       "HMAC-SHA256",
-			Encryption: encryption.NoneAlgorithm,
+			Encryption: encryption.DefaultAlgorithm,
 		},
 		ObjectFormat: object.Format{
 			Splitter: "FIXED-1M",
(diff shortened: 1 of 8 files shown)


internal/repotesting/repotesting.go
```diff
@@ -48,7 +48,7 @@ func (e *Environment) Setup(t *testing.T, opts ...func(*repo.NewRepositoryOption
 		BlockFormat: content.FormattingOptions{
 			HMACSecret: []byte{},
 			Hash:       "HMAC-SHA256",
-			Encryption: encryption.NoneAlgorithm,
```
julio-lopez: Not sure how large of an impact this will have in tests running time.
The initial thinking was to allow this for testing if needed.

repo/manifest/manifest_manager_test.go
```diff
@@ -293,7 +294,7 @@ func newManagerForTesting(ctx context.Context, t *testing.T, data blobtesting.Da
 
 	bm, err := content.NewManager(ctx, st, &content.FormattingOptions{
 		Hash:        "HMAC-SHA256-128",
```
julio-lopez: might as well change this for consistency
```suggestion
		Hash:        hashing.DefaultAlgorithm,
```

internal/repotesting/repotesting.go
jkowalski: if this is too slow we can always go back to unencrypted. I don't see any measurable difference.

Issue #312, opened 2020-03-09
Support Backblaze B2 storage


jkowalski:
Backblaze B2 is popular for its low cost storage amongst `r/datahoarder/`

https://www.reddit.com/r/DataHoarder/comments/ffj5rh/kopia_hits_050_encrypted_deduplicated_and/fjz5twf?utm_source=share&utm_medium=web2x

aksdb:
I have the PR nearly ready. Just a few cleanups left to do. (But I work on this very slowly in off-time ... so ...)

jkowalski:
btw, #373 has changed the PutBlob API to support write gathering, you may need to rebase.

jkowalski:
I've enabled testing on Travis, closing this issue.

PR #396, opened 2020-03-28
B2

aksdb:
Implements B2 as requested in #312 .

It was bit of back and forth between libraries. It seems the kinda-standard B2 library for Go (blazer) is a tiny bit overengineered and makes it very hard to deal with error cases (since you cannot access any error details without reflection). Also the code has so many layers of indirection that it was quite hard to follow what actually happens for certain requests. Therefore I went with the much simpler (in a good way) [go-backblaze](https://github.com/kothar/go-backblaze). Only downside so far: it doesn't use a `Context` to make requests. I will probably add that to that library soonish (if the author wants that).

One thing to note about B2, which should be also important when you setup a bucket for the integration tests: B2 doesn't easily support deletion of files. By default a bucket works with full versioning. To not undermine their concept, I did not implement any workarounds for real deletion (like iterating the file history and deleting version by version), I only "hide" files. Bucket rules can be set to actually delete hidden files after a specified period of time. You might want to set that for the bucket you intend to use for testing.

I also did not implement bucket creation/deletion as part of the integration test, since this would require the use of admin credentials. I think it's better to use application specific ones that only use a specific bucket. (Together with the retention rules offered by B2 that should be fine.)

Further more, I didn't implement the ability to set a custom API endpoint. I know of a single B2 mock server that could be interesting for tests, but for production usage there is no (known to me) free implementation of the B2 API (like there is with *minio* for S3). So it just doesn't seem to make sense to offer customization there.

Due to that back and forth I squashed my history into backend and UI commits, even though that garbles (well, removes) the actual timeline. But whatever :-)

### DIFF
--- a/cli/storage_b2.go
+++ b/cli/storage_b2.go
@@ -1,0 +1,30 @@
+package cli
+
+import (
+	"context"
+
+	"gopkg.in/alecthomas/kingpin.v2"
+
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/blob/b2"
+)
+
+func init() {
+	var b2options b2.Options
+
+	RegisterStorageConnectFlags(
+		"b2",
+		"a b2 bucket",
+		func(cmd *kingpin.CmdClause) {
+			cmd.Flag("bucket", "Name of the B2 bucket").Required().StringVar(&b2options.BucketName)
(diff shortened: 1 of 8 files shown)


jkowalski:
That looks perfect. Thank you so much for your contribution. 

Do we want to enable this test on Travis CI? We would need to enter some bucket credentials via secure environment variables. Please ping me on slack.kopia.io if you want to do that.

PR #420, opened 2020-04-17
GC content repair

julio-lopez:
Undelete deleted and re-referenced content.
This implementation is based on `content.Manager.ContentRewrite()`


julio-lopez:
@jkowalski PTAL

PR #421, opened 2020-04-18
content: added blob-level metadata cache

jkowalski:
Unlike regular cache, which caches segments of blobs on a per-content
basis, metadata cache will fetch and store the entire metadata blob (q)
when any of the contents in it is accessed.

Given that there are relatively few metadata blobs compared to data (p)
blobs, this will reduce the traffic to the underlying store and improve
performance of Snapshot GC which only relies on metadata contents.

### DIFF
--- a/repo/content/content_cache_metadata.go
+++ b/repo/content/content_cache_metadata.go
@@ -1,0 +1,128 @@
+package content
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"go.opencensus.io/stats"
+	"golang.org/x/sync/errgroup"
+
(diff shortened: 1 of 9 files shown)


repo/content/content_cache_metadata.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	if err == blob.ErrBlobNotFound {
+		// not found in underlying storage
+		return nil, err
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
julio-lopez, line 89: Why differentiate these two case?

julio-lopez:
Review comment can be addressed separately

PR #427, opened 2020-04-23
Support for remote repository, where all contents and manifests are fetched over HTTP(S) instead of locally manipulating blob storage

jkowalski:


jkowalski:
(this is broken into smaller commits to facilitate reviewing and provide explanation, not guaranteed that they compile individually - they should, though)

repo/manifest/manifest_entry.go
```diff
@@ -5,8 +5,8 @@ import "time"
 // EntryMetadata contains metadata about manifest item. Each manifest item has one or more labels
 // Including required "type" label.
 type EntryMetadata struct {
-	ID      ID
-	Length  int
-	Labels  map[string]string
-	ModTime time.Time
+	ID      ID                `json:"id"`
```
julio-lopez, line 8: Does this break existing (already serialized) manifests?

repo/local_config.go
```diff
@@ -12,10 +12,12 @@ import (
 
 // LocalConfig is a configuration of Kopia stored in a configuration file.
 type LocalConfig struct {
-	Storage  blob.ConnectionInfo    `json:"storage"`
-	Caching  content.CachingOptions `json:"caching"`
-	Hostname string                 `json:"hostname"`
-	Username string                 `json:"username"`
+	APIServer *APIServerInfo `json:"apiServer"`
```
julio-lopez, line 15: IIUC: a client can either directly use blob storage or go through the APIServer. Is that correct?

It makes sense to explicitly add a comment either to the struct or the fields about the intended usage.

repo/manifest/manifest_entry.go
jkowalski: no, this struct was never serialized, with this change we'll start sending it over the wire.

repo/local_config.go
jkowalski: done.

PR #429, opened 2020-04-30
[robustness testing] Wire up snapshot-store-compare behavior

redgoat650:
Connect the snapshotter, the storer, and the comparer. Invoke
the snapshotter to take/restore/delete snapshots on the repo,
the comparer to gather metadata before the snapshot and
after the restore, and the storer to save metadata for later
lookup when verifying restores.

Added a design write-up to the umbrella issue #179. Will add
updates there as each component is staged for merge.


redgoat650:
Thanks for taking a look @jkowalski. I'm sorting out a CI failure on the next PR in the chain, but I expect to open that soon. I'd estimate we've merged about 40% of the changes so far

Issue #326, opened 2020-03-10
Add testing for eventual consistency behavior of blob storage providers


jkowalski:
Idea: we can write a blob.Storage wrapper for testing that intentionally delays underlying writes and deletes with some externally configurable probabilities and latencies, we an then write a stress test that uses filesystem provider + this wrapper and have behavior similar to S3.

ntolia:
Minio has an unadvertised s3-to-s3 gateway that we might be able to leverage for this.

julio-lopez:
This can also be implemented as a provider at the storage layer that is used only for testing, similar to the `blobtesting` package.

PR #434, opened 2020-05-04
testing: added blob.Storage wrapper that simulates eventual consistency

jkowalski:
This is done by introducing N unsynchronized caches, which simulate
what frontend of a cloud storage system might do, what causes eventual
consistency behavior.

Testing for #326

### DIFF
--- a/internal/blobtesting/eventually_consistent.go
+++ b/internal/blobtesting/eventually_consistent.go
@@ -1,0 +1,190 @@
+package blobtesting
+
+import (
+	"context"
+	"io/ioutil"
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/repo/blob"
+)
+
+const ecCacheDuration = 5 * time.Second
+
+// ecFrontendCache is an instance of cache, which simulates cloud storage frontend
+// with its own in-memory state. This causes eventual consistency when a client uses
+// different instances of the cache for read and writes.
+type ecFrontendCache struct {
+	mu            sync.Mutex
+	cachedEntries map[blob.ID]*ecCacheEntry
+}
+
+func (c *ecFrontendCache) get(id blob.ID) *ecCacheEntry {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	e := c.cachedEntries[id]
+	if e != nil && !e.isValid() {
+		c.sweepLocked()
+		return nil
+	}
+
+	return e
+}
+
+func (c *ecFrontendCache) put(id blob.ID, data []byte) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.sweepLocked()
+
+	if c.cachedEntries == nil {
+		c.cachedEntries = map[blob.ID]*ecCacheEntry{}
+	}
+
+	if data != nil {
+		// clone data before storage
+		data = append([]byte(nil), data...)
+	}
+
+	c.cachedEntries[id] = &ecCacheEntry{
+		accessTime: time.Now(),
+		data:       data,
+	}
+}
+
+func (c *ecFrontendCache) sweepLocked() {
+	for k, v := range c.cachedEntries {
+		if !v.isValid() {
+			delete(c.cachedEntries, k)
+		}
+	}
+}
+
+type ecCacheEntry struct {
+	data []byte
+
+	accessTime time.Time
+}
+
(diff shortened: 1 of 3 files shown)


internal/blobtesting/eventually_consistent.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+
+	d, err := ioutil.ReadAll(data.Reader())
+	if err != nil {
+		return errors.Wrap(err, "invalid data")
+	}
+
+	// add to frontend cache
+	s.randomFrontendCache().put(id, d)
```
pkj415, line 143: Do we have to put it to the frontendCache? The behaviour would be the same even without this line since GetBlob anyway checks the underlying storage in case an object is not found in the front end cache.

pkj415: Okay, I take back the comment. The place where this would make a difference is in DeleteBlob. Nevermind.

internal/blobtesting/eventually_consistent.go
```diff
@@ -0,0 +1,190 @@
...
+	return nil
+}
+
+func (s *eventuallyConsistentStorage) ListBlobs(ctx context.Context, prefix blob.ID, callback func(blob.Metadata) error) error {
+	return s.realStorage.ListBlobs(ctx, prefix, func(bm blob.Metadata) error {
+		e := s.randomFrontendCache().get(bm.BlobID)
+		if e != nil {
+			// item recently manipulated by the cache, skip from the results with some
+			// probability
+			if rand.Float64() < s.listDropProbability {
```
pkj415, line 164: in case an item is recently manipulated, this is always have 0.25 * listDropProbability effective probability. Maybe we can check all caches to see if the item was recently manipulated so that effective probability is exactly listDropProbabililty?

internal/blobtesting/eventually_consistent.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// add to frontend cache
+	s.randomFrontendCache().put(id, d)
+
+	return nil
+}
+
+func (s *eventuallyConsistentStorage) DeleteBlob(ctx context.Context, id blob.ID) error {
+	s.randomFrontendCache().put(id, nil)
```
pkj415, line 149: consider this scenario -
1. DeleteBlob is called on an object which doesn't exist
2. PutBlob is called to create the object
3. GetBlob is called on the object

As per this code, we might get ErrBlobNotFound in the GetBlob call. Is that the correct behaviour?

internal/blobtesting/eventually_consistent.go
```diff
@@ -0,0 +1,190 @@
...
+	return s.realStorage.Close(ctx)
+}
+
+func (s *eventuallyConsistentStorage) ConnectionInfo() blob.ConnectionInfo {
+	return s.realStorage.ConnectionInfo()
+}
+
+// NewEventuallyConsistentStorage returns an eventually-consistent storage wrapper on top
+// of provided storage.
+func NewEventuallyConsistentStorage(st blob.Storage, listDropProbability float64) blob.Storage {
```
pkj415, line 184: Consider the scenario of two processes -

1. Process 1 does a DeleteBlob
2. Process 2 (assume it is a process which always starts only after Process 1) starts up and will now always sees the blob to be deleted (not eventually consistent).

To model such cases, we might have to think of something.

internal/blobtesting/eventually_consistent.go
jkowalski: I think that's valid behavior (not found), because there was a moment where a blob did not exist after creation.

internal/blobtesting/eventually_consistent.go
jkowalski: right, but this wrapper is only for unit tests where we can model whatever behaviors we need.

internal/blobtesting/eventually_consistent.go
julio-lopez, line 164: This approximates the behavior.

I'd expect that a list for a range will populate a cache, then if that same cache ends up being used, then it would return a stale view of the list for that range. But, I'm guessed the actual implementation is very nuanced probably based on consistent hashing partitions or something like that.

internal/blobtesting/eventually_consistent.go
julio-lopez: A related question / suggestion is that we may want to be able to control and enforce particular sequences.

PR #436, opened 2020-05-12
cli: improved 'snapshot delete' usage

### LINKED ISSUE #435: how to delete a snapshot ?
I use the following command to delete a snapshot
```
root@zmbox:/opt/config/kopia# kopia snapshot delete kddb8afc3321307a2bfafbc83704b019c
kopia: error: path is required, try --help
```
which the cmd
```
root@zmbox:/opt/config/kopia# kopia snapshot list /etc/dnsmasq.d
root@zmbox:/etc/dnsmasq.d
  2020-05-06 20:04:34 CST k52a5cf59b6733efc9c8c677dd78805e0 5.8 MB drwxr-xr-x files:10 dirs:2 (latest-4)
  + 1 identical snapshots until 2020-05-06 20:33:56 CST
  2020-05-06 20:37:28 CST kddb8afc3321307a2bfafbc83704b019c 5.8 MB drwxr-xr-x files:9 dirs:1 (latest-2)
  2020-05-06 20:38:49 CST kd73d7a27e46e5764101513ee844a5129 5.8 MB drwxr-xr-x files:10 dirs:2 (latest-1,annual-1,monthly-1,weekly-1,daily-1,hourly-1)
```
shows the id is correct

the version I use is below
```
root@zmbox:/opt/config/kopia# kopia --version
0.5.2 build: 4f65495d811022e65d27e376e2d94f5b905d6c8c
```

jkowalski:
New usage:

```
kopia snapshot delete manifestID... [--delete]
kopia snapshot delete rootObjectID... [--delete]
```

Fixes #435

### DIFF
--- a/cli/command_snapshot_delete.go
+++ b/cli/command_snapshot_delete.go
@@ -2,64 +2,85 @@ package cli
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 
 	"github.com/kopia/kopia/repo"
 	"github.com/kopia/kopia/repo/manifest"
+	"github.com/kopia/kopia/repo/object"
+	"github.com/kopia/kopia/snapshot"
 )
 
 var (
-	snapshotDeleteCommand      = snapshotCommands.Command("delete", "Explicitly delete a snapshot by providing a snapshot ID.")
-	snapshotDeleteID           = snapshotDeleteCommand.Arg("id", "Snapshot ID to be deleted").Required().String()
-	snapshotDeletePath         = snapshotDeleteCommand.Flag("path", "Specify the path of the snapshot to be deleted").String()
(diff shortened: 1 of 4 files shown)


julio-lopez:
@jkowalski Deleting a set of snapshots by root ID is not desirable, as it may result in inadvertently deleting many snapshots.
Perhaps deleting by root id should be either a separate `deleteAllWithRoot` command, or a flag.

Thoughts?

julio-lopez:
Also, the change in the flags is going to break things for us, we need to figure out how to stage it.

jkowalski:
> @jkowalski It makes sense to improve the user experience here.
> 
> However, deleting a set of snapshots by root ID is not desirable, as it may result in inadvertently deleting many snapshots.

This is why deletion defaults to dry run and prints every snapshot that would be deleted first. Is that not sufficient?

PR #437, opened 2020-05-12
Improvements for dealing with eventually-consistent stores (S3)

### LINKED ISSUE #326: Add testing for eventual consistency behavior of blob storage providers

jkowalski:
* added compaction log which defers deletion of old indexes until later
* added cache of local writes which is merged with backend List results

Fixes #326


repo/content/content_manager_indexes.go
```diff
@@ -0,0 +1,380 @@
...
+
+func (bm *Manager) deleteOldIndexBlobs(ctx context.Context, latestBlobID blob.ID) error {
+	allCompactionLogBlobs, err := bm.listCache.listIndexBlobs(ctx, compactionLogBlobPrefix)
+	if err != nil {
+		return errors.Wrap(err, "error listing compaction log blobs")
+	}
+
+	// look for server-assigned timestamp of the compaction log entry we just wrote as a reference.
+	// we're assuming server-generated timestamps are somewhat reasonable and time is moving
+	latestBlobTime := blobTime(allCompactionLogBlobs, latestBlobID)
```
julio-lopez, line 202: Is this really the server-side timestamp? Wouldn't the metadata for `latesBlobID` come from the `ownWritesCache` and thus have a local timestamp?

repo/content/content_manager_indexes.go
```diff
@@ -0,0 +1,380 @@
...
+
+			oldCompactedBlobs[b] = true
+		}
+	}
+
+	for cb := range oldCompactedBlobs {
+		log(ctx).Debugf("deleting compacted blob %v", cb)
+
+		if err := bm.st.DeleteBlob(ctx, cb); err != nil && err != blob.ErrBlobNotFound {
+			formatLog(ctx).Warningf("unable to delete compacted blob %v, %v", cb, err)
```
julio-lopez, line 229: Doesn't this result in a case similar to the one this patch is intended to solve?
If the corresponding compaction log entry is removed, then this index pack becomes _"valid"_ again, and thus any content (not-marked as deleted) in this index pack becomes "live" again.

julio-lopez:
@pkj415 PTAL

repo/content/content_manager_indexes.go
julio-lopez: An option here is to avoid deleting the corresponding compaction log for the index blob that could not be deleted.
Would that work?

repo/content/content_manager_indexes.go
```diff
@@ -0,0 +1,380 @@
...
+
+	formatLog(ctx).Debugf("compacting %v index blobs", len(indexBlobs))
+
+	t0 := time.Now() // allow:no-inject-time
+	bld := make(packIndexBuilder)
+
+	var logEntry compactionLogEntry
+
+	// generate random log entry ID to ensure that when it is hashed later,
+	// we don't get any collisions.
```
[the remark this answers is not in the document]
jkowalski: right, that's not really needed, removed.

repo/content/content_manager_indexes.go
```diff
@@ -0,0 +1,380 @@
...
+
+const (
+	compactionLogBlobPrefix = "m"
+	blobDeleteSafeTime      = 1 * time.Hour
+)
+
+// compactionLogEntry represents contents of compaction log entry stored in `m` blob.
+type compactionLogEntry struct {
+	// random ID which ensures that each compaction log entry is unique
+	RandomID []byte `json:"randomID"`
```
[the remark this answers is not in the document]
jkowalski: removed

repo/content/content_manager_indexes.go
jkowalski: I'm going to add support for server-side timestamps in #440 before moving forward with this PR.

repo/content/content_manager_indexes.go
jkowalski: you're right

we need to delete indexes first, wait enough time for eventual consistency in lists to stop returning them, then delete logs.

If we delete them at the same time, it's entirely possible that log will not be returned by list() but index blob will

julio-lopez: Right. There's a need for an additional window between removing the index packs and the removal of the compaction logs.

This is 2-phase GC for compaction logs ;) 🐢 🐢 all the way down.

repo/content/content_manager_indexes.go
jkowalski: done, own writes cache now supports server-side timestamps. PTAL

repo/content/index_blob_manager.go
```diff
@@ -221,63 +221,86 @@ func (m *indexBlobManagerImpl) deleteOldIndexBlobs(ctx context.Context, latestBl
...
+}
+
+func (m *indexBlobManagerImpl) findIndexBlobsToDelete(ctx context.Context, latestServerBlobTime time.Time, entries map[blob.ID]*compactionLogEntry) []blob.ID {
+	tmp := map[blob.ID]bool{}
 
 	for _, cl := range entries {
 		for _, b := range cl.InputBlobs {
 			if age := latestServerBlobTime.Sub(b.Timestamp); age < m.minIndexBlobDeleteAge {
-				log(ctx).Debugf("not deleting index blob %v, because it's too recent: %v < %v", b.BlobID, age, m.minIndexBlobDeleteAge)
+				log(ctx).Debugf("not deleting compacted index blob %v, because it's too recent: %v < %v", b.BlobID, age, m.minIndexBlobDeleteAge)
```
julio-lopez, line 258: Somewhere, before an index compaction process starts, the `minIndexBlobDeleteAge` and `minCompactionLogBlobDeleteAge` values must be validated to ensure that `minIndexBlobDeleteAge < minCompactionLogBlobDeleteAge`. This is to prevent a situation where a (relatively recent) compaction log is chosen for deletion, but the corresponding index blobs are not deleted because they happen to be too recent w.r.t. `minIndexBlobDeleteAge`.
If that has been validated upfront, then this would become an "impossibility" here, and the safe thing to do would be to return an error in that case.

jkowalski: it's already there - first line of `deleteOldIndexBlobs()`

repo/content/index_blob_manager_test.go
```diff
@@ -184,14 +189,17 @@ func TestIndexBlobManagerStress(t *testing.T) {
...
 
 		eg.Go(func() error {
-			loggedSt := logging.NewWrapper(st, t.Logf, fmt.Sprintf("actor[%v]:", actorID))
-
-			contentPrefix := fmt.Sprintf("a%v", actorID)
 			numWritten := 0
 			deletedContents := map[string]bool{}
-			ctx := testlogging.ContextWithLevelAndPrefix(t, testlogging.LevelDebug, fmt.Sprintf("actor[%v]:", actorID))
+			ctx := testlogging.ContextWithLevelAndPrefixFunc(t, testlogging.LevelDebug, func() string {
+				return fmt.Sprintf("actor[%v]@%v:", actorID, fakeTimeFunc().Format("150405.000"))
```
julio-lopez, line 201: Is logging going to cause time to advance?

repo/content/content_manager.go
```diff
@@ -755,9 +724,77 @@ func newManagerWithOptions(ctx context.Context, st blob.Storage, f *FormattingOp
 		closed:                make(chan struct{}),
 	}
 
+	if err := setupCaches(ctx, m, caching); err != nil {
+		return nil, errors.Wrap(err, "unable to set up caches")
+	}
+
 	if err := m.CompactIndexes(ctx, autoCompactionOptions); err != nil {
```
pkj415, line 731: When we compact index blobs during a maintenance run to remove "old" delete-marked content entries, we intend to iterate through all index blobs and compact them all irrespective of the number of small blobs present. But in getContentsToCompact(), we still rely on logic around the number of very small blobs and might not compact all index blobs. Is that intended behaviour?

repo/content/content_manager_indexes.go
```diff
@@ -87,48 +102,52 @@ func (bm *Manager) compactAndDeleteIndexBlobs(ctx context.Context, indexBlobs []
...
+		inputs = append(inputs, indexBlob.Metadata)
 	}
 
 	var buf bytes.Buffer
 	if err := bld.Build(&buf); err != nil {
 		return errors.Wrap(err, "unable to build an index")
 	}
 
-	compactedIndexBlob, err := bm.writePackIndexesNew(ctx, buf.Bytes())
+	compactedIndexBlob, err := bm.indexBlobManager.writeIndexBlob(ctx, buf.Bytes())
```
pkj415, line 124: A question which I had earlier but never asked - we assume that all index blobs to be compacted can be compacted into a single index blob right? Would it ever happen that we cross the limit of blob size, in case there is any?

repo/content/content_manager_indexes.go
```diff
@@ -87,48 +102,52 @@ func (bm *Manager) compactAndDeleteIndexBlobs(ctx context.Context, indexBlobs []
...
+	// compaction wrote index blob that's the same as one of the sources
+	// it must be a no-op.
 	for _, indexBlob := range indexBlobs {
-		if indexBlob.BlobID == compactedIndexBlob {
-			continue
+		if indexBlob.BlobID == compactedIndexBlob.BlobID {
+			formatLog(ctx).Debugf("compaction was a no-op")
+			return nil
 		}
+	}
```
pkj415, line 136: 1. Can this case occur only when there is just one indexBlob in indexBlobs to compact?
2. Won't PutBlob return an error when indexBlobManager.writeIndexBlob tries to write a blob with the same ID?

repo/content/list_cache.go
```diff
@@ -18,70 +18,69 @@ import (
...
+				log(ctx).Debugf("retrieved list of %v '%v' index blobs from cache", len(ci.Blobs), prefix)
+				return ci.Blobs, nil
 			}
 		} else if err != blob.ErrBlobNotFound {
 			log(ctx).Warningf("unable to open cache file: %v", err)
 		}
 	}
 
-	contents, err := listIndexBlobsFromStorage(ctx, c.st)
+	blobs, err := blob.ListAllBlobs(ctx, c.st, prefix)
```
pkj415, line 40: We are now not calling ListAllBlobsConsistent. Correct me if these statements are not true -
1. This has not affect on listing for GCS, S3 because their list APIs gives a point in time list of blob IDs. (I agree that deleted blobs might also be present in the list. But it can never happen that two blobs A and B were written one after the other, and the list returns blob B but not blob A)
2. In case of local filesystem, the case I mentioned above where blob ID of B is listed but not that of A, can occur.

repo/content/index_blob_manager.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	return bm, nil
+}
+
+func (m *indexBlobManagerImpl) getCompactionLogEntries(ctx context.Context, blobs []blob.Metadata) (map[blob.ID]*compactionLogEntry, error) {
+	results := map[blob.ID]*compactionLogEntry{}
+
+	for _, cb := range blobs {
+		data, err := m.getIndexBlob(ctx, cb.BlobID)
```
pkj415, line 190: rename getIndexBlob to something generic? Since we are fetching compaction blobs here

repo/content/index_blob_manager.go
```diff
@@ -0,0 +1,325 @@
...
+
+	return results, nil
+}
+
+func (m *indexBlobManagerImpl) deleteOldIndexBlobs(ctx context.Context, latestBlob blob.Metadata) error {
+	if m.minCompactionLogBlobDeleteAge <= m.minIndexBlobDeleteAge {
+		return errors.Errorf("configuration error - compaction log deletion age (%v) must be greater than index blob deletion age (%v)", m.minCompactionLogBlobDeleteAge, m.minIndexBlobDeleteAge)
+	}
+
+	allCompactionLogBlobs, err := m.listCache.listIndexBlobs(ctx, compactionLogBlobPrefix)
```
pkj415, line 217: ditto, make listIndexBlobs something more generic?

repo/content/content_manager_indexes.go
```diff
@@ -147,3 +166,55 @@ func (bm *Manager) addIndexBlobsToBuilder(ctx context.Context, bld packIndexBuil
...
+		}
+	}
+}
+
+func removeCompactedIndexes(ctx context.Context, m map[blob.ID]*IndexBlobInfo, compactionLogs map[blob.ID]*compactionLogEntry, markAsSuperseded bool) {
+	var validCompactionLogs []*compactionLogEntry
+
+	for _, cl := range compactionLogs {
+		// only process compaction logs for which we have found all the outputs.
+		haveAllOutputs := true
```
pkj415, line 189: This is just for safety right? When can it happen that an output index blob was not found after the corresponding compaction blob was written.

repo/content/index_blob_manager.go
```diff
@@ -0,0 +1,325 @@
...
+			return nil, errors.Wrap(err, "unable to read compaction log entry %q")
+		}
+
+		results[cb.BlobID] = le
+	}
+
+	return results, nil
+}
+
+func (m *indexBlobManagerImpl) deleteOldIndexBlobs(ctx context.Context, latestBlob blob.Metadata) error {
```
pkj415, line 212: rename the function? Since we are also deleting compaction blobs, not just index blobs.

pkj415, line 212: Would we be having anything in maintenance runs to delete old index blobs, log compaction entries?

repo/content/index_blob_manager.go
```diff
@@ -0,0 +1,325 @@
...
+
+	log(ctx).Debugf("fetching %v/%v compaction logs older than %v", len(compactionBlobs), len(allCompactionLogBlobs), compactionLogServerTimeCutoff)
+
+	compactionBlobEntries, err := m.getCompactionLogEntries(ctx, compactionBlobs)
+	if err != nil {
+		return errors.Wrap(err, "unable to get compaction log entries")
+	}
+
+	indexBlobsToDelete := m.findIndexBlobsToDelete(ctx, latestBlob.Timestamp, compactionBlobEntries)
+	compactionLogBlobsToDelete := m.findCompactionLogBlobsToDelete(ctx, latestBlob.Timestamp, compactionBlobs)
```
pkj415, line 235: What if something like this occurs -
1. Index blob A is to be deleted and is present as input in log compaction blob L.
2. This function is called after a time sufficiently late to delete both blobs.
3. Index blob A and L are deleted.
4. But in some other snapshot process which just started, index blob A was read (deletes are eventually consistent) while L was not found.

repo/content/index_blob_manager.go
```diff
@@ -0,0 +1,325 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "error listing compaction log entries")
+	}
+
+	compactionLogMetadata, err = m.ownWritesCache.merge(ctx, compactionLogBlobPrefix, compactionLogMetadata)
+	if err != nil {
+		return nil, errors.Wrap(err, "error merging local writes for compaction log entries")
+	}
+
+	storageIndexBlobs, err := m.listCache.listIndexBlobs(ctx, indexBlobPrefix)
```
pkj415, line 51: Making two calls to listIndexBlobs with different prefixes one after the other is an issue here, consider this -
1. As part of some process X, the first call to list compaction logs gave us the blobs existing.
2. The process X then stalled for some reason for a long time (hours) before listing all index blobs.
3. Another process Y came in and wrote a compaction log blob, compacting some index blobs say A and B.
4. After some time, another process Z, removed index blobs A and B.
5. Just after process Z removed A, process X resumed again to now read index blob B. This is an issue as it did not read the earlier compaction blob that was written.

Maybe we should first list all index blobs, and then all compaction log blobs.

repo/content/index_blob_manager_test.go
jkowalski: yes, but it's not really a goal to advance the time precisely here.

repo/content/index_blob_manager_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	blobMD, err := m.writeIndexBlob(testlogging.Context(t), []byte(data))
+	if err != nil {
+		t.Fatalf("failed to write index blob: %v", err)
+	}
+
+	return blobMD
+}
+
+func assetIndexBlobList(t *testing.T, m indexBlobManager, wantMD ...blob.Metadata) {
```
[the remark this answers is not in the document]
jkowalski: done

repo/content/content_manager_own_writes.go
```diff
@@ -118,7 +118,7 @@ func (d *persistentOwnWritesCache) merge(ctx context.Context, prefix blob.ID, so
 			return errors.Wrapf(err, "error unmarshaling own write cache entry %v", md.BlobID)
 		}
 
-		if age := d.timeNow().Sub(originalMD.Timestamp); age < ownWritesCacheRetention {
+		if age := d.timeNow().Sub(md.Timestamp); age < ownWritesCacheRetention {
```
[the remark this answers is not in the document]
jkowalski: done

repo/content/list_cache.go
jkowalski: I think if storage is strongly consistent (filesystem, GCS, but not S3), it's guaranteed that if you do PUT(a), PUT(b), LIST() will always return a & b.

S3 LIST on the other hand can return (in rare cases, but still):

[]
[a]
[b]
[a,b]

and for some (short) time after initial writes all of these are valid responses and they can fluctuate.

ListAllBlobsConsistent was listing until it received the same set of results, which is not really correct with S3 (because it can settle on [a] or [b] which are not really complete.

repo/content/content_manager_indexes.go
jkowalski: due to eventual consistency it's possible that:

- none will be present
- log will be present, but not index file
- index will be present but not the log
- both will be present

repo/content/content_manager_own_writes.go
```diff
@@ -118,6 +118,8 @@ func (d *persistentOwnWritesCache) merge(ctx context.Context, prefix blob.ID, so
 			return errors.Wrapf(err, "error unmarshaling own write cache entry %v", md.BlobID)
 		}
 
+		// note that we're assuming that time scale used by timeNow() is the same as used by
+		// cache storage (which is fined, since the cache is local, and not on remote FS).
```
julio-lopez, line 122:
```suggestion
		// cache storage, which is fine, since the cache is local and not on remote FS.
```

repo/content/index_blob_manager.go
jkowalski: done

repo/content/index_blob_manager.go
jkowalski: done

repo/content/content_manager_indexes.go
```diff
@@ -18,19 +18,15 @@ var autoCompactionOptions = CompactOptions{
...
 )
 
 // compactionLogEntry represents contents of compaction log entry stored in `m` blob.
 type compactionLogEntry struct {
-	// random ID which ensures that each compaction log entry is unique
-	RandomID []byte `json:"randomID"`
-
 	// list of input blob names that were compacted together.
-	InputBlobs []blob.ID `json:"inputs"`
+	InputBlobs []blob.Metadata `json:"inputMetadata"`
```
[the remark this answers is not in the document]
jkowalski: renamed

repo/content/content_manager_indexes.go
jkowalski: right now there's no limit, but we can start thinking about splitting things up when repositories get big enough.

repo/content/index_blob_manager.go
jkowalski: fixed in latest commits.

jkowalski:
I finally discovered what' wrong with the stress test - it boils down to unsafe concurrent compaction done by two actors:

The sequence:

1. A creates contentA1 in INDEX-1
2. B creates contentB1 in INDEX-2
3. A deletes contentA1 in INDEX-3
4. B does compaction, but is not seeing INDEX-3 (due to EC), so it writes INDEX-4==merge(INDEX-1,INDEX-2)
    INDEX-4 has contentA1 as active
5. A does compaction but it's not seeing INDEX-4 yet (due to EC), so it drops contentA1, writes INDEX-5=merge(INDEX-1,INDEX-2,INDEX-3)
    INDEX-5 does not have contentA1
6. C sees INDEX-5 and INDEX-5 and merge(INDEX-4,INDEX-5) contains contentA1 which is wrong, because A has been deleted (and there's no record of it anywhere in the system)

jkowalski:
I'm going to disable automatic compaction that happens when repository is first opened and move it to maintenance.

repo/content/index_blob_manager.go
jkowalski: we can add it, sure, but assuming we do semi-regular compaction things will clean themselves up.

repo/content/builder.go
```diff
@@ -107,6 +109,15 @@ func (b packIndexBuilder) Build(output io.Writer) error {
 		return errors.Wrap(err, "error writing extra data")
 	}
 
+	randomSuffix := make([]byte, randomSuffixSize)
```
julio-lopez, line 112: This may not be required.

repo/content/content_manager_indexes.go
```diff
@@ -87,48 +89,52 @@ func (bm *Manager) compactAndDeleteIndexBlobs(ctx context.Context, indexBlobs []
...
-	compactedIndexBlob, err := bm.writePackIndexesNew(ctx, buf.Bytes())
+	compactedIndexBlob, err := bm.indexBlobManager.writeIndexBlob(ctx, buf.Bytes())
 	if err != nil {
 		return errors.Wrap(err, "unable to write compacted indexes")
 	}
 
-	formatLog(ctx).Debugf("wrote compacted index (%v bytes) in %v", compactedIndexBlob, time.Since(t0)) // allow:no-inject-time
-
+	// compaction wrote index blob that's the same as one of the sources
+	// it must be a no-op.
```
julio-lopez, line 117: Sorry, I missed / didn't realize this earlier.
If the output is the same as one of the inputs, it does not necessarily mean that the operation is a no-op.
You could have a situation where the inputs =  `{n1, n2, n3}` and output = `{n1}`, when say `n2` and `n3` _"cancel"_ each other. Suppose `n2` only has a content creation for `Ci`, and `n3` only has a deletion for `Ci`, that is old enough to be discarded. And suppose `{n1}` only has an index entry for `Ck != Ci`, then the output of the compaction would be `{n1}`.
One simple way of dealing with this is removing `n1` from the inputs. Essentially, the compaction log is saying which _input_ indices are superseded. Also, this suggest that a better name for the field is "superseded" or something along those lines.

repo/content/index_blob_manager.go
```diff
@@ -0,0 +1,456 @@
...
+
+	return nil
+}
+
+func (m *indexBlobManagerImpl) findIndexBlobsToDelete(ctx context.Context, latestServerBlobTime time.Time, entries map[blob.ID]*compactionLogEntry) []blob.ID {
+	tmp := map[blob.ID]bool{}
+
+	for _, cl := range entries {
+		for _, b := range cl.InputMetadata {
+			if age := latestServerBlobTime.Sub(b.Timestamp); age < m.maxEventualConsistencySettleTime {
```
julio-lopez, line 327: The age comparison here should be done using the creation time of the compaction log and not of the individual index blobs. Essentially, the decision should be made based on **when the compaction was performed**, and thus the new output index blobs and compaction blobs were created, and not when the input blobs were originally created.

Pushed a sample change to address this at https://github.com/julio-lopez/kopia/pull/6/files

repo/content/index_blob_manager.go
```diff
@@ -0,0 +1,456 @@
...
+
+	cleanupEntries, err := m.getCleanupEntries(ctx, latestBlob.Timestamp, allCleanupBlobs)
+	if err != nil {
+		return errors.Wrap(err, "error loading cleanup blobs")
+	}
+
+	indexBlobsToDelete := m.findIndexBlobsToDelete(ctx, latestBlob.Timestamp, compactionBlobEntries)
+	compactionLogBlobsToDelayCleanup := m.findCompactionLogBlobsToDelayCleanup(ctx, compactionBlobs)
+
+	if err := m.delayCleanupBlobs(ctx, compactionLogBlobsToDelayCleanup); err != nil {
```
julio-lopez, line 297: There's an issue here with the order of creation of the cleanup blobs. I'll elaborate in a few.

julio-lopez: The creation of the delay cleanup blobs should only be done _after_ the inputs for a given compaction have been removed. At this point in time (`t_c`, then we are saying:
* All the inputs for this compaction have been deleted have been deleted prior to `t_c`, then the compaction log (and the cleanup log) can be safely deleted at `t > t_c + ec_settle_time`

repo/content/builder.go
julio-lopez: Actually, for the sake of simplifying the reasoning, it makes sense to create a new index blob , even if that incurs having to upload a new (almost identical) index blob.
If this ends up being a common case, e.g., a mostly idle repo, then we can look into optimizing it.

repo/content/index_blob_manager.go
jkowalski: good catch, fixed

repo/content/index_blob_manager.go
```diff
@@ -306,6 +300,12 @@ func (m *indexBlobManagerImpl) deleteOldBlobs(ctx context.Context, latestBlob bl
 		return errors.Wrap(err, "unable to delete compaction logs")
 	}
 
+	compactionLogBlobsToDelayCleanup := m.findCompactionLogBlobsToDelayCleanup(ctx, compactionBlobs)
```
julio-lopez, line 303: Just a minor nit: we don't need to write cleanup blobs for compaction blobs that already had a cleanup blob and will end up being deleted below (line 313)

repo/content/index_blob_manager.go
jkowalski: per offline discussion, not sure if this matters since we're filtering compaction logs to include only the "settled" ones, but fixed anyway.

repo/content/builder.go
jkowalski: ack

julio-lopez: Jarek explained why this is needed.

repo/content/content_manager_indexes.go
julio-lopez: Jarek showed another case were the output could correspond to the compaction of a _previous_ input. In that case, it is not safe to use the "mark as superseded" approach described above.

PR #440, opened 2020-05-18
blob: added support for GetMetadata() to get server-side timestamp and blob length

jkowalski:


internal/blobtesting/eventually_consistent.go
```diff
@@ -129,6 +129,21 @@ func (s *eventuallyConsistentStorage) GetBlob(ctx context.Context, id blob.ID, o
...
+
+	// see if the frontend has cached blob deleted/not exists
+	e := c.get(id)
+	if e != nil {
+		if e.data == nil {
+			return blob.Metadata{}, blob.ErrBlobNotFound
+		}
+	}
+
+	// fetch from the underlying storage.
```
julio-lopez, line 143: Shouldn't `GetMetadata` behave like `GetBlob`? That is, shouldn't the entry be added to the cache as well?

IIUC, the eventually consistent semantics in the S3 stem from metadata caches. In an implementation like S3, a `GetMetadata` would have the same effect on the metadata caches as `GetBlob`, wouldn't it?

PR #462, opened 2020-06-03
content: fixed deletion of cleanup blobs

jkowalski:


repo/content/index_blob_manager.go
```diff
@@ -311,17 +300,7 @@ func (m *indexBlobManagerImpl) deleteOldBlobs(ctx context.Context, latestBlob bl
...
-	}
-
-	if err := m.deleteBlobsFromStorageAndCache(ctx, cleanupBlobsToDelete); err != nil {
-		return errors.Wrap(err, "unable to delete cleanup blobs")
-	}
-
-	m.flushCache()
-
-	return nil
+	return m.cleanupCleanupBlobs(ctx, latestBlob.Timestamp)
```
julio-lopez, line 303: It seems that there is a subtle interaction here that could prevent both cleanup and compaction logs from being deleted.
- `delayCleanupBlobs` above may "re-create" or change the time stamp of already existing cleanup blobs.
- then, when `cleanupCleanupBlobs` executes, it will list the cleanup blobs and all the ones that were (re) created above will have a recent time stamp and will not get deleted, even though they may be old enough. It seems that the same would happen in following iterations, so the cleanup blobs would be touched again and compaction blobs would remain.
Or maybe I'm missing something.

repo/content/index_blob_manager.go
```diff
@@ -362,10 +341,10 @@ func (m *indexBlobManagerImpl) findCompactionLogBlobsToDelayCleanup(ctx context.
 }
 
 func (m *indexBlobManagerImpl) findBlobsToDelete(entries map[blob.ID]*cleanupEntry) (compactionLogs, cleanupBlobs []blob.ID) {
-	for _, e := range entries {
+	for k, e := range entries {
 		if e.age > m.maxEventualConsistencySettleTime {
 			compactionLogs = append(compactionLogs, e.BlobID)
-			cleanupBlobs = append(cleanupBlobs, e.BlobID)
+			cleanupBlobs = append(cleanupBlobs, k)
```
julio-lopez, line 347: IIUC: this was one of the issues, using the incorrect blob id for the cleanup blobs, is that correct?

repo/content/content_manager_indexes.go
```diff
@@ -31,16 +31,16 @@ func (bm *Manager) CompactIndexes(ctx context.Context, opt CompactOptions) error
 		return errors.Wrap(err, "error loading indexes")
 	}
 
-	contentsToCompact := bm.getContentsToCompact(ctx, indexBlobs, opt)
+	blobsToCompact := bm.getBlobsToCompact(ctx, indexBlobs, opt)
 
-	if err := bm.compactAndDeleteIndexBlobs(ctx, contentsToCompact, opt); err != nil {
+	if err := bm.compactAndDeleteIndexBlobs(ctx, blobsToCompact, opt); err != nil {
```
julio-lopez, line 36: nit: since this is no longer deleting index blobs, perhaps rename the function?
```suggestion
	if err := bm.compactIndexBlobs(ctx, blobsToCompact, opt); err != nil {
```

repo/content/content_manager_indexes.go
```diff
@@ -31,16 +31,16 @@ func (bm *Manager) CompactIndexes(ctx context.Context, opt CompactOptions) error
...
 		log(ctx).Warningf("error performing quick compaction: %v", err)
 	}
 
-	return nil
+	return bm.indexBlobManager.cleanup(ctx)
```
julio-lopez, line 40: Doesn't `compactAndDeleteIndexBlobs` call `registerCompaction` => `deleteOldBlobs` => `cleanupCleanupBlobs`?
Wouldn't adding this call here result in a double call to `cleanupCleanupBlobs` ?

repo/content/index_blob_manager.go
jkowalski: that's correct.

repo/content/content_manager_indexes.go
jkowalski: you're right, removed it which simplified the code at the expense of having less context, which means when we're cleaning up `l` blobs we will always keep one hour's worth of them, which I think is fine.

repo/content/content_manager_indexes.go
jkowalski: renamed

repo/content/index_blob_manager.go
jkowalski: i think you're correct, let me think about it more

jkowalski: refactored log entries to:

* have all blobIDs listed together
* include timestamp of last write

this ensures that cleanup blobs will never be overwritten.

repo/content/index_blob_manager.go
```diff
@@ -383,21 +383,17 @@ func (m *indexBlobManagerImpl) deleteBlobsFromStorageAndCache(ctx context.Contex
...
-		// determine latest storage write time of a cleanup blob
-		for _, cb := range allCleanupBlobs {
-			if cb.Timestamp.After(latestStorageWriteTimestamp) {
-				latestStorageWriteTimestamp = cb.Timestamp
-			}
+	// determine latest storage write time of a cleanup blob
+	var latestStorageWriteTimestamp time.Time
+
+	for _, cb := range allCleanupBlobs {
+		if cb.Timestamp.After(latestStorageWriteTimestamp) {
```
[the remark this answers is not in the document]
jkowalski: that's correct, we can tweak this later, I think.

repo/content/index_blob_manager.go
```diff
@@ -47,7 +47,7 @@ type compactionLogEntry struct {
 
 // cleanupEntry represents contents of cleanup entry stored in `l` blob.
 type cleanupEntry struct {
-	BlobID blob.ID `json:"blobID"`
+	BlobIDs []blob.ID `json:"blobs"`
```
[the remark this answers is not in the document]
jkowalski: done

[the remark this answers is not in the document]
jkowalski: done

PR #471, opened 2020-07-14
[Robustness] Add first robustness tests

redgoat650:
Add two tests:
- TestManySmallFiles: writes 100k files size 4k to a directory. Snapshots the data tree, restores and validates data.
- TestModifyWorkload: Loops over a simple randomized workload. Performs a series of random file writes to some random sub-directories, then takes a snapshot of the data tree. All snapshots taken during this test are restore-verified at the end.

A global test engine is instantiated in main_test.go, to be used in the robustness test suite across tests (saves time loading/saving metadata once per run instead of per test).

Updated https://github.com/kopia/kopia/issues/179 - simple test cases

### DIFF
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -1,0 +1,55 @@
+// +build darwin linux
+
+package robustness
+
+import (
+	"context"
+	"errors"
+	"log"
+	"os"
+	"testing"
+
+	engine "github.com/kopia/kopia/tests/robustness/test_engine"
+	"github.com/kopia/kopia/tests/tools/fio"
+	"github.com/kopia/kopia/tests/tools/kopiarunner"
+)
+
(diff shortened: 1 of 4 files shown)


tests/robustness/main_test.go
```diff
@@ -0,0 +1,55 @@
...
+package robustness
+
+import (
+	"context"
+	"errors"
+	"log"
+	"os"
+	"testing"
+
+	engine "github.com/kopia/kopia/tests/robustness/test_engine"
```
jkowalski, line 12: out of curiosity, why not just rename the package to avoid aliasing?

redgoat650: Good point. I don't recall having any specific reason for doing it this way. I'll address this in a future PR.

Issue #492, opened 2020-07-21
gitignore Syntax problems


darkdragon-001:
I am trying to ignore files. The following parts of the [gitignore format](https://git-scm.com/docs/gitignore#_pattern_format) don't seem to work:

- negated patterns prefixed with `!`
- absolute paths starting with `/`: I tried paths relative to system root (`/`) and repository root (`/home/USER` in my case)

jkowalski:
I'm curious what are you trying to achieve here. 

In general you can attach policies and ignore rules/dotignore files to any directory in a hierarchy, in addition to that you can mark policies to either inherit from the parent or not, so you can organize your ignores in a quite flexible way.

darkdragon-001:
I want to backup my home `~` folder but exclude `~/bin` but want to include `~/Development/bin` for example. Re-including excluded files is not supported by gitignore.

julio-lopez:
@darkdragon-001 Could you submit a PR with your suggestions to improve the documentation? Contributions are welcome.

darkdragon-001:
Actually, I would prefer Kopia implementing gitignore specification. This has the nice effects, that all `.gitignore` files in all backed up git repositories will be taken into account!

jkowalski:
Looking at the code, `!` should be supported (https://github.com/kopia/kopia/blob/aa1b24762d495337583dfd0a88e4487e6742e8d6/internal/ignore/ignore.go) but it may be missing some cases/have bugs. If you have a concrete example that's not working, please add it here or in the form of a test or even better a PR with the fix 😄 

As for absolute pathnames, they are somewhat tricky, since the code to do upload is never operating on absolute pathnames, but instead it uses virtual filesystem. While this probably __could__ be fixed, it may be more tricky since the upload code is used not just for uploads from the filesystem but also repo-to-repo migrations, so definitely needs more investigation.

darkdragon-001:
Thanks for checking and the link to the code!

I think I was testing a combination of `!` and `/path`.

Regarding absolute path names, the library should support this when setting `baseDir` correctly:
https://github.com/kopia/kopia/blob/aa1b24762d495337583dfd0a88e4487e6742e8d6/internal/ignore/ignore.go#L17

[...]

PR #496, opened 2020-07-23
Improve ignore logic.

darkdragon-001:
- Add test cases for absolute patterns
- Support globs for absolute patterns

See https://github.com/kopia/kopia/issues/492#issuecomment-662842450 for details.

### DIFF
--- a/fs/ignorefs/ignorefs_test.go
+++ b/fs/ignorefs/ignorefs_test.go
@@ -241,6 +241,38 @@ var cases = []struct {
 			"./src/some-src/f1",
 		},
 	},
+	{
+		desc: "absolut match",
+		setup: func(root *mockfs.Directory) {
+			root.Subdir("src").AddFileLines(".extraignore", []string{
+				"/sub/*.foo",
+			}, 0)
+			root.Subdir("src").AddDir("sub", 0)
+			root.Subdir("src").Subdir("sub").AddFile("a.foo", dummyFileContents, 0) // ignored by .extraignore
+			root.Subdir("src").Subdir("sub").AddFile("b.fooX", dummyFileContents, 0)
+			root.Subdir("src").Subdir("sub").AddFile("foo", dummyFileContents, 0)
+			root.Subdir("src").AddFile("c.foo", dummyFileContents, 0) // not ignored, at parent level
+		},
+		policyTree: policy.BuildTree(map[string]*policy.Policy{
+			"./src": {
+				FilesPolicy: policy.FilesPolicy{
+					DotIgnoreFiles: []string{
+						".extraignore",
+					},
+				},
+			},
+		}, policy.DefaultPolicy),
+		addedFiles: []string{
+			"./src/.extraignore",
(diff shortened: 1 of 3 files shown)


x80486:
I'm trying to come up with a `.kopiaignore` file that fits my needs, and so far, it's not "including on an exclusion". This is my example:

```shell
Desktop/
Workshop/**/.idea/
Workshop/**/.terraform/
Workshop/**/node_modules/
Workshop/**/*.iml/
Workshop/**/*.ipr/
.cache/
.config/
.gradle/
.gradle-kotlin-dsl/
.java/
.local/
.m2/
.npm/
.var/
.bash_history
.node_repl_history
# The previous lines work as expected, but as soon as I add the following lines, everything is excluded
!Workshop/**/terraform.tfstate.d/
!Workshop/**/terraform.tfstate
!.config/Cryptomator/
!.config/JetBrains/**/keymaps/
!.config/JetBrains/**/idea.key
!.local/share/bash-completion/
!.local/share/now/
```

Are you addressing something like this in this Pull Request?

darkdragon-001:
@x80486 Just pull this code, add your test cases and compile & test :wink: 

Previously *all* globs where not processed when a `/` was within pattern. So some of your patterns should work.

Quoting from the [pattern format definition](https://git-scm.com/docs/gitignore#_pattern_format)

> An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again. It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined. Put a backslash ("\") in front of the first "!" for patterns that begin with a literal "!", for example, "\!important!.txt".

at least some of your test cases are not supported.

x80486:
OK, I'll wait for this release, but I can guarantee you with `kopia-0.5.2`, what I have there doesn't work :no_good_man: ...and that's exactly what I use with `borgbackup`.

darkdragon-001:
@x80486 It would still be helpful if you construct some test cases (positive and negative ones) like I did: https://github.com/kopia/kopia/pull/496/files#diff-8527ab4f16d218759fedf0987ace90f7.

This are just short one-liners for you, but help a lot in keeping this module stable ove time.

jkowalski:
ok, I think we're ready to merge this, let's just make sure linter is clean

8176135:
Perhaps merge this PR first, and then work on the improvements for unignoring files later?

The current ignore syntax on master doesn't work for any paths starting with /, which is quite commonly used in most .gitignore files, and this branch works for that, which is a major improvement.

darkdragon-001:
> Can you also add couple smoke tests in `ignorefs_test.go`, which is one layer above.

@jkowalski I added such tests.

jkowalski:
@darkdragon-001 can you update the branch and re-push? I'll be happy to take a look and we can merge soon then work on additional cases.

jkowalski:
LGTM, Thanks!

Issue #490, opened 2020-07-19
compression performance regression at commit 272b657508329ac4035c656c2503f6f476d42e7d


jkowalski:
This happened back in March between 0.5.2 and HEAD, looks like we lost >31% of compression efficiency of `s2-default` after upgrading compression libraries.

On 200 GB sample files the compressed repository size before the change was 89.98 GB, after it jumped to 118.67 GB.

The suspect change was on 2020-03-15 and upgraded `github.com/klauspost/compress` from `v1.9.7` to `v1.10.3`

Thanks @ntolia for the investigation!

jkowalski:
I tried creating an e2e repro, and while there's a change it's only minor (0.7% which is consistent with change to compress library)

I created 2 filesystem repos and backed up identical 1000 files of 200 MB each. Total size as reported by `du -s`:

```
52944232	kopia-repo (HEAD)
52555364	kopia-repo-old (v0.5.2)
```

@ntolia can you post your repro script here?

julio-lopez:
commit 272b657508329ac4035c656c2503f6f476d42e7d

ntolia:
I created a self-contained repro here. I first created files using:
```
mkdir -p /files/largefiles
fio --nrfiles=100 --size=200G --bs=256k --iodepth=32 --dedupe_percentage=40 --buffer_compress_percentage=60 --numjobs=1 --rw=write --name=kopia-test --directory=/files/largefiles
```
These environment variables were defined:
```
export AWS_ACCESS_KEY_ID=xxx
export AWS_SECRET_ACCESS_KEY=xxx
export S3_BUCKET=kasten-kopia-test
export S3_ENDPOINT=s3.us-west-2.amazonaws.com
export S3_REGION=us-west-2 
export KOPIA_PASSWORD=totallyinsecure
```

I then ran this repro script:
```
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail
set -o xtrace

${HOME}/go/bin/kopia --version

cd $HOME
curl -L -O https://github.com/kopia/kopia/releases/download/0.5.2/kopia-0.5.2-linux-x64.tar.gz
tar zxvf kopia-0.5.2-linux-x64.tar.gz
chmod +x kopia-0.5.2-linux-x64/kopia
sudo mv kopia-0.5.2-linux-x64/kopia /usr/bin/kopia0.5.2

git clone https://github.com/kopia/kopia.git
cd kopia
make install-noui

rm -rf ~/.cache/kopia/
[...]

PR #498, opened 2020-07-24
Rolled back github.com/klauspost/compress to v1.9.7

jkowalski:
Tactical fix for #490 due to klauspost/compress#272 which caused large unexpected regression in compression efficiency for s2-default.

Will roll forward once fixed upstream or if we can find a workaround.


jkowalski:
Looks like upgrading to `v1.10.10` should fix it.

Issue #505, opened 2020-07-28
Make sure kopia passes all tests on arm and arm64 (was: Panic/segfault when creating snapshot on arm)


aksdb:
* The repository in question was created with 0.5.2.
* Snapshot is being created using kopia 0.6-rc1
* Platform is **arm**.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11f08]

goroutine 1356 [running]:
runtime/internal/atomic.goStore64(0x70a6914, 0xe95a467c, 0x1625f609)
        /home/travis/.gimme/versions/go1.14.6.linux.amd64/src/runtime/internal/atomic/atomic_arm.go:144 +0x1c
github.com/kopia/kopia/repo/content.(*cacheBase).perItemMutex(0x2df61e0, 0x9a1068, 0x2f4a310, 0x0)
        /home/travis/gopath/src/github.com/kopia/kopia/repo/content/content_cache_base.go:68 +0x120
github.com/kopia/kopia/repo/content.(*contentCacheForMetadata).getContent(0x2cb2560, 0xc93ef0, 0x304e120, 0x2cd0120, 0x21, 0x2cd0120, 0x21, 0x0, 0x0, 0xffffffff, ...)
        /home/travis/gopath/src/github.com/kopia/kopia/repo/content/content_cache_metadata.go:53 +0x74
github.com/kopia/kopia/repo/content.(*indexBlobManagerImpl).getEncryptedBlob(0x2ca6440, 0xc93ef0, 0x304e120, 0x2cd0120, 0x21, 0x3ac578, 0x2c647c0, 0x2e45fbc, 0x1, 0x101)
        /home/travis/gopath/src/github.com/kopia/kopia/repo/content/index_blob_manager.go:145 +0x70
github.com/kopia/kopia/repo/content.(*indexBlobManagerImpl).getIndexBlob(0x2ca6440, 0xc93ef0, 0x304e120, 0x2cd0120, 0x21, 0x0, 0x0, 0x2d93c40, 0x2d93c40, 0x0)
        /home/travis/gopath/src/github.com/kopia/kopia/repo/content/index_blob_manager.go:141 +0x3c
[...]

PR #506, opened 2020-07-30
Fixed arm and arm64 build.

jkowalski:
Fixed atomic alignment problems on ARMv7 mentioned in #505 

Now it's possible to build and run tests on both `armv7` and `armv8` a.k.a. `arm64` a.k.a. `aarch64`. The following are working:

```
$ make
$ make install-noui
$ make install
$ make test
$ make lint
$ make integration-test
$ make travis-setup
# slow
$ make travis-release
```

Building `site` does not work due to lack of appropriate Hugo build for ARM.
Determining whether it makes sense to build `kopia-ui` on ARM is TBD, but will certainly be interesting when new arm macbooks arrive.

Tested on 2 machines:

Raspberry PI 4 with 4 GB RAM on Ubuntu 20.04 LTS in 64-bit mode (aarch64)
Raspberry PI 4 with 2 GB RAM on Raspbian GNU/Linux 10 (buster) with 2GB RAM in 32-bit mode (armv7l)

### DIFF
--- a/repo/content/content_manager_lock_free.go
+++ b/repo/content/content_manager_lock_free.go
@@ -23,8 +23,7 @@ const indexBlobCompactionWarningThreshold = 100
 
 // lockFreeManager contains parts of Manager state that can be accessed without locking
 type lockFreeManager struct {
-	// this one is not lock-free
-	Stats Stats
+	Stats *Stats
 
 	st             blob.Storage
 	Format         FormattingOptions
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ go:
 os:
 - linux
 - osx
+arch:
+- amd64
+- arm64
 addons:
   apt:
     packages:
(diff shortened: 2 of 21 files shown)


repo/content/content_manager_lock_free.go
```diff
@@ -23,8 +23,7 @@ const indexBlobCompactionWarningThreshold = 100
 
 // lockFreeManager contains parts of Manager state that can be accessed without locking
 type lockFreeManager struct {
-	// this one is not lock-free
-	Stats Stats
+	Stats *Stats
```
julio-lopez, line 26: The reason this is being converted into a pointer is because when `lockFreeManager` is embedded into `content.Manager` then `Stats` is no longer aligned, is that correct?

The downside of making it a pointer is that it could now be `nil`. In this case, it's probably OK since this is only instantiated in `content.Manager` and the _right thing happens_.

internal/stats/countsum.go
```diff
@@ -1,4 +1,7 @@
+// +build amd64
```
julio-lopez, line 1: Should the file be renamed `countsum_amd64` ?

PR #518, opened 2020-08-05
Fix #516 - Chown not suported on Windows

### LINKED ISSUE #516: Restoring a linux snapshot on Windows fails
If I try to restore a snapshot made from my ubuntu machine to my Windows machine it fails with this error

```
.\kopia.exe restore k453a295724f6e33cfe6ade9ab97347f9 home
Restoring to local filesystem (home)...
22:00:14.072 [restore] WriteFile Desktop/***** &{{0xc0001d80e0 0xc000064000}}
kopia.exe: error: error copying: copy directory contents: copy directory contents: error setting attributes: could not change owner/group for C:\Users\****\home\Desktop/*****: chown C:\Users\****\home\Desktop/*****: not supported by windows, try --help
```

cc @redgoat650, @jkowalski

redgoat650:
Proposed fix for #516: Do not attempt to call `os.Chown` on Windows. The
command is unsupported on windows and will always return
`syscall.EWINDOWS`.

Handling for user ID and group ID on Windows already hardcodes
both values to zero. If a snapshot is restored with non-zero UID/GID
values (for instance if the snapshot was created on a different OS),
filesystem entries will be restored without changing UID/GID, resulting
in a state similar to how they would look if the snapshot was
taken on Windows in the first place.

### DIFF
--- a/snapshot/restore/local_fs_output.go
+++ b/snapshot/restore/local_fs_output.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/natefinch/atomic"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


redgoat650:
@julio-lopez @jkowalski - I believe this solution is appropriate, but looking for second opinions. What are your thoughts?

redgoat650:
CI tests failing for https://github.com/kopia/kopia/issues/519. Fix is here: https://github.com/kopia/kopia/pull/520

PR #526, opened 2020-08-12
Linter upgrade to v1.30.0

jkowalski:
Lots of fixes required, tried keeping each type of change in separate commit.

### DIFF
--- a/internal/blobtesting/asserts.go
+++ b/internal/blobtesting/asserts.go
@@ -31,7 +31,7 @@ func AssertGetBlob(ctx context.Context, t testingT, s blob.Storage, blobID blob.
 		t.Errorf("GetBlob(%v) returned %x, but expected %x", blobID, b, expected)
 	}
 
-	half := int64(len(expected) / 2) //nolint:gomnd
+	half := int64(len(expected) / 2)
 	if half == 0 {
 		return
 	}
--- a/internal/blobtesting/asserts.go
+++ b/internal/blobtesting/asserts.go
@@ -7,8 +7,9 @@ import (
 	"sort"
 	"time"
 
-	"github.com/kopia/kopia/repo/blob"
 	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/repo/blob"
 )
 
 type testingT interface {
--- a/internal/blobtesting/asserts.go
+++ b/internal/blobtesting/asserts.go
@@ -73,7 +73,7 @@ func AssertGetBlob(ctx context.Context, t testingT, s blob.Storage, blobID blob.
 	AssertInvalidOffsetLength(ctx, t, s, blobID, int64(len(expected)+1), 3)
 }
 
-// AssertInvalidOffsetLength verifies that the given combination of (offset,length) fails on GetBlob()
+// AssertInvalidOffsetLength verifies that the given combination of (offset,length) fails on GetBlob().
 func AssertInvalidOffsetLength(ctx context.Context, t testingT, s blob.Storage, blobID blob.ID, offset, length int64) {
 	if _, err := s.GetBlob(ctx, blobID, offset, length); err == nil {
 		t.Errorf("GetBlob(%v,%v,%v) did not return error for invalid offset/length", blobID, offset, length)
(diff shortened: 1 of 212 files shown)


tests/end_to_end_test/snapshot_fail_test.go
```diff
@@ -283,7 +283,7 @@ func testPermissions(e *testenv.CLITest, t *testing.T, source, modifyEntry, rest
 }
 
 func parseSnapID(t *testing.T, lines []string) string {
-	pattern := regexp.MustCompile(`Created snapshot with root [\S]+ and ID ([\S]+) in .*`)
+	pattern := regexp.MustCompile(`Created snapshot with root \S+ and ID (\S+) in .*`)
```
julio-lopez, line 286: @redgoat650 PTAL

tests/tools/kopiarunner/kopia_snapshotter.go
```diff
@@ -122,7 +122,7 @@ func (ks *KopiaSnapshotter) Run(args ...string) (stdout, stderr string, err erro
 }
 
 func parseSnapID(lines []string) (string, error) {
-	pattern := regexp.MustCompile(`uploaded snapshot ([\S]+)`)
+	pattern := regexp.MustCompile(`uploaded snapshot (\S+)`)
```
julio-lopez, line 126: @redgoat650

internal/blobtesting/asserts.go
```diff
@@ -85,7 +86,7 @@ func AssertGetBlobNotFound(ctx context.Context, t testingT, s blob.Storage, blob
 	t.Helper()
 
 	b, err := s.GetBlob(ctx, blobID, 0, -1)
-	if err != blob.ErrBlobNotFound || b != nil {
+	if !errors.Is(err, blob.ErrBlobNotFound) || b != nil {
```
julio-lopez, line 90: Isn't this in the standard `errors` package since Go 1.13?
Does the linter complain about using 1.13 specific functions?

internal/diff/diff.go
```diff
@@ -279,11 +279,12 @@ func downloadFile(ctx context.Context, f fs.File, fname string) error {
...
 	}
-	defer dst.Close() //nolint:errcheck
 
-	_, err = iocopy.Copy(dst, src)
+	if _, err := iocopy.Copy(dst, src); err != nil {
+		return err
+	}
 
-	return err
+	return dst.Close()
```
julio-lopez, line 287: Shouldn't `dst` be closed in the error case as well?

internal/blobtesting/asserts.go
jkowalski: the linter cares about "errors" package and not "github.com/pkg/errors", this was a good forcing function to standardize on the latter.

internal/diff/diff.go
jkowalski: good catch

tests/end_to_end_test/snapshot_fail_test.go
jkowalski: FWIW this change was suggested by linter, since the test is not failing it must be good.

internal/diff/diff.go
jkowalski: fixed

tests/end_to_end_test/snapshot_fail_test.go
redgoat650: Yeah this is an equivalent expression. Tested manually as well, just in case

repo/blob/b2/b2_storage.go
```diff
@@ -261,7 +261,7 @@ func toBandwidth(bytesPerSecond int) iothrottler.Bandwidth {
 	return iothrottler.Bandwidth(bytesPerSecond) * iothrottler.BytesPerSecond
 }
 
-// New creates new B2-backed storage with specified options:
+// New creates new B2-backed storage with specified options:.
```
redgoat650, line 264: May not be formatted correctly
```suggestion
// New creates new B2-backed storage with specified options.
```

PR #530, opened 2020-08-13
kopia-ui: roll back electron to v7.1.14 since it's crashing on Ubuntu

jkowalski:
Caused by cfa30f2


jkowalski:
Verified that with the rollback the build works fine on Ubuntu 20.04 LTS

PR #535, opened 2020-08-15
infra: set up github action to run quick lint on pull requests

jkowalski:


jkowalski:
I don't think this action will run on this PR.

PR #536, opened 2020-08-15
this has a linter warning

jkowalski:


jkowalski:
looks like it worked, linter failed in 1m

PR #553, opened 2020-08-23
Tools to help investigate repository structures safely

jkowalski:
- Option to connect to repository in a read-only mode which prevents any mutations.
- Option to dump contents of a particular index blob.


repo/content/content_manager_indexes.go
```diff
@@ -157,6 +157,28 @@ func (bm *Manager) addIndexBlobsToBuilder(ctx context.Context, bld packIndexBuil
 	return nil
 }
 
+// ParseIndexBlob loads entries in agiven index blob and returns them.
```
julio-lopez, line 160:
```suggestion
// ParseIndexBlob loads entries in a given index blob and returns them.
```

PR #558, opened 2020-08-26
Added endurance test which tests kopia over long time scale

jkowalski:
This creates kopia repository and simulates usage of Kopia over multiple months (using accelerated fake time) to trigger effects that are visible after long time passage (maintenance, compactions, expirations).

Fake time is injected into `kopia` binary through a HTTP endpoint that runs on localhost.
    
The test is not used part of any test suite yet but will run in post-submit mode only, preferably 24/7.

### DIFF
--- a/internal/clock/now.go
+++ b/internal/clock/now.go
@@ -1,0 +1,83 @@
+// Package clock provides indirection for accessing current time.
+package clock
+
+import (
+	"encoding/json"
+	"log"
+	"net/http"
+	"os"
+	"sync"
+	"time"
+)
+
+const refreshServerTimeEvery = 3 * time.Second
+
+// Now is overridable function that returns current wall clock time.
+var Now = time.Now // allow:no-inject-time
+
+// Since returns time since the given timestamp.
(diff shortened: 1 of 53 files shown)


internal/clock/now.go
```diff
@@ -0,0 +1,83 @@
...
+	"log"
+	"net/http"
+	"os"
+	"sync"
+	"time"
+)
+
+const refreshServerTimeEvery = 3 * time.Second
+
+// Now is overridable function that returns current wall clock time.
```
julio-lopez, line 15: To prevent this from being overridden at runtime in production, would it make sense to put this behind a build flag? such that the default implementation is:

```go
func Now() time.Time {
	return time.Now()
}
```

And an alternate version with a build flag with the one here.

```go
var Now = time.Now
```

Of course, this would require the corresponding build flags for the endurance test.

jkowalski: excellent idea

jkowalski: done

PR #563, opened 2020-08-30
content: fixed index compaction that would resurrect content entry during full maintenance

### LINKED ISSUE #483: investigate stress test failure
https://ci.appveyor.com/project/jkowalski/kopia/builds/34169820

jkowalski:
The problem was that while dropping deleted entries from the index
we were looking at index-local and not merged result.

Say index1 has content1 created at t0.
Say index2 has content1 deleted at t1 > t0.

And we're compacting and dropping index entries deleted before t2,
such that t2 > t1.

Because create and delete records are in separate index files, the
compacted file would preserve "created t0" which is wrong.

This fix changes index compaction to always look at resolved content
status.

Fixes #483 (hopefully)

### DIFF
--- a/repo/content/content_manager_indexes.go
+++ b/repo/content/content_manager_indexes.go
@@ -33,6 +33,7 @@ func (bm *Manager) CompactIndexes(ctx context.Context, opt CompactOptions) error
 	}
 
 	blobsToCompact := bm.getBlobsToCompact(ctx, indexBlobs, opt)
+	log(ctx).Debugf("blobs to compact: %v", blobsToCompact)
 
 	if err := bm.compactIndexBlobs(ctx, blobsToCompact, opt); err != nil {
 		return errors.Wrap(err, "error performing compaction")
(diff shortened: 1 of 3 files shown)


repo/content/content_manager_indexes.go
```diff
@@ -140,16 +165,6 @@ func (bm *Manager) addIndexBlobsToBuilder(ctx context.Context, bld packIndexBuil
 	}
 
 	_ = index.Iterate(AllIDs, func(i Info) error {
-		for _, dc := range opt.DropContents {
-			if dc == i.ID {
-				log(ctx).Debugf("dropping content %v", i.ID)
-				return nil
-			}
-		}
-		if i.Deleted && !opt.DropDeletedBefore.IsZero() && i.Timestamp().Before(opt.DropDeletedBefore) {
```
julio-lopez, line 149: It seems that this was assuming that there would only be a single entry for the "deleted" content, so it was enough to simply not include it in the resulting index.

repo/content/content_manager_indexes.go
```diff
@@ -93,13 +94,19 @@ func (bm *Manager) compactIndexBlobs(ctx context.Context, indexBlobs []IndexBlob
...
+		if err := bm.addIndexBlobsToBuilder(ctx, bld, indexBlob); err != nil {
 			return errors.Wrap(err, "error adding index to builder")
 		}
 
 		inputs = append(inputs, indexBlob.Metadata)
 	}
 
+	// after we built index map in memory, drop contents from it
+	// we must do it after all input blobs have been merged, otherwise we may resurrect contents.
+	dropContentsFromBuilder(ctx, bld, opt)
```
julio-lopez, line 108: This solves the case where there are multiple entries (in separate indices) for a given content (`ci`) and one of the entry with the most recent time stamp is marking the content as deleted and the compaction includes **all** the indices that have entries for `ci`.

My assumption was that during a full maintenance **all the indices** would be included, and thus it was safe to drop entries without resurrecting them.

Looking at L75 above (https://github.com/kopia/kopia/pull/563/files#diff-c8900248980c856e920adf28aa8b1019R75) it seems that even when `CompactOptions.AllIndexes == true`, there could be (maybe rare) cases where not all the indices are considered, and thus content could be resurrected from an old (relatively large) index that was not considered during the compaction. Or perhaps I am missing something.

julio-lopez: Follow up to the comment above.

Let me take that back.
It seems that during full maintenance, `getBlobsToCompact` (https://github.com/kopia/kopia/pull/563/files#diff-c8900248980c856e920adf28aa8b1019R75) would return all blobs because `CompactOptions.MaxSmallBlobs == 0` during full maintenance.

There are a number of "long-range dependencies and assumptions" in how the different code paths interact that make it harder to reason about the invariants maintained by maintenance.

PR #565, opened 2020-08-30
implemented Cache Directory Tagging Specification + CLI + UI

### LINKED ISSUE #564: Implement support for Cache Directory Tagging Specification
Hi,

I’ve already commented in another issue (https://github.com/kopia/kopia/issues/562#issuecomment-683363801) but since it’s closed I thought best to open a new one.

Would you consider implementing support for the Cache Directory Tagging Specification (https://bford.info/cachedir/)? It’s already honoured by several backup tools (tar, borg, restic, déjà dup added it recently) and other non-backup softwares create the appropriate files in their cache directories (man, fontconfig, restic itself) so it would help with interoperability.

A basic implementation could just mirror what was added in PR #524, replacing the custom file with CACHEDIR.TAG or checking for the presence of either, eventually guarding it with a policy option.

jkowalski:
Fixes #564

### DIFF
--- a/fs/ignorefs/ignorefs.go
+++ b/fs/ignorefs/ignorefs.go
@@ -56,7 +56,7 @@ type ignoreDirectory struct {
 	fs.Directory
 }
 
-func correctCacheDirSignature(ctx context.Context, f fs.File) (bool, error) {
+func isCorrectCacheDirSignature(ctx context.Context, f fs.File) (bool, error) {
 	const (
 		validSignature    = repo.CacheDirMarkerHeader
 		validSignatureLen = len(validSignature)
@@ -89,7 +89,7 @@ func (d *ignoreDirectory) skipCacheDirectory(ctx context.Context, entries fs.Ent
 
 	f, ok := entries.FindByName(repo.CacheDirMarkerFile).(fs.File)
 	if ok {
-		correct, err := correctCacheDirSignature(ctx, f)
+		correct, err := isCorrectCacheDirSignature(ctx, f)
 		if err != nil {
 			log(ctx).Debugf("unable to check cache dir signature, assuming not a cache directory: %v", err)
 			return entries
--- a/fs/ignorefs/ignorefs.go
+++ b/fs/ignorefs/ignorefs.go
@@ -58,7 +58,7 @@ type ignoreDirectory struct {
 
 func correctCacheDirSignature(ctx context.Context, f fs.File) (bool, error) {
 	const (
-		validSignature    = "Signature: 8a477f597d28d172789f06886806bc55"
+		validSignature    =  repo.CacheDirMarkerHeader
 		validSignatureLen = len(validSignature)
 	)
 
--- a/fs/ignorefs/ignorefs.go
+++ b/fs/ignorefs/ignorefs.go
@@ -10,8 +10,12 @@ import (
 
 	"github.com/kopia/kopia/fs"
 	"github.com/kopia/kopia/internal/ignore"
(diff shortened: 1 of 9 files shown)


repo/open.go
```diff
@@ -193,7 +205,14 @@ func writeCacheMarker(cacheDir string) error {
 	}
 
 	markerFile := filepath.Join(cacheDir, CacheDirMarkerFile)
-	if _, err := os.Stat(markerFile); !os.IsNotExist(err) {
+
+	st, err := os.Stat(markerFile)
+	if err == nil && st.Size() >= int64(len(cacheDirMarkerContents)) {
```
julio-lopez, line 210:
```suggestion
	if err == nil && st.Size() >= int64(len(CacheDirMarkerHeader)) {
```
It may be a good idea to check the signature header here.
Otherwise this is assuming that the header is correct and it will be skipped in `ignorefs`.
While unlikely, it may be that the contents don't match.

The implemented behavior is to overwrite the file if it is shorter or inconsistent. That's OK, just noting it.

fs/ignorefs/ignorefs.go
```diff
@@ -52,12 +56,66 @@ type ignoreDirectory struct {
 	fs.Directory
 }
 
+func correctCacheDirSignature(ctx context.Context, f fs.File) (bool, error) {
+	const (
+		validSignature    = "Signature: 8a477f597d28d172789f06886806bc55"
```
julio-lopez, line 61: For consistency:
```suggestion
		validSignature    =  repo.CacheDirMarkerHeader
```

fs/ignorefs/ignorefs.go
```diff
@@ -52,12 +56,66 @@ type ignoreDirectory struct {
 	fs.Directory
 }
 
+func correctCacheDirSignature(ctx context.Context, f fs.File) (bool, error) {
```
julio-lopez, line 59: Given the comment in `repo.OpenWithConfig`, it may make sense to move this function to `repo` and export it instead.

fs/ignorefs/ignorefs.go
```diff
@@ -52,12 +56,66 @@ type ignoreDirectory struct {
...
+	const (
+		validSignature    =  repo.CacheDirMarkerHeader
```
julio-lopez, line 61:
```suggestion
		validSignature    = repo.CacheDirMarkerHeader
```

fs/ignorefs/ignorefs.go
julio-lopez:
```suggestion
func isCorrectCacheDirSignature(ctx context.Context, f fs.File) (bool, error) {
```

fs/ignorefs/ignorefs.go
```diff
@@ -52,12 +56,66 @@ type ignoreDirectory struct {
...
+}
+
+func (d *ignoreDirectory) skipCacheDirectory(ctx context.Context, entries fs.Entries, relativePath string, policyTree *policy.Tree) fs.Entries {
+	if !policyTree.EffectivePolicy().FilesPolicy.IgnoreCacheDirectoriesOrDefault(true) {
+		return entries
+	}
+
+	f, ok := entries.FindByName(repo.CacheDirMarkerFile).(fs.File)
+	if ok {
+		correct, err := correctCacheDirSignature(ctx, f)
```
julio-lopez, line 92:
```suggestion
		correct, err := isCorrectCacheDirSignature(ctx, f)
```

jkowalski:
thank you so much for the fixes!

PR #566, opened 2020-08-30
Doc: fix create B2 repository cmd

phonovision:
Fix command line: "create" instead of "connect"


jkowalski:
Thank you for your contribution!

PR #602, opened 2020-09-10
Trivial test nits

julio-lopez:


internal/repotesting/repotesting.go
```diff
@@ -102,10 +91,6 @@ func (e *Environment) Close(ctx context.Context, t *testing.T) {
 		// should be empty, assuming Disconnect was successful
```
julio-lopez, line 91: The `Remove` above, although redundant, it is a safety check. The alternative would be to list the contents and check for an empty list.

PR #607, opened 2020-09-12
object: implemented fast concatenation of objects

jkowalski:
This is primarily to facilitate efficient parallel uploads of very large files (>1GB). Due to bottleneck of splitting which is inherently sequential, we can only one use CPU core for each Writer, which limits throughput.

### DIFF
--- a/repo/object/object_manager.go
+++ b/repo/object/object_manager.go
@@ -80,6 +80,105 @@ func (om *Manager) Open(ctx context.Context, objectID ID) (Reader, error) {
 	return om.openAndAssertLength(ctx, objectID, -1)
 }
 
+// Concatenate creates an object that's a result of concatenation of other objects. This is more efficient than reading
+// and rewriting the objects because Concatenate can efficiently merge index entries without reading the underlying
+// contents.
+//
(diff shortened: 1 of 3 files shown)


repo/object/object_manager.go
```diff
@@ -80,6 +80,112 @@ func (om *Manager) Open(ctx context.Context, objectID ID) (Reader, error) {
...
+		rd, err := om.Open(ctx, indexObjectID)
+		if err != nil {
+			return nil, 0, errors.Wrapf(err, "error opening index of %v", objectID)
+		}
+		defer rd.Close() //nolint:errcheck
+
+		ndx, err := om.flattenListChunk(rd)
+		if err != nil {
+			return nil, 0, errors.Wrapf(err, "error reading index of %v", objectID)
+		}
```
julio-lopez, line 149: Coincidentally, there's a helper for this in a PR that just got merged. #608

julio-lopez (approved):
👍 LG +1

repo/object/object_manager.go
jkowalski: thanks, rebased and updated the PR.

julio-lopez (approved):

PR #611, opened 2020-09-13
cli: ensure advanced commands are not accidentally used

jkowalski:
This prints an error when a dangerous command is used without
first setting KOPIA_ADVANCED_COMMANDS=enabled environment variable.

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -188,12 +188,9 @@ func advancedCommand() {
 	if os.Getenv("KOPIA_ADVANCED_COMMANDS") != "enabled" {
 		//nolint:errcheck
 		errorColor.Printf(`
-This command is meant for advanced users only and could be dangerous or lead to
-repository corruption when used improperly.
+This command could be dangerous or lead to repository corruption when used improperly.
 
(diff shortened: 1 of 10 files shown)


cli/app.go
```diff
@@ -184,6 +184,23 @@ func maybeRunMaintenance(ctx context.Context, rep repo.Repository) error {
...
+func advancedCommand() {
+	if os.Getenv("KOPIA_ADVANCED_COMMANDS") != "enabled" {
+		//nolint:errcheck
+		errorColor.Printf(`
+This command is meant for advanced users only and could be dangerous or lead to
+repository corruption when used improperly.
+
+Most users of Kopia should not have the need to run this command and instead should rely on
+periodic repository maintenance. See https://kopia.io/docs/maintenance/ for more information.
+
```
julio-lopez, line 196:
```suggestion
This command could be dangerous or lead to repository corruption when used improperly.

Running this command is not needed for using Kopia. Instead, most users should rely on periodic repository maintenance. See https://kopia.io/docs/maintenance/ for more information.
```

julio-lopez (approved):

PR #620, opened 2020-09-17
fuse: changed file read implementation from ReadAll() to a Handle to avoid OOMing

jkowalski:
We don't have automated tests for this but I verified this by
restoring 13GB file over fuse and memory usage never exceeded 400MB.

### DIFF
--- a/internal/fusemount/fusefs.go
+++ b/internal/fusemount/fusefs.go
@@ -6,8 +6,10 @@
 package fusemount
 
 import (
+	"io"
 	"io/ioutil"
 	"os"
+	"sync"
 
 	"bazil.org/fuse"
 	fusefs "bazil.org/fuse/fs"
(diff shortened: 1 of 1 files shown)


internal/fusemount/fusefs.go
```diff
@@ -36,6 +38,51 @@ type fuseFileNode struct {
 	fuseNode
 }
 
+var _ fusefs.NodeOpener = (*fuseFileNode)(nil)
```
julio-lopez, line 41: I'm assuming that when `NodeOpener` is implemented, then `Open` below is called and then a `Handle` is used to perform the reads, instead of `ReadAll`. Is that the case?

julio-lopez (approved):

julio-lopez:
No changes to docs, merging this.

PR #621, opened 2020-09-17
restore: support for symlinks (experimental)

### LINKED ISSUE #474: Feature Request: Support Symbolic Links During Restore
Hi, during a `kopia restore` I noticed there were lots of warnings about not creating symlinks. Here is an example:

```
23:20:39.532 [kopia/localfs] Not creating symlink "/home/rhd/src/spack/var/spack/environments/myenv/.spack-env/view/share/terminfo/z/zen30" from &{{0xc0002f2a80 0xc000180360}}                                                                                                                                                                 
```

It would be nice if kopia could also restore the symlink.

Thanks!

jkowalski:
Fixes #474


snapshot/restore/local_Fs_output_unix.go
```diff
@@ -0,0 +1,17 @@
+// +build linux darwin
```
julio-lopez, line 1: Should the file be named `local_fs_output_unix.go` ?

snapshot/restore/local_fs_output.go
```diff
@@ -73,12 +73,28 @@ func (o *FilesystemOutput) WriteFile(ctx context.Context, relativePath string, f
...
+	targetPath, err := e.Readlink(ctx)
+	if err != nil {
+		return errors.Wrap(err, "error reading link target")
+	}
+
+	log(ctx).Infof("CreateSymlink %v => %v", relativePath, targetPath)
+
+	path := filepath.Join(o.TargetPath, filepath.FromSlash(relativePath))
+
+	if err := os.Symlink(targetPath, path); err != nil {
```
julio-lopez, line 85: The only concern here is whether there should be any special handling for symlinks that point (i.e., have as target) something that is outside the tree of the snapshot. Sometimes, that's OK; sometimes it is desirable to prevent that by say, ignoring those symlinks.

julio-lopez (approved):
@jkowalski inline comments. PTAL at the one about the file name before merging.

snapshot/restore/local_Fs_output_unix.go
jkowalski: thanks!

snapshot/restore/local_fs_output.go
jkowalski: we never really look at those targets and never follow them. Perhaps this is something to change in the future.

pavannd1 (approved):
LGTM

snapshot/restore/local_fs_output.go
julio-lopez: Correct. Kopia does not look or follow those targets.

The comment is about cases where the user does not want to restore symlinks that point to something outside the tree. This is somewhat related to the functionality provided by the `--copy-unsafe-links`, `--safe-links` and `--munge-links` flags in `rsync`.

PR #639, opened 2020-09-23
Update dependencies

julio-lopez:
Combo dependency updates, since many had fallen behind.


jkowalski (approved):

go.mod
```diff
@@ -4,59 +4,53 @@ go 1.15
...
+	github.com/klauspost/pgzip v1.2.5
 	github.com/kylelemons/godebug v1.1.0
-	github.com/mattn/go-colorable v0.1.6 // indirect
-	github.com/mattn/go-ieproxy v0.0.1 // indirect
-	github.com/minio/minio v0.0.0-20200521180750-fade0562443f
-	github.com/minio/minio-go/v6 v6.0.55
-	github.com/natefinch/atomic v0.0.0-20150920032501-a62ce929ffcc
+	github.com/mattn/go-colorable v0.1.7 // indirect
+	github.com/minio/minio v0.0.0-20200922223427-0537a21b7948
+	github.com/minio/minio-go/v6 v6.0.58-0.20200612001654-a57fec8037ec
```
jkowalski, line 30: can we upgrade to v7?

also, can we get rid of minio/minio which is only used in tests.

julio-lopez: `go mod` will include dependencies even if they are only used in test or in a different platform.
I can push a separate patch to upgrade to v7, in case there are any issues with it to at least upgrade the rest.

jkowalski: The test depends on this "github.com/minio/minio/pkg/madmin" - I was wondering whether there's equivalent in v7 package.

PR #644, opened 2020-09-25
restore: improved user experience

### LINKED ISSUE #634: Restore single file error
Hi everyone. Let's say that i'm in this case number 1:

1. Source directory is this:

sourcepath
  file1.txt
  file2.txt
  file3.txt

I have successfully created a snapshot of this "sourcepath" with this command:

`kopia snapshot create sourcepath`

When i need to restore this snapshot i need to use this command

`kopia restore id targetpath` 

becouse if i use

`kopia snapshot restore id targetpath`

it returns an error. 

Case number 2:

If now i try to create a snapshot of a single file instead of a dir with this command

`kopia snapshot create sourcepath\file1.txt`

i always get an error if i want to restore the file. Any clue?

jkowalski:
* 'snapshot restore' is now the same as 'restore' and both will
  support restoring by manifest ID, root ID or root ID + subdirectory

* added support for restoring individual files

Fixes #634


cli/objref.go
```diff
@@ -67,3 +69,113 @@ func parseNestedObjectID(ctx context.Context, startingDir fs.Entry, parts []stri
...
+	}
+
+	// all snapshots have consistent metadata, pick any.
+	if areSnapshotsConsistent(mans) {
+		return mans[0], nil
+	}
+
+	// at this point we found multiple snapshots with the same root ID which don't agree on other
+	// metadata (the attributes, ACLs, ownership, etc. of the root)
+	if os.Getenv("KOPIA_REQUIRE_UNAMBIGUOUS_ROOT") != "" {
```
julio-lopez, line 118: * Should this be the default instead of "opt-in" ?
* It may make sense to explicitly pass this as a flag to the command and then as an option to the function instead of relying on relatively hidden environment variables. WDYT?

snapshot/snapshotfs/repofs.go
```diff
@@ -223,6 +223,54 @@ func SnapshotRoot(rep repo.Repository, man *snapshot.Manifest) (fs.Entry, error)
...
+	return f
+}
+
+func isPossiblyDirectoryID(oid object.ID) bool {
+	if ndx, ok := oid.IndexObjectID(); ok {
+		return isPossiblyDirectoryID(ndx)
+	}
+
+	if cid, _, ok := oid.ContentID(); ok {
+		return cid.Prefix() == "k"
```
julio-lopez, line 268: nit: Should the _well-known prefixes_ be `const`?

cli/objref.go
```diff
@@ -67,3 +69,113 @@ func parseNestedObjectID(ctx context.Context, startingDir fs.Entry, parts []stri
 
 	return e.(object.HasObjectID).ObjectID(), nil
 }
+
+func findSnapshotsByRootObjectID(ctx context.Context, rep repo.Repository, rootID object.ID) ([]*snapshot.Manifest, error) {
```
julio-lopez, line 73: nit: Does it make sense to move these helpers to the `snapshot` package?
And then export `FindSnapshotsByRootObjectID`, `FindSnapshotByRootObjectIDOrManifestID` and `FilesystemEntryFromID`
Thoughts?

cli/objref.go
jkowalski: Moved to be a CLI parameter, defaulting to false. Let me know if you have strong conviction about opt-out vs opt-in.

snapshot/snapshotfs/repofs.go
jkowalski: done

cli/objref.go
jkowalski: done, this actually helped clean up the code in a whole bunch of places.

cli/objref.go
julio-lopez: The question is what default behavior is both safer and less surprising for users.
* If the default is out-out, then it could fail with an error when there are ambiguous roots and log information on either how to allow restores of ambiguous roots, and which snapshots could be used for the restore instead of a root id.
* If the default is opt-in, then a warning could be logged when there are ambiguous roots along with info on (a) how to prevent ambiguous roots from being restored and (b) the snapshots that could be used to restore the root with different permissions.

julio-lopez (approved):
LGTM  ✅
🚢it :shipit:

Regarding the default, we can change in a separate PR if we believe it should be changed.

julio-lopez:
I did not investigate what the Windows failure was.

PR #652, opened 2020-09-30
Makefile: plug in gotestsum for better test output

jkowalski:


tools/tools.mk
```diff
@@ -202,6 +203,14 @@ endif
 
 endif
 
+gotestsum=$(TOOLS_DIR)/bin/gotestsum
+
+$(gotestsum): export GO111MODULE=off
+$(gotestsum): export GOPATH=$(TOOLS_DIR)
+$(gotestsum):
+	$(mkdir) $(TOOLS_DIR)
```
julio-lopez, line 211: This erred out on Win.

julio-lopez (approved):
Is there a way of making the default output terse, at least for the non-failure cases?

tools/tools.mk
jkowalski: yeah, fixed

PR #659, opened 2020-10-03
Fixed few minor data races

jkowalski:


snapshot/snapshotfs/upload.go
```diff
@@ -1000,6 +1007,9 @@ func (u *Uploader) Upload(
 		return nil, err
 	}
 
+	cancelScan()
```
julio-lopez, line 1010: `cancelScan()` is potentially called twice. Does not seem to be an issue, since IIUC it is idempotent.

julio-lopez (approved):
LG. `TestRestore` E2E failed on arm, not sure why.

PR #663, opened 2020-10-04
cli: 'snapshot create' switched all stderr output to use logger

jkowalski:
This allows 'kopia snap create --all --log-level=warning --no-progress'
to produce no console output unless an error is encountered.

### DIFF
--- a/cli/command_snapshot_create.go
+++ b/cli/command_snapshot_create.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"context"
-	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -67,7 +66,7 @@ func runSnapshotCommand(ctx context.Context, rep repo.Repository) error {
 
 	for _, snapshotDir := range sources {
 		if u.IsCanceled() {
-			printStderr("Upload canceled\n")
+			log(ctx).Infof("Upload canceled")
 			break
 		}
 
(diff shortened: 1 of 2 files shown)


redgoat650 (approved):

julio-lopez (approved):

cli/command_snapshot_create.go
```diff
@@ -215,11 +214,11 @@ func snapshotSingleSource(ctx context.Context, rep repo.Repository, u *snapshotf
 
 	if ds := manifest.RootEntry.DirSummary; ds != nil {
 		if ds.NumFailed > 0 {
-			errorColor.Fprintf(os.Stderr, "\nIgnored %v errors while snapshotting.", ds.NumFailed) //nolint:errcheck
+			log(ctx).Warningf("Ignored %v errors while snapshotting %v.", ds.NumFailed, sourceInfo)
 		}
 	}
 
-	printStderr("\nCreated%v snapshot with root %v and ID %v in %v\n", maybePartial, manifest.RootObjectID(), snapID, clock.Since(t0).Truncate(time.Second))
+	log(ctx).Infof("Created%v snapshot with root %v and ID %v in %v", maybePartial, manifest.RootObjectID(), snapID, clock.Since(t0).Truncate(time.Second))
```
julio-lopez, line 221: Given that this is "output" info about the operation, this one in particular should go to `stdout`. In order to completely elide it from the output we could add a `--quiet` or `--silent` flag. At some point, when structured output (JSON) is supported, the output of this operation could be something like:

```json
{
  "snapshot_id": "<snap id>"
}
```

or maybe even the entire snapshot manifest could be in the output.

PR #666, opened 2020-10-05
cli: removed confusing default on 'snapshot list' that only lists last 100 snapshots per source

jkowalski:


julio-lopez (approved):
This should not cause problems. @pavannd1 @redgoat650 can you confirm?

PR #676, opened 2020-10-10
Added support for scanning only one filesystem via files policy

eras:
So this is my very first Go development ever.. :) And as this change appeared to span more files than I originally expected, I suggest actually reading if I've done something silly there. Or possibly proposing another way of doing this if a better one exists.

I tested a couple scenarios, seemed to work as expected! Sadly I didn't find a test system for ignore scenarios, soo I didn't add any automatic tests for this. Also I did not test the HTML ui, I only blindly did the change that seemed like the same thing with another boolean policy.

The new files policy `oneFileSystem` ignores files that are mounted to other filesystems similarly to `tar`'s `--one-file-system` switch. For example, if this is enabled, backing up / should now automatically ignore `/dev`, `/proc`, etc, so the directory entries themselves don't appear in the backup. The value of the policy is `false` by default.

This is implemented by adding a non-windows-field `Device` (of type `DeviceInfo`, reflecting the implementation of `Owner`) to the `Entry` interface. `DeviceInfo` holds the `dev` and `rdev` acquired with stat (same way as with `Owner`), but in addition to that it also holds the same values for the parent directory. It would seem that doing this in some other way, ie. in `ReadDir`, would require modifying the `ReadDir` interface which seems a too large modification for a feature this small.

This change introduces a duplication of `stat` call to the files, as the `Owner` feature already does a separate call. I doubt the performance implications are noticeable, though with some refactoring both `Owner` and `Device` fields could be filled in in one go.

Filling in the field has been placed in `fs/localfs/localfs.go` where `entryFromChildFileInfo` has acquired a third parameter giving the the parent entry. From that information the `Device` of the parent is retrieved, to be passed off to `platformSpecificDeviceInfo` which does the rest of the paperwork. Other fs implementations just put in the default values.

Finally the actual check of the condition is in `ignorefs.go` function `shouldIncludeByDevice` which is analoguous to the other similarly named functions.

### DIFF
--- a/fs/ignorefs/ignorefs.go
+++ b/fs/ignorefs/ignorefs.go
@@ -28,6 +28,8 @@ type ignoreContext struct {
 	dotIgnoreFiles []string         // which files to look for more ignore rules
 	matchers       []ignore.Matcher // current set of rules to ignore files
 	maxFileSize    int64            // maximum size of file allowed
+
+	oneFileSystem bool // should we enter other mounted filesystems
 }
 
 func (c *ignoreContext) shouldIncludeByName(path string, e fs.Entry) bool {
@@ -46,6 +48,14 @@ func (c *ignoreContext) shouldIncludeByName(path string, e fs.Entry) bool {
 	}
 
 	return c.parent.shouldIncludeByName(path, e)
+}
+
+func (c *ignoreContext) shouldIncludeByDevice(e fs.Entry, parent *ignoreDirectory) bool {
+	if !c.oneFileSystem {
+		return true
+	}
+
+	return e.Device().Dev == parent.Device().Dev
 }
 
 type ignoreDirectory struct {
@@ -132,6 +142,10 @@ func (d *ignoreDirectory) Readdir(ctx context.Context) (fs.Entries, error) {
 			continue
 		}
 
+		if !thisContext.shouldIncludeByDevice(e, d) {
+			continue
+		}
+
 		if dir, ok := e.(fs.Directory); ok {
 			e = &ignoreDirectory{d.relativePath + "/" + e.Name(), thisContext, d.policyTree.Child(e.Name()), dir}
 		}
@@ -168,6 +182,7 @@ func (d *ignoreDirectory) buildContext(ctx context.Context, entries fs.Entries) 
 		onIgnore:       d.parentContext.onIgnore,
 		dotIgnoreFiles: effectiveDotIgnoreFiles,
 		maxFileSize:    d.parentContext.maxFileSize,
+		oneFileSystem:  d.parentContext.oneFileSystem,
 	}
 
 	if pol != nil {
@@ -196,6 +211,8 @@ func (c *ignoreContext) overrideFromPolicy(fp *policy.FilesPolicy, dirPath strin
 	if fp.MaxFileSize != 0 {
 		c.maxFileSize = fp.MaxFileSize
 	}
+
+	c.oneFileSystem = fp.OneFileSystemOrDefault(false)
 
 	// append policy-level rules
 	for _, rule := range fp.IgnoreRules {
--- a/fs/localfs/local_fs_nonwindows.go
+++ b/fs/localfs/local_fs_nonwindows.go
@@ -18,3 +18,14 @@ func platformSpecificOwnerInfo(fi os.FileInfo) fs.OwnerInfo {
 
 	return oi
 }
+
+func platformSpecificDeviceInfo(fi os.FileInfo) fs.DeviceInfo {
+	var oi fs.DeviceInfo
+	if stat, ok := fi.Sys().(*syscall.Stat_t); ok {
+		// not making a separate type for 32-bit platforms here..
+		oi.Dev = platformSpecificWidenDev(stat.Dev)
+		oi.Rdev = platformSpecificWidenDev(stat.Rdev)
+	}
+
+	return oi
+}
--- a/fs/localfs/local_fs_windows.go
+++ b/fs/localfs/local_fs_windows.go
@@ -9,3 +9,7 @@ import (
 func platformSpecificOwnerInfo(fi os.FileInfo) fs.OwnerInfo {
 	return fs.OwnerInfo{}
 }
+
+func platformSpecificDeviceInfo(fi os.FileInfo) fs.DeviceInfo {
+	return fs.DeviceInfo{}
+}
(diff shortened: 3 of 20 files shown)


eras:
So I found `ignorefs_test.go`, I guess I should take a look how to test this with that.

jkowalski:
This is great. I see you have some problems compiling for other platforms. Try `make gorelease` locally to speed up the process. Also `make` will run most relevant tests and linter.

eras:
So it seems my local `make` did not pass even in the `master` branch, with errors like

```
=== Failed
=== FAIL: repo/blob/rclone TestRCloneStorage (0.81s)
    tlsutil.go:31: [tls] generating new TLS certificate
    tlsutil.go:61: [tls] adding alternative IP to certificate: 127.0.0.1
    rclone_storage.go:198: [rclone] starting rclone [serve webdav /tmp/rclonetest356534592 --addr 127.0.0.1:0 --cert /tmp/kopia-rclone023562143/webdav.cert --key /tmp/kopia-rclone023562143/webdav.key --htpasswd /tmp/kopia-rclone023562143/htpasswd]
    rclone_storage.go:210: [rclone] detected webdav address: https://127.0.0.1:37451/
    rclone_storage_test.go:59: unexpected error: unexpected error when downloading non-existent blob: ReadStream 5fa/2bb/4b-2385-43de-bfb4-d1738b6b08c9.f: Authorize https://127.0.0.1:37451/: 401
    rclone_storage.go:62: [rclone] killing rclone
```

so I guess my test environment is not complete.

However, I probably fixed the issue with the mock, but didn't yet add any tests for this new functionality.

jkowalski:
BTW. If you get stuck and would like to chat, I'm available on Slack too.

eras:
Well, this nolintlint seems like a difficult obstacle, not to mention it seems a bit silly :P, what might be the solution here?

So the key problem here is that the `Dev`/`Rdev` fields are uint32 in Arm compilation and 64-bit in x86_64 and for ease-of-use the code uses `uint64` for storing them. For assigning `uint32` to `uint64` I need a cast, but this cast is ineffective in x86_64. So I disable that lint checker with `//nolint:unconvert` for the assignment BUT then this disabling itself is ineffective in the 32-bit case, therefore triggering an unlintlint linter check which cannot be (?) disabled on a line-by-line basis. Therefore I don't know a way to express this thing that works on both cases.

I would propose enabling `allow-unused` for nolintlint, because it can't know that it's unused in all compilations.

Atlernative solutions:
- make a separate `local_fs_nonwindows.go` for platforms that have different-sized Dev/Rdev (how to detect this in a robust fashion during compile time?)
- actually that's all I came up with with my limited Go understanding :).

I don't have a test yet and I'll squash those commits later.

eras:
Cleaned up the branch.

Still thinking two things:
1) This would be quite a lot simpler if instead of storing Dev/Rdev it would just check, in the scanning phase, if the parent directory is the same as the file itself and store just that one flag. But it would be less generally applicable. But I doubt there are many interesting more generally applicable problems to be solved here :).
2) The test is pretty simple.. But, fundamentally, so is the feature.

fs/ignorefs/ignorefs.go
```diff
@@ -132,6 +147,10 @@ func (d *ignoreDirectory) Readdir(ctx context.Context) (fs.Entries, error) {
 			continue
 		}
 
+		if !thisContext.shouldIncludeByDevice(e) {
```
jkowalski, line 150: This looks very good, however there's something weird about having to have Parent* together with current device information., leads to some awkward construction of those nodes.

Could we simply compare `d.Device()` with `e.Device()` here? I believe this would get rid of the need to expose Parent*, because `d` is the parent.

fs/localfs/local_fs_nonwindows.go
```diff
@@ -18,3 +18,14 @@ func platformSpecificOwnerInfo(fi os.FileInfo) fs.OwnerInfo {
 
 	return oi
 }
+
+func platformSpecificDeviceInfo(fi os.FileInfo, parentDevice *fs.DeviceInfo) fs.DeviceInfo {
```
jkowalski, line 22: we don't need parent device here, can you remove it? I believe you'll be able to revert few other changes that were required to pass this.

jkowalski:
I only have minor comments, LGTM from my part. I'm going to still add Julio to see if he has additional comments and we can merge it.

jkowalski (approved):

eras:
Noticed a copy-typo in `cli/command_policy_set.go`, fixed that as well.

I think I'm happy with the changes currently.

I can rebase this on new master if needed. I tested that it rebases cleanly.

fs/localfs/local_fs_nonwindows.go
```diff
@@ -18,3 +18,14 @@ func platformSpecificOwnerInfo(fi os.FileInfo) fs.OwnerInfo {
 
 	return oi
 }
+
+func platformSpecificDeviceInfo(fi os.FileInfo) fs.DeviceInfo {
+	var oi fs.DeviceInfo
+	if stat, ok := fi.Sys().(*syscall.Stat_t); ok {
```
julio-lopez, line 24: @eras AFAICT, this is NOT incurring a separate `stat` syscall, and instead simply reusing and casting the result of a previous call, so we may be OK here.

fs/localfs/local_fs_windows.go
```diff
@@ -9,3 +9,7 @@ import (
 func platformSpecificOwnerInfo(fi os.FileInfo) fs.OwnerInfo {
 	return fs.OwnerInfo{}
 }
+
+func platformSpecificDeviceInfo(fi os.FileInfo) fs.DeviceInfo {
```
julio-lopez, line 13: Windows allows composing arbitrary FS trees through a couple of mechanisms. I don't have much experience here, but I'm assuming that is a "rare" use case, so it is OK to not support it for now.

Refs:
* https://docs.microsoft.com/en-us/windows/win32/fileio/volume-mount-points
* https://docs.microsoft.com/en-us/windows/win32/fileio/reparse-points-and-file-operations

julio-lopez (approved):

fs/localfs/local_fs_windows.go
eras: Well, this does give me an idea of a missing feature: the one-file-system-policy should be refused in unsupported platforms with an error, or at the very least it should prompt a warning; possibly when setting the policy or when applying it.

The support could be introduced when Kopia learns more Windows tricks, such as using shadow volumes.

PR #691, opened 2020-10-25
FUSE: Passthrough options to allow other-users access mount point

ptabor:
Adds:
  - --fuse-allow-other
  - --fuse-allow-non-empty-mount
options to the ./kopia mount commands. 

The motivation for this change is to integrate kopia history (all backups) into list of exposed content from QNAP NAS
(e.g. by SMB). Without this change (--fuse-allow-other), the mount-point can be accessed only by user that executed the command (admin@).

### DIFF
--- a/internal/mount/mount_fuse.go
+++ b/internal/mount/mount_fuse.go
@@ -23,8 +23,27 @@ func (r *root) Root() (fusefs.Node, error) {
 	return r.Node, nil
 }
 
+func (mo *Options) toFuseMountOptions() []fuse.MountOption {
+	options := []fuse.MountOption{
+		fuse.ReadOnly(),
+		fuse.FSName("kopia"),
+		fuse.Subtype("kopia"),
+		fuse.VolumeName("Kopia"),
+	}
+
+	if mo.FuseAllowOther {
+		options = append(options, fuse.AllowOther())
+	}
+
+	if mo.FuseAllowNonEmptyMount {
+		options = append(options, fuse.AllowNonEmptyMount())
+	}
+
+	return options
+}
+
 // Directory mounts the given directory using FUSE.
-func Directory(ctx context.Context, entry fs.Directory, mountPoint string) (Controller, error) {
+func Directory(ctx context.Context, entry fs.Directory, mountPoint string, mountOptions Options) (Controller, error) {
 	isTempDir := false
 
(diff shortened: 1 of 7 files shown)


jkowalski:
Thanks for the contribution, looks like there's some problem with Windows build, which uses different mount mechanism (non-FUSE).

ptabor:
Fixed.

jkowalski:
I'm pretty sure go.mod changes are not required, otherwise LGTM.

jkowalski (approved):

ptabor:
Thank you for merging and I'm sorry for not cleaning up the modules. I left the work for the weekend. 

In etcd we use following piece of code to make sure modules are minimal, as part of checkers: 

```
function mod_tidy_for_module {
  # Watch for upstream solution: https://github.com/golang/go/issues/27005
  local tmpModDir
  tmpModDir=$(mktemp -d --suffix "etcd-mod")
  run cp "./go.mod" "./go.sum" "${tmpModDir}" || return 2

  local tmpFileGoModInSync=0
  diff -C 5 "${tmpModDir}/go.mod" "./go.mod" || tmpFileGoModInSync="$?"

  # Bring back initial state
  mv "${tmpModDir}/go.mod" "./go.mod"
  mv "${tmpModDir}/go.sum" "./go.sum"

  if [ "${tmpFileGoModInSync}" -ne 0 ]; then
    echo "FAIL: ${PWD}/go.mod is not in sync with 'go mod tidy'"
    return 255
  fi
}
```

internal/mount/mount_fuse.go
```diff
@@ -40,13 +59,14 @@ func Directory(ctx context.Context, entry fs.Directory, mountPoint string) (Cont
...
-	fuseConnection, err := fuse.Mount(
-		mountPoint,
+	options := append(
+		mountOptions.toFuseMountOptions(),
 		fuse.ReadOnly(),
 		fuse.FSName("kopia"),
 		fuse.Subtype("kopia"),
-		fuse.VolumeName("Kopia"),
-	)
+		fuse.VolumeName("Kopia"))
```
julio-lopez, line 67: Aren't all these options included in `toFuseMountOptions()`? Should this be simplified to simply use the options returned by `mountOptions.toFuseMountOptions()` ?

PR #703, opened 2020-12-01
cli: added support for copying (or moving) snapshot history

jkowalski:
Both source and destination can be specified using `user@host`, `@host` or `user@host:/path` where destination values override the corresponding parts of the source, so both targeted and mass copying is supported.

Supported combinations are:

```
Source:             Destination         Behavior
---------------------------------------------------
@host1              @host2              copy snapshots from all users of host1
user1@host1         @host2              copy all snapshots to user1@host2
user1@host1         user2@host2         copy all snapshots to user2@host2
user1@host1:/path1  @host2              copy to user1@host2:/path1
user1@host1:/path1  user2@host2         copy to user2@host2:/path1
user1@host1:/path1  user2@host2:/path2  copy snapshots from single path
```

When `--move` is specified, the matching source snapshots are also deleted.

### DIFF
--- a/cli/command_snapshot_copy.go
+++ b/cli/command_snapshot_copy.go
@@ -1,0 +1,190 @@
+package cli
+
+import (
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/repo"
+	"github.com/kopia/kopia/snapshot"
+)
+
+var (
+	snapshotCopyCommand     = snapshotCommands.Command("copy", "Copy snapshot history from another user or host").Alias("cp")
+	snapshotCopyDryRun      = snapshotCopyCommand.Flag("dry-run", "Do not actually copy snapshots, only print what would happen").Short('n').Bool()
+	snapshotCopyMove        = snapshotCopyCommand.Flag("move", "Delete original copied snapshots").Bool()
+	snapshotCopySource      = snapshotCopyCommand.Arg("source", "Source (user@host or user@host:path)").Required().String()
+	snapshotCopyDestination = snapshotCopyCommand.Arg("destination", "Destination (defaults to current user@host)").String()
+)
+
+// runSnapshotCopyCommand copies snapshot manifests of the specified source
+// to the respective destination. This is typically used when renaming a host,
+// switching username or moving directory around to maintain snapshot history.
+//
+// Both source and destination can be specified using user@host, @host or user@host:/path
+// where destination values override the corresponding parts of the source, so both targeted
+// and mass copying is supported.
+//
+// Source:             Destination         Behavior
+// ---------------------------------------------------
+// @host1              @host2              copy snapshots from all users of host1
+// @host1              user2@host2         (disallowed as it would potentially collapse users)
+// @host1              user2@host2:/path2  (disallowed as it would potentially collapse paths)
+//
+// user1@host1         @host2              copy all snapshots to user1@host2
+// user1@host1         user2@host2         copy all snapshots to user2@host2
+// user1@host1         user2@host2:/path2  (disallowed as it would potentially collapse paths)
+//
+// user1@host1:/path1  @host2              copy to user1@host2:/path1
+// user1@host1:/path1  user2@host2         copy to user2@host2:/path1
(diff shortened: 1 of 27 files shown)


go.mod
```diff
@@ -8,7 +8,8 @@ require (
 	contrib.go.opencensus.io/exporter/prometheus v0.2.0
 	github.com/Azure/azure-pipeline-go v0.2.3 // indirect
 	github.com/Azure/azure-storage-blob-go v0.10.0
-	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
+	github.com/alecthomas/kingpin v0.0.0-20200323085623-b6657d9477a6
```
julio-lopez, line 11: Do we need a comment here indicating that this is pulling `master`, which is newer `.v2`?
```suggestion
	// kingpin does not follow Go Modules versioning mechanism for versions > 1.
	// This is pulling from master which is newer than > v2.2.6
	github.com/alecthomas/kingpin v0.0.0-20200323085623-b6657d9477a6
```

cli/command_snapshot_copy.go
```diff
@@ -0,0 +1,190 @@
...
+// @host1              user2@host2         (disallowed as it would potentially collapse users)
+// @host1              user2@host2:/path2  (disallowed as it would potentially collapse paths)
+//
+// user1@host1         @host2              copy all snapshots to user1@host2
+// user1@host1         user2@host2         copy all snapshots to user2@host2
+// user1@host1         user2@host2:/path2  (disallowed as it would potentially collapse paths)
+//
+// user1@host1:/path1  @host2              copy to user1@host2:/path1
+// user1@host1:/path1  user2@host2         copy to user2@host2:/path1
+// user1@host1:/path1  user2@host2:/path2  copy snapshots from single path.
```
julio-lopez, line 41: It'd be good to put this comment in the command documentation instead. That way it is available as part of the help for the command.

cli/command_snapshot_copy.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/repo"
+	"github.com/kopia/kopia/snapshot"
+)
+
+var (
+	snapshotCopyCommand     = snapshotCommands.Command("copy", "Copy snapshot history from another user or host").Alias("cp")
+	snapshotCopyDryRun      = snapshotCopyCommand.Flag("dry-run", "Do not actually copy snapshots, only print what would happen").Short('n').Bool()
+	snapshotCopyMove        = snapshotCopyCommand.Flag("move", "Delete original copied snapshots").Bool()
```
julio-lopez, line 16: UI: Should `move` be a flag or a separate command that shares the flags and the implementation?

julio-lopez (commented):
This is super useful!

cli/command_snapshot_copy.go
jkowalski: done

cli/command_snapshot_copy.go
jkowalski: done

cli/command_snapshot_copy_move_history.go
```diff
@@ -0,0 +1,232 @@
...
+func snapshotExists(snaps []*snapshot.Manifest, src snapshot.SourceInfo, startTime time.Time) bool {
+	for _, s := range snaps {
+		if !startTime.Equal(s.StartTime) {
+			continue
+		}
+
+		if src != s.Source {
+			continue
+		}
+
```
julio-lopez, line 195: Should this check other fields of the snapshot manifest, in particular that the `RootEntry` fields match?

It'd be good to avoid deleting the source manifest when moving a snapshot for which there is a target manifest with same `startTime` but pointing to a different root. Deleting the source manifest in this case may result in loosing the only reference to the source's root.

It'd be good to simply say the snapshot does not exist when the root entry does not match that of the source.
Not sure whether having two snapshots at the destination with the same `startTime` may cause issues when computing the retentions.

jkowalski: done

julio-lopez (approved):

Issue #709, opened 2020-12-05
[Feature] Add support for executing scripts to produce data to be snapshotted via stdout.


julio-lopez:
See https://github.com/kopia/kopia/issues/683#issuecomment-739078359 for details.

This is a subset of #400

jkowalski:
Also https://kopia.discourse.group/t/design-additional-features-for-snapshot-creation/197/5

mschuwalow:
Are you still interested in this feature / are contributions welcome?

jkowalski:
Absolutely

Issue #683, opened 2020-10-15
[Feature] stdin support


bbigras:
Use case:
**A hosting provider only allows me to make backup as `.tar.gz` files. I would like to feed those directly to Kopia.**

With Restic I can do `gunzip -c backup.tar.gz | restic -r ~/repo backup --stdin` which is pretty simple.

> We should start by filing an issue to discuss the exact usage, since it cannot be that simple unfortunately (we need to keep track of these snapshots somehow and in Kopia snapshots are identified by user@host:path)

source: https://kopia.slack.com/archives/CH7R8L2KD/p1602654316115900

No idea if it could help but Restic allows to set a timestamp and filename like this:
`restic backup --stdin --time "2020-01-08 12:00:00" --stdin-filename backup.tar.gz`

saviodsouza:
At the moment it's not supported

There is a forum thread about the same
[Form](https://kopia.discourse.group/t/backup-standard-input/161)

eras:
Yes, this would be great!

With Ceph it would be possible to backup rbd snapshots with `rbd export images/host-root@snapshot | kopia snapshot --stdin=/rbd/images/host-root` (restic syntax seems applicable) and it would probably be just as nice as direct support for rbd.

Currently the only way to go is to have a large temporary directory—not to mention the name of directory is silly (in my case `/mnt/tmp`), because the files are only visiting the filesystem.

jkowalski:
I'm marking this as `help wanted`, because I think this would be a good small feature to implement by an external contributor - also a good learning opportunity since the feature will cut through several layers of Kopia.

julio-lopez:
An initial implementation for this feature should be relatively simple. The minor hurdles are around ensuring that the feature works well with some of the existing user expectations and the associated details of the user experience, such as the specific command line flags, their meanings, etc.

Let's simplify and clarify the _minimum_ requirements and call out or separate optional or other desired features. 

1. The input is an opaque blob, initially, assume just a stream of bytes.
2. Support for input via `stdin` from an external process or a some redirection which is external to `kopia`. That is, `kopia` is NOT launching the external process, or connecting to a socket, etc. That is out of scope. We can consider other forms of specifying the data source later.
3. A successful execution should create a _snapshot_ associated with a _kopia source_ , so other features such as restore, mount via fuse, retention policies, etc. work as expected. This probably means that the resulting object needs to have a "file name" as well.
4. (optional) the ability to specify a _"directory path"_ for the resulting file.
[...]

Issue #394, opened 2020-03-24
FR: option to run a script before and after accessing repository to create a snapshot


badrelmers:
Hi,
Is there a possibility to add an option to run a script before saving the snapshot and after saving the snapshot? 

Why I need this?
For security reasons I use a Port Knock to access my SSH server, so I need to send some packets to the server so the ssh port get open.
Now to let Kopia UI connect to my ssh automatically I need to disable the port knock completely.

Thank you.

jkowalski:
My initial instinct is that scripting at this point is probably too advanced, you can always have a script that periodically knocks your SSH, can't you? 

Having said that, I would be open to some scripting/hooks in the future, but I don't want to increase cognitive load on people now with too many advanced options just yet.

(the scripts I was thinking of would allow you to take consistent snapshot of a directory before it is uploaded or drain writes from a server for similar reason, and perhaps do some cleanup afterwards, but again - that's not something that's being worked on anytime soon)

badrelmers:
> you can always have a script (cron) that periodically knocks your SSH, can't you?

it is hard to do because I do not know when Kopia UI scheduler will run. Port Knocks generally wait few seconds to make a connection to the opened port, if not then the port is closed again.

I hope it can be done someday. thank you

jkowalski:
the SSH connection will not be totally idle between snapshots, in fact every few minutes it will need to download the index (list of `n` files) to discover if there are any new snapshots. not sure if this helps or hurts your scenario, though.

(I tried port knocking in the past but gave up exactly because of complexities like that)

badrelmers:
this is good to know, thank you.

I cannot live without the port knock (it saves me of a lot of headache), so I solve it for now with this gymnastic:
1. run an another sshd (port 922)  inside the server that listens **locally**:
`sudo /usr/sbin/sshd -d -D -p 922 -o ListenAddress=127.0.0.1 -f /etc/ssh/sshd_config_less_secure
`

2. create a permanent relay to this sshd using forward connection which open the port 19999 locally in my laptop
`ssh -N -tt -L 19999:localhost:922 -o ExitOnForwardFailure=yes -i ${ssh_key} ${ssh_user}@${ssh_IP} -p ${ssh_port}
`

3. configure Kopia UI to connect to localhost:19999 using ssh_user and ssh_key (it must be localhost , Kopia UI did not accept 127.0.0.1)


[...]

PR #722, opened 2020-12-16
Added support for actions that run before&after snapshot roots and before/after specific folders

### LINKED ISSUE #400: FR: Add configurable before-upload/after-upload scripts for directories.
Use cases:

- when backing up certain directories, we may need to instruct some daemon to flush writes and hold them in memory until directory has been saved

- another use case would be taking ZFS or BTRFS point-in-time snapshots of a directory and working off that, instead of source data that keeps changing

- custom notifications after snapshot success/failure

This could be done via policies attached to FS nodes.

jkowalski:
Fixes #400

Related: #394 #709 #683


julio-lopez (commented):
👍 LG +1

A couple of minor inline comments.

Also, a high-level question about the names / terminology?

Do we want to use the term _"hooks"_ or something else? Not saying it is wrong or suggesting something else. Merely posing the question because once this is merged, it becomes part of the API in the forms policy setting names and CLI flags, and it becomes harder to change down the road without breaking things for existing users.

cli/command_policy_set_hooks.go
```diff
@@ -0,0 +1,116 @@
...
+
+	*changeCount++
+
+	if *policySetPersistHookScript {
+		script, err := ioutil.ReadFile(value) //nolint:gosec
+		if err != nil {
+			return err
+		}
+
+		if len(script) > maxScriptLength {
```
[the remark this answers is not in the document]
jkowalski: probably not worth it, unless the concern is OOMing, which given this is CLI is not a big one.

julio-lopez (approved):

tests/end_to_end_test/snapshot_actions_test.go
```diff
@@ -219,6 +220,50 @@ func TestSnapshotActionsBeforeAfterFolder(t *testing.T) {
...
+	t.Parallel()
+
+	e := testenv.NewCLITest(t)
+
+	e.RunAndExpectSuccess(t, "repo", "create", "filesystem", "--path", e.RepoDir)
+	defer e.RunAndExpectSuccess(t, "repo", "disconnect")
+
+	var (
+		successScript      = tmpfileWithContents(t, "echo Hello world!")
+		successScript2     = tmpfileWithContents(t, "echo Hello world!")
```
julio-lopez, line 233: nit: for clarity, use the same pattern that is used for `failingScript` with `successScript2`.

That is, declare it a `string` here and set it in the `"windows"` block below.

jkowalski: done

julio-lopez (approved):

Issue #710, opened 2020-12-06
shallow restore feature [design]


rjkroege:
# TL;DR
I propose a feature to permit restores to be *shallow*: some entries in the restore are represented by placeholders instead of the full contents of the directory or file. These can be expanded to the full version on an as-needed basis. Snapshots of a shallow-restored tree would restore the complete contents (and be accessible in their entirety via `kopia mount`.)

# Why
Example use case: one's Linux computer has become a preemptible cloud node where the home directory is backed up via Kopia to cloud storage. (Perhaps because the original computer is in a lockdowned-office.) One wishes to economize and create the smallest cloud node for a given task and so creates a node whose disk is smaller than whole home directory. Use shallowrestore to fetch part of the home directory. Modify it. `cron`-invoked snaphots make sure that the *entire* home directory keeps getting preserved. Recreating the cloud node (say to get different node capabilities or because of preemption) doesn't require restoring the entire home directory: a series of shallow restores can complete much more quickly.

Can this be done without modifying Kopia? I don't think so. I explored a simple workflow of:

* restore repository in its totality
* put stuff in it
* snapshot tree

This was too slow to do while spinning up a new GCP node for my (not unreasonably-sized) home directory. And I had to allocate a disk larger than the smallest/cheapest tier.

I also tried fuse-mounting the repository and making a `unionfs` writeable fs on top. This eliminates the time to restore the repository (good) but the snapshot operation must still fetch the entire repository (slower than restoring) and Kopia cannot successfully (presumably locks or race conditions?) snapshot its own fuse mounts. After exploring the Kopia source, I thought that it would be easier (and run faster) to add a `shallowrestore` feature than fix self-snapshot.

# Proposed Command Details

## kopia shallowrestore (*new*)

```
kopia shallowrestore <source> <target-path>
```

Just like `kopia restore`, the source to be restored is specified in the form of a directory or file ID and optionally a sub-directory path.

For example:

```
kopia shallowrestore kffbb7c28ea6c34d6cbe555d1cf80faa9 d1’
kopia shallowrestore kffbb7c28ea6c34d6cbe555d1cf80faa9/subdir/subdir1 d1’
```

The first form will shallowrestore the contents of a `kffbb7c28ea6c34d6cbe555d1cf80faa9` to `d1`.
`d1` will be created by the restore command if it does not exist. The second form will
shallowrestore the contents of `subdir/subdir1` in `kffbb7c28ea6c34d6cbe555d1cf80faa9` to `d1`.

[...]

PR #725, opened 2020-12-16
Introduce a kopia shallowrestore command

rjkroege:
This CL is a prorotype for shallowrestore functionality as documented
in issue #710:

* shallowrestore restores placeholder files rather than full file
trees
* snapshot extended to create snapshots which would restore to the
full contents of the tree content referred to by the placeholder
files.

This CL remains a prototype uploaded for early feedback. Outstanding
work items are tracked in #710.


jkowalski:
I have few thoughts:

- can you split this into smaller commits for easier reviews
- there should be an opt-in flag to enable looking at xattrs when snapshotting (there could be security concerns here where somebody may plant a malicious xattr)
- we should verify that direntry in xattr was created from the current repository (ideally using some kind of HMAC we compute when writing xattr), perhaps xattr name should be tied to repository unique ID (so that two separate repositories can coexist in shallow
- we should probably verify that object in the direntry actually exists to avoid creating invalid snapshots

rjkroege:
I have cleaned up, improved and divided the patch per some earlier feedback. Please take a look.

snapshot/restore/local_fs_output.go
```diff
@@ -217,6 +223,25 @@ func (o *FilesystemOutput) createDirectory(ctx context.Context, path string) err
 		log(ctx).Debugf("Not creating already existing directory: %v", path)
 
 		return nil
+	case stat.Mode().IsRegular():
+		// TODO(rjk): This seems wrong here.
```
rjkroege, line 227: Oops. I believe that I fixed this. I'll remove `TODO` in a subsequent version.

rjkroege:
@jkowalski It's been a while since I uploaded this new version. Any comments?

jkowalski:
sorry, wasn't sure if this is ready, can you merge to the latest HEAD and re-upload?

rjkroege:
Updated to latest HEAD.

jkowalski:
i tried merging, but there are all kinds of linter errors, can you fix them and re-push

rjkroege:
@jkowalski Now this is ready for review. 😀

snapshot/manifest.go
```diff
@@ -91,6 +92,12 @@ type HasDirEntry interface {
 	DirEntry() *DirEntry
 }
 
+// HasDirEntryFromPlaceholder is implemented by objects that may have a DirEntry
```
jkowalski, line 95: can we call this `HasDirEntry` ? In principle `repofs` can also implement `HasDirEntry`, right and there are no placeholders involved.

snapshot/manifest.go
```diff
@@ -91,6 +92,12 @@ type HasDirEntry interface {
...
+// stored in the object's corresponding shallow placeholder file.
+type HasDirEntryFromPlaceholder interface {
+	DirEntryFromPlaceholder(ctx context.Context) (*DirEntry, error)
```
jkowalski, line 98: How about `DirEntry()` or perhaps `DirEntryOrNil()` ?

snapshot/restore/restore.go
```diff
@@ -66,6 +67,7 @@ type Options struct {
 	Parallel     int  `json:"parallel"`
 	Incremental  bool `json:"incremental"`
 	IgnoreErrors bool `json:"ignoreErrors"`
+	MaxDepth     int  `json:"maxdepth"`
```
jkowalski, line 70: nit: `maxDepth` 

also, perhaps call this `RestoreDirEntryAtDepth` or something similar?

I imagine we can decide to restore some files as placeholders as well (perhaps above certain sice, etc.) is that right?

snapshot/restore/restore.go
```diff
@@ -20,6 +20,7 @@ var log = logging.GetContextLoggerFunc("restore")
 type Output interface {
 	Parallelizable() bool
 	BeginDirectory(ctx context.Context, relativePath string, e fs.Directory) error
+	WriteShallowDirectory(ctx context.Context, relativePath string, e fs.Directory) error
```
jkowalski, line 23: can this be `WriteDirEntry()` and take DirEntry directly, IOW the caller would be responsible for figuring out how to get the DirEntry?

snapshot/restore/restore.go
```diff
@@ -87,8 +89,14 @@ func Entry(ctx context.Context, rep repo.Repository, output Output, rootEntry fs
 		}
 	}
 
+	// Control the depth of a restore. Default (options.MaxDepth = 0) is to restore to full depth.
+	sd := options.MaxDepth
+	if sd > 0 {
+		sd = -sd - 1
```
jkowalski, line 95: this is slightly weird logic, how about we simply default maxDepth to maxint32 or something?

snapshot/restore/restore.go
```diff
@@ -199,14 +207,22 @@ func (c *copier) copyEntryInternal(ctx context.Context, e fs.Entry, targetPath s
 	}
 }
 
-func (c *copier) copyDirectory(ctx context.Context, d fs.Directory, targetPath string, onCompletion parallelwork.CallbackFunc) error {
+func (c *copier) copyDirectory(ctx context.Context, d fs.Directory, targetPath string, depth int, onCompletion parallelwork.CallbackFunc) error {
 	atomic.AddInt32(&c.stats.RestoredDirCount, 1)
 
+	if depth == -1 {
```
jkowalski, line 213: this is confusing, can we introduce a helper that says:

func (c *copier) shouldRestoreAsDirEntry(depth, ...) {
  return depth >= options.RestoreDirEntryAtDepth
}

cli/command_shallowrestore.go
```diff
@@ -0,0 +1,147 @@
...
+directory into a new, local directory named 'd1'
+
+'shallowrestore kffbb7c28ea6c34d6cbe555d1cf80faa9 d1'
+
+Similarly, the following command will shallowrestore the contents of a
+subdirectory 'subdir/subdir2' under
+'kffbb7c28ea6c34d6cbe555d1cf80faa9' into a new, local directory named
+'sd2'
+
+'shallowrestore kffbb7c28ea6c34d6cbe555d1cf80faa9/subdir1/subdir2 sd2'
```
jkowalski, line 44: thinking out loud here: 

What if we make `shallowrestore` purely about "expanding" previously written DirEntry placeholders and add flags to trigger shallow restore to main `restore` command?

the session would be something like:

```
# initial shallow restore
kopia restore kffbb7c28ea6c34d6cbe555d1cf80faa9 d1 --shallow-restore-at-depth=3 --shallow-restore-files=*.zip`

# this creates d1/somefile and removes d1/somefile.kopia-object
kopia restore-entry d1/somefile.kopia-object`

# this creates d1/somedir/ and removes d1/somedir.kopia-object
kopia restore-entry d1/somedir.kopia-object`
```

We might be able to do the reverse - snapshot a directory/file and replace with a placeholder. Wouldn't that be cool?

jkowalski (changes requested):
looks very promising, I have some suggestions for simplifying the codebase.

fs/localfs/shallowentry.go
```diff
@@ -0,0 +1,30 @@
...
+	"context"
+	"strings"
+
+	"github.com/kopia/kopia/snapshot"
+)
+
+// Reserved suffixes to identify shallow placeholders for directories and
+// files respectively.
+const (
+	SHALLOWDIRSUFFIX  = ".kopiadir"
```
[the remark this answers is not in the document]
rjkroege: It would make the code simpler to have one suffix. I note in passing that different suffixes have the benefit of making it easier to identity directories vs files in `ls` output.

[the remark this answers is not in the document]
rjkroege: Done.

jkowalski:
One more random question/idea - would it make sense to represent directories as actual directories with a hidden placeholder file IN them?

so if I restore say `Downloads` directory it will be represented by 2 things:

```
~/Downloads/
~/Downloads/.kopia-object
```

Alternatively:

```
~/Downloads.kopia-object/
~/Downloads.kopia-object/kopia-object-info.json
```

snapshot/manifest.go
rjkroege: I spent a while trying to merge `HasDirEntryFromPlaceholder` and `HasDirEntry` and decided that maintaining the separation is nicer. Summary: 

* `HasDirEntry` is satisfied by *any* `repofs` instances that have a `DirEntry`
* `HasDirEntryFromPaceholder` is satisfied by `localfs` instances and when it returns a non-nil `*DirEntry` to the the `Uploader`, it is expected that the `*DirEntry` is from the same repository as the one being uploaded to.

I can unify them but I can't see a particularly nice way to do it. Details:

* Currently only `localfs` and `ignorefs` implement `HasDirEntryFromPlaceholder`
* `ignorefs` just delegates to its contained `fs.Directory`
* (Correct) `localfs` implementers of `HasDirEntryFromPlaceholder` only return a non-nil `*DirEntry` if the placeholder exists and contains a serialized `DirEntry`
* (per your earlier suggestion) code in `upload.go` verifies that this `DirEntry` came from the repository that it's uploading to
* during snapshot migrations, the `ignorefs` instances contain `repofs` instances that don't implement `HasDirEntryFromPlaceholder` but do implement `HasDirEntry`.
* Moreover, these instances are for a *foreign* repository. So the `DirEntry` validation fails.

So, my thoughts:

* I could push the destination repo validation into `DirEntry` for `localfs` implementations but exposing the necessary state from the `*Uploader` to the `localfs` seemed like a layering violation to me
* I could do something  with reflect in `*ignoreDirectory.DirEntry` to figure out if it contains a `repofs` instance but this seems like a type unsafe work around for not having two interfaces. (I wrote this case to prove to myself that I'm understanding what's happening here.)
* I could skip the `DirEntry` repository verification in `Uploader` during a migration but this is ugly.

snapshot/manifest.go
rjkroege: So probably not `DirEntry()` pending the resolution of the comment above? 🙂

I changed `DirEntryFromPlaceholder` to `DirEntryOrNil`.

Aside: maybe `DirEntryFromPlaceholder()` would be better as part of the `fs` package? (Because it needs to be implemented by `localfs` and `ignorefs`) Except that `DirEntry` is in `snapshot` and `snapshot` imports `fs` so `fs` can't import `snapshot`. All the ways to fix this seemed like large source changes.

snapshot/restore/restore.go
rjkroege: Rename done.

Yes. There's a TODO about directly restoring files below a certain (configurable) size. I'd prefer to add this in a subsequent PR. WDYT?

snapshot/restore/restore.go
rjkroege: Done but I'll leave it to you to decide if it's actually better this way. I'm not convinced: `WriteFile` and `WirteDirEntry` are more divergent than they were before and maybe harder to follow.

snapshot/restore/restore.go
rjkroege: You're right. It was convoluted. I tried to make it better.

snapshot/restore/restore.go
rjkroege: I changed the logic. Hopefully the code is clearer. Maybe it doesn't need the helper.

cli/command_shallowrestore.go
rjkroege: > snapshot a directory/file and replace with placeholder. Wouldn't that be cool?

Yes. My original notion for this PR (before I tried to reduce it to the smallest viable subset) included a feature like that. I'd prefer to add "collapse" in a subsequent PR.

It doesn't seem too hard to let `kopia restore` do full (the default) or shallow (via a max depth flag). (I think `--shallow-restore-at-depth` is too long to type though -- maybe it could have an abbreviation too?) In which case, the current `shallowrestore` command becomes:

```
kopia restore kffbb7c28ea6c34d6cbe555d1cf80faa9 d1 --shallow-restore-at-depth=0
```

And:

```
kopia restore-entry d1/dir-or-file.kopia-object --shallow-restore-at-depth=0
```

would expand `d1/dir-or-file.kopia-object`. The way that this is currently implemented,

```
cd d1
kopia restore-entry dir-or-file.kopia-object
```

would work equivalently. (I'm not convinced that `restore-entry` is the clearest name. `reify-kopia-object` seems more precise to me but not particularly ergonomic. I'll see about providing an abbreviation.)


Later, add this command:

```
# in d1
kopia collapse-entry dir-or-file
```

It would make sure that `dir-or-file` is in the repository and then replace `dir-or-file` with the appropriate `dir-or-file.kopia-object` if it is.

rjkroege:
> One more random question/idea - would it make sense to represent directories as actual directories with a hidden placeholder file IN them?

Yes. It resolves (even better) my comment about making directories obvious. I'll do this.

cli/command_shallowrestore.go
rjkroege: Done in f48c329

snapshot/restore/restore.go
rjkroege: Done in e206514

snapshot/restore/restore.go
rjkroege: Done in e206514

snapshot/restore/restore.go
rjkroege: Done in 0a2ce50

fs/localfs/shallowentry.go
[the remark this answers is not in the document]
rjkroege: Done in d1b4e4b

rjkroege:
@jkowalski Please take another look. I've addressed (or at least quibbled 🙂) about every comment in some fashion.

jkowalski:
can you merge latest master and fix the conflict there? should be simple.

cli/command_restore.go
```diff
@@ -105,6 +114,20 @@ func addRestoreFlags(cmd *kingpin.CmdClause) {
 	cmd.Flag("ignore-permission-errors", "Ignore permission errors").BoolVar(&restoreIgnorePermissionErrors)
 	cmd.Flag("ignore-errors", "Ignore all errors").BoolVar(&restoreIgnoreErrors)
 	cmd.Flag("skip-existing", "Skip files and symlinks that exist in the output").BoolVar(&restoreIncremental)
+	cmd.Flag("shallow-restore-at-depth", "Shallow restore the directory hierarchy starting at this level (default is to deep restore the entire hierarchy.)").Short('d').IntVar(&restoreShallowAtDepth)
```
jkowalski, line 117: nit: perhaps short flag is not needed for such an advanced feature.

jkowalski: also, perhaps just `--shallow=N` ?

cli/command_restore_entry.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/fs/localfs"
+	"github.com/kopia/kopia/repo"
+	"github.com/kopia/kopia/snapshot/restore"
+	"github.com/kopia/kopia/snapshot/snapshotfs"
+)
+
+const (
+	restoreEntryCommandHelp = `Expand one or more levels of a directory or file from shallow
```
jkowalski, line 18: It appears that this is sharing a lot of restore code and flags, could this simple be one command called `restore` that also understands `kopia-entry` arguments:

```
kopia restore --shallow=1 ksomehash /some/output
```

(writes `/some/output/some-file.kopia-entry` and `/some/output/some-subdir.kopia-entry`)

```
kopia restore /some/output/some-file.kopia-entry
```
(replaces `/some/output/some-file.kopia-entry` with `/some/output/some-file`)


```
kopia restore /some/output/some-subdir.kopia-entry --shallow=3 
```
(replaces `/some/output/some-subdir.kopia-entry` with `/some/output/some-subdir/` and restores 3 more levels?)

fs/localfs/shallow_fs.go
```diff
@@ -0,0 +1,118 @@
...
+// a directory and show up nicely in colorized ls output. A placeholder
+// for a file f1 is f1.kopia-entry.
+
+func placeholderPath(path string, et snapshot.EntryType) (string, error) {
+	switch et {
+	case snapshot.EntryTypeFile:
+		return path + SHALLOWENTRYSUFFIX, nil
+	case snapshot.EntryTypeDirectory: // Directories and regular files
+		dirpath := path + SHALLOWENTRYSUFFIX
+		if err := os.MkdirAll(dirpath, os.FileMode(dIRMODE)); err != nil {
```
jkowalski, line 28: not: weird capitalization of `dIRMODE`

jkowalski: also I think all caps is discouraged in golang, I would use `ShallowEntrySuffix` and `dirMode`

fs/localfs/local_fs.go
```diff
@@ -254,6 +255,14 @@ func (fsf *filesystemFile) Open(ctx context.Context) (fs.Reader, error) {
 	return &fileWithMetadata{f}, nil
 }
 
+func (fsf *filesystemFile) DirEntryOrNil(ctx context.Context) (*snapshot.DirEntry, error) {
```
jkowalski, line 258: it would be great to not implement this on all filesystemFiles to avoid the extra cost of Lstat() as we read the placeholder files (which most of the times won't be in use). The performance penalty is likely small (due to fs cache), but could be not insignificant at scale.

I think we could detect the special extension in `Readdir()` here and only return entries with `DirEntryOrNil` for files/subdirs that have `kopia-entry` extension.

WDYT?

fs/localfs/shallow_fs.go
```diff
@@ -0,0 +1,118 @@
...
+		return "", errors.Wrapf(err, "error writing placeholder to %q", mp)
+	}
+
+	return mp, nil
+}
+
+// ReadShallowPlaceholder returns the decoded ShallowMetadata for path if it exists
+// regardless of the placeholder type.
+func ReadShallowPlaceholder(path string) (*snapshot.DirEntry, error) {
+	originalpresent := false
```
jkowalski, line 68: i'm worried this will slow down filesystem reads when no placeholders are needed, see my other comment. It would be great to avoid this code path if we can.

jkowalski (changes requested):
This is looking great, I have a concern about performance impact in `localfs` and UX (whether it makes sense to combine `restore` and `restore-entry` into one, otherwise almost ready to LGTM.

cli/command_restore.go
rjkroege: Done.

rjkroege: In 20ff248

cli/command_restore_entry.go
rjkroege: I had something more like this suggestion in 1eb6e25. I thought that your previous suggestion above for the `restore`, `restore-entry` command separation ended up making the code both nicer to read and the command structure easier to explain.

I can change this but I'm concerned that the more complicated control flow in `command_restore.go` will end up being more complicated. WDYT?

fs/localfs/shallow_fs.go
rjkroege: Fixed in 457117b3

cli/command_restore.go
rjkroege: Done

fs/localfs/shallow_fs.go
rjkroege: done in 457117b3

cli/command_restore_entry.go
rjkroege: I had previously had less readable changes to  `command_restore.go` to make it able to do regular restores and what is currently provided by `restore-entry`. After your suggestion above to separate `restore-entry` from `restore`, I thought that the changes to `command_restore.go` ended up cleaner. I am worried that merging `restore-entry` and `restore` will make:

* the `restore` command seem more complicated (and be harder to explain)
* the logic in `command_restore.go` end up more complicated and easier to break.

I will do but am not convinced that this is actually a good idea.

fs/localfs/local_fs.go
rjkroege: Fixed in 7f232537.

fs/localfs/shallow_fs.go
rjkroege: Ditto. Also addressed in 7f232537.

cli/command_restore_entry.go
rjkroege: Fixed in 96ba13e4

rjkroege:
@jkowalski I think that I've addressed all your comments from last review. I think that a test that I added is still failing on Windows. I'll fix that. Maybe you could take a look at the rest?

jkowalski:
I guess i don't fully understand the problem with long filenames, but those should work on Windows with proper prefixing (see `atomicfile.MaybePrefixLongFilenameOnWindows`).

I think we can iron out those issues, and merge this right after 0.8.0 is out, so folks can test it and we can tweak the UX based on feedback.

rjkroege:
Thanks for the comment about `MaybePrefixLongFilenameOnWindows`. I hadn't done the right thing on windows. Rebased. @jkowalski PTAL.

jkowalski:
Looks like Windows build is still failing.

rjkroege:
Yeah. I understand why. I will fix.

rjkroege:
Apparently I don't understand why. I need to find myself a windows machine and run some experiments.

jkowalski:
BTW this PR got affected by #1046 - hopefully the merge should be easy.

jkowalski:
any updates here?

rjkroege:
I got frustrated with windows and then got distracted with work and finishing my taxes. Thanks for the reminder. I'll get this updated and freshened by early next week.

rjkroege:
@jkowalski PTAL. I think that I've correctly rebased and fixed the build.

cli/command_restore.go
```diff
@@ -19,10 +23,18 @@ import (
 )
 
 const (
-	restoreCommandHelp = `Restore a directory or file from a snapshot into the specified target path.
+	restoreCommandHelp = `Restore from placeholder or snapshot.
```
jkowalski, line 26: nit: "Restore a directory or a file" no need to bring placeholders in the first line.

cli/command_restore.go
```diff
@@ -94,6 +122,7 @@ func (c *commandRestore) setup(svc appServices, parent commandParent) {
 	cmd.Flag("ignore-permission-errors", "Ignore permission errors").Default("true").BoolVar(&c.restoreIgnorePermissionErrors)
 	cmd.Flag("ignore-errors", "Ignore all errors").BoolVar(&c.restoreIgnoreErrors)
 	cmd.Flag("skip-existing", "Skip files and symlinks that exist in the output").BoolVar(&c.restoreIncremental)
+	cmd.Flag("shallow", "Shallow restore the directory hierarchy starting at this level (default is to deep restore the entire hierarchy.)").Default(strconv.FormatInt(math.MaxInt32, 10)).Int32Var(&c.restoreShallowAtDepth)
```
jkowalski, line 125: nit: don't specify `Default()` at all - instead set the `c.restoreShallowAtDepth` to `math.MaxInt32`. This won't appear in generated docs, but will still have the default you want.

cli/command_restore.go
```diff
@@ -204,22 +242,41 @@ func printRestoreStats(ctx context.Context, st restore.Stats) {
...
+	if placeholderpath := restore.PathIfPlaceholder(c.restoreSourceID); placeholderpath != "" {
+		re, err := snapshotfs.GetEntryFromPlaceholder(ctx, rep, localfs.PlaceholderFilePath(c.restoreSourceID))
+		if err != nil {
+			return errors.Wrapf(err, "unable to get filesystem entry for placeholder %q", c.restoreSourceID)
+		}
+
+		rootEntry = re
+
+		// Depth defaults to 0 when expanding a placeholder.
+		if c.restoreShallowAtDepth == math.MaxInt32 {
```
jkowalski, line 261: perhaps introduce named constant here `unlimitedDepth` ?

fs/ignorefs/ignorefs.go
```diff
@@ -128,6 +129,18 @@ func (d *ignoreDirectory) skipCacheDirectory(ctx context.Context, entries fs.Ent
...
 
+// Make sure that ignoreDirectory implements HasDirEntryFromPlaceholder.
+var _ snapshot.HasDirEntryOrNil = &ignoreDirectory{}
+
+func (d *ignoreDirectory) DirEntryOrNil(ctx context.Context) (*snapshot.DirEntry, error) {
+	if defp, ok := d.Directory.(snapshot.HasDirEntryOrNil); ok {
+		// nolint:wrapcheck
+		return defp.DirEntryOrNil(ctx)
+	}
+	// Entry implementations don't have to have
```
jkowalski, line 140: nit: unfinished comment

jkowalski (approved):
LGTM with only minor/style comments. I don't want to keep this PR open for any more than already it is, if you don't have time to fix those remaining ones, we can also merge and fix them later.

Huge thanks for doing this and keeping up with ongoing refactorings. This is going to be a very useful feature.

rjkroege:
I'll fix these issues tomorrow.

cli/command_restore.go
rjkroege: Done.

jkowalski:
Thank you for doing this!

PR #726, opened 2020-12-16
robustness: various minor cleanups

julio-lopez:
Remove unnecessary intermediate variables
Send `SIGTERM` instead of `SIGKILL` to terminate child kopia server process
Set Pdeathsig on Linux for child kopia server process
Trivial: reduce scope of `hostFioDataPathStr` variable
Trivial: rename local variable
Trivial: Use log.Fatalln instead of log + exit(1)
Improve error message in robustness test to tell apart failure cause


redgoat650 (approved):
LGTM

PR #727, opened 2020-12-17
Makefile: various tricks to make PR builds faster

jkowalski:


Makefile
```diff
@@ -97,6 +102,12 @@ html-ui-bindata-fallback: $(go_bindata)
 kopia-ui:
 	$(MAKE) -C app build-electron
 
+# build-current-os compiles a binary for the current os/arch in the same location as goreleaser
+# kopia-ui build needs this particular location to embed the correct server binary.
+# the binary does not embed actual UI.
```
julio-lopez, line 107: > `# the binary does not embed actual UI.`

😕  Not sure I understand the comment. I appears to me as if the UI is being embedded.
- The target depends on `html-ui-bindata` which generates `htmlui_bindata.go` with an `embedhtml` build tag.
- Then this builds the kopia binary and specifies the `enbedhtml` build tag.
 I'm probably missing something.

julio-lopez (approved):

PR #730, opened 2020-12-18
'kopia policy set' code cleanup

jkowalski:


julio-lopez (approved):

cli/command_policy_set.go
```diff
@@ -95,24 +96,34 @@ func setPolicyFromFlags(ctx context.Context, p *policy.Policy, changeCount *int)
 	return nil
 }
 
-func addRemoveDedupeAndSort(ctx context.Context, desc string, base, add, remove []string, changeCount *int) []string {
+func applyPolicyStringList(ctx context.Context, desc string, val *[]string, add, remove []string, clear bool, changeCount *int) {
```
julio-lopez, line 99: nit / question: what's the rationale to make `val` and in/out parameter instead of returning the new slice as before?

cli/command_policy_set.go
```diff
@@ -456,17 +185,46 @@ func applyPolicyNumber64(ctx context.Context, desc string, val *int64, str strin
...
 	return nil
 }
 
+func applyPolicyBoolPtr(ctx context.Context, desc string, val **bool, str string, changeCount *int) error {
+	if str == "" {
+		// not changed
+		return nil
+	}
+
+	if str == inheritPolicyString || str == "default" {
```
julio-lopez, line 205:
```suggestion
	if str == inheritPolicyString || str == defaultPolicyString {
```

julio-lopez (approved):
👍 LG +1

Couple of questions / minor suggestions.

julio-lopez:
@pavannd1 FYI

cli/command_policy_set.go
jkowalski: it's somewhat easier to copy-paste, nothing else.

julio-lopez: The reason for the comment is that since a new value is always returned and the original is not really modified, it makes sense to convey that in the function signature.

PR #734, opened 2020-12-20
Upgraded linter to v1.33.0

jkowalski:
This brings an interesting new linter `wrapcheck` which ensures that all errors coming from external packages are properly wrapped before returning. This will greatly improve error messages because it provides additional context that can be used to find the actual source of error.

Wrapping every error is a subtle change, to ensure that we check for errors properly there's `errorlint` linter which disallows == and != comparisons in favor of `errors.Is()` and `errors.As()`, this makes sure that even if errors are wrapped multiple times, if the inner error is one of official Kopia errors (`{content,blob,object,manifest}.Err*`) it will be correctly detected.

This was mostly mechanical change, but please try to pay attention to typos when reviewing.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -44,6 +44,9 @@ linters:
     - whitespace
     - nlreturn
     - testpackage
+    - exhaustivestruct
+    - paralleltest
+    - tparallel
 
 run:
   skip-dirs:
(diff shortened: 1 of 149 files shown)


julio-lopez (approved):
caught a couple of typos

Issue #746, opened 2020-12-24
[Robustness] Panic in long-running robustness job


julio-lopez:
Observed in a couple of tests in a long running robustness job.
* `TestOneLargeFile`
* `TestRandomizedSmall`

Storage backend: S3
Kopia version: e03971fc

```
...
Processed 58195 (19 GB) of 93354 (35.8 GB) 617 Mbit/s (53.0%) remaining 3m38s.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8eafb5]

goroutine 51 [running]:
github.com/kopia/kopia/repo/object.(*objectReader).openCurrentChunk(0xc0027a4070, 0x0, 0x0)
	/var/lib/shippable/d189bfca-b0f3-4bc5-a45c-aebbdddb854d/build/IN/kopia_repo/gitRepo/repo/object/object_reader.go:87 +0xd5
github.com/kopia/kopia/repo/object.(*objectReader).Read(0xc0027a4070, 0xc000574000, 0x8000, 0x8000, 0x1017e40, 0x7f382f21e501, 0xc000574000)
	/var/lib/shippable/d189bfca-b0f3-4bc5-a45c-aebbdddb854d/build/IN/kopia_repo/gitRepo/repo/object/object_reader.go:63 +0x16f
io.copyBuffer(0x136bca0, 0xc003be2190, 0x7f37e47856e0, 0xc000eb0320, 0xc000574000, 0x8000, 0x8000, 0x4ad205, 0xc0018ca2d0, 0x4aa4b9)
	/usr/local/go/src/io/io.go:409 +0x12c
io.Copy(...)
	/usr/local/go/src/io/io.go:368
os.genericReadFrom(0xc0020d4028, 0x7f37e47856e0, 0xc000eb0320, 0x0, 0x0, 0x0)
	/usr/local/go/src/os/file.go:159 +0x99
os.(*File).ReadFrom(0xc0020d4028, 0x7f37e47856e0, 0xc000eb0320, 0x7f37e4785720, 0xc0020d4028, 0x1)
	/usr/local/go/src/os/file.go:153 +0x217
io.copyBuffer(0x136a6c0, 0xc0020d4028, 0x7f37e47856e0, 0xc000eb0320, 0x0, 0x0, 0x0, 0xc0000b0200, 0x5, 0x11fd49a)
	/usr/local/go/src/io/io.go:395 +0x2ff
[...]

PR #747, opened 2020-12-24
fixed cases where nil was passed to errors.Wrap() causing nil to be returned

jkowalski:
This is for #746 (there is likely a deeper issue here)

### DIFF
--- a/repo/maintenance/maintenance_schedule.go
+++ b/repo/maintenance/maintenance_schedule.go
@@ -87,7 +87,7 @@ func GetSchedule(ctx context.Context, rep MaintainableRepository) (*Schedule, er
 	}
 
 	if len(v) < c.NonceSize() {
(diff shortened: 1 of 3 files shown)


repo/maintenance/maintenance_schedule.go
```diff
@@ -87,7 +87,7 @@ func GetSchedule(ctx context.Context, rep MaintainableRepository) (*Schedule, er
 	}
 
 	if len(v) < c.NonceSize() {
-		return nil, errors.Wrap(err, "invalid schedule blob")
+		return nil, errors.Errorf("invalid schedule blob")
```
julio-lopez, line 90: nit: `errors.New("invalid schedule blob")`

julio-lopez (commented):
👍 LG +1

PR #748, opened 2020-12-25
ci: store kopia logs as GitHub artifacts

jkowalski:


.gitignore
```diff
@@ -23,3 +23,4 @@ node_modules/
 .boto
 *.log
 *~
+.logs/
```
julio-lopez, line 26:
```suggestion
.logs/

```

.github/workflows/make.yml
```diff
@@ -105,3 +105,10 @@ jobs:
       run: make travis-setup
     - name: Build
       run: make travis-release
+    - name: Store Kopia Logs
+      uses: actions/upload-artifact@v2
+      if: ${{ always() }}
+      with:
+        name: kopia-logs
```
julio-lopez, line 112: Would logs be uploaded after every build / PR / merge / release?
Any chance anything could be accidentally leaked?

jkowalski: those are logs from integration tests only, we don't have sensitive stuff there, do we?

jkowalski: Also access appears to be limited to those with write access, which seems sane:

Per: https://github.com/actions/upload-artifact#where-does-the-upload-go

> This icon will only appear for users who have write permissions to the repository.

jkowalski: My bad, this appears to be permission to delete the artifacts.

Also non-logged in users don't see the artifacts at all, but I don't have any other GH user that's not in the repo to try and see if they have read access.

julio-lopez: FWIW, I don't see the icon for the artifacts.

jkowalski:
closing this for now

Issue #731, opened 2020-12-18
Design: Add support for transactional write sessions 


jkowalski:
We currently have a rare and subtle bug where if Kopia repository client sleeps for a very long time (for example due to laptop going to sleep) in the middle of a long snapshot, it is possible for it to sleep for so long that partially written `p` or `q` blobs will be removed by garbage collection on another machine performing maintenance, because corresponding `n` blobs have not been written yet.

We would want the maintenance job to ignore blobs from in-progress write sessions made by other clients. We can actually do much more and achieve transactional behavior of client writes so other clients see them in their entirety or not at all.

Write Sessions
==============

We will introduce the concept of write sessions, identified by random strings of 32 hex characters (128 bit). Write sessions will enable transactional behaviors - until a session is complete all other repository clients will ignore blobs from it.

All blobs written during a particular write session will carry session ID in their names:

* `n{hash}-{sessionID}` - index blobs
* `m{hash}-{sessionID}`- compaction blobs
* `l{hash}-{sessionID}` - cleanup blobs
* `p{random}-{sessionID}` - data blobs
* `q{random}-{sessionID}`- metadata blobs

We will introduce new type of blob, called active session keep-alive blob `a{sessionID}-{counter}`.

Repository client that wants to write data to a repository will:

1. generate random session ID

1. write blob `a{sessionID}-0` (contents are irrelevant but we may want to embed useful data about the user@host performing the writes)

1. perform writes as usual, embedding `{sessionID}` in all newly written blob names.

1. periodically (say every hour) write new session keep-alive blob with increasing counter and remove previously written keep-alive blobs (keeping last 3 blobs or so).

   * each client will also keep track of local time when the last keep-alive blob was written and if more than say `96 hours` (4 days) have passed (because the client went to sleep/hibernation and did not wake up in 4 days) any subsequent writes will fail and the entire session will be marked as failed and its data blobs (except `a`) deleted (on a best effort).

1. at the end of a write session, the client will remove all keep-alive blobs written so far thus marking session as committed.

Repository client that reads data:

1. Will periodically (every 15 minutes or so) scan for `a` blobs and build a set of in-progress sessions (by looking at `{sessionID}s` in `a` blob names) that have been recently written.

   * Session will be considered dead if the most recent `a` blob is older than `100 hours` (larger than client timeout time by a few hours)

1. All blobs from any of the active sessions will be ignored.

1. All blobs from dead sessions (including `a` blobs) will not be ignored and are subject to garbage collection.

[...]

PR #752, opened 2020-12-30
Preliminary support for sessions

jkowalski:
Sessions allow blob GC to ignore `p` and `q` blobs that are in the process of being written but don't have a corresponding index `n` blob yet.

Content manager will write `s<hash>-<session-id>` blob first before writing any `p`, `q` or `n` blobs (which will all share the same random session id). Only after successfully writing `n` blob, we will delete the `s` blob.

The reader (blob GC) will ignore `p` and `q` blobs if there's a corresponding `s` blob with the same session ID.

This is somewhat functional now, but there's no real way to control for session lifetime (it's implicit from first content write to a subsequent Flush()) so if two logical sessions share the same content manager they will be entangled (such as in `kopia server`)

Future PRs will allow creation of much longer session lifetimes for the duration of the entire upload, so that we can create per-snapshot session and make snapshots effectively atomic even when running inside `kopia server`

Related #731

### DIFF
--- a/repo/content/blob_crypto.go
+++ b/repo/content/blob_crypto.go
@@ -14,12 +14,12 @@ import (
 )
 
 func getIndexBlobIV(s blob.ID) ([]byte, error) {
-	if len(s) < 2*aes.BlockSize {
-		return nil, errors.Errorf("blob id too short: %v", s)
-	}
-
 	if p := strings.Index(string(s), "-"); p >= 0 { // nolint:gocritic
 		s = s[0:p]
+	}
+
+	if len(s) < 2*aes.BlockSize {
+		return nil, errors.Errorf("blob id too short: %v", s)
 	}
 
 	return hex.DecodeString(string(s[len(s)-(aes.BlockSize*2):]))
@@ -68,14 +68,14 @@ func decryptFullBlob(h hashing.HashFunc, enc encryption.Encryptor, payload []byt
 	return payload, nil
 }
 
-func verifyChecksum(h hashing.HashFunc, data, contentID []byte) error {
+func verifyChecksum(h hashing.HashFunc, data, iv []byte) error {
 	var hashOutput [maxHashSize]byte
 
 	expected := h(hashOutput[:0], data)
 	expected = expected[len(expected)-aes.BlockSize:]
 
-	if !bytes.HasSuffix(contentID, expected) {
-		return errors.Errorf("invalid checksum for blob %x, expected %x", contentID, expected)
+	if !bytes.HasSuffix(iv, expected) {
+		return errors.Errorf("invalid checksum for blob %x, expected %x", iv, expected)
 	}
 
 	return nil
--- a/repo/content/blob_crypto.go
+++ b/repo/content/blob_crypto.go
@@ -1,0 +1,82 @@
+package content
+
+import (
+	"bytes"
+	"crypto/aes"
+	"encoding/hex"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/encryption"
+	"github.com/kopia/kopia/repo/hashing"
+)
+
+func getIndexBlobIV(s blob.ID) ([]byte, error) {
+	if len(s) < 2*aes.BlockSize {
+		return nil, errors.Errorf("blob id too short: %v", s)
+	}
+
+	if p := strings.Index(string(s), "-"); p >= 0 { // nolint:gocritic
+		s = s[0:p]
+	}
+
+	return hex.DecodeString(string(s[len(s)-(aes.BlockSize*2):]))
+}
+
+func encryptFullBlob(h hashing.HashFunc, enc encryption.Encryptor, data []byte, prefix blob.ID, sessionID SessionID) (blob.ID, []byte, error) {
+	var hashOutput [maxHashSize]byte
+
+	hash := h(hashOutput[:0], data)
+	blobID := prefix + blob.ID(hex.EncodeToString(hash))
+
+	if sessionID != "" {
+		blobID += blob.ID("-" + sessionID)
+	}
+
+	iv, err := getIndexBlobIV(blobID)
+	if err != nil {
+		return "", nil, err
+	}
+
+	data2, err := enc.Encrypt(nil, data, iv)
+	if err != nil {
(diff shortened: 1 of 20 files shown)


repo/content/sessions.go
```diff
@@ -0,0 +1,179 @@
...
+type SessionInfo struct {
+	ID             SessionID `json:"id"`
+	StartTime      time.Time `json:"startTime"`
+	CheckpointTime time.Time `json:"checkpointTime"`
+	User           string    `json:"username"`
+	Host           string    `json:"hostname"`
+}
+
+var (
+	sessionIDEpochStartTime   = time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC)
```
julio-lopez, line 34: Is the year intended to be `2020` ?

jkowalski: it is arbitrary, really, but we have some hardcoded timestamps before 2020 in code already.

julio-lopez: I mean, the comment below says `2020-01-01`, but `sessionIDEpochStartTime` is set to `2000-01-01`

repo/content/content_manager.go
```diff
@@ -439,6 +436,18 @@ func (bm *Manager) Flush(ctx context.Context) error {
 		bm.cond.Broadcast()
 	}()
 
+	// see if we have any packs that have failed previously
```
julio-lopez, line 452: Is this changing the existing behavior / semantics in terms of:
1. not starting a new pack if there failed ones that need to be retried;
2. the frequency of the retries?

Perhaps I'm missing something.

repo/content/blob_crypto.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/encryption"
+	"github.com/kopia/kopia/repo/hashing"
+)
+
+func getIndexBlobIV(s blob.ID) ([]byte, error) {
+	if len(s) < 2*aes.BlockSize {
+		return nil, errors.Errorf("blob id too short: %v", s)
+	}
```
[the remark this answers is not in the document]
jkowalski: good catch.

repo/content/blob_crypto.go
```diff
@@ -0,0 +1,82 @@
...
+	payload, err = enc.Decrypt(nil, payload, iv)
+	if err != nil {
+		return nil, errors.Wrap(err, "decrypt error")
+	}
+
+	// Since the encryption key is a function of data, we must be able to generate exactly the same key
+	// after decrypting the content. This serves as a checksum.
+	if err := verifyChecksum(h, payload, iv); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
jkowalski: we should be able to get rid of this when we delete CTR-based non-authenticated ciphers. Do you think it's time to do it in 0.8?

repo/content/blob_crypto.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return payload, nil
+}
+
+func verifyChecksum(h hashing.HashFunc, data, contentID []byte) error {
```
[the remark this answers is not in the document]
jkowalski: renamed to iv

repo/content/content_manager.go
jkowalski: It does change the behavior a bit:

- previously all contents were added to the index AFTER they finished writing and it could be the index blob that's currently being built or the next one

- with this change contents are pre-destined for the current index blob and if the packs fails to write, the index won't be written until the pack is retried

it should not change much in terms of timings or performance since index blobs are only flushed at the end of the session or every N minutes.

julio-lopez: I understand the change regarding the semantics for timed flushed.

The question was related to the change for retrying failed packs.
- Before a content write would be rejected (failed) if there were failed packs and the retry failed. Essentially, this provided back pressure to avoid accumulating an unbounded number of writes. That was the original motivation. Now it seems that writes would be accumulated until the next timed or explicit flush.
- And secondary, the frequency of retries is much lower, which could be desirable to avoid retrying too frequently, but at the same time if the retry period is very large, that may be undesirable.

Again, maybe I'm missing something. I'll take another deeper pass at this portion.

repo/content/blob_crypto.go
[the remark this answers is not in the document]
julio-lopez: Let's discuss that separately. Was any of the non-authenticated ciphers a default one at any point?

repo/content/content_manager.go
julio-lopez: For completeness (from a separate convo):

IIUC, `WriteContent()` calls `addToPackUnlocked` for new (not-already-existing) content.
- Previously, the retry was in `addToPackUnlocked` before adding the content to the current pack & index. It was performed for every write of new content.
- Now the retry happens conditionally in `addToPackUnlocked` only after `bm.flushPackIndexesAfter`. So, with defaults, the worst case would be 10min. Probably 5 min on average.

repo/content/blob_crypto.go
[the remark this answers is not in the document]
jkowalski: it was the default before 0.5 where very few people (if any) used Kopia.

I'm thinking we delete them and if somebody still has an old repo they can use 0.7 to migrate and move to 0.8 afterwards.

repo/content/content_manager.go
jkowalski: Pushed one more commit that will explicitly retry previously-failed packs before WriteContent() to restore the previous behavior.

repo/content/content_manager.go
```diff
@@ -176,6 +176,33 @@ func (bm *Manager) maybeFlushBasedOnTimeUnlocked(ctx context.Context) error {
...
+	fp := append([]*pendingPackInfo(nil), bm.failedPacks...)
+	for _, pp := range fp {
+		formatLog(ctx).Debugf("retry-write %v", pp.packBlobID)
+
+		if err := bm.writePackAndAddToIndex(ctx, pp, true); err != nil {
+			return errors.Wrap(err, "error writing previously failed pack")
+		}
+	}
+
+	return nil
```
julio-lopez: Any reason for not putting this in a `maybeRetryFailedPacksLocked` helper and reuse it here and below in `addToPackUnlocked` ?

julio-lopez (approved):

julio-lopez:
The armhf build is stuck. Merging nevertheless.

Issue #715, opened 2020-12-15
Empty directories upon any command with --help argument


CrendKing:
When I run `kopia` with any command and the `--help` argument, there are two empty directories ("cli-logs" and "content-logs") created at the current directory. They don't seem to be used since the real ones are in user's home directory.

OS: Windows 10 x64.
Version: 0.7.3 build: c3ade4b0ea7a1d20497fcfc4f4aec161f73c9505

CrendKing:
I went to try the `KOPIA_LOG_DIR` env var and confirmed that even with the override it still creates the two directories in the current directory.

CrendKing:
I did some digging and found that the issue comes from kingpin, issue https://github.com/alecthomas/kingpin/issues/189. The author expressed he planned to remove default command in v3.

Basically, in `Initialize()` from logfile.go, all the flags are completely empty regardless if they have default value or not, because kingpin does not process those default values as they are not needed for argument parsing. Unfortunately, lacking these default values have side effects such as the empty directories.

Simple fix would be checking if `*logDir` is empty and if yes override it with the value of `ospath.LogsDir()`. Then I was thinking, since help commands never log anything anyways, why bother to set up the logger to begin with? Why not simply check `*logDir` empty and skip `Initialize()`?

PR: https://github.com/kopia/kopia/pull/753

CrendKing:
Since PR is merged, feel free to close.

PR #753, opened 2020-12-31
Fix bug that empty log directories are created on any command with --help argument

CrendKing:
More detail in https://github.com/kopia/kopia/issues/715


CrendKing:
Since PR is merged, feel free to close.

julio-lopez:
Fixes #715

Issue #716, opened 2020-12-15
kopia mount does not work on Windows


CrendKing:
In Windows, the mount function does not seem to work. Here’s what I did:

1. `kopia snapshot list`. Pick a snapshot ID.
2. `kopia mount <snapshot_id> x:`
3. Log:
```
2020-12-14 13:14:18.596 D [manifest_manager.go:288] listing manifest contents
2020-12-14 13:14:18.602 D [mount_webdav.go:32] creating webdav server...
2020-12-14 13:14:18.604 D [mount_net_use.go:41] running c:\windows\system32\net.exe [net use x: http://127.0.0.1:3132]
2020-12-14 13:14:18.623 D [mount_net_use.go:44] net use finished with  exit status 2
2020-12-14 13:14:18.623 D [mount_webdav.go:60] web server finished with http: Server closed
2020-12-14 13:14:19.124 D [logging_storage.go:28] [STORAGE] GetBlob("kopia.maintenance",0,-1)=({2141 bytes}, <nil>) took 0s
2020-12-14 13:14:19.124 D [maintenance_run.go:80] full manintenance cycle not enabled
2020-12-14 13:14:19.124 D [maintenance_run.go:92] quick manintenance cycle not enabled
2020-12-14 13:14:19.124 D [maintenance_run.go:171] not due for maintenance
2020-12-14 13:14:19.125 D [logging_storage.go:87] [STORAGE] Close()=<nil> took 0s
2020-12-14 13:14:19.126 D [memory_tracking.go:31] in use heap 3375104 (delta 3375104 max 0) stack 524288 (delta 524288 max 0)
```
I also went to made sure `net use` itself is working by sharing a folder and net use map to a drive. Something might be wrong with webdav.

OS: Windows 10
Version: 0.7.3 build: c3ade4b0ea7a1d20497fcfc4f4aec161f73c9505

CrendKing:
Turns out this was caused by my `WebClient` service being disabled. I must have disabled it long time ago to test something. Once enabled and started, mounting works.

[...]

PR #754, opened 2020-12-31
Fix hang at unmount when Ctrl-C is pressed

CrendKing:
More details at https://github.com/kopia/kopia/issues/716

### DIFF
--- a/internal/mount/mount_webdav.go
+++ b/internal/mount/mount_webdav.go
@@ -56,6 +56,10 @@ func DirectoryWebDAV(ctx context.Context, entry fs.Directory) (Controller, error
 
 	done := make(chan struct{})
 
+	srv.RegisterOnShutdown(func() {
+		close(done)
+	})
+
 	go func() {
 		log(ctx).Debugf("web server finished with %v", srv.Serve(l))
 	}()


jkowalski:
Calling `ctrl.Done()` without reading from it has no effect, it will simply return a channel without performing any operation.

The sequence should be:

when ctrl-C is pressed we call `ctrl.Unmount()`
`ctrl.Unmount()` should eventually/asynchronously close channel returned by `ctrl.Done()` channel, somehow this is not happening.

I think `mount_webdav.go` is missing close(c.done) inside `Unmount()` or you can do something like this in line 59:

```
done := make(chan struct{})

srv.RegisterOnShutdown(func() {
		close(done)
	})
```

If you can test whether this works on Windows, and update the PR, I'll be happy to merge it.

CrendKing:
Yes, it works. Updated the PR. I also kept the hoisted common code in cli/command_mount.go, but I'm not 100% sure if that has no side effect. Also not sure if it is safe to remove the empty `<-ctrl.Done()` case. If you prefer, I can restore that file altogether.

cli/command_mount.go
```diff
@@ -86,13 +86,12 @@ func runMountCommand(ctx context.Context, rep repo.Repository) error {
...
 
+		// Reporting clean unmount in case of interrupt signal.
+		<-ctrl.Done()
+
 	case <-ctrl.Done():
-		log(ctx).Infof("Unmounted.")
-		return nil
 	}
 
-	// Reporting clean unmount in case of interrupt signal.
```
jkowalski, line 94: this entire file should be reverted.

CrendKing: Done. Sorry about that.

jkowalski:
thank you!

PR #757, opened 2021-01-01
Add potential cause of specific mount command error on Windows

CrendKing:
Before this change, error message looks like:
```
error: mount error: unable to mount webdav server as drive letter: unable to run 'net use' (): error running 'net use': exit status 2, try --help
```
After this change, when error code 2:
```
error: mount error: unable to mount webdav server as drive letter: this error is often caused by but not limited to the 'WebClient' service being unable to start: exit status 2, try --help
```
When error code other than 2:
```
error: mount error: unable to mount webdav server as drive letter: error running 'net use': exit status 123, try --help
```

This change fixes the duplicate short message about 'net use'. It removes the empty `()`, because output is empty in case of error. It adds potential cause for error code 2. Otherwise, the error message is mostly untouched.

### DIFF
--- a/internal/mount/mount_net_use.go
+++ b/internal/mount/mount_net_use.go
@@ -49,7 +49,7 @@ func netUse(ctx context.Context, args ...string) (string, error) {
 func netUseMount(ctx context.Context, driveLetter, webdavURL string) (string, error) {
 	out, err := netUse(ctx, driveLetter, webdavURL)
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to run 'net use' (%v)", out)
+		return "", errors.Wrapf(err, "unable to run 'net use' (%v), see https://kopia.io/docs/mounting/#windows for more information", out)
 	}
 
 	if driveLetter != "*" {
--- a/site/content/docs/Getting started/_index.md
+++ b/site/content/docs/Getting started/_index.md
@@ -5,13 +5,13 @@ linkTitle: "Getting Started"
 weight: 10
 ---
 
-This guide will walk you through installing Kopia, creating, managing and restoring snapshots and defining snapshot policies. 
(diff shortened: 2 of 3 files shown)


internal/mount/mount_net_use.go
```diff
@@ -43,13 +43,21 @@ func netUse(ctx context.Context, args ...string) (string, error) {
...
+		}
+	}
+
+	return string(out), errors.WithMessagef(err, errorMsg)
 }
 
 func netUseMount(ctx context.Context, driveLetter, webdavURL string) (string, error) {
 	out, err := netUse(ctx, driveLetter, webdavURL)
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to run 'net use' (%v)", out)
```
jkowalski, line 52: how about we just add to this string:

"unable to run 'net use" (%v), see https://kopia.io/docs/mounting#windows for more information"

CrendKing: Sure, it's much cleaner too. I don't see `/mounting` at the moment. Do you want me to create that page or you do it later yourself?

CrendKing: Code changed

CrendKing: I don't know why this commit fails the build.

jkowalski: it's a random failure. Would you mind adding that page under `site` so that we don't forget it?

Instructions for testing the website are in https://github.com/kopia/kopia/blob/master/BUILD.md

CrendKing: Done. Let me know if you want certain part changed.

jkowalski:
The docs page looks great, thanks!

PR #765, opened 2021-01-05
performance: improve performance of fragmented index lookups

jkowalski:
In a typical repository there will be many small indexes and few large
ones. If the number gets above ~100 or so, things get very slow.

It helps to pre-merge very small indexes in memory to reduce the number
of binary searches and mmaped IO to perform. In some extreme cases
where we have many uncompacted index segments (each with just 1-2
entries), the savings are as quite dramatic.

In one case with >100 index segments the time to run
`kopia snapshot verify` went from 10m to 0.5m after this change.

In another well-maintained repository with 1.2M contents and
about 25 segments, the time to run `kopia snapshot verify` went from
48s to 35s.

### DIFF
--- a/repo/content/committed_content_index.go
+++ b/repo/content/committed_content_index.go
@@ -1,6 +1,7 @@
 package content
 
 import (
+	"bytes"
 	"context"
 	"path/filepath"
 	"sync"
@@ -9,6 +10,11 @@ import (
 
 	"github.com/kopia/kopia/repo/blob"
 )
+
+// combineSmallIndexThreshold is the count threshold below which indexes
+// will be combined in-memory to reduce the number of segments and speed up
+// large index operations (such as verification of all contents).
+const combineSmallIndexThreshold = 100
 
 type committedContentIndex struct {
 	cache committedContentIndexCache
@@ -119,7 +125,14 @@ func (b *committedContentIndex) use(ctx context.Context, packFiles []blob.ID) (b
 		newInUse[e] = ndx
 	}
 
-	b.merged = newMerged
+	mergedAndCombined, err := combineSmallIndexes(newMerged)
+	if err != nil {
+		return false, errors.Wrap(err, "unable to combine small indexes")
+	}
+
+	log(ctx).Debugf("combined %v into %v index segments", len(newMerged), len(mergedAndCombined))
+
+	b.merged = mergedAndCombined
 	b.inUse = newInUse
 
 	if err := b.cache.expireUnused(ctx, packFiles); err != nil {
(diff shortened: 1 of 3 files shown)


repo/content/committed_content_index.go
```diff
@@ -10,6 +11,11 @@ import (
 	"github.com/kopia/kopia/repo/blob"
 )
 
+// combineSmallIndexThreshold is the count threshold below which indexes
```
julio-lopez, line 14: What about `smallIndexEntryCountThreshold`?

```suggestion
// smallIndexEntryCountThreshold is the threshold to determine whether an
// index is small. Any index with fewer entries than this threshold 
```

repo/content/committed_content_index.go
```diff
@@ -131,6 +144,46 @@ func (b *committedContentIndex) use(ctx context.Context, packFiles []blob.ID) (b
...
+	if err := b.Build(&buf); err != nil {
+		return nil, errors.Wrap(err, "error building combined in-memory index")
+	}
+
+	combined, err := openPackIndex(bytes.NewReader(buf.Bytes()))
+	if err != nil {
+		return nil, errors.Wrap(err, "error opening combined in-memory index")
+	}
+
+	return append(toKeep, combined), nil
```
julio-lopez, line 184: Wow! neat!

julio-lopez (approved):
Just a minor suggestion about a variable name for clarity.

repo/content/committed_content_index.go
```diff
@@ -10,6 +11,11 @@ import (
 	"github.com/kopia/kopia/repo/blob"
 )
 
+// combineSmallIndexThreshold is the count threshold below which indexes
+// will be combined in-memory to reduce the number of segments and speed up
+// large index operations (such as verification of all contents).
+const combineSmallIndexThreshold = 100
```
julio-lopez, line 17:
```suggestion
const smallIndexEntryCountThreshold = 100
```

repo/content/committed_content_index.go
```diff
@@ -131,6 +144,46 @@ func (b *committedContentIndex) use(ctx context.Context, packFiles []blob.ID) (b
 	return true, nil
 }
 
+func combineSmallIndexes(m mergedIndex) (mergedIndex, error) {
+	var toKeep, toMerge mergedIndex
+
+	for _, ndx := range m {
+		if ndx.ApproximateCount() < combineSmallIndexThreshold {
```
julio-lopez, line 151:
```suggestion
		if ndx.ApproximateCount() < smallIndexEntryCountThreshold {
```

Issue #272, opened 2020-02-25
[Feature] Define a Kopia CLI output format


julio-lopez:
This becomes part of the API contract with systems that interact with the Kopia CLI and parse its output.

Some commands have a well defined output, such as `blob list`, for others it is not completely clear and some of the interesting and relevant information may go instead to the logs.

First step is to review the output of the different commands, document the format for the ones that we consider are well defined, and identify the command outputs that need improvement.

As a related feature (that should probably be tracked in a separate ticket), is to add structured output for those commands, for example using JSON.

julio-lopez:
Also, whatever we decide the output of a command to be, that should go to `stdout`, and nothing else should go to `stdout`. Error and log messages should go to `stderr`

julio-lopez:
FYI @pavannd1 @redgoat650

jkowalski:
How about we support `--json` flag on all commands where it makes sense and leave non-JSON format undefined?

jkowalski:
In particular `snapshot create` could return `snapshot.Manifest` which is reasonably expected to stay unchanged and backwards compatible since it's stored in the repository itself.

julio-lopez:
It makes sense to add `--json` to provide _structured_ output.

At some point, even if we do not guarantee backwards compatibility and explicitly say it is not guaranteed, it'd be good to define what the expected output for different commands should be and have the commands follow a consistent pattern.

jkowalski:
Moving this to 0.6.

julio-lopez:
#642

julio-lopez:
We should prioritize this issue.

Independent of the output format (JSON, YAML, table, plain text, etc), the first step is to define and document the data (fields) that each command should produce on stdout.

Then we can have one or two format implementation for those. A default human-readable plain text one and perhaps JSON.

PR #767, opened 2021-01-05
Add support for kopia diff --json

eras:
This produces a basic JSON version of the textual content provided by the `kopia diff` command; most useful when combined with the use of `jq`.

Activated with a `diff`-specific `--json`-switch, so it does not try to be general in any way and the mapping from the old output to the JSON is pretty much 1:1—so there's maybe some amount of redundancy and it does not try to collect output e.g. per-file or anything. And definitely nothing as all-encompassing as what https://github.com/kopia/kopia/issues/272 aims for.

But it's still probably quite useful for analyzing the changes in the repo.

### DIFF
--- a/internal/diff/diff.go
+++ b/internal/diff/diff.go
@@ -19,6 +19,17 @@ import (
 
 var log = logging.GetContextLoggerFunc("diff")
 
+// OutputFormat to use; processed by the NewComparer function.
+type OutputFormat int
+
+const (
+	// OutputText is the basic textual line-by-line output.
+	OutputText OutputFormat = iota
+
+	// OutputJSON is the JSON object output format.
+	OutputJSON = iota
+)
+
 // Comparer outputs diff information between two filesystems.
 type Comparer struct {
 	tmpDir string
@@ -289,13 +300,20 @@ func downloadFile(ctx context.Context, f fs.File, fname string) error {
 }
 
 // NewComparer creates a comparer for a given repository that will output the results to a given writer.
-func NewComparer(out io.Writer) (*Comparer, error) {
+func NewComparer(out io.Writer, outputFormat OutputFormat) (*Comparer, error) {
 	tmp, err := ioutil.TempDir("", "kopia")
 	if err != nil {
 		return nil, errors.Wrap(err, "error creating temp directory")
 	}
 
-	output := &ComparerOutputText{out: out}
+	var output ComparerOutput
+
+	switch outputFormat {
+	case OutputText:
+		output = &ComparerOutputText{out: out}
+	case OutputJSON:
+		output = &ComparerOutputJSON{out: out}
+	}
 
 	return &Comparer{tmpDir: tmp, output: output}, nil
 }
--- a/internal/diff/diff.go
+++ b/internal/diff/diff.go
@@ -3,7 +3,6 @@ package diff
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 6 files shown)


jkowalski:
One more thought (and probably more difficult): could this use JSON nesting:

```
{
  "someDir": {
    "someFile": {
      "change": "added",
       "new": {
          "oid": "abc1235"
       },
    },
    "someOtherFile": {
      "changeType": "changed",
       "old": {
          "oid": "abc1234"
       },
       "new": {
          "oid": "abc1235"
       }
    },
   }
}
```

And so on. Many things need to be figured out here w.r.t. the format (perhaps there's an existing one we can use or something close to it)?

eras:
I figured just going with basically the same text output but as JSON wouldn't be enough ;).

The only "standard" format for representing differences I know of are `diff` and `xdelta`; I'm not aware of format suitable for representing this kind of data.

I'm of the opinion that JSON nesting is not the way to go. In particular, it is more complicated to process with simple tools, even though `jq` is able to do recursion—and it's an advanced topic with `jq`.

Something like a flat

```
{ "entries": [{
    "path/file" : {
      "change": "add"|"remove"|"update",
      "path_type": "directory"|"file"|..,
      "size_delta": 42,
      "update": ["size", "modtime", "modtimestamp"],
      "new": {
        "size": 42,
        "unix": {
          "permissions": "-rwxrwx---",
          "userid": 1001,
          "groupid": 1002,
        },
        "modtime": 1000000001,
        "modtimestamp": "2001-09-09..."
      },
      "old": {
        ...
      },
      "diff": {
        "exit_code": 1,
        "output": "..."
      }}, ...]}
```

might be more consistent, easier to process, and easier to convert to non-hierarchic formats, than the current format.

Of course, at the cost of more code :). The amount of information provided from the `Diff` module to `ComparerOutput` probably needs be increased.

If a hierarchic format is truly desirable then it could be implemented purely in the JSON output module, by building a tree out of the existing data. But it's quite unclear to me if that would be even desirable; the flat format has its benefits and I would like to keep it. Producing hierarchic output mode could be added later, with e.g. an additional switch, if the need for it rises at all. In that format i would just move add and "entries" array to the change object and possible otherwise empty intermediate objects for representing unchanged intermediate directories.

internal/diff/diff.go
```diff
@@ -20,22 +19,35 @@ import (
...
 
+// OutputFormat to use; processed by the NewComparer function.
+type OutputFormat int
+
+const (
+	// OutputText is the basic textual line-by-line output.
+	OutputText OutputFormat = iota
+
+	// OutputJSON is the JSON object output format.
+	OutputJSON = iota
```
jkowalski, line 30: nit - iota can be removed here

tests/end_to_end_test/restore_test.go
```diff
@@ -110,7 +110,7 @@ func compareDirs(t *testing.T, source, restoreDir string) {
 	testenv.AssertNoError(t, err)
 
 	if !assert.Equal(t, wantHash, gotHash, "restored directory hash does not match source's hash") {
-		cmp, err := diff.NewComparer(os.Stderr)
+		cmp, err := diff.NewComparer(os.Stderr, diff.OutputText)
```
jkowalski, line 113: Can we add a test for the JSON format?

jkowalski (changes requested):

jkowalski:
any updates here? since this PR had activity there have been a bunch of changes that caused conflicts.

jkowalski:
this has been stale for a while, please rebase and open new PR if you intend to pursue this.

PR #769, opened 2021-01-06
cli: make sure we don't run maintenance as part of read-only actions

jkowalski:


julio-lopez (approved):
LGTM  ✅

I assume the current set of actions is not comprehensive, is that correct?

One read-only action that comes to mind is `snapshot restore`

jkowalski:
> I assume the current set of actions is not comprehensive, is that correct?
> 
> One read-only action that comes to mind is `snapshot restore`

This was already marked as read-only in previous PR, so it's covered.

PR #771, opened 2021-01-06
mechanical refactoring of content manager to extract CommittedReadManager

jkowalski:
CommittedReadManager supports reading of committed contents only

(note commits are not necessarily pretty but should be easy to review)


julio-lopez (approved):
Thanks for breaking it up. It made it super easy to review.

PR #773, opened 2021-01-06
Improved .kopiaignore pattern matching

### LINKED ISSUE #492: gitignore Syntax problems

### LINKED ISSUE #571: .kopiaignore pattern differs from .gitignore
internal/ignore/ignore.go states

> Package ignore implements ignoring files based on 'gitignore' syntax.

but there seem to be some differences between .gitignore and .kopiaignore patterns. Leading slashes are treated different, at least as far I can tell from calling _kopia snapshot estimate_.

I am not sure, if the differences are intended. If they are, I would be delighted if they would be documented. In that case I'd also like to know, how to achieve the same effect as a leading slash in [.gitignore pattern format](https://git-scm.com/docs/gitignore#_pattern_format).

If the differences are intended, it is necessary to take even more care on adopting a list like https://github.com/rubo77/rsync-homedir-excludes

I'm not sure how far this is covered by #492 or #496, therefore I will give an example:

#### FS layout
```
/tmp/kopia-test/.kopiaignore
/tmp/kopia-test/1GB.mp4
/tmp/kopia-test/A/1GB.mp4
/tmp/kopia-test/A/AA/1GB.mp4
/tmp/kopia-test/A/B/A/1GB.mp4
/tmp/kopia-test/B/A/1GB.mp4
```

#### Command: kopia snapshot estimate /tmp/kopia-test

1) .kopiaignore: `/A/` or `./A/`

    kopia scans everything, estimate is about 5GB. /A/... is  **not** ignored _(Expectation: /A ignored; /B/A/ and /B/A not ignored)_

2) .kopiaignore: `A/`

    Estimate is about 1GB. /A/... and /B/A/... are ignored _(meets expectations)_

3) .kopiaignore: `B/A/`

    Estimate is about 4GB. /A/B/A/... is  **not** ignored, /B/A/... is ignored _(Expectation: /A/B/A and /B/A both ignored)_

4) .kopiaignore: `/B/A/`

    kopia scans everything, estimate is about 5GB. /B/A/... is  **not** ignored _(Expectation: /A/B/A/ not ignored, /B/A ignored)_

alphaleonis:
.kopiaignore pattern matching now (hopefully) conforms to the .gitignore specification (https://git-scm.com/docs/gitignore)

Replaced old package "ignore" with a newly written "wcmatch" that manages the globbing. This should support all the patterns that .gitignore supports.  

Some changes in ignorefs that dealt with how the patterns were matched.

This fixes #571 
Fixes #492


alphaleonis:
Just realized that I incorrectly implemented one of the rules regarding "unrooted" patterns containing directory separators.  I will fix this and push new changes in a day or so.

alphaleonis:
Should be correct now (I hope)

fs/ignorefs/ignorefs.go
```diff
@@ -25,22 +25,29 @@ type ignoreContext struct {
 
 	onIgnore []IgnoreCallback
 
-	dotIgnoreFiles []string         // which files to look for more ignore rules
-	matchers       []ignore.Matcher // current set of rules to ignore files
-	maxFileSize    int64            // maximum size of file allowed
+	dotIgnoreFiles []string                  // which files to look for more ignore rules
+	matchers       []wcmatch.WildcardMatcher // current set of rules to ignore files
```
jkowalski, line 29: any reason for the new package? can we just repurpose the old one?

internal/wcmatch/wcmatch_test.go
```diff
@@ -0,0 +1,363 @@
...
+		{"???*", "abc", true, true},
+		{"*???", "abc", true, true},
+		{"???", "abc", true, true},
+		{"*", "abc", true, true},
+		{"??", "a", false, false},
+
+		// Recursive wildcards
+		{"**", "foo", true, true},
+		{"**", "foo/bar", true, true},
+		{"/a/**/b/foo.txt", "/a/b/c/d/foo.txt", false, false},
```
jkowalski, line 131: should we add tests here for windows-style paths too with backslashes instead of slashes?

jkowalski:
This is great! Thanks for the contribution. I'd like to just make sure it works correctly on Windows. We don't have any e2e test that does ignore rules, can you try adding one here for interesting cases?

internal/wcmatch/wcmatch_test.go
jkowalski: ... unless of course the paths are always slash-separated, which I don't remember off the top of my head.

alphaleonis: Backslashes are always escape characters and cannot be used as directory separators from what it seems. So Windows style paths cannot be used in patterns, and since I developed this on Windows I assume that the paths supplied to ignorefs are also never using Windows style directory separators?

fs/ignorefs/ignorefs.go
alphaleonis: There's a reason for everything in this world, the only question whether or not it is a good reason. 😉 

Jokes aside, I started developing this as a package outside of kopia, just for wildcard matching which is where the name came from. When I moved it in it seemed more descriptive to me which is why I kept it. The matcher does globbing or wildcard matching and really has no code for "ignoring" anything, this logic is all in `ignorefs`.

But I can of course change the package name back to "ignore" if you prefer this. Just let me know.  😃

alphaleonis:
I haven't looked at the e2e tests, but I'll have a look at them and see if I can construct some test cases here with ignore rules for "snapshot create" (or are there any other commands worth testing here too?). Sounds like a good idea. I can do these in a separate PR if you like.

There was one test for ignorefs that was commented out that I uncommented as part of this PR which I'm sure you noticed.  (Also, considering I did my development on Windows I'd be more concerned that it doesn't work on linux 😉 )

alphaleonis:
The only thing, considering Windows that I have not considered in this PR is that the filesystems (normally anyway) are case-insensitive.  Now I did implement support for case-insensitive matching in wcmatch, and there are tests for it, however ignorefs currently always use case-sensitive matching.  I would like to use case insensitive matching, but then ignorefs would have to be able to determine whether or not the underlying filesystem is case-sensitive or not.  (I'm not sure if just checking whether or not we're on Windows is a good idea, since I'm assuming you can mount a case-sensitive fs on Windows as well?)  I'm not entierly sure of what the best way to accomplish this is, which is why I left it out in this PR.  Any ideas here?

internal/wcmatch/wcmatch_test.go
jkowalski: ah, ok then.

alphaleonis:
>  We don't have any e2e test that does ignore rules, can you try adding one here for interesting cases?
> 
> (Ideally, add a case that was not properly supported before and passing now, plus perhaps few simple common cases that were and should be working - maybe worth making this a separate PR with failing tests commented out and uncomment them in this PR)

See PR #774 
The uncommented tests are pushed to this PR.

jkowalski:
This is great, thank you!

PR #774, opened 2021-01-07
Created end-to-end tests verifying .kopiaignore behavior.

alphaleonis:
This is related to #571 and #773, but provided as a separate PR to include tests that did not work before PR #773.

### DIFF
--- a/tests/end_to_end_test/snapshot_create_test.go
+++ b/tests/end_to_end_test/snapshot_create_test.go
@@ -143,34 +143,34 @@ func TestSnapshotCreateWithIgnore(t *testing.T) {
 			},
 			expected: []string{},
 		},
-		{
-			desc: "ignore_all_but_text_files",
-			files: []testFileEntry{
-				{
-					Name: "/.kopiaignore",
-					Content: []string{
-						"*",
-						"!*.txt",
-						"!*/",
-					},
-				},
-				{Name: "foo/bar/file.txt"},
-				{Name: "foo/bar/file.png"},
-				{Name: "foo/file.txt"},
-				{Name: "foo/file.jpg"},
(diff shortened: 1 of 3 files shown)


tests/end_to_end_test/snapshot_create_test.go
```diff
@@ -113,3 +118,292 @@ func TestSnapshottingCacheDirectory(t *testing.T) {
...
+				".kopiaignore",
+				"file.txt",
+				"A/file.txt",
+				"A/AA/file.txt",
+				"A/B/A/file.txt",
+				"B/", // directory is empty because all contents are ignored, but the empty directory is still included.
+			},
+		},
+		{
+			desc: "ignore_rooted_vs_unrooted_5",
```
jkowalski, line 265: This looks great! Can you also add couple more tests where there are multiple `.kopiaignore` in various locations with overlapping rules:

* `./.kopiaignore` and `./A/.kopiaignore`
* `./.kopiaignore`, `./A/.kopiaignore` and `./A/B/.kopiaignore`
* `./A/.kopiaignore` and `./A/B/.kopiaignore`
* `./A/.kopiaignore` and `./B/.kopiaignore`

to make sure that those get applied correctly.

alphaleonis: Sure, we can add some of these. But I think most of these could be handled by the tests of ignorefs rather than end-to-end tests?  They are much faster to run anyway.

alphaleonis: Good idea by the way! 👍   I added some tests (commented in this PR) which led me to discovered a bug in the implementation in #773 when an ignore file in a subdirectory negated a pattern in the parent directory.

PR #776, opened 2021-01-08
Manifest manager refactoring

jkowalski:
The goal is the same as with content manager refactoring (#771) - to be able to share the single manager for keeping track of committed manifests and have per-session manifest managers.

The commits are similarly structured to minimize code movement.


repo/manifest/manifest_manager.go
```diff
@@ -153,24 +171,35 @@ func (m *Manager) Get(ctx context.Context, id ID, data interface{}) (*EntryMetad
...
 
-	for _, e := range m.pendingEntries {
-		if matchesLabels(e.Labels, labels) {
-			matches = append(matches, cloneEntryMetadata(e))
-		}
+	for _, e := range findEntriesMatchingLabels(m.pendingEntries, labels) {
+		matches = append(matches, cloneEntryMetadata(e))
+	}
+
+	committed, err := m.committed.findCommittedEntries(ctx, labels)
```
julio-lopez: Is the check below in line 208 redundant?
NM, this is addressed later.

repo/manifest/manifest_manager.go
```diff
@@ -84,67 +81,59 @@ func (m *Manager) Put(ctx context.Context, labels map[string]string, payload int
...
-	defer m.mu.Unlock()
+	e, err := m.getPendingOrCommitted(ctx, id)
+	if err != nil {
+		return nil, err
+	}
 
-	e := m.pendingEntries[id]
-	if e == nil {
-		var err error
+	return cloneEntryMetadata(e), nil
```
julio-lopez, line 98: This is not the same as before.

This is assuming that either:
* `e` points to a local copy (which it does not); or
* `*e` is not modified in any way, which currently seems to be the case but it is subtle and there is not explicit guarantee

repo/manifest/manifest_manager.go
```diff
@@ -84,67 +81,59 @@ func (m *Manager) Put(ctx context.Context, labels map[string]string, payload int
...
+		}
 	}
 
-	return &EntryMetadata{
-		ID:      id,
-		ModTime: e.ModTime,
-		Length:  len(e.Content),
-		Labels:  copyLabels(e.Labels),
-	}, nil
+	return cloneEntryMetadata(e), nil
```
julio-lopez: Same comment as above.

repo/manifest/manifest_manager.go
```diff
@@ -161,6 +150,11 @@ func findEntriesMatchingLabels(m map[ID]*manifestEntry, labels map[string]string
 
 // Find returns the list of EntryMetadata for manifest entries matching all provided labels.
 func (m *Manager) Find(ctx context.Context, labels map[string]string) ([]*EntryMetadata, error) {
+	committedMatches, err := m.committed.findCommittedEntries(ctx, labels)
```
julio-lopez, line 153: Same as above, this is implicitly assuming the entries in `commitedMatches` are not modified.

julio-lopez (commented):
👍 LG +1

See inline comments about a couple of subtle assumptions.

repo/manifest/manifest_manager.go
jkowalski: entries (pending or committed) are immutable, so I don't really see a problem here.

PR #779, opened 2021-01-11
Option to print out the commands for benchmark

icepic:
to use the results on crypto, splitter and compression.


jkowalski:
Thanks for the contribution, this is great!

PR #782, opened 2021-01-13
Cleaned up and refactored object manager

jkowalski:
- reads are now package-level functions
- object.Manager is now only responsible for writes


jkowalski:
(This is a standalone subset of #780)

julio-lopez:
👍 LG +1

PR #788, opened 2021-01-20
server: reworked authn/authz

jkowalski:
Previously authentication was done as an wrapper handler and
authorization was inlined. This change moves authn/authz handlers
inside the server and implements separate authorization module that's
individually tested.

Also fixed an issue where server users were not able to see global
or host-level policies.

### DIFF
--- a/internal/auth/authz.go
+++ b/internal/auth/authz.go
@@ -1,0 +1,77 @@
+package auth
+
+import (
+	"context"
+	"strings"
+
+	"github.com/kopia/kopia/repo"
+)
+
+// AuthorizerFunc gets the authorizations for given user.
+type AuthorizerFunc func(ctx context.Context, rep repo.Repository, username string) AuthorizationInfo
+
+// AccessLevel specifies access level when accessing repository objects.
+type AccessLevel int
+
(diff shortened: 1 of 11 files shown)


jkowalski:
Rebased on top of #780. PTAL.

internal/auth/authz.go
```diff
@@ -0,0 +1,77 @@
...
+
+type noAccessAuthorizer struct{}
+
+func (noAccessAuthorizer) ContentAccessLevel() AccessLevel { return AccessLevelNone }
+func (noAccessAuthorizer) ManifestAccessLevel(labels map[string]string) AccessLevel {
+	return AccessLevelNone
+}
+
+// NoAccess implements AuthorizationInfo which grants no permissions.
+var NoAccess AuthorizationInfo = noAccessAuthorizer{}
```
julio-lopez, line 41: nit: should this be a function to prevent accidental overwriting of the variable? Essentially, have it behave more like a constant.

```suggestion
func NoAccess() AuthorizationInfo { return noAccessAuthorizer{} }
```

julio-lopez (approved):

PR #789, opened 2021-01-21
Implemented new streaming GRPC protocol for Kopia Repository Server

jkowalski:
This has several benefits:

- sticky sessions tied to one backend which allows scale-out
- more efficient protobuf serialization
- authentication/authorization one per stream
- allows true write sessions implemented in #780

GRPC is the default but can be disabled with `--disable-grpc` when connecting and `--no-grpc` when starting the server.
Legacy API is enabled by default and can be disabled with `--disable-legacy-api` when starting the server.

This uses the new authz mechanism implemented in #788.


repo/grpc_repository_client.go
```diff
@@ -0,0 +1,532 @@
...
+	// to simplify client code.
+	if err := r.sess.Send(req); err != nil {
+		ch <- &apipb.SessionResponse{
+			Response: &apipb.SessionResponse_Error{
+				Error: &apipb.ErrorResponse{
+					Code:    apipb.ErrorResponse_CLIENT_ERROR,
+					Message: err.Error(),
+				},
+			},
+		}
```
julio-lopez, line 139: Should the channel be closed in this case, and also removed from the active requests?

repo/grpc_repository_client.go
```diff
@@ -477,56 +625,107 @@ func OpenGRPCAPIRepository(ctx context.Context, si *APIServerInfo, cliOpts Clien
...
-		cliOpts:        cliOpts,
-		sess:           sess,
-		activeRequests: make(map[int64]chan *apipb.SessionResponse),
-		nextRequestID:  1,
-	}
+		log(ctx).Debugf("establishing new GRPC streaming session")
 
-	go rr.readLoop()
+		retryPolicy := retry.Always
+		if !r.hadInnerSession && r.transparentRetries {
```
julio-lopez, line 646: where is `hadInnerSession` updated?

repo/grpc_repository_client.go
```diff
@@ -477,56 +625,107 @@ func OpenGRPCAPIRepository(ctx context.Context, si *APIServerInfo, cliOpts Clien
...
+
+func (r *grpcRepositoryClient) killInnerSession() {
+	r.innerSessionMutex.Lock()
+	defer r.innerSessionMutex.Unlock()
 
-	rr.objectFormat = object.Format{
-		Splitter: p.Splitter,
+	if r.innerSession != nil {
+		r.innerSession.cli.CloseSend() //nolint:errcheck
+		r.innerSession = nil
```
julio-lopez, line 689: Do errors need to be sent to the channels for active requests and then closed? Would active requests hang otherwise?

julio-lopez (commented):
Overall, LG.
See a few comments / questions inline.

internal/server/grpc_session.go
```diff
@@ -0,0 +1,450 @@
...
+	if err != nil {
+		log(ctx).Warningf("session handshake error: %v", err)
+		return err
+	}
+
+	return repo.DirectWriteSession(ctx, dr, opt, func(dw repo.DirectRepositoryWriter) error {
+		// channel to which workers will be sending errors, only holds 1 slot and sends are non-blocking.
+		lastErr := make(chan error, 1)
+
+		for req, err := srv.Recv(); err == nil; req, err = srv.Recv() {
```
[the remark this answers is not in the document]
jkowalski: yes, when the session terminates it's because the client got disconnected. We will log it in `defer` so no point logging two messages.

repo/grpc_repository_client.go
jkowalski: added close()

repo/grpc_repository_client.go
jkowalski: dead code

repo/grpc_repository_client.go
jkowalski: the read loop will terminate because the server will also disconnect in this case and will stuff errors into reader channels.

cli/command_repository_connect_server.go
```diff
@@ -14,6 +14,7 @@ var (
 
 	connectAPIServerURL             = connectAPIServerCommand.Flag("url", "Server URL").Required().String()
 	connectAPIServerCertFingerprint = connectAPIServerCommand.Flag("server-cert-fingerprint", "Server certificate fingerprint").String()
+	connectAPIServerDisableGRPCAPI  = connectAPIServerCommand.Flag("disable-grpc", "Use GRPC API").Bool()
```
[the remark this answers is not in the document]
jkowalski: done

repo/grpc_repository_client.go
julio-lopez: But then, that would result in a double close of the channel, both here and at the bottom of `readLoop`, which will then panic.

julio-lopez (approved):

PR #792, opened 2021-01-22
[cleanup] use t.Logf instead of log.Printf in test

julio-lopez:


jkowalski (approved):

jkowalski:
btw, we can probably create rule to prevent this.

the new linter has `forbidigo` feature for this.

PR #794, opened 2021-01-24
restore: added support for incremental restore and ignoring copy errors

jkowalski:
Created this as a hotfix for one of the users on Slack. We were trying to restore ~TB of data where a blob went missing in B2 storage (suspect pre-0.7.3 repository which did not have correct retry logic). 

This adds:

* `--incremental` - which will not attempt to download files already present in the output as long as the size and mod time are correct
* `--ignore-errors` in case a file can't be restored for any reason, print the error to output but continue copying

### DIFF
--- a/cli/command_restore.go
+++ b/cli/command_restore.go
@@ -104,7 +104,7 @@ func addRestoreFlags(cmd *kingpin.CmdClause) {
 	cmd.Flag("skip-times", "Skip times during restore").BoolVar(&restoreSkipTimes)
 	cmd.Flag("ignore-permission-errors", "Ignore permission errors").BoolVar(&restoreIgnorePermissionErrors)
 	cmd.Flag("ignore-errors", "Ignore all errors").BoolVar(&restoreIgnoreErrors)
-	cmd.Flag("incremental", "Incremental restore, do not download files that already exist in the destination").BoolVar(&restoreIncremental)
+	cmd.Flag("skip-existing", "Skip files and symlinks that exist in the output").BoolVar(&restoreIncremental)
 }
 
 func restoreOutput(ctx context.Context) (restore.Output, error) {
--- a/cli/command_restore.go
+++ b/cli/command_restore.go
@@ -77,6 +77,8 @@ var (
 	restoreSkipTimes              = false
 	restoreSkipOwners             = false
 	restoreSkipPermissions        = false
+	restoreIncremental            = false
+	restoreIgnoreErrors           = false
 )
 
 const (
@@ -101,6 +103,8 @@ func addRestoreFlags(cmd *kingpin.CmdClause) {
 	cmd.Flag("skip-permissions", "Skip permissions during restore").BoolVar(&restoreSkipPermissions)
 	cmd.Flag("skip-times", "Skip times during restore").BoolVar(&restoreSkipTimes)
 	cmd.Flag("ignore-permission-errors", "Ignore permission errors").BoolVar(&restoreIgnorePermissionErrors)
+	cmd.Flag("ignore-errors", "Ignore all errors").BoolVar(&restoreIgnoreErrors)
+	cmd.Flag("incremental", "Incremental restore, do not download files that already exist in the destination").BoolVar(&restoreIncremental)
 }
 
 func restoreOutput(ctx context.Context) (restore.Output, error) {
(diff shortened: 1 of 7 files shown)


snapshot/restore/local_fs_output.go
```diff
@@ -139,6 +167,16 @@ func fileIsSymlink(stat os.FileInfo) bool {
...
 }
 
+// SymlinkExists implements restore.Output interface.
+func (o *FilesystemOutput) SymlinkExists(ctx context.Context, relativePath string, e fs.Symlink) bool {
+	st, err := os.Lstat(filepath.Join(o.TargetPath, relativePath))
+	if err != nil {
+		return false
+	}
+
+	return (st.Mode() & os.ModeType) == os.ModeSymlink
```
julio-lopez, line 177: Should this also check whether `o.TargetPath/relativePath` points to the same target as `e` ?

cli/command_restore.go
```diff
@@ -101,6 +103,8 @@ func addRestoreFlags(cmd *kingpin.CmdClause) {
 	cmd.Flag("skip-permissions", "Skip permissions during restore").BoolVar(&restoreSkipPermissions)
 	cmd.Flag("skip-times", "Skip times during restore").BoolVar(&restoreSkipTimes)
 	cmd.Flag("ignore-permission-errors", "Ignore permission errors").BoolVar(&restoreIgnorePermissionErrors)
+	cmd.Flag("ignore-errors", "Ignore all errors").BoolVar(&restoreIgnoreErrors)
+	cmd.Flag("incremental", "Incremental restore, do not download files that already exist in the destination").BoolVar(&restoreIncremental)
```
julio-lopez, line 107: Should this be called `skip-existing` or `skip-matching` to reflect what it does? Also, expand the flag comment to describe which types of entries it applies to (regular files and symlinks) and when are entries skipped and when they are not.

julio-lopez (approved):
LG. A couple of comments, so not merging.

snapshot/restore/local_fs_output.go
jkowalski: the whole point was to avoid reading objects from repository, so we can either always read or be relatively lax here.

jkowalski: in hindsight symlink target should have been part of metadata and not stored in an object, perhaps we can fix that...

cli/command_restore.go
jkowalski: Renamed to `--skip-existing`

snapshot/restore/local_fs_output.go
julio-lopez: Then, the log message below is not accurate and could be misleading

`log(ctx).Debugf("skipping symlink %v because it already exists and is correct", targetPath)`

snapshot/restore/restore.go
```diff
@@ -91,12 +107,52 @@ func Entry(ctx context.Context, rep repo.Repository, output Output, rootEntry fs
...
+				atomic.AddInt32(&c.stats.SkippedCount, 1)
+				atomic.AddInt64(&c.stats.SkippedTotalFileSize, e.Size())
+
+				return onCompletion()
+			}
+
+		case fs.Symlink:
+			if c.output.SymlinkExists(ctx, targetPath, e) {
+				atomic.AddInt32(&c.stats.SkippedCount, 1)
+				log(ctx).Debugf("skipping symlink %v because it already exists and is correct", targetPath)
```
julio-lopez, line 133:
```suggestion
				log(ctx).Debugf("skipping symlink %v because it already exists", targetPath)
```

snapshot/restore/restore.go
```diff
@@ -91,12 +107,52 @@ func Entry(ctx context.Context, rep repo.Repository, output Output, rootEntry fs
...
+	ignoreErrors bool
 }
 
 func (c *copier) copyEntry(ctx context.Context, e fs.Entry, targetPath string, onCompletion func() error) error {
+	if c.incremental {
+		// in incremental mode, do not copy if the output already exists
+		switch e := e.(type) {
+		case fs.File:
+			if c.output.FileExists(ctx, targetPath, e) {
+				log(ctx).Debugf("skipping file %v because it already exists and is correct", targetPath)
```
julio-lopez, line 123: ?
```suggestion
				log(ctx).Debugf("skipping file %v because it already exists and metadata matches", targetPath)
```

julio-lopez (approved):

PR #798, opened 2021-01-26
blob: rewrote blob storage providers retry logic

jkowalski:
Previously we were trying to guess which errors are retriable
(often poorly, since things like DNS or HTTP errors were not retried)

This change makes it so the provider does not need built-in retry logic
and instead retrying wrapper is added which retries on all errors other
than

`blob.ErrNotFound`, `blob.ErrInvalidRange` and `blob.ErrSetTimeUnsupported`.

This required thorough testing of negative conditions that would trigger
blob.ErrInvalidLength (we already had those). Verified that tests pass
against Azure, B2, S3, GCS, Rclone, WebDAV, SFTP.

### DIFF
--- a/repo/blob/retrying/retrying_storage.go
+++ b/repo/blob/retrying/retrying_storage.go
@@ -62,18 +62,9 @@ func (s retryingStorage) DeleteBlob(ctx context.Context, id blob.ID) error {
 	return err // nolint:wrapcheck
 }
 
-// NewWrapper returns a readonly Storage wrapper that prevents any mutations to the underlying storage.
+// NewWrapper returns a Storage wrapper that adds retry loop around all operations of the underlying storage.
 func NewWrapper(wrapped blob.Storage) blob.Storage {
 	return &retryingStorage{Storage: wrapped}
-}
-
-// NonRetriable wraps an error tha is not retriable.
-type NonRetriable struct {
-	Err error
-}
-
-func (e NonRetriable) Error() string {
-	return e.Err.Error()
 }
 
 func isRetriable(err error) bool {
--- a/repo/blob/retrying/retrying_storage.go
+++ b/repo/blob/retrying/retrying_storage.go
@@ -1,0 +1,93 @@
+// Package retrying implements wrapper around blob.Storage that adds retry loop around all operations in case they return unexpected errors.
+package retrying
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/kopia/kopia/internal/retry"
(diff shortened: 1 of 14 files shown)


repo/blob/retrying/retrying_storage.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return err // nolint:wrapcheck
+}
+
+// NewWrapper returns a readonly Storage wrapper that prevents any mutations to the underlying storage.
+func NewWrapper(wrapped blob.Storage) blob.Storage {
+	return &retryingStorage{Storage: wrapped}
+}
+
+// NonRetriable wraps an error tha is not retriable.
```
julio-lopez, line 70:
```suggestion
// NonRetriable wraps an error that is not retriable.
```

repo/blob/retrying/retrying_storage.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (s retryingStorage) DeleteBlob(ctx context.Context, id blob.ID) error {
+	_, err := retry.WithExponentialBackoff(ctx, "DeleteBlob("+string(id)+")", func() (interface{}, error) {
+		return true, s.Storage.DeleteBlob(ctx, id)
+	}, isRetriable)
+
+	return err // nolint:wrapcheck
+}
+
+// NewWrapper returns a readonly Storage wrapper that prevents any mutations to the underlying storage.
```
julio-lopez, line 65: Is this comment accurate?

repo/blob/storage.go
```diff
@@ -143,3 +147,33 @@ func IterateAllPrefixesInParallel(ctx context.Context, parallelism int, st Stora
...
+func EnsureLengthExactly(b []byte, length int64) ([]byte, error) {
+	if length < 0 {
+		return b, nil
+	}
+
+	if len(b) != int(length) {
+		return nil, errors.Wrapf(ErrInvalidRange, "invalid length %v, expected %v", len(b), length)
+	}
+
+	return b[0:length], nil
```
julio-lopez, line 163: ?
```suggestion
	return b, nil
```

julio-lopez (approved):

repo/blob/retrying/retrying_storage.go
jkowalski: dead code.

repo/blob/retrying/retrying_storage.go
jkowalski: fixed.

repo/blob/storage.go
jkowalski: fixed

PR #804, opened 2021-01-28
Support for ignoring sources when using `snapshot create --all`

pavannd1:
### Overview

Added a new manual field in the scheduling policy.
While gathering all the existing sources during creation of snapshot using `--all` flag, we check the effective policy on each source and skip it if the `manual` value is set to true.

This mechanism can be used for virtual sources (`stdin`) where we don't have the respective directory/file on the local filesystem.

### DIFF
--- a/cli/command_snapshot_create.go
+++ b/cli/command_snapshot_create.go
@@ -292,19 +292,28 @@ func getLocalBackupPaths(ctx context.Context, rep repo.Repository) ([]string, er
 	for _, src := range sources {
 		// add all sources belonging to the repository user@host
 		// ignore sources that have Manual field set to true in the SchedulingPolicy
-		policyTree, err := policy.TreeForSource(ctx, rep, src)
+		includeSource, err := shouldSnapshotSource(ctx, src, rep)
 		if err != nil {
-			return nil, errors.Wrapf(err, "unable to get policy tree for source %v", src)
+			return nil, err
 		}
 
-		if src.Host == rep.ClientOptions().Hostname &&
-			src.UserName == rep.ClientOptions().Username &&
-			!policy.IsManualSnapshot(policyTree) {
+		if includeSource {
 			result = append(result, src.Path)
 		}
 	}
 
 	return result, nil
+}
+
+func shouldSnapshotSource(ctx context.Context, src snapshot.SourceInfo, rep repo.Repository) (bool, error) {
+	policyTree, err := policy.TreeForSource(ctx, rep, src)
+	if err != nil {
+		return false, errors.Wrapf(err, "unable to get policy tree for source %v", src)
+	}
+
+	return src.Host == rep.ClientOptions().Hostname &&
+		src.UserName == rep.ClientOptions().Username &&
+		!policy.IsManualSnapshot(policyTree), nil
 }
 
 func init() {
--- a/cli/command_snapshot_create.go
+++ b/cli/command_snapshot_create.go
@@ -290,7 +290,16 @@ func getLocalBackupPaths(ctx context.Context, rep repo.Repository) ([]string, er
 	var result []string
 
 	for _, src := range sources {
-		if src.Host == rep.ClientOptions().Hostname && src.UserName == rep.ClientOptions().Username {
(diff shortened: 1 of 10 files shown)


jkowalski:
can we call the policy flag `manual` and put it in `scheduling` ?

julio-lopez:
#709

pavannd1:
UI and source manager changes are WIP. I'll update the PR when ready

snapshot/policy/scheduling_policy.go
```diff
@@ -72,6 +77,49 @@ func (p *SchedulingPolicy) Merge(src SchedulingPolicy) {
...
+	if err := SetPolicy(ctx, rep, sourceInfo, p); err != nil {
+		return errors.Wrapf(err, "can't save policy for %v", sourceInfo)
+	}
+
+	return nil
+}
+
+// ValidateSchedulingPolicy returns an error if manual field is set along with scheduling fields.
+func ValidateSchedulingPolicy(p SchedulingPolicy) error {
+	if p.IntervalSeconds != 0 && p.Manual {
```
jkowalski, line 114: idea: this maybe more generic and future-proof:

if p.Manual && p != SchedulingPolicy{Manual: true} {
  return /* manual cannot be combined with other scheduling policies */
}

cli/command_snapshot_create.go
```diff
@@ -290,7 +290,16 @@ func getLocalBackupPaths(ctx context.Context, rep repo.Repository) ([]string, er
...
 	for _, src := range sources {
-		if src.Host == rep.ClientOptions().Hostname && src.UserName == rep.ClientOptions().Username {
+		// add all sources belonging to the repository user@host
+		// ignore sources that have Manual field set to true in the SchedulingPolicy
+		policyTree, err := policy.TreeForSource(ctx, rep, src)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to get policy tree for source %v", src)
+		}
+
+		if src.Host == rep.ClientOptions().Hostname &&
```
jkowalski, line 300: perhaps make a helper here - the same helper will be used in source manager.

snapshot/policy/policy.go
```diff
@@ -82,6 +82,12 @@ func MergePolicies(policies []*Policy) *Policy {
 	return &merged
 }
 
+// ValidatePolicy returns error if the given policy is invalid.
+// Currently, only SchedulingPolicy is validated.
+func ValidatePolicy(pol *Policy) error {
```
jkowalski, line 87: any reason this is not always done in SetPolicy() ?

cli/command_snapshot_create.go
pavannd1: I was testing source manager and realized that it's already skipping sources where the `nextSnapshotTime` is not set. Since we remove `interval` and `timesOfDay` fields from the scheduling policy, the `refreshStatus` will not set any `nextSnapshotTime` for these sources.
Where would the explicit handling happen if required?

jkowalski:
This is great, thank you so much!

snapshot/policy/scheduling_policy.go
```diff
@@ -111,12 +112,8 @@ func SetManual(ctx context.Context, rep repo.RepositoryWriter, sourceInfo snapsh
 
 // ValidateSchedulingPolicy returns an error if manual field is set along with scheduling fields.
 func ValidateSchedulingPolicy(p SchedulingPolicy) error {
-	if p.IntervalSeconds != 0 && p.Manual {
-		return errors.New("invalid scheduling policy: cannot set manual snapshot field with intervals")
-	}
-
-	if len(p.TimesOfDay) > 0 && p.Manual {
-		return errors.New("invalid scheduling policy: cannot set manual snapshot field with times of day")
+	if p.Manual && !reflect.DeepEqual(p, SchedulingPolicy{Manual: true}) {
```
julio-lopez, line 115: It may not matter, but this does not cover the case where `p.TimesOfDay != nil && len(p.TimesOfDay) == 0`.

julio-lopez (commented):
👍 LG +1
🎊 🎉 🎈 🏆

PR #808, opened 2021-01-29
Add password command support

### LINKED ISSUE #807: Password retrieval command
I noticed Kopia stores the repository password in `.config/kopia/repository.config.kopia-password` simply obfuscated in base64, which I find quite impacts my trust on the software.

Having the ability to run a command to get this password would allow for way safer implementations while keeping the logic simple.

Since such a command could be anything, like interacting with TPM or an Yubikey, querying from a remote service etc etc, limitless possibilities could be added.

To simplify the implementation, the `repository.config.kopia-password` file could contain a predetermined syntax (like `CMD:blablabla`) that instructs the software to run this command, outputting the file contents as is in case this syntax isn't present.

kamushadenes:
Hello there,

This adds a primitive password command support to Kopia.

Please not I haven't tested anything as this is quick hack I wrote on GH web interface, so for now I'm marking that as a draft.

Closes #807

### DIFF
--- a/repo/password.go
+++ b/repo/password.go
@@ -8,11 +8,13 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"os/user"
 	"path/filepath"
 	"runtime"
 	"strings"
 
+	"github.com/google/shlex"
 	"github.com/pkg/errors"
 	"github.com/zalando/go-keyring"
 )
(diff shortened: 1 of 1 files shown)


jkowalski:
I'd recommend you take a look at `ClientOptions` in `local_config.go` and try to add password command to the struct. This is persisted when you connect to the repository and constructed in `func connectOptions()` from the user-provided flags.

The name of the config file is passed to `repo.GetPersistedPassword()` - you can simply open it, see if there's a command specified and run use it.

jkowalski:
On second thought, your approach avoids opening the file more than once, so it definitely is simpler...

jkowalski:
On third thought - would be much simpler if we used `KOPIA_PASSWORD_COMMAND` environment variable instead. WDYT?

jkowalski:
@kamushadenes any updates here?

jkowalski:
closing for now, please reopen when ready

julio-lopez:
On UNIX-style platforms this can be done with the following without introducing any changes in kopia.
```sh
KOPIA_PASSWORD=$(command_to_get_password args) kopia repo connect ...
```

PR #809, opened 2021-01-30
Added support for user authentication using user profiles stored in the repository

jkowalski:
This is an alternative to htpasswd-based authentication. Users can be added and manipulated using CLI and are stored in a repository as manifests of type `user`.

### DIFF
--- a/internal/auth/authn_repo.go
+++ b/internal/auth/authn_repo.go
@@ -30,6 +30,9 @@ func (ac *repositoryUserAuthenticator) authenticate(ctx context.Context, rep rep
 	if rep != ac.lastRep {
 		ac.userProfiles = nil
 		ac.lastRep = rep
+
+		// ensure profiles are reloaded below
+		ac.nextRefreshTime = time.Time{}
 	}
 
 	// see if we're due for a refresh and refresh userProfiles map
@@ -44,12 +47,9 @@ func (ac *repositoryUserAuthenticator) authenticate(ctx context.Context, rep rep
 		}
 	}
 
-	u := ac.userProfiles[username]
-	if u == nil {
-		return false
-	}
-
-	return u.IsValidPassword(password)
+	// IsValidPassword can be safely called on nil and the call will take as much time as for a valid user
+	// thus not revealing anything about whether the user exists.
+	return ac.userProfiles[username].IsValidPassword(password)
 }
 
 // AuthenticateRepositoryUsers returns authenticator that accepts username/password combinations
--- a/internal/auth/authn_repo.go
+++ b/internal/auth/authn_repo.go
@@ -1,0 +1,63 @@
+package auth
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/kopia/kopia/internal/clock"
+	"github.com/kopia/kopia/internal/user"
+	"github.com/kopia/kopia/repo"
+)
+
+const defaultProfileRefreshFrequency = 10 * time.Second
+
+type repositoryUserAuthenticator struct {
+	lastRep repo.Repository
+
+	mu                          sync.Mutex
+	nextRefreshTime             time.Time
+	userProfiles                map[string]*user.Profile
+	userProfileRefreshFrequency time.Duration
+}
+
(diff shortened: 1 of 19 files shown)


internal/user/user_profile.go
```diff
@@ -0,0 +1,31 @@
...
+type Profile struct {
+	ManifestID manifest.ID `json:"-"`
+
+	Username     string `json:"username"`
+	PasswordHash string `json:"passwordHash"`
+}
+
+// SetPassword changes the password for a user profile.
+func (p *Profile) SetPassword(password string) error {
+	return p.setPasswordV1(password)
```
julio-lopez, line 19: Q: since the profile is encoded as JSON, does it make sense to explicitly encode the hashing scheme in a JSON field (for example, `"hashVersion" : "v1"`) vs. encoding it in the hash string itself? albeit slightly more verbose.

internal/auth/authn_repo.go
```diff
@@ -0,0 +1,63 @@
...
+			log(ctx).Warningf("unable to load userProfiles map: %v", err)
+		} else {
+			ac.userProfiles = newUsers
+		}
+	}
+
+	u := ac.userProfiles[username]
+	if u == nil {
+		return false
+	}
```
julio-lopez, line 50: Does this reveal whether or not a user exists?
Or should this check a fixed random password against a fixed hash that is guaranteed to fail?

internal/user/user_profile_hash_v1.go
```diff
@@ -0,0 +1,65 @@
...
+		return false
+	}
+
+	if len(data) != v1SaltLength+v1KeyLength {
+		return false
+	}
+
+	salt := data[0:v1SaltLength]
+
+	h := computePasswordHashV1(password, salt)
```
julio-lopez, line 62: Is the password checked on every request? or when a connection is established (and potentially reused for multiple requests)?
Is `scrypt.Key` going to blow up memory consumption for the server?

julio-lopez:
LG. Comments inline.

julio-lopez (commented):
LG, comments & question inline.

internal/auth/authn_repo.go
```diff
@@ -0,0 +1,63 @@
...
+
+// AuthenticateSingleUser returns an Authenticator that only allows one username/password combination.
+func (ac *repositoryUserAuthenticator) authenticate(ctx context.Context, rep repo.Repository, username, password string) bool {
+	ac.mu.Lock()
+	defer ac.mu.Unlock()
+
+	// if the server switched to serving another repository, discard cache.
+	if rep != ac.lastRep {
+		ac.userProfiles = nil
+		ac.lastRep = rep
```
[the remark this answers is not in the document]
jkowalski: good catch

internal/user/user_profile.go
jkowalski: done

internal/auth/authn_repo.go
jkowalski: good point, fixed that

internal/user/user_profile_hash_v1.go
jkowalski: the password will be checked once per stream in GRPC session and once per request in the REST case. Given that we're deprecating the REST server soon, this should practically not matter, since no user is currently using repository-based accounts.

internal/user/user_profile.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// IsValidPassword determines whether the password is valid for a given user.
+func (p *Profile) IsValidPassword(password string) bool {
+	if p == nil {
+		// if the user is invalid, return false but use the same amount of time as when we
+		// compare against valid user to avoid revealing whether the user account exists.
+		return isValidPasswordV1(password, dummyV1HashThatNeverMatchesAnyPassword)
```
julio-lopez, line 26:
```suggestion
		isValidPasswordV1(password, dummyV1HashThatNeverMatchesAnyPassword)

		return false
```

julio-lopez:
@pavannd1 PTAL when you have a chance

Issue #811, opened 2021-01-31
macOS - mount error: cannot locate OSXFUSE


panz3r:
Hi, 

When trying to mount all snapshots as per the doc on macOS Big Sur
```sh
kopia mount all ./tmp/kmnt 
```

I get the following error
```sh
kopia: error: mount error: error creating fuse connection: cannot locate OSXFUSE, try --help
```

`macFuse` is correctly installed and works with other tools.

What can I do? 
Is there a way to mount the snapshots using webDAV (as stated on #5 ) instead?

### Environment

**OS**: `macOS Big Sur (11.1)`
**kopia**: `CLI 0.7.3 build: c3ade4b0ea7a1d20497fcfc4f4aec161f73c9505`
**macFuse**: `4.0.5`

jkowalski:
It seems that something must have changed in macfuse 4.x, version 3.x worked fine, but when I upgraded I started getting the same thing...

rjkroege:
Like @panz3r, I was puzzled by this. For reference:

* MacFuse changed it's API surface [https://github.com/osxfuse/osxfuse/releases/tag/macfuse-4.0.0](https://github.com/osxfuse/osxfuse/releases/tag/macfuse-4.0.0)
* Only MacFuse 4.0.0 and up runs on Apple Silicon
* `bazil.org/fuse` no longer supports MacFuse [commit](https://github.com/bazil/fuse/commit/60eaf8f021ce00e5c52529cdcba1067e13c1c2c2)

At least on Intel Macs, probably the easiest solution is not to upgrade past macfuse 3.* and pin `bazil.org/fuse`?

panz3r:
Hi @rjkroege ,

Thanks for your clarifications, I think that this issue can then be linked to issue #453 .

rjkroege:
@jkowalski #825 is a possible short term workaround for this issue.

jkowalski:
Fixed with #829

panz3r:
Hi @jkowalski ,
Thanks for the fix, you've been really quick 👍 

Any ETA for the next release containing this fix?

PR #825, opened 2021-02-09
Working kopia mount on Apple Silicon

rjkroege:
Work around upstream issues with MacFuse and bazil.org/fuse on Apple
Silicon Macs by using WebDav on darwin arm64. Provides an (at least
temporary) fix for #811.

### DIFF
--- a/internal/mount/mount.go
+++ b/internal/mount/mount.go
@@ -23,4 +23,6 @@ type Options struct {
 	// Allows the mounting over a non-empty directory. The files in it will be shadowed by the freshly created mount.
 	// Supported only on Fuse.
 	FuseAllowNonEmptyMount bool
+	// Use WebDav even on platforms that support FUSE.
+	PreferWebDav bool
 }
--- a/internal/mount/mount.go
+++ b/internal/mount/mount.go
@@ -23,6 +23,6 @@ type Options struct {
 	// Allows the mounting over a non-empty directory. The files in it will be shadowed by the freshly created mount.
 	// Supported only on Fuse.
 	FuseAllowNonEmptyMount bool
-	// Use WebDav even on platforms that support FUSE.
-	PreferWebDav bool
+	// Use WebDAV even on platforms that support FUSE.
+	PreferWebDAV bool
 }
--- a/internal/mount/mount_posix_webdav_helper.go
+++ b/internal/mount/mount_posix_webdav_helper.go
@@ -1,0 +1,62 @@
+package mount
+
+import (
+	"context"
+	"os"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/fs"
+)
+
+type posixWedavController struct {
+	inner      Controller
+	mountPoint string
+	isTempDir  bool
+}
+
+// Directory mounts the given directory using WebDav on macos.
+func newPosixWedavController(ctx context.Context, entry fs.Directory, mountPoint string, isTempDir bool) (Controller, error) {
+	c, err := DirectoryWebDAV(ctx, entry)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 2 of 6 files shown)


jkowalski:
It appears that go-fuse now has support for more recent macfuse releases that support M1. Let me try to understand how hard it would be to port to it.

jkowalski:
> It appears that go-fuse now has support for more recent macfuse releases that support M1. Let me try to understand how hard it would be to port to it.

Can you see if #829 works on M1? I was able to test it successfully on MacFUSE 4.0.5 on Intel Mac.

BTW, I like having a webdav alternative on Mac (perhaps even Linux?), so that folks who don't have or don't want to install FUSE will still get mounting support. Instead of hardcoding this to darwin/arm64, can we make it a flag available on al `darwin` ?

rjkroege:
I can restructure this patch to make it based off a flag. Also: I'll try #829 but probably not until the weekend. Installing Fuse on an M1 is both tedious and maybe destabilizing.

internal/mount/mount_posix_webdav_helper_linux.go
```diff
@@ -0,0 +1,22 @@
...
+	if err := mount.Run(); err != nil {
+		fmt.Printf("Cowardly refusing to run with root permissions. Maybe try \"sudo /usr/bin/mount -t davfs -r %s %s\"\n", url, path)
+	}
+	return nil
+}
+
+func unmountWebDevHelper(path string) error {
+	unmount := exec.Command("/usr/bin/umount",  path)
+	if err := unmount.Run(); err != nil {
+		fmt.Printf("Cowardly refusing to run with root permissions. Maybe try \"sudo /usr/bin/umount %s\"\n",  path)
```
jkowalski, line 19: nit: use log(ctx).Warningf, also why not use the error at all?

internal/mount/mount_posix_webdav_helper.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return nil
+}
+
+func (c posixWedavController) MountPath() string {
+	return c.mountPoint
+}
+
+func (c posixWedavController) Done() <-chan struct{} {
+	// TODO(rjk): Do I need to do something else?
```
jkowalski, line 60: done channel should close when the mount is unmounted, this is fine, I guess.

internal/mount/mount.go
```diff
@@ -23,4 +23,6 @@ type Options struct {
 	// Allows the mounting over a non-empty directory. The files in it will be shadowed by the freshly created mount.
 	// Supported only on Fuse.
 	FuseAllowNonEmptyMount bool
+	// Use WebDav even on platforms that support FUSE.
+	PreferWebDav bool
```
jkowalski, line 27: nit: prefer `preferWebDAV` in golang

jkowalski (commented):
mostly LGTM, still needs couple linter fixes and we can merge.

internal/mount/mount_posix_webdav_helper_linux.go
rjkroege: Fixed.

internal/mount/mount_posix_webdav_helper.go
rjkroege: 🙂 Removed the TODO then.

internal/mount/mount.go
rjkroege: Changed to `PreferWebDAV`. I left it as an exported name because it's set from `cli/command_mount.go`.

rjkroege:
I left the review fixes as a separate commit to make it easier to see what I'd done. I'm not confidant that I understand how this should be squashed into a single commit before merging.

jkowalski:

PR #831, opened 2021-02-13
cache: streamlined flags and cache handling

jkowalski:
- ensure we do one full sweep before closing if cache has been modified - before we would do periodic sweep every minute which would not kick in for very short snapshots, which Kopia does very frequently. This leads to build-up of metadata cache entries (q blobs) that never get cleaned until some long session.
- deprecated some caching-related flags, now cache is always on or off with no way to disable it per invocation.
- reduced default list cache duration from 10min to 30s
- moved blob-list cache to separate subdirectory
- cleaned up cache info output to include blob-list cache parameters
- removed ability to disable cache for per-context (this was only used in 'snapshot verify' codepath)
- added ability to partially clear individual caches via CLI


julio-lopez (approved):
Minor comment that can be addressed separately. Will merge.

PR #835, opened 2021-02-15
Removed manifest manager refresh + server improvements

jkowalski:
This does 4 things as separate commits, but clumping them together since they depend on each other:

1. removed the need for manifest manager refresh. Instead content manager is exposing revision counter that changes on every mutation and every time new index blob is added - manifest manager will refresh itself when the underlying content manager revision has changed.

2. refactored server initialization - needed for test below.

3. added unit tests for the repository server API (both REST and GRPC). This actually uncovered the problem requiring manifest refresh fix. 

4. once the tests were in place, added optimization to avoid sending content payload to the server if the server already has the content with a given ID.

### DIFF
--- a/internal/server/grpc_session.go
+++ b/internal/server/grpc_session.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"encoding/json"
+	"net/http"
 	"runtime"
 	"strings"
 	"sync"
@@ -49,10 +50,6 @@ func (s *Server) authenticateGRPCSession(ctx context.Context) (string, error) {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
 		return "", status.Errorf(codes.PermissionDenied, "metadata not found in context")
-	}
-
-	if s.authenticator == nil {
-		return "", status.Errorf(codes.PermissionDenied, "no authenticator")
 	}
 
 	if u, h, p := md.Get("kopia-username"), md.Get("kopia-hostname"), md.Get("kopia-password"); len(u) == 1 && len(p) == 1 && len(h) == 1 {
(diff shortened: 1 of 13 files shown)


repo/manifest/committed_manifest_manager.go
```diff
@@ -258,15 +247,19 @@ func (m *committedManifestManager) ensureInitialized(ctx context.Context) error
 	m.lock()
 	defer m.unlock()
 
-	if m.initialized {
+	rev := m.b.Revision()
```
julio-lopez, line 250: It may be worth renaming `ensureInitialized` to `maybeRefresh` or something along those lines

internal/server/server_test.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if got, want := len(snaps), wantCount; got != want {
+		t.Fatalf("unexpected number of snapshots: %v, want %v", got, want)
+	}
+}
+
+func must(t *testing.T, err error) {
```
julio-lopez, line 274: nit: what about `mustNoError`, `noError`, `checkNoError`, `verifyNoError`, `ensureNoError` or similar?

internal/server/grpc_session.go
```diff
@@ -51,10 +52,6 @@ func (s *Server) authenticateGRPCSession(ctx context.Context) (string, error) {
 		return "", status.Errorf(codes.PermissionDenied, "metadata not found in context")
 	}
 
-	if s.authenticator == nil {
```
julio-lopez, line 54: Why is this no longer needed?

julio-lopez (approved):

PR #841, opened 2021-02-18
Virtual File System to support stdin source

pavannd1:
### Overview

Added a separate package to implement an in-memory filesystem. This package has helpers to support adding files with `stdin` source.

NOTE: This is based off of the existing `mockfs` package with minor enhancements. We could refactor and reuse a lot of common code. It would involve removing panics and changing dependent tests currently using `mockfs`.

### Test Plan

```
=== RUN   TestAddDir
    directory_test.go:17: Add root directory
    directory_test.go:21: Add sub-directory d1
    directory_test.go:26: Add duplicate sub-directory d1
    directory_test.go:30: Add sub-directory with invalid name /d2
--- PASS: TestAddDir (0.00s)
=== RUN   TestAddAllDirs
    directory_test.go:36: Add root directory
    directory_test.go:40: Add a directory: root/d1
    directory_test.go:46: Add a sub-dir under an existing directory: root/d1/d2
    directory_test.go:52: Add third/fourth level dirs: root/d1/d3/d4
    directory_test.go:59: Add a directory under a file (expect failure): root/f1/d6
--- PASS: TestAddAllDirs (0.00s)
=== RUN   TestAddFile
    directory_test.go:68: Add root directory
    directory_test.go:72: Add file with stdin source: root/f1
    directory_test.go:80: Add file with stdin source at third level: root/d1/f2
    directory_test.go:94: Add file with content at third level: root/d2/f3
--- PASS: TestAddFile (0.00s)
=== RUN   TestAddFileWithStdinSource
    directory_test.go:126: Add a root directory
    directory_test.go:137: Add a file with stdin source
    directory_test.go:142: Open virtual file
    directory_test.go:150: Read virtual file
--- PASS: TestAddFileWithStdinSource (0.00s)
=== RUN   TestNewDirectory
    virtualfs_test.go:25: Root Directory success
    virtualfs_test.go:25: Root directory with `/` in the name
--- PASS: TestNewDirectory (0.00s)
PASS
ok  	github.com/kopia/kopia/fs/virtualfs	0.209s
```

### DIFF
--- a/fs/virtualfs/directory.go
+++ b/fs/virtualfs/directory.go
@@ -1,0 +1,177 @@
+package virtualfs
+
+import (
+	"context"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/fs"
+)
+
+// Directory is a mock in-memory implementation of fs.Directory.
+type Directory struct {
+	entry
+
+	children fs.Entries
+}
+
+var _ fs.Directory = (*Directory)(nil)
+
+// AddDir adds a directory with a given name and permissions.
+func (imd *Directory) AddDir(name string, permissions os.FileMode) (*Directory, error) {
+	subdir := &Directory{
+		entry: entry{
+			name: name,
+			mode: permissions | os.ModeDir,
+		},
+	}
+
+	if err := imd.addChild(subdir); err != nil {
+		return nil, err
+	}
+
+	return subdir, nil
+}
+
+// AddAllDirs creates under imd, all the necessary directories in the pathname, similar to os.MkdirAll.
+func (imd *Directory) AddAllDirs(pathname string, permissions os.FileMode) (subdir *Directory, err error) {
+	p, missing, err := imd.resolveDirs(pathname)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, n := range missing {
+		if p, err = p.AddDir(n, permissions); err != nil {
+			return nil, errors.Wrapf(err, "unable to add sub directory '%s'", n)
+		}
+	}
+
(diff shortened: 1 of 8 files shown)


jkowalski:
i don't think I understand the intended use here. Is this for testing or production?

jkowalski:
I wasn't initially sure what we're trying to achieve so I spoke with Julio about it.

I think we can set the first goal of simple stdin source, which will be backed up as directory with a single virtual file whose contents will come from stdin and later expand to support more virtual files, etc.

There are 3 building blocks we need:

1. New `fs.StreamingFile` interface (which is slightly different than `fs.File` because it does not have a size, etc.) and implementation:

```
func StreamingFileFromReader(r io.Reader) fs.StreamingFile {}
```

2. A `fs.Directory` implementation that's in memory (but should probably be create-only and immutable so no Add(), Remove(), etc.):

```
func StaticDirectory(entries fs.Entries) fs.Directory { }
```

3. The biggest thing is adding support for processing `fs.StreamingFile` in the `snapshotfs.Uploader` (which will be slightly different than regular `fs.File` because in order to determine the size we must drain the reader, so we can't use caching, etc.)

Now to support uploading a single file from stdin (and make it show up as a virtual file in a directory) you can feed the following to the uploader:

```
u := snapshotfs.NewUploader(...)
u.Upload(
  fs.StaticDirectory([]fs.Entries{
    "virtual-file": fs.StreamingInputFromReader(os.Stdin),
  }))
```

Later we can evolve this in a number of directions.

pavannd1:
Thanks for the feedback. Will simplify this.

pavannd1:
Used a different approach for this in #862. Closing this PR.

PR #848, opened 2021-02-21
lint: fixed vet-time-inject and replaced with forbidigo linter

jkowalski:
added `faketime.NewClockTimeWithOffset` and used that to fix flaky
`TestDeleteUnreferencedBlobs` test.

### DIFF
--- a/internal/faketime/faketime.go
+++ b/internal/faketime/faketime.go
@@ -4,6 +4,8 @@ package faketime
 import (
 	"sync/atomic"
 	"time"
+
+	"github.com/kopia/kopia/internal/clock"
 )
 
 // Frozen returns a function that always returns t.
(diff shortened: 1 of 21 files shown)


internal/faketime/faketime.go
```diff
@@ -53,3 +55,28 @@ func (t *TimeAdvance) Advance(dt time.Duration) time.Time {
...
+
+// ClockTimeWithOffset allows controlling the passage of time. Intended to be used in
+// tests.
+type ClockTimeWithOffset struct {
+	offset time.Duration
+}
+
+// NewClockTimeWithOffset creates a ClockTimeWithOffset with the given start time.
+func NewClockTimeWithOffset(offset time.Duration) *ClockTimeWithOffset {
+	return &ClockTimeWithOffset{}
```
julio-lopez, line 67: @jkowalski Was this supposed to be the following?

```go
	return &ClockTimeWithOffset{offset: offset}
```

PR #850, opened 2021-02-23
Path lock utility for coordination between robustness engine actions

redgoat650:
Add a utility to ensure path-based synchronization between goroutines. If a path is locked, a subsequent Lock will block if the requested path is the same, or a child/parent (recursive), of the locked path.

This assists with coordination between asynchronous robustness engine actions that may rely on the underlying data directory remaining unchanged. For example:

- between gathering a filesystem fingerprint and taking a snapshot.
- when one WriteFilesAtDepth command has traversed into a directory that another goroutine has picked for deletion.

### DIFF
--- a/tests/robustness/pathlock/path_lock.go
+++ b/tests/robustness/pathlock/path_lock.go
@@ -57,7 +57,7 @@ func (l *lock) Unlock() {
 }
 
 // busyCounter is for unit testing, to determine whether a Lock has been
-// called and blocked
+// called and blocked.
 var busyCounter uint64
 
 // Lock will lock the given path, preventing concurrent calls to Lock
--- a/tests/robustness/pathlock/path_lock.go
+++ b/tests/robustness/pathlock/path_lock.go
@@ -1,0 +1,160 @@
+// Package pathlock defines a PathLocker interface and an implementation
+// that will synchronize based on filepath.
+package pathlock
+
+import (
+	"path/filepath"
+	"strings"
+	"sync"
+	"sync/atomic"
+)
+
+// Locker is an interface for synchronizing on a given filepath.
+// A call to Lock a given path will block any asynchronous calls to Lock
+// that same path, or any parent or child path in the same sub-tree.
+// For example:
+// 	- Lock path /a/b/c
+//		- Blocks a Lock call for the same path /a/b/c
+//		- Blocks a Lock call for path /a/b or /a
+//		- Blocks a Lock call for path /a/b/c/d
+//		- Allows a Lock call for path /a/b/x
+//		- Allows a Lock call for path /a/x
+type Locker interface {
+	Lock(path string) (Unlocker, error)
+}
+
+// Unlocker unlocks from a previous invocation of Lock().
+type Unlocker interface {
+	Unlock()
+}
+
+var _ Locker = (*pathLock)(nil)
+
+// pathLock is a path-based mutex mechanism that allows for synchronization
+// along subpaths. A call to Lock will block as long as the requested path
+// is equal to, or otherwise in the path of (e.g. parent/child) another path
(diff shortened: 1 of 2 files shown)


tests/robustness/pathlock/path_lock.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// unlock will unlock the given path. It is assumed that Lock
+// has already been called, and that unlock will be called once
+// and only once with the exact path provided to the Lock function.
+func (pl *pathLock) unlock(path string) {
+	pl.mu.Lock()
+	defer pl.mu.Unlock()
+
+	close(pl.lockedPaths[path])
```
julio-lopez, line 144: Since closing a `nil` channel panics, is `pl.lockedPaths[path]` guaranteed to be in the map? that is, not return `nil`? It seems that the invariant holds ATM since `unlock` is only called from `lock.Unlock()`, assuming there are no other callers now, nor in the future.

tests/robustness/pathlock/path_lock.go
```diff
@@ -0,0 +1,160 @@
...
+	delete(pl.lockedPaths, path)
+}
+
+// isInPath is a helper to determine whether one path is
+// either the same as another, or a child path (recursively) of it.
+func isInPath(path1, path2 string) (bool, error) {
```
julio-lopez, line 150: Is there an expected relationship between these path parameters? Are there a more descriptive names for `path1` and `path2` that describes the (expected) relationship?
In other words, which of these is the one that is expected to contain the other one? or, if the function returns true, what is the relationship between path1 and path2? is it that path1 contains path2 or that path2 contains path1 ?

tests/robustness/pathlock/path_lock.go
```diff
@@ -0,0 +1,160 @@
...
+	relFP, err := filepath.Rel(path2, path1)
```
julio-lopez, line 151: Wondering whether the check is simply whether one is the prefix of the other one?
Or maybe I'm missing something.

julio-lopez (approved):
@redgoat650 Minor comments that can be addressed separately.
Merging.

tests/robustness/pathlock/path_lock.go
redgoat650: Yep good point. My thought on that is that a panic may be an appropriate mechanism: Currently the way we would hit that case is calling `pl.Unlock()` twice in a row, which is somewhat analogous to calling `Unlock()` on an unlocked `sync.Mutex`, which also panics. However in this case instead of getting a helpful message like "`sync: unlock of unlocked mutex`" we'd see something unhelpful like `close of nil channel`.

Maybe the best way to handle it is with a nil check and a helpful custom panic message.

tests/robustness/pathlock/path_lock.go
redgoat650: For all intents and purposes you're probably right. However technically `Rel` can handle some filesystem quirks that a simple prefix check wouldn't, for example if we happen to be handling a filesystem where paths aren't case sensitive (even though, yes, in reality we probably won't be anytime soon, and even if we did the caller (in this case the engine code) should already be passing in the same paths with whatever expected case already).

The only reason I came up with that example, by the way, is that I was poking around the `path/filepath` package after reading your comment and found `func HasPrefix(p, prefix string) bool`, which sounds almost exactly like what you suggested. Until I read the description 😄 :

> HasPrefix exists for historical compatibility and should not be used.
Deprecated: HasPrefix does not respect path boundaries and does not ignore case when required.

tests/robustness/pathlock/path_lock.go
redgoat650: Good point yes, that isn't very descriptive as written.

PR #863, opened 2021-03-04
Publish image

julio-lopez:
Includes support for Go 1.16

Also, updates base Alpine image to 3.13


julio-lopez:
opened by accident

PR #864, opened 2021-03-04
Fix robustness engine i/o limit test flake

redgoat650:
Instead of a flaky timing check, calculate the amount of data that was actually written to the fio file tree.


julio-lopez (approved):
Thanks @redgoat650

PR #865, opened 2021-03-04
Path protection between robustness engine FileWriter and Snapshotter

redgoat650:
Protect filesystem subtrees from concurrent manipulation during critical windows if engine actions are called asynchronously. This change provides coordination between the `Snapshotter` and the `FileWriter`. For example, the `FileWriter` should be blocked from perturbing the same directory tree if a Gather-Snapshot is taking place along that tree simultaneously. This will ensure the fingerprint data accumulated during the `Gather` phase will correspond unambiguously to the data included in the snapshot.


tests/tools/fio/fio.go
```diff
@@ -62,8 +64,23 @@ type Runner struct {
...
+
+var _ pathlock.Locker = (*InactivePathLocker)(nil)
+
+// Lock implements the pathlock.Locker interface.
+func (l *InactivePathLocker) Lock(lockPath string) (pathlock.Unlocker, error) {
+	return l, nil
+}
+
+// Unlock satisfies the pathlock.Unlocker interface.
+func (l *InactivePathLocker) Unlock() {}
```
julio-lopez, line 82: nit: `NullPathLocker` follows the convention for other no-op structs in the repo

```suggestion
// NullPathLocker satisfies the pathlock.Locker interface but is a no-op.
type NullPathLocker struct{}

var _ pathlock.Locker = (*NullPathLocker)(nil)

// Lock implements the pathlock.Locker interface.
func (l *NullPathLocker) Lock(lockPath string) (pathlock.Unlocker, error) {
	return l, nil
}

// Unlock satisfies the pathlock.Unlocker interface.
func (l *NullPathLocker) Unlock() {}
```

tests/tools/fio/fio.go
```diff
@@ -136,6 +153,7 @@ func NewRunner() (fr *Runner, err error) {
 				}.WithDirectory(fioWriteBaseDir),
 			},
 		},
+		PathLock: &InactivePathLocker{},
```
julio-lopez, line 156: Missed this one

```suggestion
		PathLock: &NullPathLocker{},
```

julio-lopez (approved):

redgoat650:
Thanks @julio-lopez

PR #870, opened 2021-03-07
Improvements and cleanups for connecting to kopia server

jkowalski:
I was manually testing the robustness of the user experience of connecting to kopia repository server. I tried emulating mistakes users might make along the way and verifying that the UX makes sense.

* The biggest issue I've noticed that we were not properly setting up cache when `kopia repository connect server` was used.
* Another notable change is that `kopia cache` commands will be allowed for server-based repositories.
* ~~Also tightened validation of usernames to enforce that they must be lowercase and matching some sane regex.~~ (moved to #873)
* `KOPIA_CACHE_DIRECTORY` environment variable can now be used to globally override the cache path for kopia (both for create/connect and when opening the repository)
* Also made some errors when connecting to the server more descriptive and actually include useful information.

### DIFF
--- a/repo/api_server_repository.go
+++ b/repo/api_server_repository.go
@@ -5,10 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/url"
-	"os"
-	"path/filepath"
 	"time"
 
 	"github.com/pkg/errors"
@@ -40,7 +37,8 @@ type apiServerRepository struct {
 	omgr         *object.Manager
 	wso          WriteSessionOptions
 
-	contentCache *cache.PersistentCache
+	isSharedReadOnlySession bool
+	contentCache            *cache.PersistentCache
 }
 
 func (r *apiServerRepository) APIServerURL() string {
@@ -148,6 +146,7 @@ func (r *apiServerRepository) NewWriter(ctx context.Context, opt WriteSessionOpt
 
 	w.omgr = omgr
 	w.wso = opt
+	w.isSharedReadOnlySession = false
 
 	return w, nil
 }
@@ -195,6 +194,11 @@ func (r *apiServerRepository) WriteContent(ctx context.Context, data []byte, pre
 		return "", errors.Wrapf(err, "error writing content %v", contentID)
 	}
 
+	if prefix != "" {
+		// add all prefixed contents to the cache.
+		r.contentCache.Put(ctx, string(contentID), data)
+	}
+
 	return contentID, nil
 }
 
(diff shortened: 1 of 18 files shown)


repo/local_config.go
```diff
@@ -88,26 +90,34 @@ type repositoryObjectFormat struct {
...
+		}
+	}
 
-// Save writes the configuration to the specified writer.
-func (lc *LocalConfig) Save(w io.Writer) error {
-	b, err := json.MarshalIndent(lc, "", "  ")
+	b, err := json.MarshalIndent(lc2, "", "  ")
 	if err != nil {
-		return nil
+		return errors.Wrap(err, "error writing config file")
```
julio-lopez, line 109: nit: _"error creating config file contents"_ or something to differentiate it from the actual `write` operation.

repo/local_config.go
```diff
@@ -88,26 +90,34 @@ type repositoryObjectFormat struct {
...
 	}
 
-	_, err = w.Write(b)
+	if err = os.MkdirAll(filepath.Dir(filename), 0o700); err != nil {
+		return errors.Wrap(err, "unable to create config directory")
+	}
 
-	return errors.Wrap(err, "error saving local config")
+	return atomicfile.Write(filename, bytes.NewReader(b))
```
julio-lopez, line 116: Should the error be wrapped?
`errors.Wrap` does the right thing when `err == nil`

repo/grpc_repository_client.go
```diff
@@ -586,10 +591,17 @@ func (r *grpcRepositoryClient) UpdateDescription(d string) {
 }
 
 func (r *grpcRepositoryClient) Close(ctx context.Context) error {
+	if r.omgr == nil {
```
julio-lopez, line 594: is this safe with concurrency?

repo/api_server_repository.go
```diff
@@ -204,8 +208,17 @@ func (r *apiServerRepository) UpdateDescription(d string) {
 }
 
 func (r *apiServerRepository) Close(ctx context.Context) error {
-	if err := r.omgr.Close(); err != nil {
-		return errors.Wrap(err, "error closing object manager")
+	if r.omgr != nil {
```
julio-lopez, line 211: Is `Close()` expected to be called concurrently?

repo/grpc_repository_client.go
jkowalski: Close() is meant to be called without concurrency

repo/api_server_repository.go
jkowalski: Close() is meant to be called without concurrency

PR #873, opened 2021-03-07
user: validate that usernames in repository are all-lowercase formatted as user@hostname

jkowalski:


internal/user/user_manager.go
```diff
@@ -97,10 +99,30 @@ func GetUserProfile(ctx context.Context, r repo.Repository, username string) (*P
 	return p, nil
 }
 
+// all-lowercase subset of RFC 1123 without domain name (no dots allowed).
+var validHostnameRegexp = regexp.MustCompile(`^[a-z0-9]([a-z0-9\-]*[a-z0-9])*$`)
+
+// valid user is a superset of valid hostname (it allows _ and .)
+var validUserRegexp = regexp.MustCompile(`^[a-z0-9]([a-z0-9\-_.]*[a-z0-9])*$`)
```
julio-lopez, line 106: @pavannd1 PTAL

pavannd1: Need to verify if this would break some existing assumptions. I'll take a look and update tomorrow 👍🏼

jkowalski: FWIW we actually don't need to be as strict as the RFCs, basically just ensure no troublesome characters sneak in and all names are lowercase, so `[a-z0-9\-_.]+` might be sufficient for both.

pavannd1: Okay, it would be good if we allow `_` and `.`. In the current model, we use dots in the generated hostname. I was looking to start using these commands to manage users and allowing dots will make the transition smoother.

jkowalski:
PTAL - simplified and relaxed regexp to allow underscore, dash and period in both username and password.

internal/user/user_manager_test.go
```diff
@@ -51,16 +51,66 @@ func TestUserManager(t *testing.T) {
...
 	}
 }
 
+func TestValidateUsername_Valid(t *testing.T) {
+	cases := []string{
+		"foo@bar",
+		"foo.foo@bar",
+		"foo.foo@bar.bar",
+		"foo_foo@bar",
+		"foo_foo@bar_bar",
```
julio-lopez, line 70: Maybe add a case with a combination of `.` and `_` in the hostname. By inspection the RE allows it.

```suggestion
		"foo_foo@bar_bar",
		"foo_foo@bar_bar.baz",
		"foo_foo@bar.bar_baz",
		"foo_foo@bar.bar.baz",
```

julio-lopez (approved):
LGTM
@pavannd1 ?

pavannd1 (approved):
LGTM. Thanks!

PR #889, opened 2021-03-15
Add Arch-specific installation instructions

eh8:


julio-lopez:
👍 LG
@jkowalski should approve though.

PR #895, opened 2021-03-18
testing: ensure tests are releasing all buffer pools to reduce memory usage, we had huge leaks

jkowalski:


internal/buf/pool.go
```diff
@@ -158,6 +168,23 @@ func NewPool(ctx context.Context, segmentSize int, poolID string) *Pool {
 // Close closes the pool.
 func (p *Pool) Close() {
 	close(p.closed)
+
+	activePoolsMutex.Lock()
+	delete(activePools, p)
+	activePoolsMutex.Unlock()
+}
+
+// ActivePools returns the number of active activePools.
```
julio-lopez, line 177:
```suggestion
// ActivePools returns the set of active activePools.
```

julio-lopez (approved):

PR #897, opened 2021-03-20
Dockerfile: switched to distroless, specified defaults environment variables for containerized kopia

jkowalski:


tools/docker/Dockerfile
```diff
@@ -1,6 +1,23 @@
 FROM alpine
 ARG TARGETARCH
-RUN apk add --no-cache --verbose ca-certificates && adduser -D kopia && addgroup kopia kopia
+
+# install CA certificates and create user/group 'kopia' with IDs 51515:51515
+RUN apk add --no-cache --verbose ca-certificates && \
```
jeremyje, line 5: What are the certs for?

tools/docker/Dockerfile
```diff
@@ -1,6 +1,23 @@
 FROM alpine
```
jeremyje, line 1: Consider distroless:static-nonroot if this is just a go binary

jeremyje (changes requested):

tools/docker/Dockerfile
```diff
@@ -1,6 +1,23 @@
...
 ARG TARGETARCH
-RUN apk add --no-cache --verbose ca-certificates && adduser -D kopia && addgroup kopia kopia
+
+# install CA certificates and create user/group 'kopia' with IDs 51515:51515
+RUN apk add --no-cache --verbose ca-certificates && \
+  addgroup -g 51515 kopia && \
+  adduser -u 51515 -H -D -G kopia kopia && \
+  mkdir -p /logs /cache /config && \
```
jeremyje, line 8: Kopia should be doing the directory creations in place of it needs them.

jeremyje, line 8: Kopia should be doing the directory creations in place of it needs them.

tools/docker/Dockerfile
```diff
@@ -1,6 +1,23 @@
...
+  chown kopia:kopia /logs /cache /config
```
jeremyje, line 9: Unless there's good reason I'd not create a bunch of directories in the root. 

The script here can be simplified (ideally no RUN statements). That enables miltiarch building

jeremyje (changes requested):

tools/docker/Dockerfile
jkowalski: mostly connecting to S3, Azure and other services on the internet.

tools/docker/Dockerfile
jkowalski: it actually uses one Unix command to set TTY parameters, I thought about it, perhaps if we can remove this dependency, we can move to distroless.

tools/docker/Dockerfile
jkowalski: this is already multiarch for 3 linux archs, Windows won't work, though.

tools/docker/Dockerfile
jkowalski: unless those are under `/home/kopia` (which is ugly and I don't want to create it) it won't have permissions.

jkowalski:
Switched to `gcr.io/distroless/static:nonroot`, PTAL @jeremyje

tools/docker/Dockerfile
jkowalski: fixed that.

tools/docker/Dockerfile
jkowalski: got rid of all `RUN`

jeremyje (approved):

PR #898, opened 2021-03-20
Terminal password prompt - switched to golang.org/x/term

jkowalski:
This required massive upgrade of dependencies for some convoluted reason.


julio-lopez (commented):
LGTM  ✅

PR #913, opened 2021-03-26
blob: fix puts for object-lock-enabled S3 stores

### LINKED ISSUE #711: Can't create a repository on a S3 bucket with object lock and default retention
Hi 

I can't create a repository on a bucket with object lock enabled and default retention.

michael-3-locked is a bucket where I enable object lock and set a default retention period of 1 day.

```
kopia repository create s3 \
        --bucket=michael-3-locked \
        --access-key=AKIAWWWWWWW \
        --secret-access-key=YYYYYYXXXXXXXX

Initializing repository with:
  block hash:          BLAKE2B-256-128
  encryption:          AES256-GCM-HMAC-SHA256
  splitter:            DYNAMIC-4M-BUZHASH
kopia: error: cannot initialize repository: unable to write format blob: unable to write format blob: Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters, try --help
```

It seems that when you need to put object with lock parameters you need this extra Content-MD5 header.

julio-lopez:
Fixes #711

Send content MD5 when object locking retention is enabled.

From the S3 documentation:

>
> If you configure a default retention period on a bucket, requests to upload objects in such a bucket must include the Content-MD5 header.
>
> The Content-MD5 header is required for any request to upload an object with a retention period configured using Amazon S3 Object Lock.
>

Ref:
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-overview.html#object-lock-bucket-config
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html

### DIFF
--- a/repo/blob/s3/s3_storage.go
+++ b/repo/blob/s3/s3_storage.go
@@ -27,7 +27,8 @@ const (
 type s3Storage struct {
 	Options
 
-	cli *minio.Client
+	cli     *minio.Client
+	sendMD5 bool
 
 	downloadThrottler *iothrottler.IOThrottlerPool
 	uploadThrottler   *iothrottler.IOThrottlerPool
@@ -114,7 +115,8 @@ func (s *s3Storage) PutBlob(ctx context.Context, b blob.ID, data blob.Bytes) err
 	}
 
 	uploadInfo, err := s.cli.PutObject(ctx, s.BucketName, s.getObjectNameString(b), throttled, int64(data.Length()), minio.PutObjectOptions{
-		ContentType: "application/x-kopia",
+		ContentType:    "application/x-kopia",
+		SendContentMd5: s.sendMD5,
 	})
 
 	if errors.Is(err, io.EOF) && uploadInfo.Size == 0 {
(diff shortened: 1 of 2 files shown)


jkowalski:
I tried this on Wasabi:

```
failed to open repository: cannot open storage: could not get object locking config for bucket XXX: expected element type <ObjectLockConfiguration> but have <ListBucketResult>
ERROR: open repository: unable to open repository: cannot open storage: could not get object locking config for bucket XXX: expected element type <ObjectLockConfiguration> but have <ListBucketResult>
```

julio-lopez:
@jkowalski 
It seems that Wasabi does not implement the object locking API in an S3-compatible manner.
Some options here:
- Ignore the error and assume the MD5 is not needed
- Special case the error, but this may end up being different for other "S3-compatible" providers.
- Try to determine the S3-provider type (ugh!)

Thoughts?

jkowalski:
can we detect an error at Put() and retry with MD5 and remember this for the remainder of this session? error detection does not need to be perfect - basically if we get request error, retry with MD5 and if that fixes the issue, stick to it.

julio-lopez:
> basically if we get request error, retry with MD5 and if that fixes the issue, stick to it

With this approach, any unrelated request error forces to unnecessarily send the MD5 for the rest of the "blob storage" lifecycle.

Added a patch to only send MD5 when a specific error with the following `er.Message` is observed:
`Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters`

```
minio.ErrorResponse{XMLName:xml.Name{Space:"", Local:"Error"}, Code:"InvalidRequest", Message:"Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters",
// ...
```

jkowalski:
> Added a patch to only send MD5 when a specific error with the following `er.Message` is observed:
> `Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters`

I would search for `strings.Contains(strings.ToLower(er.Message), "content-md5")` to eliminate impact of message changes. Also, would be good to confirm against top S3-compatible providers.

jkowalski (approved):

PR #914, opened 2021-03-26
s3: refactored testing to support multiple S3-compatible providers

jkowalski:
The credentials are passed via JSON matching s3.Options struct
Fixed some verification issue with Wasabi.


repo/blob/s3/s3_storage_test.go
```diff
@@ -107,6 +118,47 @@ func getEnv(name, defValue string) string {
...
+
+	value := getEnvOrSkip(tb, envName)
+
+	var o Options
+	if err := json.NewDecoder(strings.NewReader(value)).Decode(&o); err != nil {
+		tb.Skipf("invalid credentials JSON provided in '%v'", envName)
+	}
+
+	if o.Prefix != "" {
+		tb.Fatalf("options providd in '%v' must not specify a prefix", envName)
```
julio-lopez, line 132:
```suggestion
		tb.Fatalf("options provided in '%v' must not specify a prefix", envName)
```

repo/blob/s3/s3_storage_test.go
```diff
@@ -402,18 +457,22 @@ func createMinioSessionToken(t *testutil.RetriableT, minioEndpoint, kopiaUserNam
 	return *result.Credentials.AccessKeyId, *result.Credentials.SecretAccessKey, *result.Credentials.SessionToken
 }
 
-func cleanupOldData(ctx context.Context, t *testutil.RetriableT, options *Options) {
+func cleanupOldData(ctx context.Context, tb testing.TB, options *Options, cleanupAge time.Duration) {
```
julio-lopez, line 460: question / nit: Should the parameter name `tb` remain the same as before => `t`

julio-lopez (approved):

repo/blob/s3/s3_storage_test.go
jkowalski: the linter flagged it, so I just fixed it.

julio-lopez: I see :(

PR #921, opened 2021-04-03
nit: output snapshot ID when --json is used

jkowalski:


snapshot/manifest.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // Manifest represents information about a single point-in-time filesystem snapshot.
 type Manifest struct {
-	ID     manifest.ID `json:"-"`
+	ID     manifest.ID `json:"id"`
```
julio-lopez, line 18: `omitempty` ?

PR #922, opened 2021-04-03
filesystem: support ~ in repository path, require absolute paths

### LINKED ISSUE #918: handling of directory paths
When you create a repo in your current working directory, without specifying an absolute path. 
eg. `kopia repo create filesystem --path=kopia-repo`
Then when connected to the repo and you change your working directory, trying to do operations on it will give a error stating the directory doesn't exist. 
So kopia should use absolute paths to reference repositories by default (also in the repo config file).

Another thing i found on linux is the ~ alias for the home directory is not seen correctly.
`kopia repo create filesystem --path=~/kopia-repo` will create a `~` directory with `kopia-repo` as a subdirectory in your current location, instead of doing it under `/home/username/kopia-repo`

jkowalski:
Fixes #918

### DIFF
--- a/internal/ospath/ospath.go
+++ b/internal/ospath/ospath.go
@@ -2,7 +2,9 @@
 package ospath
 
 import (
+	"os"
 	"path/filepath"
+	"strings"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


internal/ospath/ospath.go
```diff
@@ -19,3 +21,21 @@ func ConfigDir() string {
 func LogsDir() string {
 	return filepath.Join(userLogsDir, "kopia")
 }
+
+// ResolveUserFriendlyPath replaces ~ in a path with a home directory.
+func ResolveUserFriendlyPath(path string, relativeToHome bool) string {
+	home, _ := os.UserHomeDir()
+	if home != "" && strings.HasPrefix(path, "~") {
+		return home + path[1:]
```
julio-lopez, line 29: This should be:

```go
	if home != "" && strings.HasPrefix(path, "~/") {
		return home + path[2:]
```

PR #941, opened 2021-04-04
cli: output usage to stdout but errors to stderr

### LINKED ISSUE #940: stderr instead of stdout for version and --help
I wonder about the decision that Kopia outputs its help and the result of `--version` on stderr instead of stdout. I stumbled over this while creating some helper scripts. I know how to work around but can't remember any other tool to output those to stderr. Isn't stderr supposed to be used for error messages "only"?

jkowalski:
Fixes #940


julio-lopez (commented):
👍 LG +1

PR #943, opened 2021-04-06
Handle forbidden responses for missing directories

### LINKED ISSUE #551: WebDav backend not working with Synology Webdav
I have set up WebDav on Synology. I can connect and use the share using net use https:/....
When I setup the repository in the UI and pres Connect Repository, the screen changes to "Setup Repository", and I can press Initialize Repository. But first I have to fill the empty Confirm Repository Passowrord. After confirming Repository Password and pressing "Initialize Repository", I get a messagebox with error "failed to create repository..". In the share, a file, kopia.repository.f is created.
Kopia-1

I have tried to find som log files, to show the dialogue between kopia and synology, but havn't found it. If I can get some help to provide needed information, I would like to repeat the process to get the needed information

miquella:
Some WebDAV implementations respond with 403 Forbidden when a missing directory is encountered.

To handle this situation, any time an error is encountered while writing a blob, we first try to create requisite directories and if that fails, we return the error.

Additionally, the original failed write call is immediately tried again (rather than using the retry logic).

Fixes #551.

### DIFF
--- a/repo/blob/webdav/webdav_storage.go
+++ b/repo/blob/webdav/webdav_storage.go
@@ -119,23 +119,30 @@ func (d *davStorageImpl) PutBlobInPath(ctx context.Context, dirPath, filePath st
 	b := buf.Bytes()
 
 	return retry.WithExponentialBackoffNoValue(ctx, "WriteTemporaryFileAndCreateParentDirs", func() error {
-		err := d.translateError(d.cli.Write(tmpPath, b, defaultFilePerm))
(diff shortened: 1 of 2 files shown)


miquella:
This same test has failed in other PRs recently, so I'm guessing I didn't cause this failure. But I'm happy to dig in more if there's something suspect.

jkowalski:
This is great, thank you!. Don't worry about the flake, I'm fixing it in #944

PR #951, opened 2021-04-07
cli: consistent naming for delete command and aliases

julio-lopez:


jkowalski:
Thanks for doing this, as a separate thing we might want to unify the confirmation flags (variously called `--delete`, `--no-dry-run` or absent) depending on the case.

PR #956, opened 2021-04-09
Tagging of kopia snapshots

bathina2:


jkowalski:
i took a quick look - you're on the right track but there's no need to pass tags all the way to the uploader - you can set them just before saving the manifest, similar to how description is saved.

cli/command_snapshot_create.go
```diff
@@ -105,6 +112,18 @@ func runSnapshotCommand(ctx context.Context, rep repo.RepositoryWriter) error {
 	return errors.Errorf("encountered %v errors:\n%v", len(finalErrors), strings.Join(finalErrors, "\n"))
 }
 
+func getTags() (map[string]string, error) {
+	tags := map[string]string{}
+	for _, tagkv := range *snapshotCreateTags {
+		tagFields := strings.Split(tagkv, ":")
```
jkowalski, line 118: nit: you can also do strings.SplitN(tagkv, ":", 2) this will keep the 2nd part unsplit, which allows it to have `:`

PR #979, opened 2021-04-13
encryption: removed old, non-authenticated encryption methods

jkowalski:


repo/content/index_blob_manager_test.go
```diff
@@ -748,7 +748,7 @@ func newIndexBlobManagerForTesting(t *testing.T, st blob.Storage, localTimeNow f
 	t.Helper()
 
 	p := &FormattingOptions{
-		Encryption: encryption.DeprecatedNoneAlgorithm,
```
julio-lopez, line 751: Q: does it make sense to keep "None" for testing? and perhaps have linked only for tests, if possible.

julio-lopez (approved):

repo/content/index_blob_manager_test.go
jkowalski: I was actually thinking of re-adding non-encrypted but with a simple checksum such as crc32. Since NONE was not authenticated at all, that would have to be a whole new format (like `NO-ENCRYPTION-CRC32` or something)

PR #981, opened 2021-04-13
cli: include parameters in maintenance info JSON output

julio-lopez:


cli/command_maintenance_info.go
```diff
@@ -25,7 +25,16 @@ func runMaintenanceInfoCommand(ctx context.Context, rep repo.DirectRepository) e
 	}
 
 	if jsonOutput {
-		printStdout("%s\n", jsonBytes(s))
+		mi := struct {
```
jkowalski, line 28: can we define a named structure here and add an e2e test that parses it?

julio-lopez: addressed feedback

PR #983, opened 2021-04-13
Allow to compile Kopia cli on OpenBSD

ledeuns:
Only CLI is available and mount is not supported.


fs/localfs/local_fs_openbsd.go
```diff
@@ -0,0 +1,15 @@
+package localfs
```
jkowalski, line 1: doesn't `local_fs_nonwindows.go` work here?

jkowalski: you probably just need to tweak build flags:

`local_fs_64bit.go` by adding  !openbsd`
`local_fs_32bit.go` by adding `openbsd`

jkowalski:
also, can you add `.goreleaser.yml` entry similar to freebsd

fs/localfs/local_fs_openbsd.go
ledeuns: Doing that yields : 
```
# github.com/kopia/kopia/fs/localfs
fs/localfs/local_fs_nonwindows.go:26:12: undefined: platformSpecificWidenDev
fs/localfs/local_fs_nonwindows.go:27:13: undefined: platformSpecificWidenDev
```

ledeuns: Thank you, it is fixed.

julio-lopez (commented):
👍 LGTM, but @jkowalski should approve.

.goreleaser.yml
```diff
@@ -26,6 +26,7 @@ archives:
       linux: linux
       windows: windows
       freebsd: freebsd-experimental
+      openbsd: openbsd-experimental
```
jkowalski, line 29: you also want this in `goos` section above. test with `make goreleaser`

ledeuns: `make goreleaser` fails due to `goreleaser` not having an OpenBSD release package :)

PR #988, opened 2021-04-17
mechanical field renames

jkowalski:


internal/grpcapi/repository_server.proto
```diff
@@ -7,7 +7,7 @@ package kopia_repository;
 // corresponds to content.Info
 message ContentInfo {
   string id = 1;
-  uint32 length = 2;
+  uint32 packed_length = 2;
```
julio-lopez, line 10: this is non-breaking because kopia is the only consumer of the proto files ;)

julio-lopez (approved):

PR #991, opened 2021-04-18
Remove "KB" label from processed files count

teepr:
In the Kopia UI "Processed Files" has a "KB" label on the value, which seems incorrect.


jkowalski:
Thanks for the fix!

PR #992, opened 2021-04-18
Be consistent in $(MAKE) usage

ledeuns:
This allows to do `gmake install` on OpenBSD and not fail on `make html-ui` (GNU make is not BSD make)


jkowalski:
Thanks for doing this!

PR #995, opened 2021-04-19
User must install npm manually on OpenBSD

ledeuns:
Node official binary is not released for OpenBSD. User must install it manually from OpenBSD ports or from packages with `pkg_add node`. This patches `tools/tools.mk` to check this in case we run on OpenBSD.

This is the last diff to get a working `make install` on OpenBSD (CLI and embedded UI)

_kopia-ui_ application won't be available anytime soon as it is an Electron app.

### DIFF
--- a/tools/tools.mk
+++ b/tools/tools.mk
@@ -120,19 +120,17 @@ endif
 
 # put NPM in the path
 PATH:=$(node_dir)$(path_separator)$(PATH)
-OPENBSD_NPM_PATH=/usr/local/bin/npm
+ifeq ($(GOOS),$(filter $(GOOS),openbsd freebsd))
+npm=/usr/local/bin/npm
+endif
 
 $(npm):
-
 ifeq ($(GOOS),openbsd)
-# Node do not provide OpenBSD package. User must install from port/package beforehand.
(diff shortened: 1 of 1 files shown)


tools/tools.mk
```diff
@@ -120,8 +120,20 @@ endif
 
 # put NPM in the path
 PATH:=$(node_dir)$(path_separator)$(PATH)
+OPENBSD_NPM_PATH=/usr/local/bin/npm
 
 $(npm):
```
jkowalski, line 125: That will work, but i'm wondering whether it would make more sense to redefine `npm` variable itself to point at `/usr/local/bin/npm` (and not provide the `$(npm): ` rule). Because it's a dependency of targets that require it, if the file is present it will just work, otherwise it will say something like `no rule to make /usr/local/bin/npm`. WDYT?

Should we do the same thing for `freebsd` (not sure how compatible those are)?

ledeuns: You are right, this will be better. I will bake another diff.
I don't know about FreeBSD, I might check later.

ledeuns:
Tested with OpenBSD and FreeBSD.

PR #1013, opened 2021-04-25
Inodes

anicolao:
NOT FOR COMMIT.

Here is the basic sketch of how I think you wanted inode info added to fs.Entry. As I've never written any production go code I needed plenty of Rob's help to put this tiny patch together and I wanted to make sure it was more or less what you were thinking before I got too far along.

We can debate whether or not the same thing should be done for the other types of local_fs entries (directories, symlinks). If directories are handled properly they could be a big win; symlinks I see no value to this change as it won't make anything materially faster for anyone.

As for building the cache in memory, let's debate that after I have a first cut of it. I'd really rather the cache was an index on the `prevEntries []fs.Entries` list that gets passed to `findCachedEntry` so that I could have this work the same way the existing metadata caching works. Would you rather debate that here in the PR or on the thread in the forum?

Thanks
osric

### DIFF
--- a/fs/localfs/local_fs.go
+++ b/fs/localfs/local_fs.go
@@ -95,6 +95,16 @@ type filesystemFile struct {
 	filesystemEntry
 }
 
+type filesystemFilePosix struct {
+	filesystemFile
+
+	inode uint64
+}
+
+func (e *filesystemFilePosix) GetInode() uint64 {
+	return e.inode
+}
+
 type filesystemErrorEntry struct {
 	filesystemEntry
 	err error
@@ -223,13 +233,17 @@ type fileWithMetadata struct {
 	*os.File
 }
 
+func newFileEntry(fi os.FileInfo, parentDir string) fs.Entry {
+	return platformSpecificNewFileEntry(fi, parentDir)
+}
+
 func (f *fileWithMetadata) Entry() (fs.Entry, error) {
 	fi, err := f.Stat()
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to stat() local file")
 	}
 
-	return &filesystemFile{newEntry(fi, filepath.Dir(f.Name()))}, nil
+	return newFileEntry(fi, filepath.Dir(f.Name())), nil
 }
 
 func (fsf *filesystemFile) Open(ctx context.Context) (fs.Reader, error) {
@@ -265,7 +279,7 @@ func NewEntry(path string) (fs.Entry, error) {
 		return &filesystemSymlink{newEntry(fi, filepath.Dir(path))}, nil
 
 	case 0:
-		return &filesystemFile{newEntry(fi, filepath.Dir(path))}, nil
+		return newFileEntry(fi, filepath.Dir(path)), nil
 
 	default:
 		return &filesystemErrorEntry{newEntry(fi, filepath.Dir(path)), fs.ErrUnknown}, nil
@@ -295,7 +309,7 @@ func entryFromChildFileInfo(fi os.FileInfo, parentDir string) fs.Entry {
 		return &filesystemSymlink{newEntry(fi, parentDir)}
 
 	case 0:
-		return &filesystemFile{newEntry(fi, parentDir)}
+		return newFileEntry(fi, parentDir)
 
 	default:
 		return &filesystemErrorEntry{newEntry(fi, parentDir), fs.ErrUnknown}
@@ -305,6 +319,8 @@ func entryFromChildFileInfo(fi os.FileInfo, parentDir string) fs.Entry {
 var (
 	_ fs.Directory  = &filesystemDirectory{}
 	_ fs.File       = &filesystemFile{}
+	_ fs.File       = &filesystemFilePosix{}
+	_ fs.HasInode   = &filesystemFilePosix{}
 	_ fs.Symlink    = &filesystemSymlink{}
 	_ fs.ErrorEntry = &filesystemErrorEntry{}
 )
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -78,6 +78,11 @@ type ErrorEntry interface {
 // ErrEntryNotFound is returned when an entry is not found.
 var ErrEntryNotFound = errors.New("entry not found")
 
+// HasInode is an optional interface implemented by Entry instances that can report their inode.
+type HasInode interface {
+	GetInode() uint64
+}
+
 // ReadDirAndFindChild reads all entries from a directory and returns one by name.
 // This is a convenience function that may be helpful in implementations of Directory.Child().
 func ReadDirAndFindChild(ctx context.Context, d Directory, name string) (Entry, error) {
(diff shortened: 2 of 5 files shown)


fs/localfs/local_fs.go
```diff
@@ -223,13 +234,25 @@ type fileWithMetadata struct {
 	*os.File
 }
 
+func newFileEntry(fi os.FileInfo, parentDir string) fs.Entry {
```
jkowalski, line 237: i think this will need to be moved to OS-specific file, since on Windows there's no `*syscall.Stat_t` at all.

OTOH, on linux/darwin you should be able to assume `fi.Sys()` is always `*syscall.Stat_t`

anicolao: It's your codebase so you can have it how you want it, but the presence or absence of an inode is not a dependency on the platform but on the filesystem. It is in principle possible for the libraries provided on windows to allow reading/writing ext4 and for go to implement *syscall.Stat_t in that case and this code will work on all platforms where the filesystem has the property of being based on inodes and the libraries provided by go are written to correctly report it.

On NTFS, the File ID is pretty much analogous to the inode number and there is an argument to be made that go should support reporting that as the inode and then this would even benefit filesystems where it's not, strictly speaking, an inode number yet serves the same function and has similar enough semantics. if `syscall.Stat_t` used `GetFileInformationByHandle` on Windows, I believe the behaviour would be OK.

So in my view, this implementation is the correct way from `kopia`'s perspective, and then it is up to the go standard library authors to decide what makes sense re: `syscall.Stat_t` in their view.

jkowalski: I was just making a comment that this won't compile on Windows, because `syscall.Stat_t` does not exist there. Look at the `Build / Make (windows-latest)` check. That's how Golang does it - `syscall` package is GOOS-specific.

I think NTFS FileID makes sense to use in this context, it's just getting that information requires different code on Windows than on Linux.

Tip: you can try cross-compiling for Windows `GOOS=windows GOARCH=amd64 go build`

anicolao: Oh! Well I think that's quite wrong of go but that's another matter. Thanks for the tip on cross-compiling, I can see that you're right and it doesn't build with those settings. 

I've not the slightest idea what the setup is here for os-specific code. Looks like there are files named _windows.go and _linux.go, but no such examples in the fs subtree. I'll need a clue as to how to do it so that it fits with the intended build system(s).

jkowalski: There are already windows and non-windows specific files for localfs. Same directory. In golang naming conventions matter, which is probably controversial but works quite well.

anicolao: Doh! I ran `find fs -name '*linux*'` instead of just looking at all the files, and failed to find anything for linux. But I see now it's named `_nonwindows.go`. I'll rewrite this later tonight or tomorrow.

anicolao: OK this I think is an OK fix to the platform problem. Now where should we discuss building the cache?

I would like to do a similar thing to what I've done to local_fs to repofs.go. That is, if the repofs is a snapshot of a filesystem whose entries `HasInode`, then the repofs version of it should also `HasInode`. This will be ignored in any case, as references to the inodes in snapshot uploads will be guarded by a flag that's default off. 

However based on our other discussion I assume you don't like that approach...

jkowalski: This is great so far.

I don't understand the use of storing inodes in repository, we have a file identifier already `oid` which is globally unique. If we want to preserve hardlinks on restore, we can do the reverse process there - build a map of `oid` to inodes of restored files and if we find `oid` reuse we do a hard-link to a previously-written file.

As for building the cache, there are different ways of doing this - I would keep some sort of LRU cache with bounded size that's configurable with a default of zero, but somebody might want to set it to a high value like 10M. The cache/map would be `(dev,inode)-> *snapshot.DirEntry` if we find a match, we double check metadata (size, mtime) and if so, reuse the dir entry.

anicolao: The use of having the inodes in the repo dirEntry is that we can have a cache hit before hashing anything, just like with filenames. So if the user does:
```
$ mv bigtree bigtreeNewHome
```
and snapshots, the snapshot will be instant due to device/inode/metadata hit, as though the pathnames hadn't all changed. The inodes are not useful for restore, and I guess I don't know what the `oid` is -- does it correspond to something in my filesystem?

In my particular case, it's also that I want to run a snapshot on one of my already backed up time machine directories with the goal of populating the inode information quickly. On a small example, this is the difference bettween the initial snapshot hashing everything which takes 90 minutes and having all cache hits which takes 10 minutes. I haven't timed it on my large example.

anicolao: [is oid the internal name of the hash? Then that is not useful for avoiding hashing a newly found file at upload time, while a stored inode is]

anicolao: Also, presumably whatever mechanism solves the cache size problem for filenames also solves it for inodes. That is, if the entire repository snapshot is not loaded into ram, but only the `dirEntry`s that correspond to the currently uploading directory, in a similar way only the `dirEntry` whose inode matches the currently uploading `dirEntry` inode needs to be in memory and so there's no need for a new caching mechanism, just another cache like the existing prevEntries sorted by inode instead of filename. I do not really understand hwo the existing prevEntries cache works, but it would be subject to exactly the same problems as the proposed inode-based cache.

jkowalski: I don't think I understand how this works. Want to hop on slack so we can chat interactively?

jkowalski: I think what you're looking for is some lightweight machine-specific inode->oid cache. prevEntries will not suffice here, because it's only for the current directory, so would not handle anything but trivial renames in one dir.

When it comes to memory usage, we only keep in-memory prevEntries for the directories that are currently "on the stack" during upload, which avoids bloat.

jkowalski:
any updates to this PR?

anicolao:
I'm sorry I got distracted because my schedule suddenly got very full (from now until about June 22). I do mean to get back to this and clean it up as we discussed on slack, but most likely we'll see no progress until at least this coming weekend and possibly longer.

[Meanwhile the for loop approach with retraversing the same filesystem tree inside the various timestamped Time Machine directories worked well enough to unblock the backup task - though that did wind up taking almost 70h of total runtime to complete - reducing the pressure to hack this for my specific case and increasing the pressure to do something generally useful for all use cases.]

jkowalski:
this has been stale for a long time, closing, please rebase and reopen if you intend to pursue this.

anicolao:
Sorry I do plan to return to this. Let's see if I can get to it over the Christmas holiday, and leave it closed until I do some more work on it.

seaniedan:
I would love someone to work on this (I don't know 'Go'!). It's not feasible to use Kopia for backing up Time Machine or other backup tools which use lots of hardlinks, but I'd like to.

PR #1030, opened 2021-04-28
Tagging of kopia snapshots and listing of snapshots by tag

### LINKED ISSUE #950: Ability to tag snapshots and list snapshots by tag.
Tagging snapshots would be a useful way to manage the snapshots

bathina2:
Fixes #950

### DIFF
--- a/cli/command_snapshot_create.go
+++ b/cli/command_snapshot_create.go
@@ -113,6 +113,8 @@ func runSnapshotCommand(ctx context.Context, rep repo.RepositoryWriter) error {
 
 func getTags(tagStrings []string) (map[string]string, error) {
 	numberOfPartsInTagString := 2
+	// tagKeyPrefix is the prefix for user defined tag keys.
+	tagKeyPrefix := "tag:"
 
 	tags := map[string]string{}
 
@@ -122,7 +124,7 @@ func getTags(tagStrings []string) (map[string]string, error) {
 			return nil, errors.New("Invalid tag format. Requires <key>:<value>")
 		}
 
-		key := snapshot.TagKeyPrefix + parts[0]
+		key := tagKeyPrefix + parts[0]
 		if _, ok := tags[key]; ok {
 			return nil, errors.Errorf("Duplicate tag <key> found. (%s)", parts[0])
 		}
--- a/cli/command_snapshot_create.go
+++ b/cli/command_snapshot_create.go
@@ -20,7 +20,6 @@ import (
 const (
 	maxSnapshotDescriptionLength = 1024
 	timeFormat                   = "2006-01-02 15:04:05 MST"
-	numberOfPartsInTagString     = 2
 )
 
 var (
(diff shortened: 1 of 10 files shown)


cli/command_snapshot_create.go
```diff
@@ -20,6 +20,7 @@ import (
 const (
 	maxSnapshotDescriptionLength = 1024
 	timeFormat                   = "2006-01-02 15:04:05 MST"
+	numberOfPartsInTagString     = 2
```
jkowalski, line 23: nit: that type of constant is very local and makes sense to define in the function or next to it, not at the top of the file. Perhaps even not have a constant but //nolint:gomnd because it's always going to be 2 anyway.

snapshot/manager.go
```diff
@@ -31,6 +31,13 @@ const (
 	loadSnapshotsConcurrency = 50 // number of snapshots to load in parallel
 )
 
+var unavailableKeyMap = map[string]struct{}{
```
jkowalski, line 34: instead of restricting, can we have label keys prefixed with `tag:` so all key/value pairs are effectively allowed?

```
{
   "type": "snapshot",
   "username": "alice",
   "hostname": "wonderland",
   "path": "/rabbit/hole",
   "tag:foo": "bar",
   "tag:baz": "qux",
}
```

snapshot/manager.go
```diff
@@ -184,6 +207,10 @@ func ListSnapshotManifests(ctx context.Context, rep repo.Repository, src *Source
 		labels = sourceInfoToLabels(*src)
 	}
 
+	for key, value := range tags {
+		labels[key] = value
```
jkowalski, line 211:
```
labels["tag:"+key] = value
```

see my other comment

jkowalski (changes requested):

snapshot/manager.go
julio-lopez: @jkowalski That also works. I had initially suggested the validation to prevent clobbering _"reserved"_ labels.

snapshot/manifest.go
```diff
@@ -28,6 +28,8 @@ type Manifest struct {
 	RootEntry *DirEntry `json:"rootEntry"`
 
 	RetentionReasons []string `json:"-"`
+
+	Tags map[string]string `json:"tags,omitempty"`
```
julio-lopez, line 32: @jkowalski Q: does it make sense to store the tags here as well (besides the manifest labels)? This is needed mainly to facilitate displaying them when listing snapshots.
Also, if the `tag:` prefix is added to the labels, then it makes sense to strip it from the tags here. Thoughts?

jkowalski: Yes, I would store them in the snapshot manifest (for display purposes) and create corresponding labels when saving snapshots for indexing purposes.

jkowalski:
this is great, thank you!

jkowalski:
looks like e2e tests are failing

bathina2:
AH, looks like I added the the tag an extra time. I call `getTags` on both paths.

PR #1033, opened 2021-04-29
Add documentation for nginx reverse proxy

josvo:
Hello

This PR adds some documentation for how to configure nginx as reverse proxy with kopia server. 
Might not be perfect, but it's a start. :smile: 
Everything I tested works with this setup: Connect repository, backup, restore, browse, ...


jkowalski:
thank you!

PR #1042, opened 2021-05-01
Update full maintenance automation explanation

eh8:
Full maintenance is now enabled by default in >=0.8.0


jkowalski:
Thank you!

PR #1046, opened 2021-05-03
Major CLI refactoring

jkowalski:
This removes almost all global variables in the `cli` package in order to make commands unit-testable, this will let us move some tests from integration tests to unit tests where we can do error injection, etc that's not possible to do in integration tests.

The PR has been structured into several commits:

* 1st commit is completely mechanical and introduces structures for all commands, where each flag is a field and they are all bound inside the `setup()` method to CLI flags. The `app` global variable and shared flags remain global.

* subsequent commits remove all (but 2, those are TBD) global variables.

Next PRs will add actual command-level unit tests.


cli/command_acl.go
```diff
@@ -7,11 +7,11 @@ type commandServerACL struct {
 	list   commandACLList
 }
 
-func (c *commandServerACL) setup(parent commandParent) {
+func (c *commandServerACL) setup(app appServices, parent commandParent) {
```
julio-lopez, line 10: Could `app` be instead `as` or `svc`, since it is not the app, but instead the services used by the app?

cli/app.go
```diff
@@ -43,15 +43,21 @@ type appServices interface {
 	repositoryWriterAction(act func(ctx context.Context, rep repo.RepositoryWriter) error) func(ctx *kingpin.ParseContext) error
 	maybeRepositoryAction(act func(ctx context.Context, rep repo.Repository) error, mode repositoryAccessMode) func(ctx *kingpin.ParseContext) error
 
+	repositoryConfigFileName() string
+	getProgress() *cliProgress
+}
+
+type coreAppServices interface {
```
julio-lopez, line 44: _core_ suggests that it's the basic building block and thus a subset (and not a superset) of `appServices`

julio-lopez (commented):
🥇Thanks for doing this.

See inline comments

jkowalski:
I did the renames, PTAL.

cli/app.go
```diff
@@ -31,46 +31,78 @@ var (
...
-	blobCommands        = app.Command("blob", "Commands to manipulate BLOBs.").Hidden()
-	indexCommands       = app.Command("index", "Commands to manipulate content index.").Hidden()
-	benchmarkCommands   = app.Command("benchmark", "Commands to test performance of algorithms.").Hidden()
-	maintenanceCommands = app.Command("maintenance", "Maintenance commands.").Hidden().Alias("gc")
-	sessionCommands     = app.Command("session", "Session commands.").Hidden()
-	userCommands        = serverCommands.Command("users", "Manager repository users").Alias("user")
-	aclCommands         = serverCommands.Command("acl", "Manager server access control list entries")
 )
 
+type TheApp struct {
```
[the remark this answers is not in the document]
jkowalski: there was the variable called `app`, now it's gone so I renamed `TheApp` -> `App`

cli/command_acl.go
jkowalski: renamed to `svc` everywhere

cli/app.go
jkowalski: renamed to `advancedAppServices`

PR #1051, opened 2021-05-04
robustness: use testing.T for parallel robustness tests

stevejoachim:
Update multi-client robustness test harness to use the parallel `Run` functionality included in the `testing` package.

This fixes an issue where the robustness test could get stuck waiting for a failed test to complete.

### DIFF
--- a/tests/robustness/multiclient_test/framework/harness.go
+++ b/tests/robustness/multiclient_test/framework/harness.go
@@ -6,12 +6,13 @@ import (
 	"context"
 	"errors"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
 	"path"
-	"sync"
 	"syscall"
+	"testing"
 
 	"github.com/kopia/kopia/tests/robustness/engine"
 	"github.com/kopia/kopia/tests/robustness/fiofilewriter"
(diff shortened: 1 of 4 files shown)


tests/robustness/multiclient_test/framework/harness.go
```diff
@@ -207,31 +208,48 @@ func (th *TestHarness) Engine() *engine.Engine {
...
-
-			if cleanup {
-				th.snapshotter.CleanupClient(ctx)
-			}
+func (th *TestHarness) Run( // nolint:thelper
+	ctxs []context.Context,
+	t *testing.T, cleanup bool,
+	f func(context.Context, *testing.T),
+) {
+	t.Helper()
```
jkowalski, line 216: t.Helper() seems counter-productive here as it will hide the details about which line failed, I would ignore the linter error instead

jkowalski: (same thing applies in other places)

stevejoachim: Good point, thanks for reviewing! I just updated with the suggested changes.

julio-lopez:
CI failures fixed in #1059

PR #1053, opened 2021-05-06
cli: plumbed through 'textOutput' which controls stdout/stderr writers

jkowalski:
This is mostly for testability.


julio-lopez (commented):
🥇Thanks for doing this.

PR #1066, opened 2021-05-12
[do not merge] robustness: use testing.T for parallel robustness tests

stevejoachim:
This is a duplicate of #1051, but the changes are based on the commit right before https://github.com/kopia/kopia/commit/d2288c443f89ba80c3c7a0ab7160097866655158. This is an attempt to test out if that commit is causing the `ci-integration-test` failures that we are seeing on #1051. With some local testing, I think I was able to reproduce the failure and bisect to that commit but want to confirm by running it on the CI system on GitHub.


julio-lopez:
CI failure fixed in #1059

PR #1076, opened 2021-05-16
Support for content-level compression

jkowalski:
This adds experimental support for compressing data at content manager level as opposed to object manager.

This will re-wire the snapshot creation pipeline from `read -> split -> compress -> hash -> store`  to `read -> split -> hash -> compress -> store`

This helps address a long-standing problem where different minor versions of compression libraries produce slightly different compressed outputs that hash to completely different values, which can limit deduplication in some cases.

It will also enable some future exciting scenarios like re-compressing contents after the fact, for example during maintenance.

This feature is not enabled by default and must be enabled by passing `--index-version=2` at repository creation time, which also enables richer per-content bookkeeping.

Because of this change you wouldn’t be able to tell if an object is compressed or not by looking at its ID (previously we had `Z` prefix for compressed objects so `Z8dc8296448adf39244fd8242feef1c7b` was compressed but `8dc8296448adf39244fd8242feef1c7b` was not). `kopia content list -c` will show compression information on a per-content basis and `kopia content stats` has been enhanced to provide per-compression statistics.

Also since compression will be done after hashing, it has so be done server-side, thus the bandwidth usage and CPU utilization between kopia client to kopia server may change. 

Other than that those two, this should be a largely-invisible change.

# IMPORTANT:

This feature should be considered experimental until `Kopia v0.9`is released  - backwards compatibility is not guaranteed until then.


cli/command_snapshot_verify.go
```diff
@@ -61,7 +61,7 @@ type verifier struct {
 func (v *verifier) progressCallback(ctx context.Context, enqueued, active, completed int64) {
 	maybeTimeRemaining := ""
 
-	if est, ok := v.tt.Estimate(float64(active), float64(completed)); ok {
+	if est, ok := v.tt.Estimate(float64(completed), float64(enqueued)+float64(completed)); ok {
```
julio-lopez, line 64: Is this related to the _content-level compression_ functionality? could it be merged independently?

cli/show_utils.go
```diff
@@ -90,3 +92,15 @@ func convertTimezone(ts time.Time) time.Time {
 		return ts
 	}
 }
+
+func formatCompressionRatio(original, compressed int64) string {
```
julio-lopez, line 96: nit: this is really space savings or compression percentage. Compression ratio would be `original/compressed`, sometimes noted as `(original/compressed):1`

https://en.wikipedia.org/wiki/Data_compression_ratio

repo/content/committed_read_manager.go
```diff
@@ -336,9 +336,15 @@ func (sm *SharedManager) addRef() {
 
 // release removes a reference to the shared manager and destroys it if no more references are remaining.
 func (sm *SharedManager) release(ctx context.Context) error {
+	if atomic.LoadInt32(&sm.closed) != 0 {
```
julio-lopez, line 339: This is a bug fix, right? Can we merge it ahead by itself in a separate PR?

repo/content/content_manager_test.go
```diff
@@ -46,8 +46,7 @@ func TestMain(m *testing.M) { testutil.MyTestMain(m) }
 func TestContentManagerEmptyFlush(t *testing.T) {
 	ctx := testlogging.Context(t)
 	data := blobtesting.DataMap{}
-	keyTime := map[blob.ID]time.Time{}
-	bm := newTestContentManager(t, data, keyTime, nil)
+	bm := newTestContentManager(t, data)
```
julio-lopez, line 50: it seems that this test refactoring could be (or have been) merged on its own upfront.

tests/end_to_end_test/compression_test.go
```diff
@@ -56,11 +56,42 @@ func TestCompression(t *testing.T) {
...
+		}
+	} else {
+		// with content-level compression we're looking for a content with compression.
+		lines := e.RunAndExpectSuccess(t, "content", "ls", "-c")
+		found := false
+
+		for _, l := range lines {
+			if strings.HasPrefix(l, entries[0].ObjectID) {
+				require.Contains(t, l, "pgzip")
+				found = true
```
julio-lopez, line 77: nit: `break` here?

cli/command_content_list.go
```diff
@@ -88,65 +88,50 @@ func (c *commandContentList) run(ctx context.Context, rep repo.DirectRepository)
...
+	c.out.printStdout("%v length %v packed %v %v %v\n",
+		b.GetContentID(),
+		maybeHumanReadableBytes(c.human, int64(b.GetOriginalLength())),
+		maybeHumanReadableBytes(c.human, int64(b.GetPackedLength())),
+		c.compressionInfoStringString(b),
+		c.deletedInfoString(b),
 	)
+}
 
+func (c *commandContentList) deletedInfoString(b content.Info) string {
```
julio-lopez, line 113: nit: is the receiver needed?
same below.

julio-lopez (commented):
+1: LG 👍
minor comments

cli/command_snapshot_verify.go
jkowalski: done

repo/content/committed_read_manager.go
jkowalski: done

repo/content/content_manager_test.go
jkowalski: done

cli/command_content_list.go
jkowalski: removed

tests/end_to_end_test/compression_test.go
jkowalski: done

PR #1077, opened 2021-05-16
compression: added 'lz4' and 'deflate' methods

jkowalski:


julio-lopez (commented):
🥇Thanks for doing this.

PR #1113, opened 2021-06-02
Add context to robustness store interface

stevejoachim:
Add context to robustness store interface. This is needed for the new persister to use the Kopia API, as many of those functions require a context to be passed in.


stevejoachim:
Meant to open this in `kastenhq/kopia`

PR #1126, opened 2021-06-09
Add a flag to not shallowrestore small files

rjkroege:
When doing a shallow restore, small files might take up less size than
storing the DirectoryEntry metadata. Add a minimum file size flag that
where files below that size will be written directly instead being
represented with shallow placeholders. This improves on #710.


jkowalski:
Thanks!

PR #1129, opened 2021-06-11
Fixed bug in .kopiaignore pattern matching for patterns containing a star.

### LINKED ISSUE #961: Ignore rules puzzle me
This is what's in the filesystem:

```
/var/www/users/user1/logs/some.log
/var/www/users/user1/www/index.html
/var/www/users/user2/logs/another.log
/var/www/users/user2/www/index.html
```

I create a snapshot of `/var/www`

When I add an ignore rule like this:

```
/users/*/logs/
```

Kopia will not exclude anything.

When I add an ignore rule like this:

```
/users/user1/logs/
```

Kopia will exclude the logs of user1.

When I add an ignore rule like this:

```
/users/**/logs/
```

Kopia will exclude the logs of all users.

Why won't it explode all user logs with the single `*` rule?

P.S.: Kopia 0.8.0

alphaleonis:
Integrated some additional test cases to cover this bug.
Fixes #961

### DIFF
--- a/internal/wcmatch/wcmatch.go
+++ b/internal/wcmatch/wcmatch.go
@@ -382,7 +382,8 @@ func doMatch(tokens []token, text []rune, ignoreCase bool) matchResult {
 					return wcNoMatch
 				}
 
-				t.pos = slashIndex
+				// Skip all characters up to the upcoming directory sep.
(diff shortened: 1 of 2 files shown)


jkowalski:
thank you!

PR #1154, opened 2021-06-22
Multiplaces

rjkroege:
Fixes a usability problem noted in #710: `kopia restore a.kopiaentry b.kopiaentry` will now expand both placeholders.

### DIFF
--- a/cli/command_restore.go
+++ b/cli/command_restore.go
@@ -91,7 +91,6 @@ followed by the path of the directory for the contents to be restored.
 	unlimitedDepth = math.MaxInt32
 )
 
-// TODO(rjk): I need to push this concept down through the stack.
 type restoreSourceTarget struct {
 	source        string
 	target        string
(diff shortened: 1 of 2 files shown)


rjkroege:
@jkowalski ptal.

The failures look like flake to me. At least the the Test-UI failure seems like an issue completely unrelated to the PR. If the windows failure is not flake, then it seems very subtle.

snapshot/restore/restore.go
```diff
@@ -92,12 +99,14 @@ func Entry(ctx context.Context, rep repo.Repository, output Output, rootEntry fs
...
 	}
 
-	// Control the depth of a restore. Default (options.MaxDepth = 0) is to restore to full depth.
+	// Control the depth of a restore. Default (options.RestoreDirEntryAtDepth = maxint) is to restore to full depth.
 	currentdepth := int32(0)
 
-	c.q.EnqueueFront(ctx, func() error {
-		return errors.Wrap(c.copyEntry(ctx, rootEntry, "", currentdepth, options.RestoreDirEntryAtDepth, func() error { return nil }), "error copying")
-	})
+	for _, ep := range roots {
```
jkowalski, line 105: i think you need to capture ep for the closure:

```
ep := ep
```

Otherwise you'll be processing the last item in the slice.

snapshot/restore/restore.go
```diff
@@ -75,8 +81,9 @@ type Options struct {
 	Cancel           chan struct{} // channel that can be externally closed to signal cancelation
 }
 
-// Entry walks a snapshot root with given root entry and restores it to the provided output.
-func Entry(ctx context.Context, rep repo.Repository, output Output, rootEntry fs.Entry, options Options) (Stats, error) {
+// Entries restores one or more EntryPair objects using the provided output
+// by walking the EntryPair.root to the EntryPair.dest.
+func Entries(ctx context.Context, rep repo.Repository, output Output, roots []EntryPair, options Options) (Stats, error) {
```
jkowalski, line 86: Do we need do do that? I think this complicates the code by a lot, and it's not obvious to me what the gain is.

I think you're trying to reuse the single Output for multiple restores - would it be simpler to run all this in a sequence and have one `Output` for each restore? In other words - encapsulate all the change in the CLI layer only.

rjkroege: I thought of doing that and went down this rabbit hole of a more complicated approach so that the shallow expansions could all get queued up for parallel execution. The way that you suggest would definitely be simpler. I can do that.

snapshot/restore/restore.go
rjkroege: Oops. But probably moot given your other comment. 🙂

jkowalski:
Thanks for doing this!

PR #1155, opened 2021-06-22
robustness: reduce memory consumption

stevejoachim:
This change introduces a new `robustness.Persister` implementation with the goal of reducing the memory consumption of the test as the collection of snapshot metadata grows over time.

The existing persister pulls all snapshot metadata from object storage at the time the test engine is initialized and loads the data into memory. It then pushes all metadata when the engine is cleaned up.

The new persister pulls and pushes metadata for individual snapshots as needed throughout the test. This way, as the size of metadata for all snapshots grows over time, an individual test run will only keep metadata in memory as needed for the snapshots under test.

### DIFF
--- a/tests/robustness/checker/checker.go
+++ b/tests/robustness/checker/checker.go
@@ -98,11 +98,11 @@ func (ssMeta *SnapshotMetadata) IsDeleted() bool {
 }
 
 // GetSnapshotMetadata gets the metadata associated with the given snapshot ID.
-func (chk *Checker) GetSnapshotMetadata(snapID string) (*SnapshotMetadata, error) {
+func (chk *Checker) GetSnapshotMetadata(ctx context.Context, snapID string) (*SnapshotMetadata, error) {
 	chk.mu.RLock()
 	defer chk.mu.RUnlock()
 
-	return chk.loadSnapshotMetadata(snapID)
+	return chk.loadSnapshotMetadata(ctx, snapID)
 }
 
 // GetLiveSnapIDs gets the list of snapshot IDs being tracked by the checker's snapshot store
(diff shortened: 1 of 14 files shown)


julio-lopez:
Already reviewed by @redgoat650 and @carlbraganza
@stevejoachim CI is failing on mac and ARM64. 

@jkowalski FYI: We'll +2 and merge once the failures are addressed. Please let us know if there are any concerns.

jkowalski:
The failures are known flakes which have recently gotten worse, i'm slowly working on improving test and fixing that.

stevejoachim:
@jkowalski @julio-lopez I think there's an issue with how codecov is calculating the patch coverage.

I added a commit 84c0e3c to increase coverage of the new code in this PR, but the `codecov/patch` number didn't change. It still reports `63.49% of diff hit (target 65.56%)`. See the [before](https://github.com/kopia/kopia/pull/1155/checks?check_run_id=2899068688) and [after](https://github.com/kopia/kopia/pull/1155/checks?check_run_id=2899120762) codecov reports on GitHub actions.

I would have expected it to increase because the results of running 
`go test -v -race -coverpkg=./tests/robustness/snapmeta,./tests/tools/kopiaclient ./tests/robustness/snapmeta`
before and after commit 84c0e3c are 46.6% and 48.8%, respectively. I also used `go tool cover -html=<...>` to confirm that the additional covered lines are part of this PR diff.

It seems like this is because the packages in the `tests` directory are not used for calculating the coverage profile that gets sent to codecov. That makes sense, but then it seems like `codecov/patch` should also ignore a PR if the modified lines are not in the packages that we use to calculate coverage.

julio-lopez (approved):
Thanks @stevejoachim We'll merge and iterate.

PR #1160, opened 2021-06-30
Always restore .kopiaignore

rjkroege:
Always restore .kopiaignore even in shallow restores.

* fixes a bug with the handling of .kopiaignore placeholders (as noted in #710)
* keeps .kopiaignore files applicable when adding files to a partially reified restore tree

### DIFF
--- a/cli/command_restore.go
+++ b/cli/command_restore.go
@@ -17,6 +17,8 @@ import (
 	"github.com/kopia/kopia/internal/timetrack"
 	"github.com/kopia/kopia/internal/units"
 	"github.com/kopia/kopia/repo"
+	"github.com/kopia/kopia/snapshot"
+	"github.com/kopia/kopia/snapshot/policy"
 	"github.com/kopia/kopia/snapshot/restore"
 	"github.com/kopia/kopia/snapshot/snapshotfs"
 )
@@ -113,6 +115,7 @@ type commandRestore struct {
 	restoreIgnoreErrors           bool
 	restoreShallowAtDepth         int32
 	minSizeForPlaceholder         int32
+	dotIgnoreFiles                []string
 
 	restores []restoreSourceTarget
 }
@@ -148,7 +151,7 @@ const (
 	restoreModeTgz           = "tgz"
 )
 
-// constructTargetPairs builds the sourceIdPathPairs array for this
+// constructTargetPairs builds the restoreSourceTarget array for this
 // command for the two forms of command: expansion of one or more
 // placeholders or restoring of a single source to a single destination.
 func (c *commandRestore) constructTargetPairs() error {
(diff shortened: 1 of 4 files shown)


rjkroege:
@jkowalski PTAL. Another fix for a small issue in shallow restores.

snapshot/restore/shallow_fs_output.go
```diff
@@ -64,6 +64,12 @@ func (o *ShallowFilesystemOutput) WriteFile(ctx context.Context, relativePath st
 		return o.FilesystemOutput.WriteFile(ctx, relativePath, f)
 	}
 
+	// Always restore .kopiaignore files so that they correctly apply to
+	// later snapshots.
+	if f.Name() == ".kopiaignore" {
```
jkowalski, line 69: Unfortunately `.kopiaignore` is not hardcoded, by settable by policies and there could be several filenames here.

You probably want to restore all files listed in `FilesPolicy.DotIgnoreFiles` effective for the directory we're restoring to (so when the directory is later snapshotted, we will consider the correct ones). This is actually quite complicated in terms of layering (output probably should not depend on source policies), so we may need to introduce some helper indirection

jkowalski (changes requested):

snapshot/restore/shallow_fs_output.go
rjkroege: Thanks for the insight. I will adjust.

rjkroege:
@jkowalski 

I updated this. I changed it on the assumption that I would want to restore the `DotIgnoreFiles` in the shallow tree in anticipation of the policy in effect during a subsequent `snapshot create`. Please let me know what you think.

cli/command_restore.go
```diff
@@ -343,6 +375,12 @@ func (c *commandRestore) run(ctx context.Context, rep repo.Repository) error {
 			rootEntry = re
 		}
 
+		// TODO(rjk): error handling... plumbing, etc.
+		// I don't think that
+		if err := c.getDotIgnoreFilesFromPolicy(ctx, rep, rstp); err != nil {
```
jkowalski, line 380: So this should work but only partially, because one can have dotignore files as child policies. For example:

foo@bar:/home policy to ignore `.kopiaignore`
foo@bar:/home/jarek policy to ignore `.jarekignore` (which may or may not inherit `.kopiaignore`)

(that means that `/home/rjk` will have `.kopiaignore` while `/home/jarek` will have `.kopiaignore` and `.jarekignore`

IOW i think you need to plumb through `*policy.Tree` instead of statically reading the policy at the root just like during upload.

snapshot/restore/restore.go
```diff
@@ -65,11 +65,12 @@ func (s *Stats) clone() Stats {
...
-	Incremental            bool  `json:"incremental"`
-	IgnoreErrors           bool  `json:"ignoreErrors"`
-	RestoreDirEntryAtDepth int32 `json:"restoreDirEntryAtDepth"`
-	MinSizeForPlaceholder  int32 `json:"minSizeForPlaceholder"`
+	Parallel               int      `json:"parallel"`
+	Incremental            bool     `json:"incremental"`
+	IgnoreErrors           bool     `json:"ignoreErrors"`
+	RestoreDirEntryAtDepth int32    `json:"restoreDirEntryAtDepth"`
+	MinSizeForPlaceholder  int32    `json:"minSizeForPlaceholder"`
+	DotIgnoreFiles         []string `json:"ignoreDotFiles,omitempty"`
```
jkowalski, line 73: is this still needed?

tests/end_to_end_test/shallowrestore_test.go
```diff
@@ -770,6 +781,12 @@ func findFileDir(t *testing.T, shallow string) (dirinshallow, fileinshallow stri
 			continue
 		}
 
+		// .kopiaignore files are always restored and will be real in a shallow
+		// tree. Skip them.
+		if fi.Name() == ".kopiaignore" {
```
jkowalski, line 786: you probably need more test cases taking child policies into account.

jkowalski (changes requested):

jkowalski:
this has been stale for a while, please rebase and open new PR if you intend to pursue this.

PR #1183, opened 2021-07-10
localfs: adjust build flags to support more archs

daniel-eys:
Encountered while trying to update kopia to 0.8.4 in Void Linux, see https://github.com/void-linux/void-packages/pull/30146.
```
# github.com/kopia/kopia/fs/localfs
fs/localfs/local_fs_nonwindows.go:26:41: cannot use stat.Dev (type uint64) as type int32 in argument to platformSpecificWidenDev
fs/localfs/local_fs_nonwindows.go:27:42: cannot use stat.Rdev (type uint64) as type int32 in argument to platformSpecificWidenDev
```

~Should I also add goarch 386 to `.goreleaser.yml`?~


q66:
this should be extended to include at least the following extra archs: ppc64le, ppc64, s390x

daniel-eys:
Hi @jkowalski would you mind having a look into this? Downstream distros currently need to patch around this build issue. Would be nice to have this upstream, thanks!

jkowalski:
I can probably merge this, but Kopia has never been tested on these architectures, so YMMV.

PR #1196, opened 2021-07-13
repo: limit the duration of kopia.repository caching to 15 minutes

jkowalski:
also added flags to override caching duration and disable completely

### DIFF
--- a/cli/command_repository_connect.go
+++ b/cli/command_repository_connect.go
@@ -54,6 +54,9 @@ type connectOptions struct {
 	connectReadonly               bool
 	connectDescription            string
 	connectEnableActions          bool
+
+	formatBlobCacheDuration time.Duration
+	disableFormatBlobCache  bool
 }
 
 func (c *connectOptions) setup(cmd *kingpin.CmdClause) {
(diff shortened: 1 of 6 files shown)


cli/command_repository_connect.go
```diff
@@ -69,6 +72,16 @@ func (c *connectOptions) setup(cmd *kingpin.CmdClause) {
 	cmd.Flag("readonly", "Make repository read-only to avoid accidental changes").BoolVar(&c.connectReadonly)
 	cmd.Flag("description", "Human-readable description of the repository").StringVar(&c.connectDescription)
 	cmd.Flag("enable-actions", "Allow snapshot actions").BoolVar(&c.connectEnableActions)
+	cmd.Flag("repository-format-cache-duration", "Duration of kopia.repository format blob cache").Hidden().DurationVar(&c.formatBlobCacheDuration)
+	cmd.Flag("disable-repository-format-cache", "Disable caching of kopia.repository format blob").Hidden().BoolVar(&c.disableFormatBlobCache)
+}
+
+func (c *connectOptions) getFormatBlobCacheDuration() time.Duration {
+	if c.disableFormatBlobCache {
+		return -1
```
julio-lopez, line 81: Could 0 be used here? and simply change the default behavior from caching for 15 minutes to no caching for already existing repo connections? How often is the format blob consulted after open?

julio-lopez (approved):

cli/command_repository_connect.go
jkowalski: Yeah, I did not want to suddenly change this. We can probably revisit. This blob is only read once on each `Open()`

PR #232, opened 2020-02-16
Kopia UI - lots of new features

jkowalski:
Added ability to create and connect to repository from the UI.

See https://youtu.be/QNZ-xRuHdVo for a screencast

PR #1198, opened 2021-07-14
robustness: serialize client authorization (#232)

stevejoachim:
The robustness tests have been failing intermittently with client authentication errors. I believe this is caused by trying to authorize several clients in parallel.

At a high level, the steps for client authorization include running server user add and server refresh. I think we are running into issues by having multiple clients adding users and refreshing the server in parallel.

This PR serializes client authorization to avoid this potential issue.


jkowalski:
LGTM, but I don't think I understand why this helps. If there is one server and client always refreshes AFTER adding user then the last refresh will always cover all user adds. If this does not help, let's dig deeper.

PR #1199, opened 2021-07-17
Imm/pit versioned store

julio-lopez:


jkowalski:
Thanks for doing this,

My worry here is that by adding abstractions early we may make it harder to actually implement this feature. It could also be that particular choice abstractions will force some less optimal solution than what could be achieved without it.

I would start small by focusing on the S3 first - simply add `PointInTime` parameter to `s3.Options` and implement `GetBlob`, `ListBlobs` and `GetMetadata` directly using minio client without any internal wrappers or interfaces (`VersionedStorage`, the `pit` package), only protected by `if` statements, which I'm hoping will be simple enough.

Alternatively embed `s3Storage` into a wrapper struct - this will give you access to all internal minio stuff, throttlers, etc. without having to go through `VersionedStorage`)

```
type s3VersionedStore struct {
  s3Storage
}

func (s *s3VersionedStore) GetBlob(ctx context.Context, b blob.ID, offset, length int64) ([]byte, error) {
}

func (s *s3VersionedStore) ListBlobs(ctx context.Context, prefix blob.ID, callback func(blob.Metadata) error) error {
}

func (s *s3VersionedStore) GetMetadata(ctx context.Context, b blob.ID) (blob.Metadata, error) {
}
```

Then in `New()`:

```
if (opt.PointInTime != nil) {
  return readonly.NewWrapper(&s3VersionedStore{&s3Storage{...}})
} else {
  return &s3Storage{...}
}
```

Later, when we see enough similarities between S3, GCS, Azure and B2 implementations, we should absolutely move common code to some form of helper library and/or wrapper.

julio-lopez:
While the implementation is equivalent after refactoring, the diff for the change is significant.
Opened #1259 instead with the changes and tests, and it is based of current master.

PR #1209, opened 2021-07-24
force-hash: Allow fractional percentages

moben:
I was passing `--force-hash=0.1` for a while before I realized that this is actually parsed as `0` and the rest is silently discarded.

This could of course be turned into an error but personally I believe that fractional percentages are useful here, especially if you create snapshots very frequently.

I run hourly snapshots and `--force-hash=0.1` if I did my math right this should result in a ~0.5% chance that a given file is not hashed within a month, which feels good enough for my regular background snapshots and keeps the load down quite a bit.


jkowalski:
I've been meaning to do that for a while, thank you!

PR #1286, opened 2021-09-11
content: fixed repo upgrade version

jkowalski:
Previously upgrade would enable epoch manager and index v2 but would
not set the version of the format itself. Everything worked fine
but it would not protect from old kopia opening the repository.


.github/workflows/compat-test.yml
```diff
@@ -0,0 +1,30 @@
+name: Compatibility Test
+on:
+  push:
+    branches: [ master ]
+    tags:
+      - v*
+  pull_request: {}
```
julio-lopez, line 7: @jkowalski Is this intended to run on PRs? Why not just `pull_request:` ?

PR #1293, opened 2021-09-15
cli: added support for passing bare filename for --config-file parameter

jkowalski:
When --`config-file` is passed as a filename without any directory
(absolute or relative) it is resolved in OS-specific
config path.

For example on macOS:

`--config-file foo.config`

resolves to:

`~/Library/Application Support/kopia/foo.config`


julio-lopez (commented):
👍 LG +1 

Not completely sure about automatically adding `.config` when no extension is specified. What is the motivation for disallowing config filenames without an extension?

jkowalski:
> 👍
> LG +1
> 
> Not completely sure about automatically adding `.config` when no extension is specified. What is the motivation for disallowing config filenames without an extension?

Good point. Removed the extension magic.

PR #1316, opened 2021-09-26
add lifecycle example to storage tiers

### LINKED ISSUE #1255: Add documentation how to support tiering on S3 repositories
#1143 introduced some documentation, which is hard for a user to grasp. It would be great if someone could add a working AWS [Lifecycle configuration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/intro-lifecycle-rules.html) to illustrate how to support tiering e.g. on S3.

zyradyl:
This would close #1255


jkowalski:
FYI: We had a conversation with @CrendKing on Slack and the plan is to implement support inside s3 provider to automatically apply the provided storage tier.

zyradyl:
ahh okay, i will go ahead and close this out then. thanks for the heads up!

Issue #908, opened 2021-03-23
Before and After actions


Hoffenbar:
The new actions are great! Unfortunally, in conjunction with snapshots, I've not found a tool yet which creates and mounts a VSS snaphot and then giving control back. E.g., shadowtask and shadowspan are ceating and mounting a snapshot, then running a external program, and then deleting the snapshot (mount). This way it does not work well with kopia actions. I have to snapshot the whole disk, then start e.g. kopia-ui, but for this I don't need actions....

jkowalski:
This VSS is horribly under-documented and it seems that CLI tooling required is not even present on Windows 10 (server does have it, though). I would really be prefer to have 1st-class support for this inside Kopia, but I'm not an expert in this area so I'm not even going to attempt it. Contributions are welcome & encouraged.

Assuming you want to go down the road of something like `shadowspawn`, here's a really bad idea to get this to work in the meantime, using only materials in the room:

We need three scripts:

## The `waiter` script

 * write own process identifier (`pid`) to some well-known file, say `c:\temp\waiter-%KOPIA_SNAPSHOT_ID%.txt`
 * sleep forever waiting to be externally killed

## The `before` script:

  * find (how?) an unused drive letter to mount to (say `Q:`)
  * start `shadowspawn %KOPIA_SOURCE_PATH% %DRIVE_LETTER%: waiter` in the background but don't wait for it
  * wait for `c:\temp\waiter-%KOPIA_SNAPSHOT_ID%.txt` to become available (say no more than 60 seconds). This is an indication that `waiter` script is running and %DRIVE_LETTER% has been mounted
  * echo "KOPIA_SNAPSHOT_PATH=%DRIVE_LETTER%"

## The `after` script:

  * read `c:\temp\waiter-%KOPIA_SNAPSHOT_ID%.txt` and kill the process whose ID is in the file.
  * This will cause `shadowspawn` to unmount the %DRIVE_LETTER% and clean up the snapshot
  * delete `c:\temp\waiter-%KOPIA_SNAPSHOT_ID%.txt`

jkowalski:
Yeah, that was horrible. I did some research and it may be possible to write a proper script, though (cmd or powershell should be possible)

. Here's what I found so far:

Step 1:

```
wmic shadowcopy call create Volume=C:\
```

this prints a line containing something like:


```
        ShadowID = "{EF415FB9-F23A-477D-9A68-C8BDD0BB4221}";
```


Step 2:

```
[...]

PR #1319, opened 2021-09-29
Add example of Windows shadow copy to Actions doc

CrendKing:
Issue https://github.com/kopia/kopia/issues/908.

Just tested locally and works as expected. However, I have some questions:

1. Is there a way to show the actual error message if the action fails for some reason? For example, it could be useful if user knows the snapshot is failed due to Windows unable to take shadow copy in before_action.ps1.
2. When the snapshot is redirected to the shadow copy with the `KOPIA_SNAPSHOT_PATH=<path>` output in before action, I notice that the ignore rules defined for that directory no longer works. Is it expected?

For example, suppose I have C:\test\abc\. In the policy of C:\test, I ignore `abc/`. When snapshot C:\test directly, abc isn't backed up. However, if I take shadow copy and instead snapshot D:\VSS\a4823kd294\test, abc is backed up. I tried several variants such as `**/abc/` and they don't work.

3. The existing example for ZFS, we have `echo KOPIA_SNAPSHOT_PATH: /mnt/$KOPIA_SNAPSHOT_ID` in the before action. The doc above says `KOPIA_SNAPSHOT_PATH=<new-directory>` is the syntax for redirection. I also tried the colon style in my VSS script and didn't work. I haven't tested if the colon style works in Linux. Should we change it to `echo KOPIA_SNAPSHOT_PATH=/mnt/$KOPIA_SNAPSHOT_ID` anyway for sake of consistency?


jkowalski:
> The existing example for ZFS, we have `echo KOPIA_SNAPSHOT_PATH: /mnt/$KOPIA_SNAPSHOT_ID` in the before action. The doc above says `KOPIA_SNAPSHOT_PATH=<new-directory>` is the syntax for redirection. I also tried the colon style in my VSS script and didn't work. I haven't tested if the colon style works in Linux. Should we change it to `echo KOPIA_SNAPSHOT_PATH=/mnt/$KOPIA_SNAPSHOT_ID` anyway for sake of consistency?

Yes please, this is a typo `=` needs to be used.

jkowalski:
> When the snapshot is redirected to the shadow copy with the `KOPIA_SNAPSHOT_PATH=<path>` output in before action, I notice that the ignore rules defined for that directory no longer works. Is it expected?
> 
> For example, suppose I have C:\test\abc. In the policy of C:\test, I ignore `abc/`. When snapshot C:\test directly, abc isn't backed up. However, if I take shadow copy and instead snapshot D:\VSS\a4823kd294\test, abc is backed up. I tried several variants such as `**/abc/` and they don't work.

That is a bug. Filed #1346

Issue #759, opened 2021-01-03
Add support for s3 storage classes


olfway:
Amazon has different storage classes with different prices

https://aws.amazon.com/s3/storage-classes/

but it seems there is no way to configure s3 storage class in kopia

acl1704:
Are lifecycle rules insufficient for this operation?

olfway:
I guess it's possible to do it with lifecycle rules but it will be easier to specify it in kopia

andrewhampton:
I just discovered kopia on Hacker News today and this was my first question. I have a script that does daily backups using the glacier storage class. Will the delayed access of the glacier storage class work with kopia? In other words, does it store files in s3 that it needs immediate access to?

acl1704:
> 
> 
> I just discovered kopia on Hacker News today and this was my first question. I have a script that does daily backups using the glacier storage class. Will the delayed access of the glacier storage class work with kopia? In other words, does it store files in s3 that it needs immediate access to?

Glacier will work but it'll take more time. You'd need to restore to hot storage before being able to run repository maintenance operations or make a new snapshot. In the past I did just that on Scaleway: restore the whole bucket, run snapshot creation and maintenance, and let lifecyles move it back to glacier. But AWS-specific charges for restoring glacier objects may not be cost effective as another provider.

julio-lopez:
https://aws.amazon.com/s3/storage-classes/intelligent-tiering/ may work for most cases.

PR #1335, opened 2021-10-02
Support setting AWS S3 storage class for all types of blobs

CrendKing:
Recently I started a discussion thread at https://kopia.discourse.group/t/local-metadata-cache-for-incremental-backup/623/ for asking support of storing Kopia backups in AWS S3 Glacier storage classes. Currently there is no way in Kopia to store data into S3 in storage class other than STANDARD, and the recommendation was to [use lifecycle rule](https://github.com/kopia/kopia/issues/759#issuecomment-757327456). While it ultimately works, lifecycle approach comes with several shortcomings. For example, the lifecycle event only triggers at 0:00 UTC, so users will likely have to wait for hours for that to happen. Managing lifecycle rules is also a cost.

Since the library Kopia uses to interact with AWS, minio-go, already supports specifying storage class at `PutObject`, I ask myself why not make use of it. Further, in order to utilize the [cheap price of the Glacier storage classes](https://aws.amazon.com/s3/pricing/) (as low as $1/month/TB) without disrupting normal Kopia operations (because Glacier is write-only), this PR adds option for using a different storage class just for "p" data blobs.

ATM I haven't updated docs of this change, because I'm not sure you guys feel about it. If you guys think it's useful, let me know and I'll continue.

### DIFF
--- a/cli/command_repository_create.go
+++ b/cli/command_repository_create.go
@@ -101,7 +101,7 @@ func (c *commandRepositoryCreate) ensureEmpty(ctx context.Context, s blob.Storag
 func (c *commandRepositoryCreate) runCreateCommandWithStorage(ctx context.Context, st blob.Storage) error {
 	err := c.ensureEmpty(ctx, st)
 	if err != nil {
-		//return errors.Wrap(err, "unable to get repository storage")
+		return errors.Wrap(err, "unable to get repository storage")
 	}
 
 	options := c.newRepositoryOptionsFromFlags()
--- a/cli/command_repository_create.go
+++ b/cli/command_repository_create.go
@@ -101,7 +101,7 @@ func (c *commandRepositoryCreate) ensureEmpty(ctx context.Context, s blob.Storag
 func (c *commandRepositoryCreate) runCreateCommandWithStorage(ctx context.Context, st blob.Storage) error {
 	err := c.ensureEmpty(ctx, st)
 	if err != nil {
-		return errors.Wrap(err, "unable to get repository storage")
+		//return errors.Wrap(err, "unable to get repository storage")
 	}
 
 	options := c.newRepositoryOptionsFromFlags()
--- a/cli/storage_s3.go
+++ b/cli/storage_s3.go
@@ -22,6 +22,8 @@ func (c *storageS3Flags) setup(_ storageProviderServices, cmd *kingpin.CmdClause
 	cmd.Flag("access-key", "Access key ID (overrides AWS_ACCESS_KEY_ID environment variable)").Required().Envar("AWS_ACCESS_KEY_ID").StringVar(&c.s3options.AccessKeyID)
 	cmd.Flag("secret-access-key", "Secret access key (overrides AWS_SECRET_ACCESS_KEY environment variable)").Required().Envar("AWS_SECRET_ACCESS_KEY").StringVar(&c.s3options.SecretAccessKey)
 	cmd.Flag("session-token", "Session token (overrides AWS_SESSION_TOKEN environment variable)").Envar("AWS_SESSION_TOKEN").StringVar(&c.s3options.SessionToken)
+	cmd.Flag("storage-class", "Storage class for all blobs").Default("STANDARD").StringVar(&c.s3options.StorageClass)
+	cmd.Flag("storage-class-data", "Overrides storage class for data blobs when set").Default("").StringVar(&c.s3options.StorageClassDataBlob)
 	cmd.Flag("prefix", "Prefix to use for objects in the bucket").StringVar(&c.s3options.Prefix)
 	cmd.Flag("disable-tls", "Disable TLS security (HTTPS)").BoolVar(&c.s3options.DoNotUseTLS)
 	cmd.Flag("disable-tls-verification", "Disable TLS (HTTPS) certificate verification").BoolVar(&c.s3options.DoNotVerifyTLS)
--- a/cli/storage_s3.go
+++ b/cli/storage_s3.go
@@ -22,8 +22,6 @@ func (c *storageS3Flags) setup(_ storageProviderServices, cmd *kingpin.CmdClause
 	cmd.Flag("access-key", "Access key ID (overrides AWS_ACCESS_KEY_ID environment variable)").Required().Envar("AWS_ACCESS_KEY_ID").StringVar(&c.s3options.AccessKeyID)
 	cmd.Flag("secret-access-key", "Secret access key (overrides AWS_SECRET_ACCESS_KEY environment variable)").Required().Envar("AWS_SECRET_ACCESS_KEY").StringVar(&c.s3options.SecretAccessKey)
 	cmd.Flag("session-token", "Session token (overrides AWS_SESSION_TOKEN environment variable)").Envar("AWS_SESSION_TOKEN").StringVar(&c.s3options.SessionToken)
-	cmd.Flag("storage-class", "Storage class for all blobs").Default("STANDARD").StringVar(&c.s3options.StorageClass)
-	cmd.Flag("storage-class-data", "Overrides storage class for data blobs when set").Default("").StringVar(&c.s3options.StorageClassDataBlob)
 	cmd.Flag("prefix", "Prefix to use for objects in the bucket").StringVar(&c.s3options.Prefix)
 	cmd.Flag("disable-tls", "Disable TLS security (HTTPS)").BoolVar(&c.s3options.DoNotUseTLS)
 	cmd.Flag("disable-tls-verification", "Disable TLS (HTTPS) certificate verification").BoolVar(&c.s3options.DoNotVerifyTLS)
--- a/repo/blob/s3/s3_storage.go
+++ b/repo/blob/s3/s3_storage.go
@@ -37,6 +37,7 @@ type s3Storage struct {
 
 	downloadThrottler *iothrottler.IOThrottlerPool
 	uploadThrottler   *iothrottler.IOThrottlerPool
+	storageConfig     *StorageConfig
 }
 
 func (s *s3Storage) GetBlob(ctx context.Context, b blob.ID, offset, length int64, output *gather.WriteBuffer) error {
@@ -142,10 +143,7 @@ func (s *s3Storage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes) (ve
 		return versionMetadata{}, errors.Wrap(err, "AddReader")
 	}
 
-	var storageClass = s.Options.StorageClass
-	if s.Options.StorageClassDataBlob != "" && b[0:1] == "p" {
-		storageClass = s.Options.StorageClassDataBlob
-	}
+	storageClass := s.storageConfig.getStorageClassForBlobID(b)
 
 	uploadInfo, err := s.cli.PutObject(ctx, s.BucketName, s.getObjectNameString(b), throttled, int64(data.Length()), minio.PutObjectOptions{
 		ContentType:    "application/x-kopia",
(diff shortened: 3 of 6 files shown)


jkowalski (commented):
The implementation in the blob provider looks right, but we need to figure out the UX/flags story:

One thing bothers me here - this flag is per connection so all connections from all clients will need to remember to specify it and if they forget we'll get a mix of storage classes for new blobs. Is that ok? Desired?

I'm wondering whether it makes sense to persist this configuration in the bucket itself. Basically have a well-known object (say `.storageconfig`) which would contain those kinds of per-blob options that should be shared between clients.

It could be a non-encrypted JSON with contents resembling something like this (that particular example is probably too fancy, just showing potential for future expansion):

```json
{
   "blobOptions": [
     {"prefix":"p", "storageClass": "GLACIER" },
     {"prefix":"q", "storageClass": "NEARLINE" }
  ]
}
```

The same basic idea was recently implemented in the `filesystem` repository in the form of `.shards` file.

What do you think? Too much?

cli/storage_s3.go
```diff
@@ -22,6 +22,8 @@ func (c *storageS3Flags) setup(_ storageProviderServices, cmd *kingpin.CmdClause
 	cmd.Flag("access-key", "Access key ID (overrides AWS_ACCESS_KEY_ID environment variable)").Required().Envar("AWS_ACCESS_KEY_ID").StringVar(&c.s3options.AccessKeyID)
 	cmd.Flag("secret-access-key", "Secret access key (overrides AWS_SECRET_ACCESS_KEY environment variable)").Required().Envar("AWS_SECRET_ACCESS_KEY").StringVar(&c.s3options.SecretAccessKey)
 	cmd.Flag("session-token", "Session token (overrides AWS_SESSION_TOKEN environment variable)").Envar("AWS_SESSION_TOKEN").StringVar(&c.s3options.SessionToken)
+	cmd.Flag("storage-class", "Storage class for all blobs").Default("STANDARD").StringVar(&c.s3options.StorageClass)
```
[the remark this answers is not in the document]
CrendKing: When you say "bucket-level default", do you mean some sort of lifecycle rule? I don't think you can set a value in a bucket and all future Put automatically goes into that class. If there is, can you share doc? That'd be the perfect solution.

repo/blob/s3/s3_storage.go
```diff
@@ -142,9 +142,15 @@ func (s *s3Storage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes) (ve
 		return versionMetadata{}, errors.Wrap(err, "AddReader")
 	}
 
+	var storageClass = s.Options.StorageClass
+	if s.Options.StorageClassDataBlob != "" && b[0:1] == "p" {
```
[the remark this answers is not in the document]
CrendKing: An abstraction for blob type is most welcomed. You want me to do it?

And sorry I don't know much about golang. I saw a few places in the repo using `[0:1]` so I copied.

CrendKing:
> .storageconfig

I think bucket-wise storage class makes way more sense than connection-wise. So this config should be created at the `repository create` time right? And then we do not need to store the storage class info in the local .config file.

repo/blob/s3/s3_storage.go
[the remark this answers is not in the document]
jkowalski: > > .storageconfig
> 
> I think bucket-wise storage class makes way more sense than connection-wise. So this config should be created at the `repository create` time right? And then we do not need to store the storage class info in the local .config file.

Perhaps we can even start simpler - how about we won't have any flags (for now), and the provider would read the `.storageconfig` if it exists in the bucket and use it to apply the storage classes? This should unblock your testing and we can figure out the UX later - for now to change the rule one can simply update the file in the bucket, WDYT?

jkowalski:
> > .storageconfig
> 
> I think bucket-wise storage class makes way more sense than connection-wise. So this config should be created at the `repository create` time right? And then we do not need to store the storage class info in the local .config file.

Reposting in the right thread:

Perhaps we can even start simpler - how about we won't have any flags (for now), and the provider would read the .storageconfig if it exists in the bucket and use it to apply the storage classes? This should unblock your testing and we can figure out the UX later - for now to change the rule one can simply update the file in the bucket, WDYT?

CrendKing:
Give me some time to test this out.

jkowalski:
> Give me some time to test this out.

Of course.

CrendKing:
BTW, a few days ago you were asking about how restore works in the context of Glacier class. For the record, since Glacier is write-only, to read object one has to initiate a "restore" request to S3. "Standard" request takes about 12 hours to complete but more expensive. "Bulk" request takes longer but cheaper.

I don't think we need to support "restore" natively in Kopia. User can easily do that in AWS console. However, It'd be helpful if we could print out all the required blob IDs should user want to restore partially from a snapshot. This way user can selectively request AWS just for those blobs, saving money.

repo/blob/s3/s3_storage.go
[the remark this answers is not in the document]
CrendKing: BTW, changing to `b[0]` yields this error during build: `invalid operation: b[0] == "p" (mismatched types byte and untyped string)`

CrendKing:
Implemented the read-only use case per discussion. The problem is, currently it requires the bucket to be completely empty to create repository there, but without cli argument, we can't pass the initial .storageconfig, so either I have to comment out that throw, or we have to accept the initial few files (such as `kopia.repository`) will always on the default storage class.

cli/command_repository_create.go
```diff
@@ -101,7 +101,7 @@ func (c *commandRepositoryCreate) ensureEmpty(ctx context.Context, s blob.Storag
 func (c *commandRepositoryCreate) runCreateCommandWithStorage(ctx context.Context, st blob.Storage) error {
 	err := c.ensureEmpty(ctx, st)
 	if err != nil {
-		return errors.Wrap(err, "unable to get repository storage")
+		//return errors.Wrap(err, "unable to get repository storage")
```
jkowalski, line 104: that's probably not intentional, revert?

repo/blob/s3/s3_storage_config.go
```diff
@@ -0,0 +1,58 @@
...
+	return errors.Wrap(json.NewDecoder(r).Decode(p), "error parsing JSON")
+}
+
+// Save saves the parameters to the provided writer.
+func (p *StorageConfig) Save(w io.Writer) error {
+	return errors.Wrap(json.NewEncoder(w).Encode(p), "error writing JSON")
+}
+
+// Clone returns a clone of sharding parameters.
+func (p *StorageConfig) Clone() *StorageConfig {
```
jkowalski, line 38: clone is probably unused at this point and can be removed?

repo/blob/s3/s3_storage.go
```diff
@@ -306,13 +311,28 @@ func newStorage(ctx context.Context, opt *Options) (*s3Storage, error) {
...
 		Options:           *opt,
 		cli:               cli,
 		sendMD5:           0,
 		downloadThrottler: downloadThrottler,
 		uploadThrottler:   uploadThrottler,
-	}, nil
+		storageConfig:     &StorageConfig{},
+	}
+
+	if _, err = s.GetMetadata(ctx, ConfigName); err == nil {
```
jkowalski, line 323: one round-trip to the server (GetMetadata) can be avoided if you do:

err := s.GetBlob(ConfigName)
if err == nil {
  // Load()
} else if !errors.Is(err, blob.ErrBlobNotFound) {
  return errors.Wrap(err, "error retrieving storage config...")
}

jkowalski (changes requested):
it's almost ready to go, just needs some cleanups.

paul-ri:
> it's almost ready to go, just needs some cleanups.

Hi! I just stumbled on this pull request and added a high level comment on the forum this weekend as I didn't want to add noise in here:
https://kopia.discourse.group/t/local-metadata-cache-for-incremental-backup/623/6

It's about the need for deletion: are these changes supporting cases where a blob ("p" file) needs deleted on Glacier? Is this out of scope? Or am I getting something wrong in how Kopia works? Cheers!

repo/blob/s3/s3_storage.go
jkowalski: BTW you also want to add logic to skip `ConfigName` inside `ListBlobs()` - this way the isEmpty check will be fine. Basically the provider will pretend to be empty if only .storageconfig file is in the bucket.

cli/command_repository_create.go
CrendKing: This is the problem I mentioned above. If user wants to have this .storageconfig in the bucket before the repository is created, so the even the initial blobs are under the storage class treatment, this error here will prevent repository from being created, because it's not empty. Of course, we could tell users "only create .storageconfig AFTER repository is created", but I feel it unnecessarily complicate our story.

Is there a way to just ignore the .storageconfig when checking empty?

repo/blob/s3/s3_storage_config.go
CrendKing: Done

repo/blob/s3/s3_storage.go
CrendKing: Was my original plan, but I didn't know how to check error type. Thanks.

cli/command_repository_create.go
jkowalski: you can change `ListBlobs` and not invoke the callback for this one file. This is probably the simplest way to hide provider-internal files.

jkowalski:
> Implemented the read-only use case per discussion. The problem is, currently it requires the bucket to be completely empty to create repository there, but without cli argument, we can't pass the initial .storageconfig, so either I have to comment out that throw, or we have to accept the initial few files (such as `kopia.repository`) will always on the default storage class.
> 
> I tried to specify the file via cli argument, load its content to initialize, so that the bucket can start empty, then write the file to the bucket after initialization. The problem is that `initialize()` is oblivious of the specialization of the blob storage, and I don't find any appropriate callback in the storage interface. I guess that's why you meant to "concern the UX later".
> 
> Another question. The `PrefixAndShards` struct seems to be a good case for being a hash map (with the prefix as key). Why did you make it an array?

This allows all prefixes of arbitrary length including 0 or 2, which can have some cool applications. 

In kopia 0.9.0 the indexes use two-letter prefixes for new epoch-based indexes (`xe`, `xn`, `xw`, `xr` and `xs`) so that's already a use case.

CrendKing:
@paul-ri Just tested the deletion case. It works.

Here is the .storageconfig I used for test:
```json
{
   "blobOptions": [
     { "prefix":"p", "storageClass": "DEEP_ARCHIVE" },
     { "prefix":"q", "storageClass": "REDUCED_REDUNDANCY" },
     { "prefix":"x", "storageClass": "STANDARD_IA" }
  ]
}
```
Looks like this after snapshoting one file:
before

`snapshot delete` and maintenance:
```
> .\kopia --delete da065521a79e4b4b6142f1712b5f1e92
Deleting snapshot 524e3d3c2795f5d7e1141b80d6d752e5 of <redacted> at 2021-10-03 16:46:04 PDT...
default (64KiB) - allocated 36 chunks freed 35 alive 1 max 5 free list high water mark: 4

> .\kopia maintenance run --full --safety=none
Running full maintenance...
Looking for active contents...
  Processed 0 contents, discovered 0...
Looking for unreferenced contents...
Rewriting contents from short packs...
Found safe time to drop indexes: 2021-10-03 16:47:13.6952277 -0700 PDT m=+1.975550001
Dropping contents deleted before 2021-10-03 16:47:13.6952277 -0700 PDT m=+1.975550001
Looking for unreferenced blobs...
Deleted total 4 unreferenced blobs (17.2 KB)
Cleaned up 0 logs.
Finished full maintenance.
default (64KiB) - allocated 73 chunks freed 72 alive 1 max 9 free list high water mark: 8
```
After:
after

jkowalski:
> It's about the need for deletion: are these changes supporting cases where a blob ("p" file) needs deleted on Glacier? Is this out of scope? Or am I getting something wrong in how Kopia works? Cheers!

This is just a first step - fully figuring out delete / restore story for glacier will require a lot of thinking and likely a lot of code. I think we'll need some way to have two-phase restore and two-phase maintenance where first phase requests things to be available and second actually restores/rewrites/deletes them.

cli/command_repository_create.go
CrendKing: Neat! Done.

jkowalski:
This is LGTM. I will merge once the tests pass.

PR #1337, opened 2021-10-02
cli: linux: read KOPIA_USE_KEYRING env

pawitp:
On Linux, Keyring is disabled by default and `--use-keyring` must be specified on every command-line invocation, which gets quite annoying.

This PR allows enabling keyring from environment variable instead.

Example
```
$ ./kopia snapshot list
Enter password to open repository:
$ export KOPIA_USE_KEYRING=1
$ ./kopia snapshot list
[list of snapshots]
```

P.S. It would be great if you can add a `hacktoberfest-accepted` label so this PR can count for Hacktoberfest. This works even if the main repo doesn't have a `hacktoberfest` topic.

### DIFF
--- a/cli/password_linux.go
+++ b/cli/password_linux.go
@@ -5,5 +5,5 @@ import (
 )
 
 func (c *App) setupOSSpecificKeychainFlags(app *kingpin.Application) {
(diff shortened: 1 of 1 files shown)


jkowalski:
Thanks for the contribution, also added `hacktoberfest` topic to the repo.

PR #1341, opened 2021-10-02
cachefs: make cache read O(1) instead of O(n)

### LINKED ISSUE #1135: Large directory listings on `kopia mount` are very slow (1 entry per second)
(I'm currently benchmarkiing various kopia operations.)

Directory listing on `kopia mount` is slow, e.g. when using `find`: `getdents64()` syscall on a large directory takes 20 seconds per 18 entries.

`strace -T` output lines look like:

```c
getdents64(/kopiamount/.../largedir, /* 18 entries */, 32768) = 1584 <20.273433>
```

Both the original source data to be backed up, and the kopia repository, are on localhost CephFS (kopia `--path` repository). On the source data, `find` runs on `255010` dirents in 8 seconds -- **35000x faster.**

For reference, `cat`ing a large file from `kopia mount` runs at 50 MB/s. At least in theory, large directories are also 

The snapshot that includes this directory is partial, and the `find` is under a `... (checkpoint) ...` in the `kopia mount` -- not sure if that matters. Using kopia v0.8.4 on Linux.

Is there something that could speed up directory listings on `kopia mount`?

Thanks!

pawitp:
Move the wrapping of fs.Entry before caching rather than doing it every time we try to read from the cache.

Fixes #1135

### DIFF
--- a/fs/cachefs/cache.go
+++ b/fs/cachefs/cache.go
@@ -80,7 +80,7 @@ func (c *Cache) remove(e *cacheEntry) {
 // Loader provides data to be stored in the cache.
 type Loader func(ctx context.Context) (fs.Entries, error)
 
-// EntryWrapper allows an fs.Entry to be modified before inserting into the cache
+// EntryWrapper allows an fs.Entry to be modified before inserting into the cache.
 type EntryWrapper func(entry fs.Entry) fs.Entry
 
 // Readdir reads the contents of a provided directory using ObjectID of a directory (if any) to cache
--- a/fs/cachefs/cache.go
+++ b/fs/cachefs/cache.go
@@ -80,13 +80,16 @@ func (c *Cache) remove(e *cacheEntry) {
 // Loader provides data to be stored in the cache.
 type Loader func(ctx context.Context) (fs.Entries, error)
 
+// EntryWrapper allows an fs.Entry to be modified before inserting into the cache
+type EntryWrapper func(entry fs.Entry) fs.Entry
+
 // Readdir reads the contents of a provided directory using ObjectID of a directory (if any) to cache
 // the results.
-func (c *Cache) Readdir(ctx context.Context, d fs.Directory) (fs.Entries, error) {
+func (c *Cache) Readdir(ctx context.Context, d fs.Directory, w EntryWrapper) (fs.Entries, error) {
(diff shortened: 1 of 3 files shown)


jkowalski:
You have minor linter failures (`make lint`) other than that I think it's the right fix. Thanks for your contribution.

jkowalski:
BTW when the cache was first created `fs.Directory` did not support `Child()` which is why `Lookup` is using `Readdir()` + `FindByName()`. 

Currently Child() can be used for lookups by name, but I just took a look and it would not help much because the actual implementation inside `repositoryDirectory` still is implemented in terms of `Readdir()`:

```
func (rd *repositoryDirectory) Child(ctx context.Context, name string) (fs.Entry, error) {
	// nolint:wrapcheck
	return fs.ReadDirAndFindChild(ctx, rd, name)
}
```

Just a though - would it make sense (as a next PR, if you're interested) to also improve caching in `repofs` to avoid reloading the directory there? This might help with restore performance, if I'm not mistaken.

In general thanks for tackling large-N directory sizes. This is great stuff! I personally don't have such examples which is why it was never implemented properly.

pawitp:
I'll need to look more into the code to understand how it's being used.

A question that came from a quick look is if caching a potentially huge directory in `repofs` makes sense or not from a memory usage point of view when we already have a caching layer to do that.

The fact that `Child()` can be slower than `Readdir()` + `FindByName()` is due to caching implementation detail is also quite concerning as someone may change the code to use `Child()` in the future and re-introduce the performance issue.

jkowalski:
In fact as part of working on reducing memory footprint I was planning to do exactly that (switch to using Child()). 

BTW can we add some test here that asserts the number of backend reads for this scenario?

pawitp:
Can you clarify which scenario you're talking about?

This PR does not change the number of backend read at all, as the inefficiency is from re-wrapping all the objects not from backend calls.

If you're talking about the number of reads all the way from `fusefs.go`, I'm not sure how to test that.

Maybe we should have a performance (integration) test that test huge directories. It's a bit too big in scope for me though.

jkowalski:
sounds good, we can merge this one for now, i agree we need some integration test with huge directories, i'm working on setting something like that up (along with other shapes and sizes).

PR #1351, opened 2021-10-03
add conditional to KopiaUI makefile to build macos arm64 and x64 versions

zyradyl:
hello!

i noticed in activity monitor that KopiaUI was running under the Intel architecture, which struck me as odd so i did some investigating. it seems that even though the embedded app is universal, the main app is x86_64 only, which causes it to call the intel version of the embedded library.

```
/Applications/KopiaUI.app/Contents
❯ file MacOS/KopiaUI
MacOS/KopiaUI: Mach-O 64-bit executable x86_64

/Applications/KopiaUI.app/Contents
❯ file Resources/server/kopia
Resources/server/kopia: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
Resources/server/kopia (for architecture x86_64):       Mach-O 64-bit executable x86_64
Resources/server/kopia (for architecture arm64):        Mach-O 64-bit executable arm64
```

i added a couple new lines to the Makefile to create an arm64 release. i tried self signing the resulting app but gatekeeper still argued with me so i wasn't able to test, but running `file` revewaled the KopiaUI file was now arm64.

there is a way to [make universal electron apps](https://github.com/electron/universal) but it would bring the app size up to 700mb or so, which seems excessive.

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -48,6 +48,11 @@ unexport CSC_KEY_PASSWORD
 
 endif
 
+# build x86_64 and apple silicon binaries
+ifeq ($(GOOS),darwin)
+electron_builder_flags+=--x64 --arm64
+endif
+
 dev: node_modules/.up-to-date
 	$(npm) $(npm_flags) run dev
 


jkowalski:
We need to be careful here - that may mess up the auto-update behavior. Let's merge this and see what kinds of artifacts are produced in `kopia-test-builds`.

zyradyl:
ahh i hadn't considered that. i can switch over to the test build once its done and see if i run into any issues? i'd be interested in trying to fix anything that pops up, it just might take me a bit to figure it out. :)

jkowalski:
Looks like https://github.com/kopia/kopia-test-builds/releases/tag/v20211003.0.164029 is out - can you test it out? I just merged another unrelated change that will produce another build in about 30 minutes...

Can you test the following scenarios:

- download arm64 v20211003.0.164029, check for updates, should get the next version of arm64
- download amd64 v20211003.0.164029, check for updates, should get the next version of amd64

Can you let me know which architecture is the `kopia` server running on for each of them?

zyradyl:
on it, also fiddling with trying to make a large universal binary in the event something breaks. the apple silicon version is running well right now, and is triggering the arm64 kopia server too.

zyradyl:
autoupdate works successfully on arm64, and amd64.

the above is from arm64, below is amd64.



while fiddling around with universal binaries, i found out that because the electron app packages the universal server binary, electron-builder fails to merge them (duplicate files). when trying to separate them back out and use `dist_darwin_${arch}` in `package.json`, i realized go calls 64bit Intel amd64, and node looks for x64, so i'm stuck there.

but, it seems like the individual binaries work properly, so i don't think we need to go down that path?

jkowalski:
Amazing, thanks for your help. I'm going to cut rc4 and hopefully 0.9.0 sometime next week.

PR #1352, opened 2021-10-03
Remove unnecessary subdirectory from Windows archive

CrendKing:
A few months ago, I opened https://github.com/kopia/scoop-bucket/issues/1. Didn't seem to gain any traction, so here's my attempt to address the issue.

The issue is that Kopia is the only app I've seen that after the Scoop installation, still has version number towards its binary. For example, if my Scoop app is at C:\Scoop, kopia.exe is set at C:\Scoop\apps\kopia\current\kopia-0.9.0-rc3-windows-x64\kopia.exe. Most other apps would have C:\Scoop\apps\kopia\current\kopia.exe instead. The extra subdirectory is unnecessary.

The problem of this version-based subdirectory is that user can no longer have a stable path to the executable after updates. The stable path is the reason why we have the "current" junction point in Scoop. One may argue that since we already have shim, why still care about the actual path to executable. Well, the shim is not always perfect. It is extra command line layer, which could mess up with things. For example, I have some application that loads an application in a special way with special [STARTUPINFO](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-startupinfow). The shim would drop all the settings.

The fix is simple: just don't package the directory when archiving. I only changed the structure for Windows zip, because I don't know if changing it across the board would trigger errors in other publishing means. Also updated the scoop template.

### DIFF
--- a/tools/scoop-kopia.json.template
+++ b/tools/scoop-kopia.json.template
@@ -1,15 +1,15 @@
 {
     "version": "VERSION",
+    "description": "Fast and secure open source backup.",
+    "homepage": "https://kopia.io/",
+    "license": "Apache-2.0",
     "architecture": {
         "64bit": {
             "url": "https://github.com/SOURCE_REPO/releases/download/vVERSION/kopia-VERSION-windows-x64.zip",
             "bin": [
-                "kopia-VERSION-windows-x64/kopia.exe"
+                "kopia.exe"
             ],
             "hash": "HASH_WINDOWS_AMD64"
         }
-    },
-    "homepage": "https://kopia.io/",
-    "license": "Apache-2.0",
-    "description": "Fast and secure open source backup."
+    }
 }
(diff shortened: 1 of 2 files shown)


tools/scoop-kopia.json.template
```diff
@@ -1,15 +1,15 @@
...
             "bin": [
-                "kopia-VERSION-windows-x64/kopia.exe"
+                "kopia.exe"
             ],
             "hash": "HASH_WINDOWS_AMD64"
         }
-    },
-    "homepage": "https://kopia.io/",
-    "license": "Apache-2.0",
-    "description": "Fast and secure open source backup."
```
jkowalski, line 14: I took a look at scoop tutorial - could we add `"extract_dir": "kopia-VERSION-windows-x64"` ? I really would like to avoid changing the zip file structure as it has benefits.

CrendKing: You are right. Let's do that.

jkowalski:
Awesome!

Issue #1320, opened 2021-09-30
Configurable page size for htmlui?


CrendKing:
Currently the page size of htmlui is [hardcoded at 10](https://github.com/kopia/kopia/blob/d9c1bd0758321eabf501dee50d62034d566f4a68/htmlui/src/Table.js#L66). With lots of backed up items and a large display, it's often to see 10+ pages and huge amount of wasted screen space. I tested a larger page size locally and that drastically improved my user experience. Ideally there could be either `server start` argument or even in-UI setting to change that, but I don't know how to.

For example, if we pursue option one, could we somehow create a temporary "input" json file (e.g. `{ "pageSize": 20 }`) from `server start`, and ask the react app to consume it?

jkowalski:
Can we hardcore it to a larger value?

CrendKing:
It might solve my immediate need, but too large the value would cause the UI to have to scroll up/down on smaller devices. You must've put the page size at 10 originally for a reason, right?

Since there is no one size of clothes for all people, I thought making it configurable is better.

millecentdix:
Hi, a selector on the UI to view 10, 30 or more will be welcome :)

zyradyl:
this can probably be closed now 😃

PR #1355, opened 2021-10-05
add simple selector for #1320

zyradyl:
another small PR for #1320, it's nothing pretty but it does work :)

i also merged the individual page bar into the area between the arrows to keep things a bit neater.

before (taken from KopiaUI):

after:

new bar + selector:

options:

resized page:

example of what it looks like when there are more than 20 pages (had to tweak page size for a demo):

### DIFF
--- a/htmlui/src/Table.js
+++ b/htmlui/src/Table.js
@@ -59,7 +59,8 @@ export default function MyTable({ columns, data }) {
     gotoPage,
     nextPage,
     previousPage,
-    state: { pageIndex },
+    setPageSize,
+    state: { pageIndex, pageSize },
   } = useTable({
     columns,
     data,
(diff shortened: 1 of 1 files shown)


htmlui/src/Table.js
```diff
@@ -59,11 +59,12 @@ export default function MyTable({ columns, data }) {
...
     nextPage,
     previousPage,
-    state: { pageIndex },
+    setPageSize,
+    state: { pageIndex, pageSize },
   } = useTable({
     columns,
     data,
-    initialState: { pageSize: 10 },
+    initialState: { pageSize: 3 },
```
jkowalski, line 67: 10 perhaps?

htmlui/src/Table.js
```diff
@@ -80,13 +81,14 @@ export default function MyTable({ columns, data }) {
...
+      {paginationItems(pageOptions.length, pageIndex+1, gotoPage)}
       <Pagination.Next onClick={() => nextPage()} disabled={!canNextPage} />
       <Pagination.Last onClick={() => gotoPage(pageCount - 1)} disabled={!canNextPage} />
     </Pagination>
-
-    &nbsp;
     <Pagination size="sm" variant="dark">
-      {paginationItems(pageOptions.length, pageIndex+1, gotoPage)}
+    <select value={pageSize} onChange={e => {setPageSize(Number(e.target.value))}}>
+      {[10, 20, 30, 40, 50].map(pageSize => (<option key={pageSize} value={pageSize}>Show {pageSize}</option>))}
```
jkowalski, line 90: would it hurt to add 100 or something?

jkowalski:
This is very cool, we can longer-term work on keeping this persistent across windows and repos, but it's a great start. I have just minor comments.

htmlui/src/Table.js
zyradyl: not at all, i will do that now!

htmlui/src/Table.js
zyradyl: oops, this was put to three just for that last screenshot and i forgot to change it back.

Issue #228, opened 2020-02-14
Test a wide range of unicode characters in file names


julio-lopez:
Currently, many of synthetic source FS trees in tests have a narrow character set for file and directory names. For example, e2e tests construct files using hex digits in the file names.

fiorm:
is this done ? or still open for contributions  ? @julio-lopez

ntolia:
> is this done ? or still open for contributions ? @julio-lopez

@fiorm I can confirm this is still open.

julio-lopez:
Yes, this is still open.

PR #1356, opened 2021-10-05
modify randomName() to generate unicode filenames - #228

zyradyl:
hello, i was looking through the "good first issue" tag to try and find something i could wrap my head around, and i found #228. i'm _very_ new to go so i'm sure this could be done in a better way, but i thought i would give it a shot. i ran through `make test` `make ci-integration-test` and `make ci-test` and everything passed. 

let me know if there's anything i can improve on/any changes to be made. the downside of this is it eliminates the hexadecimal filenames, but i suppose those characters are _technically_ in this set. 😅


jkowalski:
I think this is slightly too aggressive, macOS tests failed - probably because of filesystem limitations - Linux and Windows are ok.

zyradyl:
i kinda expected that to happen, but i am surprised it happened on Mac since that's what i ran my make commands on. i'll have to look at the logs a bit later to satisfy my curiosity 😃 

the other option i was considering was picking maybe 4 languages with different character sets (maybe Russian/Arabic/Japanese/English?) and having it switch between them to generate file names? i think this is probably a bit more realistic than testing the entirety of printable unicode.

that will probably be my project for tonight unless you'd prefer a different process?

Issue #1350, opened 2021-10-03
New fresh repo on OneDrive, using rclone backend and this shows up...


xxxliqu1dxxx:
New fresh repo on OneDrive, using rclone backend and this shows up after my first snapshot:

![image](https://user-images.githubusercontent.com/83290244/135768419-4a94e787-0885-465a-8108-5f85f6767495.png)

Following the prompts I see this:
![image](https://user-images.githubusercontent.com/83290244/135768630-fa70fc9e-ad47-4c6e-bd36-53520e572c69.png)

And then as expected, this shows "as normal"... all kopia files:
![image](https://user-images.githubusercontent.com/83290244/135768658-a3a5e75a-040c-44b8-aa02-e16c3b23ce0d.png)

Another view with the rclone ls command
![image](https://user-images.githubusercontent.com/83290244/135768455-00cd2df8-e2a8-4b77-8602-4ebba42a77c9.png)

I uploaded stuff to OneDrive before and this never showed up. I am not sure if this is new to OneDrive or caused by RC3. As mentioned these are "new OneDrive accounts" which are "Fresh", empty... so perhaps that's the issue, but maybe someone here knows what's wrong.

I am sure this is a false positive as like I said the only stuff on there are kopia files, but I wonder if perhaps OneDrive triggers that ransomware stuff due to the way files are named with Kopia... is there a way to change the filenames or directory structure to avoid this situation? I know there's a --flat parameter which I tried, and this did not seem to do anything for me...

In any case, I will continue to play with this, but I found it interesting that it triggered this false positive. 

Thanks in advance if you can provide an answer :)

xxxliqu1dxxx:
If I create the repo within an rclone crypt, this does not seem to trigger that OneDrive "protection"... so I will use that workaround for now.

zyradyl:
out of curiosity, when you use rclone crypt what do the files in OneDrive look like?

it seems like this is [intended behavior](https://support.microsoft.com/en-us/office/ransomware-detection-and-recovering-your-files-0d90ec50-6bfd-40f4-acc7-b8c12c73637f) at least on OneDrive's end. in particular, the .f file extension would be "strange" to 365, so it gets flagged.

i tried looking for more information on the specifics of the detection, but all i can find is references to mass-deletions, or strange filenames and extensions.

xxxliqu1dxxx:
> 
> 
> out of curiosity, when you use rclone crypt what do the files in OneDrive look like?
> 

[...]

PR #1357, opened 2021-10-06
replace hardcoded instances of '.f' with CompleteBlobSuffix

zyradyl:
while looking through and tinkering with things for #1350, i noticed that some of the test routines rely on a hardcoded value of `CompleteBlobSuffix = ".f"`. this just switches those out with a reference to the variable.

it's very possible the hardcode was intentional, especially considering that changing the suffix would mean needing to implement an upgrade mechanism, but i thought i'd submit a PR just in case it was an oversight. 😅

### DIFF
--- a/cli/command_blob_shards_modify_test.go
+++ b/cli/command_blob_shards_modify_test.go
@@ -20,38 +20,38 @@ func TestBlobShardsModify(t *testing.T) {
 
 	// verify default sharding is 3,3
 	require.FileExists(t, filepath.Join(env.RepoDir, someQBlob[0:3], someQBlob[3:6], someQBlob[6:]+sharded.CompleteBlobSuffix))
-	require.FileExists(t, filepath.Join(env.RepoDir, "kopia.repository.f"))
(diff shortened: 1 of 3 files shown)


cli/command_blob_shards_modify_test.go
```diff
@@ -20,38 +20,38 @@ func TestBlobShardsModify(t *testing.T) {
 
 	// verify default sharding is 3,3
 	require.FileExists(t, filepath.Join(env.RepoDir, someQBlob[0:3], someQBlob[3:6], someQBlob[6:]+sharded.CompleteBlobSuffix))
-	require.FileExists(t, filepath.Join(env.RepoDir, "kopia.repository.f"))
+	require.FileExists(t, filepath.Join(env.RepoDir, "kopia.repository"+sharded.CompleteBlobSuffix))
```
jkowalski, line 23: i think it's ok to hardcode constants like that otherwise we're not ever sure what the filenames are asserted.

zyradyl: ahh, that's very true and something i hadn't considered. alright, i'll keep this as my little branch to fiddle with and close this out. sorry for all the PRs lately 😅

PR #1360, opened 2021-10-06
refactor: move from io/ioutil to io and os package

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since kopia has upgraded to Go 1.17 (#1256), this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

PR #1361, opened 2021-10-06
refactor: move from ioutil.ReadDir to os.ReadDir

Juneezee:
This PR is an addition to #1360 (see https://github.com/kopia/kopia/pull/1360#discussion_r723279310).

Things to note:
1. `os.ReadDir` is not a direct replacement for `ioutil.ReadDir`, unlike all other functions in `io/ioutil`
2. `os.ReadDir` returns a slice of [os.DirEntry](https://golang.org/pkg/os/#DirEntry), while `ioutil.ReadDir` returns a slice of [fs.FileInfo](https://golang.org/pkg/io/fs/#FileInfo). The biggest difference is the lack of `Size()`, `ModTime()`, and `Sys()` method in the `os.DirEntry` interface.
3. To get the `Size()`, `ModTime()`, or `Sys()` of a file or folder from `os.DirEntry`, we need to call `Info()` first and do an additional error handling.

The `os.ReadDir` was proposed in https://github.com/golang/go/issues/41467. According to the [documentation for `ioutil.ReadDir`](https://pkg.go.dev/io/ioutil#ReadDir) starting from Go 1.16, it is recommended to use `os.ReadDir` as it is a more efficient and correct implementation.

### DIFF
--- a/fs/localfs/local_fs.go
+++ b/fs/localfs/local_fs.go
@@ -128,8 +128,8 @@ type entryWithError struct {
 	err   error
 }
 
-func toDirEntryOrNil(basename, prefix string) (fs.Entry, error) {
-	fi, err := os.Lstat(prefix + basename)
+func toDirEntryOrNil(dirEntry os.DirEntry, prefix string) (fs.Entry, error) {
+	fi, err := os.Lstat(prefix + dirEntry.Name())
 	if err != nil {
 		if os.IsNotExist(err) {
 			return nil, nil
@@ -153,7 +153,7 @@ func (fsd *filesystemDirectory) Readdir(ctx context.Context) (fs.Entries, error)
 	var entries fs.Entries
 
 	// read first batch of directory entries using Readdir() before parallelization.
-	firstBatch, firstBatchErr := f.Readdirnames(numEntriesToReadFirst)
+	firstBatch, firstBatchErr := f.ReadDir(numEntriesToReadFirst)
 	if firstBatchErr != nil && !errors.Is(firstBatchErr, io.EOF) {
 		return nil, errors.Wrap(firstBatchErr, "unable to read directory entries")
 	}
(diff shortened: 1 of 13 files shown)


Juneezee:
There are more stuff to change than I expected, such as the `ReadDir` method signature of `sharded.Impl` interface. Will fix them ASAP

repo/blob/filesystem/filesystem_storage.go
```diff
@@ -231,7 +230,19 @@ func (fs *fsImpl) ReadDir(ctx context.Context, dirname string) ([]os.FileInfo, e
...
+		fi, err := e.Info()
+		if err != nil {
+			// nolint:wrapcheck
+			return nil, err
+		}
+
+		fileInfos = append(fileInfos, fi)
+	}
+
+	return fileInfos, nil
```
Juneezee, line 245: This is an easier workaround to return the results as `[]os.FileInfo`. Changing the method signature from `ReadDir(ctx context.Context, path string) ([]os.FileInfo, error)` to `ReadDir(ctx context.Context, path string) ([]os.DirEntry, error)` will require massive refactoring of other storage implementations (`davStorageImpl`, `fsImpl`, and `sftpImpl`) as well.

jkowalski:
That windows failure is a real one. I encountered it before when trying to refactor this exact same thing.

It appears that windows directory times are different when returned by ReadDir() vs Lstat(). I did not investigate further. 

This is probably a test issue related to launching actions.

jkowalski:
So as I was saying - with this change Windows will be failing 100% - there's some subtle difference between os.Lstat() and os.ReadDir() result, (which could be insignificant but the test is failing nonetheless).

fs/localfs/local_fs.go
```diff
@@ -128,8 +128,8 @@ type entryWithError struct {
 	err   error
 }
 
-func toDirEntryOrNil(basename, prefix string) (fs.Entry, error) {
-	fi, err := os.Lstat(prefix + basename)
+func toDirEntryOrNil(dirEntry os.DirEntry, prefix string) (fs.Entry, error) {
+	fi, err := os.Lstat(prefix + dirEntry.Name())
```
Juneezee, line 132: > So as I was saying - with this change Windows will be failing 100% - there's some subtle difference between os.Lstat() and os.ReadDir() result, (which could be insignificant but the test is failing nonetheless).

@jkowalski The Windows integration test should pass now. I have tested this on my fork repository Actions workflow.

I tried changing `os.Lstat()` into `os.Stat()` and the Windows integration test passed as well. There is definitely something difference between `dirEntry.Info()` and `os.Stat() / os.Lstat()`

jkowalski:
thank you!

PR #1374, opened 2021-10-12
KopiaUI: add cli flag framework and starting flags

### LINKED ISSUE #367: feature request: add config flag to kopia UI
Hi,
is it possible to add some basic cli flags to Kopia GUI, so we can deploy it easier. 
I mean to have the flag --config-file at less.
it would be nice to have an option to pass the flags to kopia cli.
thank you

baka-yuki:
this got a bit larger than i anticipated it being, and at this point i could use another set of eyes on it.

this modifies the UI to be able to parse specified cli flags in addition to the normal electron flags that can be passed. to start with, i added `--portable`, `--cache-directory`, `--log-dir`, `--help`, and `--config-file` to match the kopia cli, and to close #367.

this opens up a couple new features:
- calling the UI executable with just `--portable` will create the portable directory structure if it does not already exist.
- calling the UI executable with just `--config-file` will start the UI with connection(s) to the specified repository(ies). the cache/log location will be the same as the non-portable app.
- combining the two will put the cache-directory and the log-directory for each repository right next to the .config file.
- calling `--log-dir` will overwrite the log directory, period. so the UI will read configs from either portable/standard locations, but write log files to the specified location.
- calling `--cache-directory` by itself doesn't accomplish anything, because of the potential to mess up configs in the standard location
- there's a help output now that i tried to make match the cli

because the cache-directory is located in the config file, i wrote an update routine. this could probably be moved into the main binary, but i wasn't sure if that was desired/wanted there.

i also fixed misc. linter issues that i noticed.

i'm gonna leave this as a draft for the time being, no rush on reviewing it.

### DIFF
--- a/app/public/cli.js
+++ b/app/public/cli.js
@@ -1,0 +1,149 @@
+const log = require("electron-log")
+const fs = require('fs');
+const path = require('path');
+
+const { app } = require("electron")
+
+const cliArgs = process.argv
+const configFileSuffix = ".config";
+
+let hasCliConfigFlag = false;
+let hasPortableFlag = false;
+let hasLogDirFlag = false;
+let hasCacheDirFlag = false;
+
+function invalidConfigFile(file) {
+    log.error(`invalid configuration file passed on command line: ${file}`);
+    console.log(`Error: Invalid configuration file passed on command line.\
+                 File Name: ${file}\
+                 Valid configuration files end in ${configFileSuffix}.\
+                 Exiting...`);
+    app.exit(1);
+}
+
+function configFileNotFound(file) {
+    log.error(`unable to locate configuration file: ${file}`);
+    console.log(`Error: Unable to verify existance of configuration file.\
+                 File Name: ${file}\
+                 Please check that the path is correct and try again.\
+                 Exiting...`);
+    app.exit(1);
+}
+
+// this isn't pretty but it doesn't require any additional dependencies.
+function outputHelp() {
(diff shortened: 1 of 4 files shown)


zyradyl:
closing per discussion on slack -- thanks!

EarthlingDavey:
@zyradyl hi, this sounds useful. Did you abandon the PR?

zyradyl:
after speaking with Jarek for a bit on slack we decided this was not the way forward for the UI. instead i'm planning to add an in UI way to convert from a local layout to a portable layout. however, i'm starting with adding deletion to the UI as it is the more requested feature. hoping to get the first parts of that knocked out and in a PR this weekend. once i'm all done i'll come back to the portable/local issue.

until then i think the branch is still in my fork? you would have to rebase it a bit and fix the things Jarek pointed out above, but custom versions could me made. sorry for the inconvenience 

-z

EarthlingDavey:
Thanks for replying so quickly. It's awesome to see such activity on the repo it is appreciated and when I am in a position to support this project, I will! 

BTW. I learned of kopia a few weeks ago. But only implemented it over the past few days. Everything went well 👍

I understand your explanation, and sure, it makes sense to prioritise more requested features. 

I have actually made a work around, I'm on macOS and I wanted my configs to be on an encrypted drive. I simply made 2 symlinks for the kopia and kopia-ui directories :)

But, I'll follow the repo progress closely, if support is added for --args to an official release then I'll use that :)

Cheers ✌🏼

PR #1375, opened 2021-10-12
Add doc for Amazon S3 .storageconfig

CrendKing:
Let me know how to improve. Thanks.


jkowalski:
Can you add more of a warning to `Special notes for using the Glacier storage classes` - specifically mention this will require a lot of manual effort to restore and is not recommended for most users.

CrendKing:
What about now?

PR #1376, opened 2021-10-12
logging: switched from go-logging to zap

### LINKED ISSUE #1345: Reduce logging memory footprint
After recent efforts to reduce memory usage, one of the top items in the profile is now logging due to garbage produced by lots of small strings.

We should consider:

- logging less
- switching to https://github.com/rs/zerolog or https://github.com/uber-go/zap or another no-allocation logger

jkowalski:
This is much more efficient in terms of memory allocations and speeds up backup due to less GC pressure.
There are some minor differences in behavior mostly in terms of console colors, but hopefully those are not something folks rely upon.

Fixes #1345


jkowalski:
> Any sense for how much the memory consumption difference is?

I'll run a benchmark once this is merged.

jkowalski:
I did precise pprof with sampling rate of 1:

```
kopia --profile-memory=1 --no-auto-maintenance --profile-dir=/tmp/profile-dir snap create <path>
```

on a directrory that is 204.4 MB files:49176 dirs:5360

```
before: 7345907 objects, 745.79MB total
after: 4706910 objects, 497.78MB total
```

PR #1420, opened 2021-10-19
Migrate: pick latest instead of oldest snapshot with --latest-only

### LINKED ISSUE #1414: Snapshot migrate --latest-only migrates oldest instead of latest snapshot
Hi,

Well, the title basically says it all. I was using `kopia snapshot migrate --all --latest-only` and noticed that only the oldest snapshot of each source is migrated.

Thanks for creating and maintaining this super-useful piece of software.

Best regards
Robert

bob1de:
Fix #1414


jkowalski:
golang requires particular canonical file formatting, in particular `len(s)-1` without spaces. I'm guessing you did you use `gofmt` to format the file after changing. 

In any case I recommend using some proper code editor which formats automatically - Visual Studio Code is very good and magically does all of this.

bob1de:
Alright, using gofmt now.

PR #1452, opened 2021-10-29
Use blob.OutputBuffer in blob.Reader interface instead of internal gather.WriteBuffer

dzaninovic:
Exported blob.Storage interface can't be implemented as GetBlob() is using internal gather.WriteBuffer as an output argument type.

My use case is to intercept all Kopia calls to blob.Storage so I can recreate new underlying blob.Storage on the fly if AWS session expires in the middle of a long operation like creating the snapshot.

I am also using this blob.Storage wrapper to log and keep statistics of all calls to blob.Storage.

Proposed changes change the GetBlob() argument type to blob.OutputBuffer.

make command ran all tests successfully

Forum discussion reference:
https://kopia.discourse.group/t/implementing-repo-blob-reader-interface/706/4

### DIFF
--- a/internal/blobtesting/eventually_consistent.go
+++ b/internal/blobtesting/eventually_consistent.go
@@ -12,7 +12,6 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/kopia/kopia/internal/clock"
-	"github.com/kopia/kopia/internal/gather"
 	"github.com/kopia/kopia/repo/blob"
 )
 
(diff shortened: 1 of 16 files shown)


jkowalski:
I think it would be simpler to add interface like this:

```
// OutputBuffer is implemented by *gather.WriteBuffer
type OutputBuffer interface {
	io.Writer

	Reset()
	Length() int
}
```

and change `*gather.WriteBuffer` to `storage.OutputBuffer` everywhere.

Also, if we renamed `Length` to `Len`, I believe `bytes.Buffer` would match this interface as well, which might be convenient if gather.WriteBuffer is not needed.

dzaninovic:
Good idea, that would also give me access to Length() from the wrapper.
I will work on implementing that.

dzaninovic:
I made the change you suggested but had to add ToByteSlice() and Append() to the interface as they were needed by eventuallyConsistentStorage and mapStorage implementations.

I looked into bytes.Buffer but that is not an interface so change would be more complicated.

jkowalski:
I don't think we need Append and ToByteSlice - I tried adding a commit to this PR but don't have permissions to your branch, Can you add this:

https://github.com/jkowalski/kopia/commit/36d30b3b5f1f916e95493ca7552e6612f56624a6

dzaninovic:
I cherry-picked your commit and rebased this branch to latest master.
All tests make ran are passing.

jkowalski:

BTW. do you want to perhaps sync up on Slack, since you're consuming Kopia in some novel ways, I'd like to make sure we stay in touch as there may be low-level changes that might break you going forward (as the golang API is not really stable yet).

dzaninovic:
Thank you for merging the pull request!
I joined Slack and will monitor the channels for updates.

PR #1498, opened 2021-11-07
server: add API endpoints for source, snapshot, batch deletion

zyradyl:
as per our slack discussions, i just set up the endpoints and added some base logic. i probably could have made it a bit simpler for step 1 but it got away from me a bit, haha.

tested everything with `curl` on a server started with `--insecure --without-password` and had the expected results

let me know what you think 😄

### DIFF
--- a/internal/server/api_snapshots.go
+++ b/internal/server/api_snapshots.go
@@ -2,10 +2,15 @@ package server
 
 import (
 	"context"
+	"encoding/json"
 	"net/http"
 	"net/url"
 
+	"github.com/gorilla/mux"
+	"github.com/pkg/errors"
+
 	"github.com/kopia/kopia/internal/serverapi"
+	"github.com/kopia/kopia/repo/manifest"
 	"github.com/kopia/kopia/snapshot"
 	"github.com/kopia/kopia/snapshot/policy"
 )
@@ -78,4 +83,46 @@ func convertSnapshotManifest(m *snapshot.Manifest) *serverapi.Snapshot {
 	}
 
 	return e
+}
+
+func (s *Server) handleSnapshotDelete(ctx context.Context, r *http.Request, body []byte) (interface{}, *apiError) {
+	mid := mux.Vars(r)["snapshotID"]
+	rep := s.rep
+
+	// logic here should eventually be similar to that in commandSnapshotDelete run
+	// for now just assume the oid is a manifest ID and handle simple not found errors
+	_, err := snapshot.LoadSnapshot(ctx, rep, manifest.ID(mid))
+	if err != nil {
+		log(ctx).Errorf("error loading snapshot %v", mid)
+		return nil, requestError(serverapi.ErrorNotFound, "snapshot ID not found")
+	}
+
+	log(ctx).Debugw("would delete snapshot with id", "id", mid)
+
+	return nil, internalServerError(errors.Errorf("deletion not implemented"))
+}
+
+func (s *Server) handleSnapshotDeleteBatch(ctx context.Context, r *http.Request, body []byte) (interface{}, *apiError) {
+	var req serverapi.BatchSnapshotDeleteRequest
+
+	rep := s.rep
+
+	if err := json.Unmarshal(body, &req); err != nil {
+		return nil, requestError(serverapi.ErrorMalformedRequest, "malformed request body")
+	}
+
+	if len(req.SnapshotManifestIds) == 0 {
+		return nil, requestError(serverapi.ErrorMalformedRequest, "missing list of snapshots")
+	}
+
+	// make sure all manifest ids are valid
+	for _, id := range req.SnapshotManifestIds {
+		_, err := snapshot.LoadSnapshot(ctx, rep, manifest.ID(id))
+		if err != nil {
+			log(ctx).Errorf("error loading snapshot %v", id)
+			return nil, requestError(serverapi.ErrorMalformedRequest, "invalid manifest id in snapshotManifestIds")
+		}
+	}
+
+	return nil, internalServerError(errors.Errorf("deletion not implemented"))
 }
--- a/internal/serverapi/serverapi.go
+++ b/internal/serverapi/serverapi.go
@@ -147,6 +147,11 @@ type CreateSnapshotSourceResponse struct {
 	SnapshotStarted bool `json:"snapshotted"` // whether snapshotting has been started
 }
 
+// BatchSnapshotDeleteRequest contains a list of snapshot IDs to be deleted.
+type BatchSnapshotDeleteRequest struct {
+	SnapshotManifestIds []string `json:"snapshotManifestIDs"`
+}
+
 // Snapshot describes single snapshot entry.
 type Snapshot struct {
 	ID               manifest.ID          `json:"id"`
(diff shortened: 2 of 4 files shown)


internal/serverapi/serverapi.go
```diff
@@ -147,6 +147,31 @@ type CreateSnapshotSourceResponse struct {
 	SnapshotStarted bool `json:"snapshotted"` // whether snapshotting has been started
 }
 
+// DeletionResponseSingle contains response of attempted deletion of either snapshot or source.
+type DeletionResponseSingle struct {
+	Type    string `json:"type"`    // did we attempt to delete a snapshot or a source
+	Deleted bool   `json:"deleted"` // whether the object was deleted
+	Error   string `json:"error"`   // any errors
```
jkowalski, line 154: errors are not normally returned inside the response but instead are serialized in ErrorResponse.

internal/serverapi/serverapi.go
```diff
@@ -147,6 +147,31 @@ type CreateSnapshotSourceResponse struct {
...
+type DeletionResponseBatch struct {
+	Items []*DeletionStatus `json:"items"` // status of each item we attempted to delete
+}
+
+// DeletionStatus is used to return information about each item that had deletion attempted.
+type DeletionStatus struct {
+	Type    string `json:"type"`    // was the item we attempted to delete a source or a snapshot
+	ID      string `json:"content"` // ID for the item, this would be a path for a source or a manifest ID for a snapshot
+	Deleted bool   `json:"deleted"` // was the item deleted
+	Error   string `json:"error"`   // any errors
```
jkowalski, line 167: same thing there.

internal/serverapi/serverapi.go
```diff
@@ -147,6 +147,31 @@ type CreateSnapshotSourceResponse struct {
...
+}
+
+// DeletionResponseBatch contains response of attempted batch deletion.
+type DeletionResponseBatch struct {
+	Items []*DeletionStatus `json:"items"` // status of each item we attempted to delete
+}
+
+// DeletionStatus is used to return information about each item that had deletion attempted.
+type DeletionStatus struct {
```
jkowalski, line 163: do we need this response at all? Empty seems like it may suffice

internal/serverapi/serverapi.go
```diff
@@ -147,6 +147,31 @@ type CreateSnapshotSourceResponse struct {
...
+	Type    string `json:"type"`    // was the item we attempted to delete a source or a snapshot
+	ID      string `json:"content"` // ID for the item, this would be a path for a source or a manifest ID for a snapshot
+	Deleted bool   `json:"deleted"` // was the item deleted
+	Error   string `json:"error"`   // any errors
+}
+
+// BatchSnapshotDeleteRequest contains a list of snapshot IDs to be deleted.
+type BatchSnapshotDeleteRequest struct {
+	Snapshots []string `json:"snapshots"`
```
jkowalski, line 172: Can we rename this to `snapshotManifestIds` ? The notion of snapshot ID is slightly ambiguous.

internal/serverapi/serverapi.go
jkowalski: Also, do we need this response at all? `Empty` may be sufficient.

internal/serverapi/serverapi.go
```diff
@@ -147,6 +147,31 @@ type CreateSnapshotSourceResponse struct {
...
+// DeletionResponseSingle contains response of attempted deletion of either snapshot or source.
+type DeletionResponseSingle struct {
+	Type    string `json:"type"`    // did we attempt to delete a snapshot or a source
+	Deleted bool   `json:"deleted"` // whether the object was deleted
+	Error   string `json:"error"`   // any errors
+}
+
+// DeletionResponseBatch contains response of attempted batch deletion.
+type DeletionResponseBatch struct {
+	Items []*DeletionStatus `json:"items"` // status of each item we attempted to delete
```
jkowalski, line 159: do we anticipate this failing in any non-trivial ways (we don't really have locks on snapshots so this will either always succeed or always fail)? Thus I would say we may not need this response type either (at least for now)?

internal/server/api_sources.go
```diff
@@ -127,3 +129,31 @@ func (s *Server) handleSourcesCreate(ctx context.Context, r *http.Request, body
...
+	}
+
+	// logic here should be similar to commandSnapshotDelete snapshotDeleteSources
+	// for now just output the path that would be deleted
+	log(ctx).Debugw("would delete source with path", "path", sourceInfo.Path)
+
+	resp := serverapi.DeletionResponseSingle{
+		Type:    "source",
+		Deleted: false,
+		Error:   "deletion not yet fully implemented",
```
jkowalski, line 155: How about `return nil, errors.Errorf("deletion not implemented")` ?

internal/server/api_sources.go
```diff
@@ -127,3 +129,31 @@ func (s *Server) handleSourcesCreate(ctx context.Context, r *http.Request, body
 
 	return resp, nil
 }
+
+func (s *Server) handleSourcesDelete(ctx context.Context, r *http.Request, body []byte) (interface{}, *apiError) {
+	b64path := mux.Vars(r)["sourcePath"]
+
+	path, err := base64.URLEncoding.DecodeString(b64path)
```
jkowalski, line 136: is there a reason for base64? wouldn't uri encode be sufficient?

internal/server/api_snapshots.go
```diff
@@ -79,3 +84,86 @@ func convertSnapshotManifest(m *snapshot.Manifest) *serverapi.Snapshot {
...
+		Items: []*serverapi.DeletionStatus{},
+	}
+
+	for _, idstr := range req.Snapshots {
+		item := &serverapi.DeletionStatus{
+			Type: "snapshot",
+			ID:   idstr,
+		}
+
+		id, err := object.ParseID(idstr)
```
jkowalski, line 140: I think deletion is relatively simple, we may be better off not introducing per-item responses

if you parse all IDs early on and fail on any errors, you can simply return `Empty` on success.

internal/server/api_snapshots.go
```diff
@@ -79,3 +84,86 @@ func convertSnapshotManifest(m *snapshot.Manifest) *serverapi.Snapshot {
 
 	return e
 }
+
+func (s *Server) handleSnapshotDelete(ctx context.Context, r *http.Request, body []byte) (interface{}, *apiError) {
+	oidstr := mux.Vars(r)["snapshotID"]
+	rep := s.rep
+
+	oid, err := object.ParseID(oidstr)
```
jkowalski, line 92: this is actually not correct - you're parsing the object ID then casting to manifest.ID, we should always treat those as manifest IDs (which don't require any parsing).

jkowalski: BTW. I see what you're trying to do - have fuzzy notion of "object ID", but this is an API and I think we should be precise and always require manifest IDs - the only caller will be the UI and they will already have the manifest ID ready.

jkowalski (changes requested):
we may be unnecessarily making the API too much CLI-like where it tries to guess user intention - I would recommend making it stricter, fail harder, only support manifest IDs on input and not return per-item responses (which are a pain to deal with, anyway).

internal/serverapi/serverapi.go
zyradyl: yep, have made the change.

internal/server/api_sources.go
zyradyl: uri encode seems to cause the server to attempt to parse the string as part of the path, and respond accordingly, which can generate either no response (leading `/`) or a 404 not found error.

jkowalski: as discussed on slack, can we switch to passing user/host/path as query string parameters and not URL path, just like `/sources/cancel`, `/sources/upload` and few others.

jkowalski:
Closing, this was implemented in #1730

Issue #1478, opened 2021-11-04
Replace Debugf() with structural logging using Debugw()


jkowalski:
#1475 introduced `Debugw()` method which can be used to emit structural logs.

In many places we still use `Debugf()`. This issue tracks refactoring of remaining `Debugf()` to `Debugw()`, so that we can eventually get rid of `Debugf()` and have all logs be fully structural.

PR #1501, opened 2021-11-07
refactor: switch to debugw where possible

zyradyl:
i had a bit of time this morning so i thought i would give this a go. the problem now is the tests are failing and i'm not quite sure why. in particular, upload_test.go, fails due to not finding "uploading" but i don't think i modified that?

i labeled the json fields based on context clues in the surrounding code, but i'm sure i missed a few things, and i also realized near the end i probably should have done this by module instead of all at once. 

anyways, let me know your thoughts. 😄 

related #1478

### DIFF
--- a/cli/auto_upgrade.go
+++ b/cli/auto_upgrade.go
@@ -25,7 +25,7 @@ func maybeAutoUpgradeRepository(ctx context.Context, r repo.Repository) error {
 		return nil
 	}
 
-	log(ctx).Debugf("Setting default maintenance parameters...")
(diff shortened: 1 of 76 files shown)


jkowalski:
can you perhaps split this into individual PRs layer-by-layer (repo/blob, repo/content, repo/manifest, repo/object, etc?

there are actually tests that verify log format, and those are failing, perhaps revert affected files and make a separate PR just for those as well?

zyradyl:
i knew that was coming as soon as i realized the size of it. absolutely can do. will probably be sometime next weekend though, apologies for the delay!

PR #1505, opened 2021-11-09
Upgraded linter to 1.43.0

### LINKED ISSUE #1477: Upgrade linter to 1.43.0
There are some new good linters introduced in this version, notably varnamelen, ireturn, contextcheck, tenv and they all will require some code changes and/or adding exceptions.

See the output: https://gist.github.com/jkowalski/a626fd6a23496dd8701832e22eb3c992

jkowalski:
Fixes #1477 

NOTE: the last commit (`linter: fixed nilnil linter violations (non-mechanical)`) is slightly more tricky than others and requires more careful review.


fs/localfs/shallow_fs.go
```diff
@@ -63,21 +63,21 @@ func WriteShallowPlaceholder(path string, de *snapshot.DirEntry) (string, error)
 	return mp, nil
 }
 
-func dirEntryFromPlaceholder(path string) (*snapshot.DirEntry, error) {
+func dirEntryFromPlaceholder(path string) (*snapshot.DirEntry, bool, error) {
```
julio-lopez, line 66: While the rationale seems reasonable, it also seems redundant. It seems that the `bool` return value would always be `false` when `error != nil`, and it could either be `true` or `false` depending on the first return value, most likely `false` when `v1 == nil`.

Does this pattern, and thus this linter add real value? what am I missing here?

repo/content/committed_content_index.go
```diff
@@ -57,19 +57,15 @@ func (c *committedContentIndex) getContent(contentID ID) (Info, error) {
...
-			return nil, ErrContentNotFound
-		}
-
-		return info, nil
+	if err != nil {
+		return nil, err
 	}
 
-	if err == nil {
+	if c.shouldIgnore(info) {
```
julio-lopez, line 64: Is `shouldIgnore(info)` false when `info == nil` ?

repo/content/merged.go
```diff
@@ -65,6 +65,10 @@ func (m mergedIndex) GetInfo(id ID) (Info, error) {
 	for _, ndx := range m {
 		i, err := ndx.GetInfo(id)
 		if err != nil {
+			if errors.Is(err, ErrContentNotFound) {
```
julio-lopez, line 68: nit: use a `switch` here?

```go
switch {
case errors.Is(err, ErrContentNotFound):
	continue
case err != nil:
	return nil, errors.Wrapf(err, "error getting id %v from index shard", id)
}
```

repo/content/index.go
```diff
@@ -19,7 +19,7 @@ type packIndex interface {
 
 	ApproximateCount() int
 
-	GetInfo(contentID ID) (Info, error)
+	GetInfo(contentID ID) (Info, error) // may return ErrContentNotFound
```
julio-lopez, line 22: Q: what's the motivation for returning `ErrContentNotFound`? Is it due to the "nil nil" linter? just curiosity, not saying it should not be done.

This has or may have a slight change in semantics, and would be a good candidate to have put in a separate PR by itself to make it easier to reason through it.

julio-lopez (commented):
+1: LG 👍

fs/localfs/shallow_fs.go
jkowalski: the pattern this enforces is:

value, err := someFunction()
if err != nil {
   // 
}

now value is assumed to be != nil.

By adding boolean we're making it explicit that there's more than one thing to look at (not just dirEntry, but also the boolean).

repo/content/committed_content_index.go
jkowalski: at this point info won't be nil, that's due to the change in `merged.GetInfo()`

repo/content/index.go
jkowalski: I'm going to suppress the linter and move this commit out to separate PR.

julio-lopez (approved):

Issue #1502, opened 2021-11-09
Implement upload and download throttling using wrapper provider


jkowalski:
Currently upload and download throttling is done at the provider level, which means it's not uniformly supported. For example filesystem, webdav and sftp do not support it, while all 4 cloud providers do support it.

We can make this more uniform by implementing a wrapper provider (like `readonly`, `logging`, `retrying`, etc.), which would do rate limiting in terms of bytes sent and received before/after sending the request to the underlying provider.

PR #1510, opened 2021-11-11
blob: added throttling wrapper

jkowalski:
Throttler is only mocked for unit test purposes and not implemented yet.

Related #1502

### DIFF
--- a/repo/blob/throttling/throttling_storage.go
+++ b/repo/blob/throttling/throttling_storage.go
@@ -1,0 +1,97 @@
+// Package throttling implements wrapper around blob.Storage that adds throttling to all calls.
+// Throttling is performed for both operations and byte counts (upload and download separately).
+package throttling
+
+import (
+	"context"
+	"time"
+
+	"github.com/kopia/kopia/repo/blob"
+)
+
+// assume we will need to download ~20 MB for blobs of unknown length, we will refund the difference
+// if we guess wrong or acquire more.
+const unknownBlobAcquireLength = 20000000
+
+// Throttler implements throttling policy by blocking before certain operations are
+// attempted to ensure we don't exceed the desired rate of operations/bytes uploaded/downloaded.
+type Throttler interface {
+	BeforeOperation(ctx context.Context, op string)
+
+	// BeforeDownload acquires the specified number of downloaded bytes
+	// possibly blocking until enough are available.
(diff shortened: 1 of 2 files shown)


repo/blob/throttling/throttling_storage.go
```diff
@@ -0,0 +1,97 @@
...
+func (s *throttlingStorage) SetTime(ctx context.Context, id blob.ID, t time.Time) error {
+	s.throttler.BeforeOperation(ctx, "SetTime")
+	return s.Storage.SetTime(ctx, id, t) // nolint:wrapcheck
+}
+
+func (s *throttlingStorage) PutBlob(ctx context.Context, id blob.ID, data blob.Bytes) error {
+	s.throttler.BeforeOperation(ctx, "PutBlob")
+	s.throttler.BeforeUpload(ctx, int64(data.Length()))
+
+	return s.Storage.PutBlob(ctx, id, data) // nolint:wrapcheck
```
Shrekster, line 86: nit; if PutBlob() fails early then we may sometimes not return the unused upload quota. It might help to have PutBlob() return the number of bytes written but I think this is okay being taxing on errors because that is not a happy-path. Plus, it may not be possible to get the uploaded bytes through every SDK.

repo/blob/throttling/throttling_storage_test.go
```diff
@@ -0,0 +1,146 @@
...
+		"inner.PutBlob",
+	}, m.activity)
+
+	m.Reset()
+	require.NoError(t, wrapped.GetBlob(ctx, "blob1", 0, -1, &tmp))
+	require.Equal(t, []string{
+		"BeforeOperation(GetBlob)",
+		"BeforeDownload(20000000)", // length is unknown, we assume 20MB
+		"inner.GetBlob",
+		"ReturnUnusedDownloadBytes(19999993)", // refund all but 7 bytes
```
Shrekster, line 95: Very nice! Would be interesting to cover how this activity trail would be with faulty storage as well.

Shrekster (approved):
LGTM! Nice activity trail coverage. Some nits about faulty storage behavior.

Shrekster:
Another note here, @julio-lopez . This is throttling per api server / kopia instance. If we need throttling across multiple instances to broadly limit the data pipe usage the nusing multiple servers then we might need to coordinate between those servers individual throttler interface.

PR #1511, opened 2021-11-11
storage api put-blob retention options

Shrekster:
In order to provide blob-retention setting on each blob put operation we introduce put-blob API changes at the storage interface level. These settings allow us to supply retention configuration. For now, just AWS S3 storage support has been added that respects these options but the same concept can be extended to other supported storage vendors in the future.

### DIFF
--- a/internal/ownwrites/ownwrites.go
+++ b/internal/ownwrites/ownwrites.go
@@ -122,11 +122,11 @@ func (s *CacheStorage) ListBlobs(ctx context.Context, prefix blob.ID, cb func(bl
 }
 
 // PutBlob implements blob.Storage and writes markers into local cache for all successful writes.
-func (s *CacheStorage) PutBlob(ctx context.Context, blobID blob.ID, data blob.Bytes) error {
-	err := s.Storage.PutBlob(ctx, blobID, data)
+func (s *CacheStorage) PutBlob(ctx context.Context, blobID blob.ID, data blob.Bytes, opts blob.StoragePutBlobOptions) error {
+	err := s.Storage.PutBlob(ctx, blobID, data, opts)
 	if err == nil && s.isCachedPrefix(blobID) {
 		// nolint:errcheck
-		s.cacheStorage.PutBlob(ctx, prefixAdd+blobID, markerData)
+		s.cacheStorage.PutBlob(ctx, prefixAdd+blobID, markerData, opts)
 	}
 
 	// nolint:wrapcheck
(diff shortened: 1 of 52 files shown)


repo/blob/gcs/gcs_storage.go
```diff
@@ -97,7 +97,11 @@ func translateError(err error) error {
 	}
 }
 
-func (gcs *gcsStorage) PutBlob(ctx context.Context, b blob.ID, data blob.Bytes) error {
+func (gcs *gcsStorage) PutBlob(ctx context.Context, b blob.ID, data blob.Bytes, opts blob.StoragePutBlobOptions) error {
+	if opts.RetentionPeriod != 0 || opts.RetentionMode != "" {
```
jkowalski, line 101: how about a helper `opts.HasRetentionOptions()` ?

internal/ownwrites/ownwrites.go
```diff
@@ -138,7 +138,7 @@ func (s *CacheStorage) DeleteBlob(ctx context.Context, blobID blob.ID) error {
 	err := s.Storage.DeleteBlob(ctx, blobID)
 	if err == nil && s.isCachedPrefix(blobID) {
 		// nolint:errcheck
-		s.cacheStorage.PutBlob(ctx, prefixDelete+blobID, markerData)
+		s.cacheStorage.PutBlob(ctx, prefixDelete+blobID, markerData, blob.StoragePutBlobOptions{})
```
jkowalski, line 141: can we shorten the name to `blob.PutOptions` ?

repo/blob/storage.go
```diff
@@ -17,6 +17,18 @@ var ErrSetTimeUnsupported = errors.Errorf("SetTime is not supported")
 // ErrInvalidRange is returned when the requested blob offset or length is invalid.
 var ErrInvalidRange = errors.Errorf("invalid blob offset or length")
 
+// ErrBlobRetentionUnsupported is returned by implementations of Storage that
+// don't support setting blob-retention times.
+var ErrBlobRetentionUnsupported = errors.Errorf("setting blob-retention is not supported")
```
jkowalski, line 22: do we expect callers to react to these errors programmatically in any way? if not (and they are always bubble up to the first caller and become ), they should not be public.

repo/blob/gcs/gcs_storage.go
Shrekster: good idea, will do

repo/blob/s3/s3_storage_test.go
```diff
@@ -178,6 +178,70 @@ func TestS3StorageAWSSTS(t *testing.T) {
...
+	options := &Options{
+		Endpoint:        getEnv(testEndpointEnv, awsEndpoint),
+		AccessKeyID:     getEnvOrSkip(t, testAccessKeyIDEnv),
+		SecretAccessKey: getEnvOrSkip(t, testSecretAccessKeyEnv),
+		BucketName:      getEnvOrSkip(t, testBucketEnv),
+		Region:          getEnvOrSkip(t, testRegionEnv),
+	}
+
+	createBucket(t, options)
+	testStorage(t, options, false)
```
Shrekster, line 194: I forgot to plumb the retention options through tests. Will fix this and push.

repo/blob/storage.go
Shrekster: Not in the context fo this PR. Having an error exposed can help with improved error response at CLI/UI. Better to keep this unexposed for now and expose it when needed on incremental PRs.

jkowalski (approved):

Issue #1503, opened 2021-11-09
Make upload and download throttling universally configurable


jkowalski:
Currently upload and download throttle values are specified at `create` or `connect` time using provider-specific flags and thus is not uniformly across providers.

When we implement #1502, should be able to deprecate provider-specific flags for controlling throttles in favor and make them universal and changeable:

```
$ kopia repo create --max-upload-speed X --max-download-speed Y 
$ kopia repo connect --max-upload-speed X --max-download-speed Y
$ kopia repo set-client --max-upload-speed X --max-download-speed Y
```

xxxliqu1dxxx:
Would it make sense to consider the use case of bandwidth vs time of day?
I.e. between 9am to 5pm = 1MB/s, but 5pm to 9am = 10MB/s

PR #1512, opened 2021-11-13
throtting: implemented a Throttler based on token bucket and configur…

### LINKED ISSUE #1502: Implement upload and download throttling using wrapper provider

jkowalski:
This unifies throttling for all providers.

Fixes #1502 
Related #1503


repo/blob/throttling/throttler_test.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		wg.Wait()
+
+		numSeconds := timer.Elapsed().Seconds()
+		actualRate := float64(*total) / numSeconds
+
+		// make sure the rate is less than target with some tiny margin of error
+		require.Less(t, actualRate, wantRate*1.05)
+		require.Greater(t, actualRate, wantRate*0.9)
```
julio-lopez, line 133: will this be flaky?

repo/blob/throttling/token_bucket.go
```diff
@@ -0,0 +1,124 @@
...
+type tokenBucket struct {
+	name  string
+	now   func() time.Time
+	sleep func(ctx context.Context, d time.Duration)
+
+	mu                   sync.Mutex
+	lastTime             time.Time
+	numTokens            float64
+	maxTokens            float64
+	addTokensPerTimeUnit float64
```
julio-lopez, line 24: is there any case where `addTokensPerTimeUnit != maxTokens` ?

julio-lopez (commented):
+1: LG 👍

repo/blob/throttling/throttler_test.go
jkowalski: i ran it enough times - 5% buffer appears sufficient but we can raise it if it flakes out.

repo/blob/throttling/token_bucket.go
jkowalski: not really, removed for now

PR #1513, opened 2021-11-14
blob: changed default shards from {3,3} to {1,3}

jkowalski:
Turns out for very large repository around 100TB (5M blobs),
we end up creating millions of directories which is way too much
and slows down listing. Currently each leaf directory only has a handful
of files.

Simple sharding of {1,3} should work much better and will end up creating
directories with meaningful shard sizes - 12 K files per directory
should not be too slow and will reduce the overhead of listing by
4096 times.

The change is done in a backwards-compatible way and will respect
custom sharding (.shards) file written by previous 0.9 builds
as well as older repositories that don't have the .shards file (which
we assume to be {3,3}).

The layout produces repositories looking like:

```
/.shards
/_/log/_20211114104830_a3a5_1636915710_1636915723_1_ff05424d9d270febdb2c504f28d30db6.f
/_/log/_20211114104830_a3a5_1636915723_1636915727_2_d5ab683b3057c5a193895e7db0d217ad.f
/_/log/_20211114104830_3c01_1636915710_1636915710_1_d094b4064422442d634c700b8eb75aeb.f
/kopia.repository.f
/kopia.maintenance.f
/q/a85/70517963bee3521d69e7dec844cb7-saa5e639eaff4d1dc10a.f
/q/865/3a42e8eea6223cc6c045b12cab871-saa5e639eaff4d1dc10a.f
/q/85e/837f10b6238e74a2afbcf82e66a72-sa0f28471f1d7e0bb10a.f
/x/n0_/30e2817ffa969600081aa4cf314a6a3e-sa0f28471f1d7e0bb10a-c1.f
/x/n0_/b76c7150880f7580bf819bb006b25f41-saa5e639eaff4d1dc10a-c1.f
/p/797/91bc7fec987458460f2f5322b2fb3-saa5e639eaff4d1dc10a.f
/p/d12/8caec69e1912052a962db0f0e338d-saa5e639eaff4d1dc10a.f
/p/ed9/3745d2216cf512711558c5987b599-saa5e639eaff4d1dc10a.f
/p/0f9/c974fd398319b49844b3081a5b2e0-saa5e639eaff4d1dc10a.f
/p/36e/a270cffeab38b5b04a1ec4ab9c498-saa5e639eaff4d1dc10a.f
/p/95d/f7c1b24f1b10d980e73f6ad8e2a9c-saa5e639eaff4d1dc10a.f
```

### DIFF
--- a/cli/storage_filesystem.go
+++ b/cli/storage_filesystem.go
@@ -38,7 +38,7 @@ func (c *storageFilesystemFlags) setup(_ storageProviderServices, cmd *kingpin.C
 	cmd.Flag("list-parallelism", "Set list parallelism").Hidden().IntVar(&c.options.ListParallelism)
 }
 
-func (c *storageFilesystemFlags) connect(ctx context.Context, isNew bool) (blob.Storage, error) {
+func (c *storageFilesystemFlags) connect(ctx context.Context, isCreate bool) (blob.Storage, error) {
 	fso := c.options
 
 	fso.Path = ospath.ResolveUserFriendlyPath(fso.Path, false)
@@ -64,16 +64,8 @@ func (c *storageFilesystemFlags) connect(ctx context.Context, isNew bool) (blob.
 		fso.DirectoryShards = []int{}
 	}
 
-	if isNew {
-		log(ctx).Debugf("creating directory for repository: %v dir mode: %v", fso.Path, fso.DirectoryMode)
-
-		if err := os.MkdirAll(fso.Path, fso.DirectoryMode); err != nil {
-			log(ctx).Errorf("unable to create directory: %v", fso.Path)
-		}
-	}
-
 	// nolint:wrapcheck
-	return filesystem.New(ctx, &fso)
+	return filesystem.New(ctx, &fso, isCreate)
 }
 
 func getIntPtrValue(value string, base int) *int {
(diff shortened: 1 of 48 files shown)


cli/storage_filesystem.go
```diff
@@ -59,21 +59,24 @@ func (c *storageFilesystemFlags) connect(ctx context.Context, isNew bool) (blob.
...
+	if flat {
+		return []int{}
+	}
 
-		if err := os.MkdirAll(fso.Path, fso.DirectoryMode); err != nil {
-			log(ctx).Errorf("unable to create directory: %v", fso.Path)
-		}
+	// when creating a repository for version 1, use fixed {3,3} sharding,
+	// otherwise old client can't read it.
+	if formatVersion == 1 {
```
julio-lopez, line 75: I'm confused here about the format version and compatibility matrix.

julio-lopez (commented):
+1: 👍

LG overall.

Confused about the format and backwards client compatiblity.

cli/storage_filesystem.go
jkowalski: it's enforced in the compat tests. repository created with `--format-version=1` must be readable by 0.8.4.

PR #1522, opened 2021-11-18
Change Mbit/s units to MB/s (base-10)

CrendKing:
Almost all numbers with units in Kopia uses byte (either base-2 or base-10) as the data unit. However, there are two cases use bit, which feels unnecessary and inconsistent. For example, a typical `kopia repository sync-to rclone` output looks like this:
```
Synchronizing repositories:
  Source:      Filesystem: <redacted>
  Destination: RClone <redacted>
Looking for BLOBs to synchronize...
  Found 241 BLOBs in the destination repository (2.4 GB)
  Found 244 BLOBs (2.5 GB) in the source repository, 3 (67.8 MB) to copy
  Found 0 BLOBs to delete (0 B), 241 in sync (2.4 GB)
Copying...
  Copied 2 blobs (45.2 MB), Speed: 14.9 Mbit/s, ETA: 3s (2021-11-17 22:20:38 PST)
```
The data sizes are all byte-based, but speed is suddenly changed to bit-based. Feels inconsistent. And most users aren't used to bits.

Is there a particular reason why it's bit-based to begin with? If not, shall we change it?

### DIFF
--- a/cli/command_repository_sync.go
+++ b/cli/command_repository_sync.go
@@ -246,7 +246,7 @@ func (c *commandRepositorySyncTo) runSyncBlobs(ctx context.Context, src blob.Rea
 
 				if est, ok := tt.Estimate(float64(bytesCopied), float64(totalBytes)); ok {
 					eta = fmt.Sprintf("%v (%v)", est.Remaining, formatTimestamp(est.EstimatedEndTime))
-					speed = units.BytesPerSecondsString(est.SpeedPerSecond) //nolint:gomnd
+					speed = units.BytesPerSecondsString(est.SpeedPerSecond)
 				}
 
 				c.outputSyncProgress(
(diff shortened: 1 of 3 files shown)


jkowalski:
LGTM, although I think in networking context the throughput is usually expressed in bits per second for some reason (that's why we have '1 Gbit/s' links and not '125 MB/s' links).

@julio-lopez what do you think?

julio-lopez:
Yes, historically in the networking context `bps` (bits/s) is used.
However, in the storage context, the unit for transfer bandwidth and throughput is bytes/s.

Using bytes/s in this context seems appropriate.

CrendKing:
I agree that when we are talking about network bandwidth, it's mostly bit/s. However, if it is the actual speed, I think it's better to consistently use the same unit as we already use for copying files between local disks.

It's preferential. Feel free to reject this.

julio-lopez:
We can always revert ;)

Issue #1440, opened 2021-10-23
Kopia documentation missing?


petersjostrom:
https://kopia.io/docs/reference/command-line/common/
This page has no sub-pages about "kopia snapshot...", but there are pages if you search, like "snapshot create", "snapshot list", etc.
Seems the documentation lists only the first X sub-pages, the rest are not indexed and visible?
I can't find any info on snapshot commands on the documentation or site either.

CrendKing:
There is a script that goes through all CLI commands in source and generate a doc page for each. There are currently 80 "common commands". However, the source code for the sidebar is hardcoded to [only display the first 50](https://github.com/kopia/kopia/blob/2394b420b02eaf301b3eb4e0f0626cd124f60469/site/themes/docsy/layouts/partials/sidebar-tree.html#L32). And since "`s`napshot" arrives late in the alphabet, they are truncated.

The fix would be easy. Just raise the number. But then the sidebar may become too long. I wonder maybe a better solution would be not listing commands in sidebar, but in a generated page, which provides links to each command's page. This way the page is OK to be long and exhaustive. This does require some rewriting though.

Does hugo allow a page to exist without being listed?

CrendKing:
Attempt to fix the doc: https://github.com/kopia/kopia/pull/1525

PR #1525, opened 2021-11-20
Cleanup cli2md to fix issue #1440

CrendKing:
As issue #1440 states, I can see several problems of the generated doc pages:

1. The top-level [Common Commands](https://kopia.io/docs/reference/command-line/common/) is very cluttered. This is because `cli2md` puts all the command's help string to the page, not accounting for how long the string could be (some such as `restore` are very long). This PR proposes to only put the summary of each help string, e.g. the first line, to the page.
2. The sidebar currently only shows [the first 50 pages in that doc level](https://github.com/kopia/kopia/blob/2394b420b02eaf301b3eb4e0f0626cd124f60469/site/themes/docsy/layouts/partials/sidebar-tree.html#L32). But there are more than 50 CLI commands. The inconsistency between the summary page content and sidebar confuses users. Since the summary page already lists all commands, this PR proposes to hide the commands in the sidebar.

### DIFF
--- a/site/cli2md/cli2md.go
+++ b/site/cli2md/cli2md.go
@@ -236,13 +236,27 @@ func generateSubcommands(w io.Writer, dir, sectionTitle string, cmds []*kingpin.
 		}
 
 		if first {
-			fmt.Fprintf(w, "\n### %v\n\n", sectionTitle)
+			sectionTitleCutoffIndex := len(sectionTitle)
+			if sectionTitle[sectionTitleCutoffIndex-1] == '.' {
+				sectionTitleCutoffIndex -= 1
+			}
+
+			fmt.Fprintf(w, "\n### %v\n\n", sectionTitle[:sectionTitleCutoffIndex])
 
 			first = false
 		}
 
+		helpSummaryCutoffIndex := strings.Index(c.Help, "\n")
+		if helpSummaryCutoffIndex == -1 {
+			helpSummaryCutoffIndex = len(c.Help)
+		}
+		if c.Help[helpSummaryCutoffIndex-1] == '.' {
+			helpSummaryCutoffIndex -= 1
+		}
+
 		subcommandSlug := strings.Replace(c.FullCommand, " ", "-", -1)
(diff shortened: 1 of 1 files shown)


site/cli2md/cli2md.go
```diff
@@ -236,13 +236,27 @@ func generateSubcommands(w io.Writer, dir, sectionTitle string, cmds []*kingpin.
 		}
 
 		if first {
-			fmt.Fprintf(w, "\n### %v\n\n", sectionTitle)
+			sectionTitleCutoffIndex := len(sectionTitle)
```
jkowalski, line 239: is this `strings.TrimSuffix(..., ".")` ?

site/cli2md/cli2md.go
```diff
@@ -236,13 +236,27 @@ func generateSubcommands(w io.Writer, dir, sectionTitle string, cmds []*kingpin.
...
+			if sectionTitle[sectionTitleCutoffIndex-1] == '.' {
+				sectionTitleCutoffIndex -= 1
+			}
+
+			fmt.Fprintf(w, "\n### %v\n\n", sectionTitle[:sectionTitleCutoffIndex])
 
 			first = false
 		}
 
+		helpSummaryCutoffIndex := strings.Index(c.Help, "\n")
```
jkowalski, line 249: strings.TrimSuffix() again?

jkowalski:
You have linter failures, can you run `make lint` and fix the issues until it comes out clean?

site/cli2md/cli2md.go
jkowalski: sorry, my bad this is something more like `strings.Split(..., "\n")[0]`

site/cli2md/cli2md.go
CrendKing: Nice! Thanks.

site/cli2md/cli2md.go
CrendKing: I chose to not use Split because I thought this has slight performance benefit due to only needing the first token. But I admit I didn't do any benchmark to prove. So if you prefer concise code, I'll change.

CrendKing: Never mind. Found `strings.SplitN`

PR #1533, opened 2021-11-23
Replace google with gcs in examples

kido5217:
Google Cloud Storage repository creation and connection should use "gcs" instead of "google" in cli examples.


jkowalski:

Issue #1524, opened 2021-11-19
kopia snapshot restore is slow(er than expected)


0x53A:
## Theory:

with a hot cache, ``kopia snapshot restore`` should be about as fast as ``robocopy``.

## System:

```
kopia: 0.9.5 build: b4619ffacfb7ba387755b2e714a7dffe9244721d from: kopia/kopia

Windows 10 2004 x64
Lenovo P1 Laptop with i7-8850H@2.6GHz (6C/12T)
VM (unfortunately I can't currently test on bare-metal, but I know the results are similar)
20GB RAM, fixed to VM (no dynamic memory)
fixed size VHDX file on a Samsung 980 Pro 2TB
both cache and restore output are on the same partition
```

## Setup:

```
.\kopia.exe repository create azure --container #### --storage-account #### --storage-key #####
.\kopia.exe policy set --global --compression pgzip
```

Then I uploaded a snapshot with an on-disk size of 13.1 GB, 66k files, 8k folders, containing mostly binary files (exe, dll, etc)

----------

## Test:

```powershell
.\kopia repository connect from-config --token ####
.\kopia cache set --content-cache-size-mb=20000 --metadata-cache-size-mb=20000

# run it once to make sure the cache is hot
.\kopia snapshot restore #### .\snapshot

# run it a second time
(Measure-Command { .\kopia snapshot restore #### .\snapshot | Out-Default }).TotalSeconds

# for comparison, run a robocopy
(Measure-Command { robocopy /MIR /NP /NFL /NDL .\snapshot .\snapshot2 | Out-Default }).TotalSeconds
```

## Result:

In Taskmanager, I can see that the disk is fully utilized, CPU is utilized about one third to half. Disk IO sometimes spikes to ~400 MB/s, but mosty stays below 100 MB/s or even less

![image](https://user-images.githubusercontent.com/4236651/142680703-27f40d1a-c0f5-464a-a13b-f6a1f706bac4.png).

-------------

In comparison, robocopy is mostly managing a steady ~200MB/s to ~600MB/s (with some dips to ~120 MB/s), with 5% CPU utilization. Now, the CPU difference is not a big surprise since kopia does need to decompress the files on-the-fly.

![image](https://user-images.githubusercontent.com/4236651/142681704-5c9e46e3-52c5-4771-98c0-3fbcc4035797.png)

The second kopia restore took 290 seconds.
The first robocopy took 156 seconds.

[...]

PR #1534, opened 2021-11-23
[snapshot restore] use non-atomic writes

0x53A:
fixes https://github.com/kopia/kopia/issues/1524

---------

This is the branch I used for the last test.

Now, for the design, I'm not too happy with passing the parameter __everywhere__, especially since it's currently hardcoded to ``false`` anyway.

Since the function ``copyFileContent`` seems to be only used by ``kopia restore``, the easiest solution would be to always use non-atomic writes here?


snapshot/restore/local_fs_output.go
```diff
@@ -91,11 +91,11 @@ func (o *FilesystemOutput) Close(ctx context.Context) error {
 }
 
 // WriteFile implements restore.Output interface.
-func (o *FilesystemOutput) WriteFile(ctx context.Context, relativePath string, f fs.File) error {
+func (o *FilesystemOutput) WriteFile(ctx context.Context, relativePath string, f fs.File, ensureFileIsWrittenAtomic bool) error {
```
jkowalski, line 94: instead of the flag you can use a field on FilesystemOutput itself. Just make sure the default value of false is reasonable (which it is in this case).

snapshot/restore/local_fs_output.go
```diff
@@ -299,7 +299,30 @@ func (o *FilesystemOutput) createDirectory(ctx context.Context, path string) err
...
+		return errors.Errorf("unable to create file %q: %v", targetPath, err)
+	}
+
+	// ensure we always close f. Note that this does not conflict with  the
+	// close below, as close is idempotent.
+	defer f.Close()
+
+	name := f.Name()
+
+	if _, err := io.Copy(f, r); err != nil {
```
jkowalski, line 314: iocopy.JustCopy() will generally manage buffers better.

snapshot/restore/local_fs_output.go
```diff
@@ -299,7 +299,30 @@ func (o *FilesystemOutput) createDirectory(ctx context.Context, path string) err
 	}
 }
 
-func (o *FilesystemOutput) copyFileContent(ctx context.Context, targetPath string, f fs.File) error {
+func write(targetPath string, r fs.Reader) error {
+	f, err := os.Create(targetPath)
+	if err != nil {
+		return errors.Errorf("unable to create file %q: %v", targetPath, err)
```
jkowalski, line 305: Use `errors.Wrap()` or in this case return `err` directly since filesystem errors already include path in the text.

snapshot/restore/local_fs_output.go
0x53A: iocopy.JustCopy: 96.9 seconds
io.Copy: 106 seconds
iocopy.JustCopy: 95 seconds
io.Copy: 122 seconds

after these 4 runs my results went a bit all over the place, maybe heat throttling or something, but iocopy.JustCopy definitely seems to perform much better

0x53A:
Thanks! I think that should be mostly it, I didn't test the server part though.

Should I rename ``write`` and/or move it to a different file?

jkowalski:
Looks like one more failure:

  restore_test.go:488: invalid mode on /tmp/kopia-test1102773437/fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff/restored-5: -rw-r--r--, want -rw-------

0x53A:
Thanks for pointing out the exact error!

So it's a bit weird, I think that's a side effect of ``atomic`` using ``ioutil.TempFile``, which uses ``os.CreateTemp``, which sets the mode to ``0600``, whereas ``os.Create`` sets it to ``0666``.

Both values are hardcoded, so the only option is to explicitly change it afterwards.

snapshot/restore/local_fs_output.go
```diff
@@ -315,6 +315,13 @@ func write(targetPath string, r fs.Reader) error {
 
 	name := f.Name()
 
+	// note: atomic.WriteFile uses os.CreateTemp internally, which sets the mode to 0600.
```
jkowalski, line 318: How about:

```
os.OpenFile(name, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o600)
```

0x53A: yeah that's much better, thanks!

jkowalski:
Looks like linter is troubling you, make sure to run `make lint` before updating the PR. If that comes out clean, you're generally good to go.

0x53A:
Yeah the last few changes I just did in the github UI, and didn't know how to invoke the linter.

Linter is now clean, let's see what happens:
image

Assuming this is OK, can you rebase on merge, or should I manually rebase?

jkowalski:
i can rebase, no problem.

jkowalski:

PR #1536, opened 2021-11-24
Add doc for sharding

CrendKing:
Also, I wonder if it's good idea to only do `[3, 3]` sharding for local file system repos, i.e. `repository create s3` would always do `[]`?


jkowalski:
We recently switched to [1,3] as the default, so the example should probably be updated.

See #1513

CrendKing:
I changed to a way that eludes actual value. Hope it works.

PR #1589, opened 2021-12-15
azure: migrated provider to newer client library, remove gocloud.dev dependency

jkowalski:


internal/gather/gather_bytes.go
```diff
@@ -114,26 +114,57 @@ func (b Bytes) ReadAt(p []byte, off int64) (n int, err error) {
 	return len(p), b.AppendSectionTo(bytes.NewBuffer(p[:0]), int(off), len(p))
 }
 
-// Reader returns a reader for the data.
-func (b Bytes) Reader() io.Reader {
-	b.assertValid()
+type bytesReadSeekCloser struct {
```
julio-lopez, line 117: `bytes.Reader` already implements `io.ReadSeeker`. It's only missing `Close()`. Why not extend `bytes.Reader` with just `Close()`? What am I missing?

julio-lopez (commented):
+1: LG 👍

See question inline

CI failed due to Azure license. Needs to be checked.

internal/gather/gather_bytes.go
jkowalski: there's still a case of MultiReader, which don't implement io.ReadSeeker unfortunately.

.wwhrd.yml
```diff
@@ -14,3 +14,6 @@ allowlist:
   - MPL-2.0
   - CC0-1.0
   - ISC
+
+exceptions:
+ - github.com/Azure/azure-sdk-for-go/sdk/internal/...
```
julio-lopez, line 19: nit: add EOL

PR #1590, opened 2021-12-15
versioned map mock storage for blob-retention unit-testing

Shrekster:
This internal mock will help test the blob-retention changes in an upcoming PR. This PR adds a versioned in-memory blob-store to kopia which allows setting retention times and delete-markers on put and delete api calls respectively. This will allow unit-testing various versioned buckets storage features (like blob-retention) without the need for external environmental and infrastructure dependencies.

### DIFF
--- a/internal/blobtesting/versionedmap.go
+++ b/internal/blobtesting/versionedmap.go
@@ -1,0 +1,248 @@
+package blobtesting
+
+import (
+	"bytes"
+	"context"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/internal/clock"
+	"github.com/kopia/kopia/repo/blob"
+)
+
+type entry struct {
+	value          []byte
+	mtime          time.Time
+	retentionTime  time.Time
+	isDeleteMarker bool
+}
+
+type versionedEntries map[blob.ID][]*entry
+
+type versionedMapStorage struct {
+	data    versionedEntries
+	timeNow func() time.Time
+	mutex   sync.RWMutex
+}
+
+func (s *versionedMapStorage) getLatestByID(id blob.ID, writeIntent bool) (*entry, error) {
+	versions, ok := s.data[id]
+	if !ok {
+		return nil, blob.ErrBlobNotFound
+	}
+
+	// get the latest version and if it is a delete marker then simulate
+	// not-found
+	e := versions[len(versions)-1]
+	if e.isDeleteMarker {
+		return nil, blob.ErrBlobNotFound
+	}
+
+	if writeIntent && !e.retentionTime.IsZero() && e.retentionTime.After(s.timeNow()) {
+		return nil, errors.New("cannot alter object before retention period expires")
+	}
+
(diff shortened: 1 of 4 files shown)


jkowalski (approved):

internal/blobtesting/versionedmap.go
```diff
@@ -0,0 +1,248 @@
...
+	// get the latest version and if it is a delete marker then simulate
+	// not-found
+	e := versions[len(versions)-1]
+	if e.isDeleteMarker {
+		return nil, blob.ErrBlobNotFound
+	}
+
+	if writeIntent && !e.retentionTime.IsZero() && e.retentionTime.After(s.timeNow()) {
+		return nil, errors.New("cannot alter object before retention period expires")
+	}
```
julio-lopez, line 47: I'm not sure I understand this part. This seems to be related to object locking.
Can this be done in the caller and then remove the `writeIntent` parameter?

The operation that should fail when the retention has not elapsed is deleting a blob version. `PutBlob` and `DeleteBlob` should succeed even if the retention has not elapsed.

internal/blobtesting/versionedmap.go
```diff
@@ -0,0 +1,248 @@
...
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/internal/clock"
+	"github.com/kopia/kopia/repo/blob"
+)
+
+type entry struct {
```
julio-lopez, line 17: It seems that the intent is to implement not only versioning, but object locking as well. It'd make sense to rename the data structs, functions and files to reflect that.
=> `objectLockingMap` or similar.

internal/blobtesting/versionedmap.go
```diff
@@ -0,0 +1,248 @@
...
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	// ths will prevent stacking (if latest is already a marker) or trailing
+	// (if entry does not exist) delete-markers
+	if _, err := s.getLatestByID(id, false); err != nil {
+		// no error if already deleted
+		if errors.Is(err, blob.ErrBlobNotFound) {
+			return nil
+		}
```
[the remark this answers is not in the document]
Shrekster: I think it is some 2XX status.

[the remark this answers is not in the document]
Shrekster: 204 (No Content), doesn't matter whether the object exists or not

https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html

internal/blobtesting/versionedmap.go
Shrekster: > Can this be done in the caller and then remove the writeIntent parameter?

Yes. The `writeIntent` allowed reuse of the same check. If `DeleteBlob` should succeed then we can move this to `PutBlob` because it is not a common check elsewhere.

internal/blobtesting/versionedmap.go
Shrekster: Will do.

julio-lopez (approved):

PR #1595, opened 2021-12-17
blob: replaced blob.Storage.SetTime() method with blob.PutOptions.SetTime

jkowalski:
This was only used for `kopia repo sync-to` and got replaced with
an equivalent blob.PutOptions.SetTime, which when set to non-zero time
will attempt to set the modification time on a file.

Since some providers don't support changing modification time, we
are able to emulate it using per-blob metadata (on B2, Azure and GCS),
sadly S3 is still unsupported, because it does not support returning
metadata in list results.

Also added PutOptions.GetTime, which when set to not nil, will
populate the provided variable with actual time that got assigned
to the blob.

Added tests that verify that each provider supports GetTime
and SetTime according to this spec.

### DIFF
--- a/cli/command_repository_sync.go
+++ b/cli/command_repository_sync.go
@@ -30,9 +30,8 @@ type commandRepositorySyncTo struct {
 	repositorySyncDestinationMustExist bool
 	repositorySyncTimes                bool
 
-	lastSyncProgress       string
-	syncProgressMutex      sync.Mutex
-	setTimeUnsupportedOnce sync.Once
+	lastSyncProgress  string
+	syncProgressMutex sync.Mutex
 
 	out textOutput
 }
@@ -303,19 +302,25 @@ func (c *commandRepositorySyncTo) syncCopyBlob(ctx context.Context, m blob.Metad
 		return errors.Wrapf(err, "error reading blob '%v' from source", m.BlobID)
 	}
 
-	if err := dst.PutBlob(ctx, m.BlobID, data.Bytes(), blob.PutOptions{}); err != nil {
-		return errors.Wrapf(err, "error writing blob '%v' to destination", m.BlobID)
+	opt := blob.PutOptions{}
+	if c.repositorySyncTimes {
+		opt.SetModTime = m.Timestamp
 	}
 
-	if c.repositorySyncTimes {
-		if err := dst.SetTime(ctx, m.BlobID, m.Timestamp); err != nil {
-			if errors.Is(err, blob.ErrSetTimeUnsupported) {
-				c.setTimeUnsupportedOnce.Do(func() {
-					log(ctx).Infof("destination repository does not support setting time")
-				})
-			}
+	if err := dst.PutBlob(ctx, m.BlobID, data.Bytes(), opt); err != nil {
+		if errors.Is(err, blob.ErrSetTimeUnsupported) {
+			// run again without SetModTime, emit a warning
+			opt.SetModTime = time.Time{}
 
-			return errors.Wrapf(err, "error setting time on destination '%v'", m.BlobID)
+			log(ctx).Warnf("destination repository does not support preserving modification times")
+
+			c.repositorySyncTimes = false
+
+			err = dst.PutBlob(ctx, m.BlobID, data.Bytes(), opt)
+		}
+
+		if err != nil {
+			return errors.Wrapf(err, "error writing blob '%v' to destination", m.BlobID)
 		}
 	}
 
--- a/internal/timestampmeta/timestampmeta.go
+++ b/internal/timestampmeta/timestampmeta.go
@@ -1,0 +1,30 @@
+// Package timestampmeta provides utilities for preserving timestamps
+// using per-blob key-value-pairs (metadata, tags, etc.)
+package timestampmeta
+
+import (
+	"strconv"
+	"time"
+)
+
+// ToMap returns a map containing single entry representing the provided time or nil map
+// if the time is zero. The key-value pair map should be stored alongside the blob.
+func ToMap(t time.Time, mapKey string) map[string]string {
+	if t.IsZero() {
+		return nil
+	}
+
+	return map[string]string{
+		mapKey: strconv.FormatInt(t.UnixNano(), 10), // nolint:gomnd
+	}
+}
+
+// FromValue attempts to convert the provided value stored in metadata into time.Time.
+func FromValue(v string) (t time.Time, ok bool) {
+	nanos, err := strconv.ParseInt(v, 10, 64) // nolint:gomnd
+	if err != nil {
+		return time.Time{}, false
+	}
+
+	return time.Unix(0, nanos), true
+}
--- a/repo/blob/azure/azure_storage.go
+++ b/repo/blob/azure/azure_storage.go
@@ -20,7 +20,7 @@ import (
 const (
 	azStorageType = "azureBlob"
 
-	timeMapKey = "Kopiamtime" // this must be capital letter followed by lowercase, don't ask
+	timeMapKey = "Kopiamtime" // this must be capital letter followed by lowercase, to comply with AZ tags naming convention.
 )
 
 type azStorage struct {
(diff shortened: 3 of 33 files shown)


internal/timestampmeta/timestampmeta.go
```diff
@@ -0,0 +1,30 @@
+// Package timestampmeta provides utilities for preserving timestamps
+// using per-blob key-value-pairs (metadata, tags, etc.)
+package timestampmeta
```
Shrekster, line 3: I think is is more of a Kopia "encoding" package where more of similar future encoding/decoding methods could live.

repo/blob/azure/azure_storage.go
```diff
@@ -112,13 +122,24 @@ func (az *azStorage) PutBlob(ctx context.Context, b blob.ID, data blob.Bytes, op
...
-	return translateError(err)
-}
+	resp, err := bc.Upload(ctx, data.Reader(), ubo)
+	if err != nil {
+		return translateError(err)
+	}
+
+	if opts.GetModTime != nil {
+		if opts.SetModTime.IsZero() {
+			*opts.GetModTime = *resp.LastModified
```
Shrekster, line 136: wouldn't it be better to rely on `resp.LastModified` always whether or not `opts.SetModTime` was defined ?

repo/blob/azure/azure_storage.go
```diff
@@ -11,12 +12,15 @@ import (
...
 	"github.com/kopia/kopia/internal/iocopy"
+	"github.com/kopia/kopia/internal/timestampmeta"
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/blob/retrying"
 )
 
 const (
 	azStorageType = "azureBlob"
+
+	timeMapKey = "Kopiamtime" // this must be capital letter followed by lowercase, don't ask
```
Shrekster, line 23:
```suggestion
	timeMapKey = "Kopiamtime" // this must be capital letter followed by lowercase to comply with AZ tags naming convention
```

repo/blob/azure/azure_storage.go
```diff
@@ -144,16 +165,53 @@ func (az *azStorage) ListBlobs(ctx context.Context, prefix blob.ID, callback fun
...
+			}
+
+			// see if we have 'Kopiamtime' metadata, if so - trust it.
+			if t, ok := timestampmeta.FromValue(it.Metadata.Kopiamtime); ok {
+				bm.Timestamp = t
+			} else {
+				// fall back to using last modified time.
+				t, err := time.Parse(time.RFC1123, it.Properties.LastModified)
+				if err != nil {
+					return errors.Wrap(err, "invalid timestamp")
```
Shrekster, line 211:
```suggestion
					return errors.Wrap(err, "invalid timestamp: %q", it.Properties.LastModified)
```

repo/blob/b2/b2_storage.go
```diff
@@ -149,13 +158,20 @@ func (s *b2Storage) PutBlob(ctx context.Context, id blob.ID, data blob.Bytes, op
...
+		return translateError(err)
+	}
 
-	return translateError(err)
-}
+	if opts.GetModTime != nil {
+		if opts.SetModTime.IsZero() {
+			*opts.GetModTime = time.Unix(0, fi.UploadTimestamp*1e6)
+		} else {
+			*opts.GetModTime = opts.SetModTime
```
Shrekster, line 170: nit; suggestion to use the time from the response either way

repo/blob/filesystem/filesystem_storage.go
```diff
@@ -188,6 +188,21 @@ func (fs *fsImpl) PutBlobInPath(ctx context.Context, dirPath, path string, data
 			}
 		}
 
+		if t := opts.SetModTime; !t.IsZero() {
+			if chtimesErr := fs.osi.Chtimes(path, t, t); err != nil {
+				return errors.Wrap(chtimesErr, "can't change file times")
```
Shrekster, line 193:
```suggestion
				return errors.Wrapf(chtimesErr, "can't change file %q times", path)
```

repo/blob/filesystem/filesystem_storage.go
```diff
@@ -188,6 +188,21 @@ func (fs *fsImpl) PutBlobInPath(ctx context.Context, dirPath, path string, data
...
+			}
+		}
+
+		if t := opts.GetModTime; t != nil {
+			fi, err := fs.osi.Stat(path)
+			if err != nil {
+				return errors.Wrap(err, "can't get mod time")
```
Shrekster, line 200:
```suggestion
				return errors.Wrapf(err, "can't get mod time for file %q", path)
```

cli/command_repository_sync.go
```diff
@@ -303,19 +302,25 @@ func (c *commandRepositorySyncTo) syncCopyBlob(ctx context.Context, m blob.Metad
...
-				})
-			}
+	if err := dst.PutBlob(ctx, m.BlobID, data.Bytes(), opt); err != nil {
+		if errors.Is(err, blob.ErrSetTimeUnsupported) {
+			// run again without SetModTime, emit a warning
+			opt.SetModTime = time.Time{}
+
+			log(ctx).Warnf("destination repository does not support preserving modification times")
+
+			c.repositorySyncTimes = false
```
Shrekster, line 317: nit; nothing t todo here, just a mention

could be "undefined behavior" for a bit but should be good eventually ...

repo/blob/filesystem/osinterface_mock_test.go
```diff
@@ -98,6 +100,14 @@ func (osi *mockOS) Stat(fname string) (fs.FileInfo, error) {
 	return osi.osInterface.Stat(fname)
 }
 
+func (osi *mockOS) Chtimes(fname string, atime, mtime time.Time) error {
+	if atomic.AddInt32(&osi.chtimesRemainingErrors, -1) >= 0 {
+		return &os.PathError{Op: "chtimes", Err: errors.Errorf("underlying problem")}
+	}
```
Shrekster, line 106:
```suggestion
	if atomic.AddInt32(&osi.chtimesRemainingErrors, -1) <= 0 {
		return &os.PathError{Op: "chtimes", Err: errors.Errorf("too many Chtime() failures")}
	}
```

repo/blob/sftp/sftp_storage.go
```diff
@@ -202,15 +201,22 @@ func (s *sftpImpl) PutBlobInPath(ctx context.Context, dirPath, fullPath string,
 			return errors.Wrap(err, "unexpected error renaming file on SFTP")
 		}
 
-		return nil
-	})
-}
+		if t := opts.SetModTime; !t.IsZero() {
+			if chtimesErr := sftpClientFromConnection(conn).Chtimes(fullPath, t, t); err != nil {
+				return errors.Wrap(chtimesErr, "can't change file times")
```
Shrekster, line 206:
```suggestion
				return errors.Wrapf(chtimesErr, "can't change file %q times", fullPath)
```

Shrekster: same suggestion on other fs-op failures, having the error and also the full path is helpful on returned errors

repo/blob/azure/azure_storage.go
jkowalski: done

repo/blob/azure/azure_storage.go
jkowalski: done, applied this everywhere

repo/blob/b2/b2_storage.go
jkowalski: done

repo/blob/sftp/sftp_storage.go
jkowalski: i think all these errors already mention the path.

Shrekster:
Would be good to also drop the `SetTime` from `internal/blobtesting/object_locking_map.go` in latest master.

LGTM otherwise! Thanks!

julio-lopez (commented):
+1: LG 👍

PR #1596, opened 2021-12-17
feat(repository): persistence for blob-retention configuration

Shrekster:
This PR adds:

- an authenticated repository blob `kopia.retention` when the repository is created with blob-retention options.
- updates the change-password mechanism to rewrite the retention blob
- adds plumbing for retention options to be supplied to index and pack blobs puts


internal/repotesting/repotesting.go
```diff
@@ -70,7 +76,14 @@ func (e *Environment) setup(tb testing.TB, version content.FormatVersion, opts .
 		}
 	}
 
-	st := blobtesting.NewMapStorage(blobtesting.DataMap{}, nil, openOpt.TimeNowFunc)
+	var st blob.Storage
+	if opt.RetentionPeriod == 0 && opt.RetentionMode == "" {
```
redgoat650, line 80: might not be too important or relevant, but just wondering if this should be `||` instead of `&&`, to match the behavior of `retentionBlob`'s `IsNull` method?

Shrekster: Yeah, will make it consistent with IsNull.

cli/command_repository_create.go
```diff
@@ -2,8 +2,10 @@ package cli
 
 import (
 	"context"
+	"time"
 
 	"github.com/alecthomas/kingpin"
+	"github.com/minio/minio-go/v7"
```
jkowalski, line 8: that is not a desireable dependency, can we hardcode the constant instead?

repo/retention_blob.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// RetentionBlobID is the identifier of a BLOB that describes BLOB retention
+// settings for the repository.
+const RetentionBlobID = "kopia.retention"
```
jkowalski, line 12: perhaps `kopia.blobcfg` so we can add more things there?

repo/blob/beforeop/beforeop.go
```diff
@@ -0,0 +1,73 @@
...
+type (
+	callback          func() error
+	onGetBlobCallback func(id blob.ID) error
+	onPutBlobCallback func(id blob.ID, opts *blob.PutOptions) error // allows mutating the put-options
+)
+
+type beforeOp struct {
+	blob.Storage
+	onGetMetadata, onDeleteBlob callback
+	onGetBlob                   onGetBlobCallback
```
jkowalski, line 19: only PutBlob is used in this PR, I presume the rest is for the upgrade PR?

repo/format_block.go
```diff
@@ -17,6 +17,7 @@ import (
 )
 
 const (
+	aes256GcmEncryption             = "AES256_GCM"
 	defaultFormatEncryption         = "AES256_GCM"
```
jkowalski, line 21: nit: use defaultFormatEncryption = aes256GcmEncryption 

also, Gcm should be spelled GCM per Golang convention.

repo/retention_blob.go
```diff
@@ -0,0 +1,78 @@
...
+
+type retentionBlob struct {
+	Mode   string        `json:"mode,omitempty"`
```
jkowalski, line 15: for future proofing, can this be `retentionMode` and `retentionPeriod` ?

repo/retention_blob.go
Shrekster: done

repo/blob/beforeop/beforeop.go
Shrekster: correct

repo/retention_blob.go
Shrekster: done, also renamed the struct

repo/format_block.go
Shrekster: I tried this earlier, but the linter was getting confused, not yet sure why. Might have to file a linter bug.

cli/command_repository_create.go
Shrekster: Good point. We can make some constants yes. Will do.

repo/retention_blob.go
Shrekster: f65715a

cli/command_repository_create.go
Shrekster: done

jkowalski:
Thank you!

PR #1600, opened 2021-12-18
htmlui: moved HTML UI to separate repository

jkowalski:
The source code for htmlui is now in github.com/kopia/htmlui
GitHub Actions will drop compiled builds in github.com/kopia/htmluibuild
where they are now used as a go module dependency.

This greatly simplifies the build and improves the security,
because Kopia will be consuming pre-built htmlui.

This also means kopia can now installed with embedded UI using:
`go install github.com/kopia/kopia@latest`


internal/server/htmlui_embed.go
```diff
@@ -1,13 +1,12 @@
-// +build embedhtml
```
julio-lopez, line 1: Let's bring back a way of building without the embedded HTML. We can default to including the HTML UI, and have a build flag to build without the embedded HTML.

PR #1615, opened 2021-12-22
s3/test: refactor retention tests

julio-lopez:
Refactors retention tests to coalesce a couple of tests. It has the following effects:

- serializes the tests;
- reuses the bucket and reduces bucket creation calls.

Also, checks whether a bucket exists before attempting to create it to reduce the number of bucket creation calls.

### DIFF
--- a/repo/blob/s3/s3_storage_test.go
+++ b/repo/blob/s3/s3_storage_test.go
@@ -186,7 +186,7 @@ func TestS3StorageRetentionUnlockedBucket(t *testing.T) {
 	t.Parallel()
 
 	// skip the test if AWS creds are not provided
-	options := &Options{
+	options := Options{
 		Endpoint:        getEnv(testEndpointEnv, awsEndpoint),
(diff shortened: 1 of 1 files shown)


repo/blob/s3/s3_storage_test.go
```diff
@@ -526,18 +509,26 @@ func createClient(tb testing.TB, opt *Options) *minio.Client {
...
-	if err := cli.MakeBucket(context.Background(), opt.BucketName, minio.MakeBucketOptions{
+	ctx := context.Background()
+
+	if loc, err := cli.GetBucketLocation(ctx, opt.BucketName); err == nil {
+		tb.Log("found bucket", opt.BucketName, "in location", loc)
+
+		return
+	}
+
+	if err := cli.MakeBucket(ctx, opt.BucketName, minio.MakeBucketOptions{
```
Shrekster, line 531: nit; requesting a one-liner here reasoning why we are doing a get-bucket instead of handling status-exists from make-bucket.

Shrekster (approved):
Thank you!

Issue #1560, opened 2021-12-02
Maintenance on S3 leaking threads


raisinbear:
Hi,

I asked about this earlier today on the Kopia forum: [Post](https://kopia.discourse.group/t/maintenance-run-crashing-minio-s3-backend/786)

In short: I suspect that `kopia maintenance run` is leaking threads when iterating over unreferenced blobs. It reliably crashes my low-key minio server by eventually spawning (or rather trying to) > 11466 threads, which is the OS hard limit. Repo size is roughly 400 GB and 14000 blobs, so probably nothing extraordinary.
I did not encounter any other kopia command that brings the number of minio threads above 100 if at all.

For completeness sake, this is the kopia error reported to stdout once the minio server side OS interferes:
`2021-12-01T09:23:10.794863Z ERROR kopia/cli error deleting unreferenced blobs: error looking for unreferenced blobs: error iterating blobs: Get "[ DOMAIN ]/kopia/?delimiter=%2F&encoding-type=url&fetch-owner=true&list-type=2&prefix=p7": net/http: timeout awaiting response headers`

Thanks a lot for maintaining this awesome project!

potatoru:
Sorry, but I have the same issue with my minio setup.

PR #1629, opened 2021-12-29
logging: log maximum concurrency level when talking to backend storage

jkowalski:
Trying to understand if we're doing something wrong for #1560

### DIFF
--- a/repo/blob/logging/logging_storage.go
+++ b/repo/blob/logging/logging_storage.go
@@ -3,6 +3,7 @@ package logging
 
 import (
 	"context"
+	"sync/atomic"
 
 	"github.com/kopia/kopia/internal/timetrack"
 	"github.com/kopia/kopia/repo/blob"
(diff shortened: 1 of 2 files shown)


repo/blob/logging/logging_storage.go
```diff
@@ -3,19 +3,41 @@ package logging
...
 	logger logging.Logger
 }
 
+func (s *loggingStorage) beginConcurrency() {
+	v := atomic.AddInt32(&s.concurrency, 1)
+
+	if mv := atomic.LoadInt32(&s.maxConcurrency); v > mv {
+		if atomic.CompareAndSwapInt32(&s.maxConcurrency, mv, v) {
+			s.logger.Debugw(s.prefix+"concurrency level reached",
+				"maxConcurrency", mv)
```
julio-lopez, line 28: minor:

Is there a _small_ race here?

Suppose:
* the state is `concurrency == maxConcurrency`, for example `7`;
* there are two routines `g1` and `g2` concurrently running `beginConcurrency`;
* `g1` gets `v == 8` and `g2` gets `v == 9`;
* both `g1` and `g2` get `mv == 7` and both attempt to update `maxConcurrency`;
* `g1`'s update succeeds and now `maxConcurrency == 8`, but `concurrency > maxConcurrency`

This may self correct if there is a higher concurrency value down the road.

julio-lopez (commented):
@jkowalski PTAL

PR #1653, opened 2022-01-12
fix(security): prevent cross-site request forgery in the UI website

jkowalski:
This fixes a [cross-site request forgery (CSRF)](https://en.wikipedia.org/wiki/Cross-site_request_forgery)
vulnerability in self-hosted UI for Kopia server.

The vulnerability allows potential attacker to make unauthorized API calls against a running Kopia server. It requires an attacker to trick the user into visiting a malicious website while also logged into a Kopia website.

The vulnerability only affected self-hosted Kopia servers with UI. The following configurations were not vulnerable:

* Kopia Repository Server without UI
* KopiaUI (desktop app)
* command-line usage of `kopia`

All users are strongly recommended to upgrade at the earliest convenience.

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/alecthomas/kingpin"
 	"github.com/fatih/color"
+	"github.com/gorilla/mux"
 	"github.com/mattn/go-colorable"
 	"github.com/pkg/errors"
 
@@ -425,21 +426,21 @@ func (c *App) maybeRepositoryAction(act func(ctx context.Context, rep repo.Repos
 			defer gather.DumpStats(ctx)
 
 			if c.metricsListenAddr != "" {
-				mux := http.NewServeMux()
-				if err := initPrometheus(mux); err != nil {
+				m := mux.NewRouter()
+				if err := initPrometheus(m); err != nil {
 					return errors.Wrap(err, "unable to initialize prometheus.")
 				}
 
 				if c.enablePProf {
-					mux.HandleFunc("/debug/pprof/", pprof.Index)
-					mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
-					mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
-					mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
-					mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+					m.HandleFunc("/debug/pprof/", pprof.Index)
+					m.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+					m.HandleFunc("/debug/pprof/profile", pprof.Profile)
+					m.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+					m.HandleFunc("/debug/pprof/trace", pprof.Trace)
 				}
 
 				log(ctx).Infof("starting prometheus metrics on %v", c.metricsListenAddr)
-				go http.ListenAndServe(c.metricsListenAddr, mux) // nolint:errcheck
+				go http.ListenAndServe(c.metricsListenAddr, m) // nolint:errcheck
 			}
 
 			memtrack.Dump(ctx, "before openRepository")
--- a/cli/command_server_start.go
+++ b/cli/command_server_start.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"contrib.go.opencensus.io/exporter/prometheus"
+	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
 	prom "github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
@@ -30,10 +31,11 @@ type commandServerStart struct {
 	co connectOptions
 
 	serverStartHTMLPath string
-	serverStartUI       bool
 
+	serverStartUI                  bool
 	serverStartLegacyRepositoryAPI bool
 	serverStartGRPC                bool
+	serverStartControlAPI          bool
 
 	serverStartRefreshInterval time.Duration
 	serverStartInsecure        bool
@@ -63,6 +65,8 @@ type commandServerStart struct {
 
 	logServerRequests bool
 
+	disableCSRFTokenChecks bool // disable CSRF token checks - used for development/debugging only
+
 	sf  serverFlags
 	svc advancedAppServices
 	out textOutput
@@ -75,6 +79,7 @@ func (c *commandServerStart) setup(svc advancedAppServices, parent commandParent
 
 	cmd.Flag("legacy-api", "Start the legacy server API").Default("true").BoolVar(&c.serverStartLegacyRepositoryAPI)
 	cmd.Flag("grpc", "Start the GRPC server").Default("true").BoolVar(&c.serverStartGRPC)
+	cmd.Flag("control-api", "Start the control API").Default("true").BoolVar(&c.serverStartControlAPI)
 
 	cmd.Flag("refresh-interval", "Frequency for refreshing repository status").Default("300s").DurationVar(&c.serverStartRefreshInterval)
 	cmd.Flag("insecure", "Allow insecure configurations (do not use in production)").Hidden().BoolVar(&c.serverStartInsecure)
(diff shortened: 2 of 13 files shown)


cli/command_server_start.go
```diff
@@ -222,7 +222,29 @@ func (c *commandServerStart) run(ctx context.Context, rep repo.Repository) error
...
+		srv.SetupControlAPIHandlers(m)
+	}
+
+	if c.serverStartUI {
+		if c.serverStartHTMLPath != "" {
+			srv.ServeStaticFiles(m, http.Dir(c.serverStartHTMLPath))
+		} else {
+			srv.ServeStaticFiles(m, server.AssetFile())
+		}
+	}
```
julio-lopez, line 244: Any reason for not including this block in the first `if c.serverStartUI {` above?
Are there any implicit dependencies?

julio-lopez (commented):
@Shrekster should take a look.

internal/server/server_authz_checks.go
```diff
@@ -2,16 +2,66 @@ package server
...
 	"net/http"
 
 	"github.com/kopia/kopia/internal/auth"
 )
 
+const kopiaSessionCookie = "Kopia-Session-Cookie"
+
+func (s *Server) generateCSRFToken(sessionID string) string {
+	h := hmac.New(sha256.New, s.authCookieSigningKey)
+	io.WriteString(h, sessionID) //nolint:errcheck
```
Shrekster, line 19: nit; better to panic-on-err if you believe that is an impossible edge / would indicate memory issues. there is nothing productive to be done when that'd happen.

```suggestion
	if _, err := io.WriteString(h, sessionID); err != nil {
	    panic("io.WriteString() failed: " + err.Error())
	}
```

internal/server/server_authz_checks.go
```diff
@@ -2,16 +2,66 @@ package server
...
+
+	ctx := r.Context()
+	path := r.URL.Path
+
+	sessionCookie, err := r.Cookie(kopiaSessionCookie)
+	if err != nil {
+		log(ctx).Warnf("missing or invalid session cookie for %v", path)
+
+		return false
+	}
```
Shrekster, line 37: nit; It does simplify a few things if `validateCSRFToken` returns a boolean always. However, in this particular scenario it'll eat some precious debugging information when systems are under load/memory-pressure.

I'd be desirable if `validateCSRFToken` returns false only for `ErrNoCookie`. For anything else `validateCSRFToken` should probably wrap the error with some context and return it along with `false`.

internal/server/server_authz_checks.go
```diff
@@ -2,16 +2,66 @@ package server
...
+	if s.options.DisableCSRFTokenChecks {
+		return true
+	}
+
+	ctx := r.Context()
+	path := r.URL.Path
+
+	sessionCookie, err := r.Cookie(kopiaSessionCookie)
+	if err != nil {
+		log(ctx).Warnf("missing or invalid session cookie for %v", path)
```
Shrekster, line 34: Best to also include the error in the log.

```suggestion
		log(ctx).Warnf("missing or invalid session cookie for %v: %v", path, err)
```

internal/server/server_authz_checks.go
```diff
@@ -2,16 +2,66 @@ package server
...
+
+		return false
+	}
+
+	validToken := s.generateCSRFToken(sessionCookie.Value)
+
+	token := r.Header.Get("X-Kopia-Csrf-Token")
```
Shrekster, line 41: `X-Kopia-Csrf-Token` is a good candidate to be a const

cli/app.go
```diff
@@ -425,21 +426,21 @@ func (c *App) maybeRepositoryAction(act func(ctx context.Context, rep repo.Repos
...
-					mux.HandleFunc("/debug/pprof/", pprof.Index)
-					mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
-					mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
-					mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
-					mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+					m.HandleFunc("/debug/pprof/", pprof.Index)
+					m.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+					m.HandleFunc("/debug/pprof/profile", pprof.Profile)
+					m.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+					m.HandleFunc("/debug/pprof/trace", pprof.Trace)
```
Shrekster, line 439: nit; Probably should also specify the methods allowed for additional validation, for example:

```
m.HandleFunc("/debug/pprof/symbol", pprof.Symbol).Methods("GET", "POST")
```

cli/app.go
```diff
@@ -425,21 +426,21 @@ func (c *App) maybeRepositoryAction(act func(ctx context.Context, rep repo.Repos
...
 					return errors.Wrap(err, "unable to initialize prometheus.")
 				}
 
 				if c.enablePProf {
-					mux.HandleFunc("/debug/pprof/", pprof.Index)
-					mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
-					mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
-					mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
-					mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+					m.HandleFunc("/debug/pprof/", pprof.Index)
```
Shrekster, line 435: nit; might be slightly cleaner to create sub-router for pprof:

```go
pr := m.PathPrefix("/debug/pprof").
	Methods("POST", "GET").
	Schemes("http", "https").
	Subrouter()
pr.HandleFunc("/", pprof.Index)
pr.HandleFunc("/cmdline", pprof.Cmdline)
pr.HandleFunc("/profile", pprof.Profile)
pr.HandleFunc("/symbol", pprof.Symbol)
pr.HandleFunc("/trace", pprof.Trace)
```

cli/command_server_start.go
```diff
@@ -104,6 +109,7 @@ func (c *commandServerStart) setup(svc advancedAppServices, parent commandParent
 	cmd.Flag("ui-preferences-file", "Path to JSON file storing UI preferences").StringVar(&c.uiPreferencesFile)
 
 	cmd.Flag("log-server-requests", "Log server requests").Hidden().BoolVar(&c.logServerRequests)
+	cmd.Flag("disable-csrf-token-checks", "Disable CSRF token").Hidden().BoolVar(&c.disableCSRFTokenChecks)
```
Shrekster, line 112: Wondering which approach would be better: "disabling CSRF checks in tests" vs "test-api to generate a CSRF token" for the test calls to the server ?

Right now it seems disabling is okay. Will think more about this ...

internal/server/server_authz_checks.go
jkowalski: I don't see why this matters, CSRF token that is missing, malformed or invalid should have the same effect, right?

cli/app.go
jkowalski: this is the default and I would expect built-in handlers to take care of validation if they don't handle some particular method.

cli/app.go
jkowalski: in my opinion it's easier to code-search for those strings if they are explicit and not mangled. For the same reason I use "/api/v1" prefix explicitly everywhere.

cli/command_server_start.go
jkowalski: Disabling is only used when developing the UI in a dev server using a proxy, so very rare.

internal/server/server_authz_checks.go
jkowalski: done

internal/server/server_authz_checks.go
jkowalski: done

internal/server/server_authz_checks.go
jkowalski: done

cli/command_server_start.go
jkowalski: no reason, done.

Shrekster (approved):
LGTM

cli/app.go
Shrekster: fair enough

internal/server/server_authz_checks.go
Shrekster: missing would throw `ErrNoCookie`

malformed / invalid would be something we probably should try not to return a boolean for, that in my head is indicative of a third state which can be "I don't know".

not a deal breaker though, so please don't block on this comment

PR #1654, opened 2022-01-12
feat(providers): expand PutBlob API to allow for idempotent puts

adowair:
Today, executing a PutBlob operation will always overwrite a blob if it exists. It would be helpful for some customer workflows to be able to write blobs only if they don't exist (i.e. don't overwrite/recreate blobs). This PR expands the PutBlob API with an additional option `DoNotRecreate`, which toggles this feature. Note that not all backends support idempotent creates, so this PR also implements error logic to rejects such PutBlobs if they can't be executed.

### DIFF
--- a/internal/providervalidation/providervalidation.go
+++ b/internal/providervalidation/providervalidation.go
@@ -31,18 +31,21 @@ type Options struct {
 	NumGetMetadataWorkers int
 	NumListBlobsWorkers   int
 	MaxBlobLength         int
+
+	SupportIdempotentCreates bool
 }
 
 // DefaultOptions is the default set of options.
 // nolint:gomnd,gochecknoglobals
 var DefaultOptions = Options{
-	MaxClockDrift:           3 * time.Minute,
-	ConcurrencyTestDuration: 30 * time.Second,
-	NumPutBlobWorkers:       3,
-	NumGetBlobWorkers:       3,
-	NumGetMetadataWorkers:   3,
-	NumListBlobsWorkers:     3,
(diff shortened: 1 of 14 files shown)


internal/providervalidation/providervalidation.go
```diff
@@ -183,6 +186,15 @@ func ValidateProvider(ctx context.Context, st blob.Storage, opt Options) error {
 		return errors.Wrap(err, "error validating concurrency")
 	}
 
+	log(ctx).Infof("Validating blob recreation...")
+
+	if !opt.SupportIdempotentCreates {
+		err := st.PutBlob(ctx, "dummy_id", gather.FromSlice([]byte{99}), blob.PutOptions{DoNotRecreate: true})
+		if err == nil {
```
jkowalski, line 193: can we introduce and expect `ErrUnsupportedPutBlobOption` here?

jkowalski:
Is this only going to work for GCS? What about S3-like providers?

Shrekster:
> Is this only going to work for GCS? What about S3-like providers?

Currently only GCS, yes. If/When S3 has support for write-once then we can extend.

internal/providervalidation/providervalidation.go
adowair: Yes, thanks for the suggestion @jkowalski. Added this new type new and old opts.

jkowalski (approved):

adowair:
Thank you @jkowalski, I will also investigate other storage providers and add support in future PRs as is possible.

PR #1656, opened 2022-01-14
feat(ui): embed version info in index.html

jkowalski:


ledeuns:
If `BuildVersion` is not set, UI will display `vv0-unofficial`. Note the double 'v'. I don't know if you should remove the 'v' in the default `BuildVersion` value or remove the 'v' from the UI.

PR #1667, opened 2022-01-16
Use gpg instead of apt-key

jorgenman:
Use gpg to install signing key. `apt-key` is deprecated and will be removed after Ubuntu 22.04.


jkowalski:

PR #1668, opened 2022-01-17
Fail when opening an uninitialized repo

jorgenman:
Fix for #9.

This is my first work in this project and my first time using Go, so hopefully this is the right way to go about this. Let me know if not and I'm happy to fix it.

It seemed to make sense to do this check whenever opening a "direct" repo, not just when creating a snapshot - I couldn't think of a situation where you'd want to have an opened repo that didn't exist.

### DIFF
--- a/repo/open.go
+++ b/repo/open.go
@@ -159,6 +159,17 @@ func openDirect(ctx context.Context, configFile string, lc *LocalConfig, passwor
 		return nil, errors.Wrap(err, "cannot open storage")
 	}
 
+	var tmp gather.WriteBuffer
+	defer tmp.Close()
+
+	if err = st.GetBlob(ctx, FormatBlobID, 0, -1, &tmp); err != nil {
+		if errors.Is(err, blob.ErrBlobNotFound) {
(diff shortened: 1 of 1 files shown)


jorgenman:
Okay, looks like I failed to run some of the tests before I pushed. I can get them now with `make all` - is that the best thing to run to make sure I don't miss anything? I read the Build Architecture page, but it didn't discuss build targets for unit tests, etc.

jkowalski:
Sorry for late response. The fix is unfortunately somewhat more complicated. Kopia will generally verify that the `kopia.repository` blob exists, except it doe sit every 15 minutes or so due to caching.

This change is effectively disabling caching of `kopia.repository` blob. The same thing can be achieved today using :

```
$ kopia repository set-client --disable-repository-format-cache
```

BTW @Shrekster is going to change how this works in his #1622 (which has a lot of other implications), so let's discuss this further to decide which PR to merge and how they interact. I think we should split #1622 into pieces, merge them and revisit this issue. 

I'm gonna keep this open for now.

PR #1672, opened 2022-01-21
Fix #982 add --disable-progress option

### LINKED ISSUE #982: Add --no-progress to kopia sync-to
I am synching my local repo to a Wasabi S3 bucket daily. I am capturing the output from the cron job and have that one sent as an e-mail to myself. The output of the sync-to command is reallly… lengthy and I am actually not interested in that output.

How about adding the --no-progress option to this command, so it only outputs any errors ans not the whole progress?

fhriley:


cli/cli_progress.go
```diff
@@ -20,13 +20,13 @@ const (
...
 type progressFlags struct {
-	enableProgress         bool
+	disableProgress        bool
 	progressUpdateInterval time.Duration
 	out                    textOutput
 }
 
 func (p *progressFlags) setup(svc appServices, app *kingpin.Application) {
-	app.Flag("progress", "Enable progress bar").Hidden().Default("true").BoolVar(&p.enableProgress)
+	app.Flag("disable-progress", "Disable progress bar").BoolVar(&p.disableProgress)
```
jkowalski, line 29: I don't think we should remove existing flag. Also `--no-progress` is automatically added since this is a boolean flag, so perhaps all we need is better documentation, like this:

```
app.Flag("progress", "Enable progress bar (--no-progress to disable)").Default("true").BoolVar(&p.enableProgress)
```

fhriley: To me it doesn't make sense to have `--progress`. Why would I ever specify `--progress`? It defaults to true.

jkowalski: Thats correct. But it also adds no-progress which is what folks use today.

jkowalski: Also, your change will add `--no-disable-progress` :)

So how about we keep `progress` as-is for backwards compatibility (and hidden) and add this new flag which is arguably much more discoverable. Progress will be disabled if `disableProgress==true || progress == false`

PR #1675, opened 2022-01-22
feat(providers): treat token expiration errors as non-retryable

onkarbhat:
When an AWS session token expires, kopia should treat token expiration errors as non-retryable errors. This will ensure that transfers fail fast and the error is reported back to the user sooner.


repo/blob/storage.go
```diff
@@ -17,6 +17,14 @@ var ErrSetTimeUnsupported = errors.Errorf("SetTime is not supported")
 // ErrInvalidRange is returned when the requested blob offset or length is invalid.
 var ErrInvalidRange = errors.Errorf("invalid blob offset or length")
 
+// TokenExpiredErrStr is the error string returned by the provider
+// when a token has expired.
+const TokenExpiredErrStr = "The provided token has expired"
+
+// ErrTokenExpired is returned when the token used for
+// authenticating with a storage provider has expired.
+var ErrTokenExpired = errors.Errorf(TokenExpiredErrStr)
```
jkowalski, line 26: nit: can we call this `ErrInvalidCredentials` which is possible for username/password too when the password changes. Please adjust method names everywhere.

repo/blob/storage.go
```diff
@@ -17,6 +17,14 @@ var ErrSetTimeUnsupported = errors.Errorf("SetTime is not supported")
 // ErrInvalidRange is returned when the requested blob offset or length is invalid.
 var ErrInvalidRange = errors.Errorf("invalid blob offset or length")
 
+// TokenExpiredErrStr is the error string returned by the provider
+// when a token has expired.
+const TokenExpiredErrStr = "The provided token has expired"
```
jkowalski, line 22: nit: does this need to be public?

repo/blob/storage.go
jkowalski: (would be also great to adjust other providers' code if we can cleanly detect credential issues there)

repo/blob/storage.go
onkarbhat: Yes. The const is defined in the blob package . It is used in two places: 
-  In the same blob package with the const ErrInvalidCredentials 
- In the s3 package, in the function added in this PR isInvalidCredentials()

repo/blob/storage.go
onkarbhat: >  can we call this ErrInvalidCredentials which is possible for username/password too when the password changes. Please adjust method names everywhere.

Done. 

> (would be also great to adjust other providers' code if we can cleanly detect credential issues there)

Can I follow up on this in future PRs ?

julio-lopez (approved):
We can merge and iterate. Thanks @onkarbhat

PR #1685, opened 2022-01-29
fix(providers): fixed b2 provider failure

jkowalski:
There was a small regression in #1654 - only in negative tests.

Also removed unnecessary retries on ErrUnsupportedPutBlobOption
+ switched provider test to run once per day, instead of 12 times/day.


julio-lopez (approved):

julio-lopez:
FYI @adowair

adowair:
Thanks for raising this @jkowalski ! 🙏

PR #1688, opened 2022-01-29
fix(providers): fixed DoNotRecreate and tests for gcs

jkowalski:
Also simplified the validation test suite, which will check whether the provider supports `DoNotRecreate` or properly rejects it without external configuration.

@julio-lopez There were a number of suble issues there all stacked up behind "empty conditions" failure. They could only have been caught by running the test against actual GCS and were not easy to spot in the code review. 

We need to figure out how to run provider tests (which require credentials) as part of pull requests to catch those kinds of things early. There's an obvious security issue for untrusted PRs, but perhaps there's a way we can trigger these tests with access to secrets as one-off after eyeballing that PR can be trusted?

@adowair FYI


jkowalski:
I guess one strategy could be to require all PRs that touch `./repo/blob/` to be merged into `test/providers` branch first, which is where we can run tests with secrets after merge although this will get awkward for refactorings that don't change behavior.

PR #1697, opened 2022-02-01
refactor(cli): An in-memory storage mock setup for CLI tests

Shrekster:
This PR extends the cli App services interface to allow customizing the tests environments to append a custom storage provider at runtime. This will allow appending an in-memory storage provider for CLI tests that otherwise depend upon the local filesystem access. The PR also adds a reconnect-able storage provider which tests can use to connect to using the CLI:

```sh
kopia repo connect in-memory --UUID=XYZ
```

The UUID of the storage can be accessed using the reconnect able-storage ConnectionInfo() api. This refactor allows us to do these things:

- allows us to use more complex storage mocks like versioned / object locking maps to test retention settings for example
- potentially speed up CLI tests
- avoid filesystem leaving cruft from previous test crashes

### DIFF
--- a/cli/storage_inmemory.go
+++ b/cli/storage_inmemory.go
@@ -1,0 +1,28 @@
+package cli
+
+import (
+	"context"
+
+	"github.com/alecthomas/kingpin"
+
+	"github.com/kopia/kopia/internal/repotesting"
+	"github.com/kopia/kopia/repo/blob"
+)
+
+// StorageInMemoryFlags is in-memory storage initialization flags for cli
+// setup.
+type StorageInMemoryFlags struct {
+	options repotesting.ReconnectableStorageOptions
+}
+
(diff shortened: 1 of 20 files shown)


cli/storage_inmemory.go
```diff
@@ -0,0 +1,28 @@
+package cli
```
jkowalski, line 1: this will bring in `repotesting` dependency into the kopia binary which is undesired. Can you move this to a test-only package location?

jkowalski: other than this, LGTM

jkowalski (approved):

cli/storage_inmemory.go
Shrekster: good point, yes

Shrekster: done in 232d764825432a1b381c78dcbc0922b594dc4a89

PR #1713, opened 2022-02-02
fix(providers): fixed rclone connection in KopiaUI

jkowalski:
v0.10.3 introduced a regression where RClone connections did not
work in KopiaUI but were ok in the CLI.

The root cause was #1691 which caused storage context to be closed
after opening the repository.

Added test that verifies that storage does not rely on the context
remaining open.

### DIFF
--- a/repo/blob/azure/azure_storage_test.go
+++ b/repo/blob/azure/azure_storage_test.go
@@ -110,12 +110,16 @@ func TestAzureStorage(t *testing.T) {
 
 	ctx := testlogging.Context(t)
 
-	st, err := azure.New(ctx, &azure.Options{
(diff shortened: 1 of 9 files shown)


repo/blob/azure/azure_storage_test.go
```diff
@@ -138,13 +142,18 @@ func TestAzureStorageSASToken(t *testing.T) {
 
 	ctx := testlogging.Context(t)
 
-	st, err := azure.New(ctx, &azure.Options{
+	// use context that gets canceled after storage is initialize,
```
Shrekster, line 145: nit; initialize -> initialized

PR #1718, opened 2022-02-03
build(ci): added rclone 1.57.0 to Docker container images

jkowalski:
/bin/rclone is configured to use config in /app/rclone/rclone.conf
so the user can pass -v ~/.config/rclone:/app/rclone


site/content/docs/Installation/_index.md
```diff
@@ -244,6 +244,7 @@ In order to run Kopia in a container, you must:
 * (recommended) mount `/app/cache` directory in which Kopia will be keeping a cache of downloaded data
 * (optional) mount `/app/logs` directory in which Kopia will be writing logs
 * mount any data directory used for locally-attached repository
+* mount `/app/rclone` directory in which RClone will look for `rclone.conf` file
```
julio-lopez, line 247: optionally, no? this should only be required when one is using an rclone backend, right?

julio-lopez (approved):

site/content/docs/Installation/_index.md
jkowalski: done

julio-lopez (approved):

PR #1727, opened 2022-02-04
feat(general): Upgrade Lock intent definitions

Shrekster:
Upgrade lock intent will eventually be added to the Kopia format blob on the repository. This will be watched by all the Kopia clients for communicating the upgrade proposal. The Lock intent definition described In this PR sets the various upgrade parameters (drain timeouts, polling interval etc) which will allow the upgrade owner to start executing the readers/writers draining protocol.

Design REF: https://docs.google.com/document/d/1zIghQc4i4vJSANjSDBOvjzMiqnZSEp1cCasxSUkltSk/edit#heading=h.khston2oxmxy

### DIFF
--- a/repo/content/content_upgrade_lock.go
+++ b/repo/content/content_upgrade_lock.go
@@ -21,24 +21,6 @@ type UpgradeLock struct {
 	MaxPermittedClockDrift time.Duration `json:"maxPermittedClockDrift,omitempty"`
 }
 
-// NewUpgradeLock constructs a new lock object for repository format upgrades.
-func NewUpgradeLock(
-	now time.Time,
-	ownerID string,
-	advanceNotice, ioDrainTimeout, statusPollInterval, maxPermittedClockDrift time.Duration,
-	oldFormatVersion FormatVersion,
-) *UpgradeLock {
-	return &UpgradeLock{
-		OwnerID:                ownerID,
-		CreationTime:           now,
-		AdvanceNoticeDuration:  advanceNotice,
-		IODrainTimeout:         ioDrainTimeout,
-		StatusPollInterval:     statusPollInterval,
-		MaxPermittedClockDrift: maxPermittedClockDrift,
-		OldFormatVersion:       oldFormatVersion,
-	}
-}
-
 // Update upgrades an existing lock intent. This method controls what mutations
 // are allowed on an upgrade lock once it has been placed on the repository.
 func (l *UpgradeLock) Update(other *UpgradeLock) (*UpgradeLock, error) {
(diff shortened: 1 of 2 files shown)


repo/content/content_upgrade_lock.go
```diff
@@ -0,0 +1,160 @@
...
+	CreationTime           time.Time     `json:"creationTime,omitempty"`
+	AdvanceNoticeDuration  time.Duration `json:"advanceNoticeDuration,omitempty"`
+	IODrainTimeout         time.Duration `json:"ioDrainTimeout,omitempty"`
+	StatusPollInterval     time.Duration `json:"statusPollInterval,omitempty"`
+	OldFormatVersion       FormatVersion `json:"oldFormatVersion,omitempty"`
+	MaxPermittedClockDrift time.Duration `json:"maxPermittedClockDrift,omitempty"`
+}
+
+// NewUpgradeLock constructs a new lock object for repository format upgrades.
+func NewUpgradeLock(
```
jkowalski, line 25: does this function add much value? by calling it we're effectivelly hiding the meaning of various parameters. I would prefer explicit initialization + validate

repo/content/content_upgrade_lock.go
```diff
@@ -0,0 +1,160 @@
...
+// UpgradeLock represents the intent to lock a kopia repository for upgrade
+// related maintenance activity. This signals a request for exclusive access to
+// the repository. The lock object is set on the Kopia repository format blob
+// 'kopia.repository' and must be respected by all clients accessing the
+// repository.
+type UpgradeLock struct {
+	OwnerID                string        `json:"ownerID,omitempty"`
+	CreationTime           time.Time     `json:"creationTime,omitempty"`
+	AdvanceNoticeDuration  time.Duration `json:"advanceNoticeDuration,omitempty"`
+	IODrainTimeout         time.Duration `json:"ioDrainTimeout,omitempty"`
```
jkowalski, line 18: instead of time.Duration can this be a point in time (`IODrainTime time.Time`) ? Would it makes this simpler to reason about?

repo/content/content_upgrade_lock.go
Shrekster: Constructor style pattern is what I prefer, but I'd avoid having a mix of coding patterns in a single project. Consistency is more important, so would switch to init + validate pattern if that is what we follow all across Kopia.

repo/content/content_upgrade_lock.go
Shrekster: I'd prefer that we always compute the `UpgradeTime()` from config parameters in the lock instead of pre-determining it. This makes it easier to alter parameters later (for example, postponing advance notice) and reason about why the `UpgradeTime()` is what it is at any given point which is good for debugging.

julio-lopez:
@Shrekster See comments above

Issue #90, opened 2019-08-18
[FR]Google Drive support


yurikoles:
Hi @jkowalski,

Could you please implement native Google Drive support, because `gcsf` seems unreliable?

yurikoles:
Because Google Drive prices seems more attractive than Google Cloud Storage.

eras:
Does the rclone support introduced by #545 satisfy this request?

zyradyl:
@yurikoles i'm going to close this under the assumption #545 resolves it. feel free to open a new issue if you're running into problems with the rclone backend! 😄

julio-lopez:
Contributions are welcome.

Quoting Jarek's comment from the #kopia Slack channel:

> it absolutely could be done, we can probably talk about bringing in {Google,One}Drive as first-class providers. Building a provider requires implementing about 5 non-trivial methods (GetBlob, GetMetadata, PutBlob, DeleteBlob, ListBlobs) and some trivial ones.
A full-fledged provider is only about 300-500 lines of code and there’s reasonably extensive test suite to validate basic properties of each provider that are required by Kopia + maybe 100 lines of code for the UI and special CLI flags, so definitely a good starter project that requires little knowledge of how the rest of kopia works.

xkxx:
I can try to work on this. I want to know if anyone else is working on this feature concurrently?

jkowalski:
nobody is working on it, AFAIK, go ahead!

xkxx:
# Implementation Plan

Code directory: `kopia/repo/blob/gdrive`
Package: `gdrive`

The implementation will be loosely based on the GCS backend, but with GDrive APIs. We will integrate with the official go client for GDrive `google.golang.org/api/drive/v3`. 

## Background

Google Drive itself needs no introduction. Some challenges arise from the Drive architecture and the permission model. These are:

**Permissions**: Google no longer gives out blanket Drive access to all developers. To do this, they would have to go through a security review. Since we expect each user to create their own developer account to use with Kopia, this is not realistic. Instead, [Google recommends](https://developers.google.com/drive/api/v3/about-auth) `drive.file` or `drive.appdata` permission scopes, which give access to specific files or a hidden per-application directory, respectively. I'll experiment with `drive.file` scope first and see if it meets Kopia's needs.

[...]

PR #1731, opened 2022-02-06
feat(providers): Add Google Drive support.

xkxx:
Hi @jkowalski,

I'm requesting early review for this feature before I finish up with tests and documentation. This PR implements https://github.com/kopia/kopia/issues/90 and passes `validate-provider`.

The implementation strategy is the one we discussed in https://github.com/kopia/kopia/issues/90#issuecomment-1025507405:
* FileIdCache: A BlobId -> FileId map is put in place to speed up file look ups.
* GetBlob / GetBlobMetadata: Use exact name match with aggressive retrying (100ms x 3 times). 
* ListBlobs: Use name prefix search with a cap of 26 characters (Drive limitation). Additionally, we search through FileIdCache for any local matches in case the Drive-side index is not stable yet.

I would like your input on:
* The general implementation strategy - is it sensible? Don't worry about go readability right now as I can clean it up after the big picture is set.
* Right now, FileIdCache is an ephemeral and uncapped cache. The fact is that this cache is effectively append-only since the id mapping is stable. Would it be better to turn it into a LRU cache or persist it on disk?

List of changes in PR:
* `repo/blob/gdrive`: Implement blob storage in GDrive.
* `cli/`: Implement `gdrive` storage subcommands.
* `internal/retry/retry.go`: Fix a bug affecting gdrive where `Errorf` is used instead of `Wrapf`, obscuring the error chain.

Thanks,

Kexiang

### DIFF
--- a/repo/blob/gdrive/file_id_cache.go
+++ b/repo/blob/gdrive/file_id_cache.go
@@ -10,13 +10,13 @@ const (
 	changeLogCacheSize = 1 << 8
 )
 
-// FileIDCache is a cache for managing the association of blobID -> fileID.
-type FileIDCache struct {
-	// Map of blobID -> *CacheEntry.
+// fileIDCache is a cache for managing the association of blobID -> fileID.
+type fileIDCache struct {
+	// Map of blobID -> *cacheEntry.
 	Blobs sync.Map
 	// Record of recent cache changes.
 	// It's stored as a synchronized circular buffer.
-	ChangeLog [changeLogCacheSize]ChangeEntry
+	ChangeLog [changeLogCacheSize]changeEntry
 	// ChangeLogIdx indicates the next location to write to.
 	// The log entry range is [ChangeLogIdx+1, ChangeLogIdx-1].
 	ChangeLogIdx int
@@ -24,8 +24,8 @@ type FileIDCache struct {
 	ChangeLogMut sync.RWMutex
 }
 
-// CacheEntry is a blob cache entry.
-type CacheEntry struct {
+// cacheEntry is a blob cache entry.
+type cacheEntry struct {
 	// Guards access to the entry. Must be taken before accessing fields.
 	Mut sync.Mutex
 	// blobID. Is always populated.
@@ -34,8 +34,8 @@ type CacheEntry struct {
 	FileID string
 }
 
-// ChangeEntry is a blob change entry.
-type ChangeEntry struct {
+// changeEntry is a blob change entry.
+type changeEntry struct {
 	// blobID. If empty, the entry has not been written to yet.
 	BlobID blob.ID
 	// Associated fileID. If empty, the blob is deleted.
@@ -45,7 +45,7 @@ type ChangeEntry struct {
 // Lookup finds the cache entry for blobID.
 // The entry may be read or mutated.
 // The callback is guaranteed to have exclusive access to the entry.
-func (cache *FileIDCache) Lookup(blobID blob.ID, callback func(entry *CacheEntry) (interface{}, error)) (interface{}, error) {
+func (cache *fileIDCache) Lookup(blobID blob.ID, callback func(entry *cacheEntry) (interface{}, error)) (interface{}, error) {
 	entry := cache.getEntry(blobID)
 
 	entry.Mut.Lock()
@@ -58,19 +58,19 @@ func (cache *FileIDCache) Lookup(blobID blob.ID, callback func(entry *CacheEntry
 }
 
 // Internal method for retrieving an entry.
-func (cache *FileIDCache) getEntry(blobID blob.ID) *CacheEntry {
-	loaded, _ := cache.Blobs.LoadOrStore(blobID, &CacheEntry{
+func (cache *fileIDCache) getEntry(blobID blob.ID) *cacheEntry {
+	loaded, _ := cache.Blobs.LoadOrStore(blobID, &cacheEntry{
 		Mut:    sync.Mutex{},
 		BlobID: blobID,
 		FileID: "",
 	})
 
-	return loaded.(*CacheEntry) //nolint:forcetypeassert
+	return loaded.(*cacheEntry) //nolint:forcetypeassert
 }
 
 // BlindPut blindly mutates the association for a blobID.
-func (cache *FileIDCache) BlindPut(blobID blob.ID, fileID string) {
-	_, _ = cache.Lookup(blobID, func(entry *CacheEntry) (interface{}, error) {
+func (cache *fileIDCache) BlindPut(blobID blob.ID, fileID string) {
+	_, _ = cache.Lookup(blobID, func(entry *cacheEntry) (interface{}, error) {
 		entry.FileID = fileID
 		return nil, nil
 	})
@@ -78,11 +78,11 @@ func (cache *FileIDCache) BlindPut(blobID blob.ID, fileID string) {
 
 // RecordBlobChange records a newly created or deleted blob.
 // An empty fileID signals that the blob is deleted.
-func (cache *FileIDCache) RecordBlobChange(blobID blob.ID, fileID string) {
+func (cache *fileIDCache) RecordBlobChange(blobID blob.ID, fileID string) {
 	cache.ChangeLogMut.Lock()
 
 	i := cache.ChangeLogIdx
-	cache.ChangeLog[i] = ChangeEntry{
+	cache.ChangeLog[i] = changeEntry{
 		BlobID: blobID,
 		FileID: fileID,
 	}
@@ -92,7 +92,7 @@ func (cache *FileIDCache) RecordBlobChange(blobID blob.ID, fileID string) {
 }
 
 // VisitBlobChanges iterates through newly created or deleted blobs.
-func (cache *FileIDCache) VisitBlobChanges(callback func(blobID blob.ID, fileID string)) {
+func (cache *fileIDCache) VisitBlobChanges(callback func(blobID blob.ID, fileID string)) {
 	cache.ChangeLogMut.RLock()
 
 	for i := circularBufferNext(cache.ChangeLogIdx); i != cache.ChangeLogIdx; i = circularBufferNext(i) {
@@ -106,9 +106,9 @@ func (cache *FileIDCache) VisitBlobChanges(callback func(blobID blob.ID, fileID 
 }
 
 // Clear resets the file ID cache.
-func (cache *FileIDCache) Clear() {
+func (cache *fileIDCache) Clear() {
 	cache.Blobs = sync.Map{}
-	cache.ChangeLog = [changeLogCacheSize]ChangeEntry{}
+	cache.ChangeLog = [changeLogCacheSize]changeEntry{}
 }
 
 func circularBufferNext(curr int) int {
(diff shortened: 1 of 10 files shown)


repo/blob/gdrive/gdrive_storage.go
```diff
@@ -0,0 +1,460 @@
...
+	getMetadataFields  = "name,id,mimeType,size,modifiedTime"
+)
+
+type gdriveStorage struct {
+	Options
+
+	client      *drive.Service
+	folderId    string
+	// Caches blob id -> drive file id associations.
+	fileIdCache map[blob.ID]string
```
jkowalski, line 49: can you define a separate struct type for the fileIDCache so we can test unit-test all the cache operations?

repo/blob/gdrive/gdrive_storage.go
```diff
@@ -0,0 +1,460 @@
...
+  // googleapi.Field values.
+	listIdFields       = "files(name,id)"
+	listMetadataFields = "files(name,id,mimeType,size,modifiedTime)"
+	getMetadataFields  = "name,id,mimeType,size,modifiedTime"
+)
+
+type gdriveStorage struct {
+	Options
+
+	client      *drive.Service
```
jkowalski, line 46: I don't know if drive API has something that prevents mocking, but I'm hoping `client` be an interface includes just the methods of `*drive.FilesService` we actually use?

This would allow us to write fake implementation of this interface and unit-test most of this provider without calling actual Google drive, because unlike others it has some pretty non-trivial logic?

I realize this is can be much harder than the current approach (so it perhaps can be done in separate PR, once this early version is merged), but I think it's important to prevent regressions here.

repo/blob/gdrive/gdrive_storage.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+func (gdrive *gdriveStorage) GetBlob(ctx context.Context, b blob.ID, offset, length int64, output blob.OutputBuffer) error {
+	if offset < 0 {
+		return blob.ErrInvalidRange
+	}
+
+	fileId, err := gdrive.GetFileId(ctx, b)
+	if err != nil {
+		return err
```
jkowalski, line 59: in general prefer `errors.Wrap()` over `return err` - this helps with tracking down sources of errors.

repo/blob/gdrive/gdrive_storage.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+func (gdrive *gdriveStorage) ListBlobs(ctx context.Context, prefix blob.ID, callback func(blob.Metadata) error) error {
+	// Tracks blob matches in cache but not returned by API.
+	unvisitedIds := make(map[blob.ID]bool)
+
+	consumer := func(files *drive.FileList) error {
+		for _, file := range files.Files {
+			blobId := toBlobId(file.Name)
+			gdrive.fileIdCache[blobId] = file.Id
```
jkowalski, line 181: cache must be thread/goroutine safe, so you'd need a mutex inside cache implementation - simple map access will generally panic on concurrent access attempts.

repo/blob/gdrive/gdrive_storage.go
```diff
@@ -0,0 +1,460 @@
...
+		return err
+	}
+
+	err = gdrive.client.Files.Delete(fileId).Context(ctx).Do()
+	return translateError(err)
+}
+
+func (gdrive *gdriveStorage) ListBlobs(ctx context.Context, prefix blob.ID, callback func(blob.Metadata) error) error {
+	// Tracks blob matches in cache but not returned by API.
+	unvisitedIds := make(map[blob.ID]bool)
```
jkowalski, line 176: n/b: this is probably fine for most use cases, but maps like this can get really big for large repositories potentially using lots of RAM.

repo/blob/gdrive/gdrive_storage.go
```diff
@@ -0,0 +1,460 @@
...
+				return err
+			}
+		}
+
+		return nil
+	}
+
+	query := fmt.Sprintf("'%s' in parents and mimeType = '%s'", gdrive.folderId, gdriveMimeType)
+	if prefix != "" {
+		query = fmt.Sprintf("'%s' in parents and name contains '%s' and mimeType = '%s'", gdrive.folderId, capPrefix(prefix), gdriveMimeType)
```
jkowalski, line 205: should it be `contains`? is there 'startsWith' or something like that?

jkowalski:
Overall strategy LGTM, i did not comment on linter errors since those are already included.

I would also recommend cloing `gcs_storage_test.go` and running tests locally to make sure the required semantics is met by this provider. When you do that, expect all credentials to be passed via environment variables (see `provider-tests.yaml`) - this way we will be able to automate test runs against this provider on CI/CD and you can easily pass those when developing locally.

repo/blob/gdrive/gdrive_storage.go
xkxx: Looks like FilesService doesn't mind being stored on its own. Done

repo/blob/gdrive/gdrive_storage.go
xkxx: That's a good idea. I also used the chance to add synchronization so that all the concurrent tests pass now.

repo/blob/gdrive/gdrive_storage.go
xkxx: Thanks for the tip. I went through and added Wrapf to all the places where we interface with drive api.

repo/blob/gdrive/gdrive_storage.go
xkxx: Done.

repo/blob/gdrive/gdrive_storage.go
xkxx: I decided to split this up to two caches:
- Blobs: This cache tracks blob -> file mapping and is unbounded.
- ChangeLog: This cache tracks recent blob creations / deletions and is limited to 2^8 entries. Only this cache is iterated through for ListBlobs.

repo/blob/gdrive/gdrive_storage.go
xkxx: So according to the drive api `contains` acts like `startswith` for file names. I added a comment here to clarify this.

xkxx:
Thanks for the review!

I added a new commit with these changes:
- Added a synchronized FileIdCache. As you pointed out, this is necessary for concurrent repo access. This cache has two parts:
  - Blobs: This tracks individual blob -> file mappings. Individual rows have their own mutex locks. Since all lookups can involve writes, this is a mutex not rwmutex.
  - ChangeLog: This tracks recent changes for ListBlobs. One rwmutex governs this.
- Added tests and make sure all of these pass locally.
- Fix lint and style issues.

repo/blob/gdrive/gdrive_storage.go
```diff
@@ -0,0 +1,556 @@
...
+
+	fileID, err := gdrive.GetFileID(ctx, b)
+	if err != nil {
+		return errors.Wrapf(err, "get file id in GetBlob(%s)", b)
+	}
+
+	req := gdrive.client.Get(fileID)
+	req.Header().Set("Range", toRange(offset, length))
+	res, err := req.Context(ctx).Download()
+
```
xkxx, line 67: There's an issue with the linter setup:
- gofumpt wants to remove it
- wsl wants to add it back

jkowalski: i think you need to add empty line above 66

xkxx: That fixed it. I apparently misread the error output. Thanks!

repo/blob/gdrive/file_id_cache.go
```diff
@@ -0,0 +1,130 @@
...
+	ChangeLog [changeLogCacheSize]ChangeEntry
+	// ChangeLogIdx indicates the next location to write to.
+	// The log entry range is [ChangeLogIdx+1, ChangeLogIdx-1].
+	ChangeLogIdx int
+	// Guards access to ChangeLog.
+	ChangeLogMut sync.RWMutex
+}
+
+// CacheEntry is a blob cache entry.
+type CacheEntry struct {
```
jkowalski, line 28: nit, can you make this non-exported (lowercase)?

repo/blob/gdrive/file_id_cache.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/kopia/kopia/repo/blob"
+)
+
+const (
+	changeLogCacheSize = 1 << 8
+)
+
+// FileIDCache is a cache for managing the association of blobID -> fileID.
+type FileIDCache struct {
```
jkowalski, line 14: nit, can you make this non-exported (lowercase)?

repo/blob/gdrive/file_id_cache.go
```diff
@@ -0,0 +1,130 @@
...
+	// Guards access to the entry. Must be taken before accessing fields.
+	Mut sync.Mutex
+	// blobID. Is always populated.
+	BlobID blob.ID
+	// Associated fileID. If empty, the blob doesn't have a cached fileID.
+	FileID string
+}
+
+// ChangeEntry is a blob change entry.
+type ChangeEntry struct {
```
jkowalski, line 38: nit, can you make this non-exported (lowercase)?

jkowalski:
I tested the PR on my local machine and it appears to be working against test gdrive and credentials I generated, which is great. I added couple CI-related changes to your branch.

I'm going to change this PR to merge into `test/providers` and we will merge to `master` when the tests pass.

One more thing we should do before merging - ensure the test can run in parallel, as of now we use shared folder, can the test be modified to create a subfolder with a random name and run with it, so that two tests don't collide with each other?

repo/blob/gdrive/file_id_cache.go
xkxx: Done. Thanks for teaching me on go :)

repo/blob/gdrive/file_id_cache.go
xkxx: Done

repo/blob/gdrive/file_id_cache.go
xkxx: Done

xkxx:
Sounds great! I added the changes on test isolation. Once this is merged, I can work on updating the documentation.

jkowalski:
I ran this locally and appears to all work, I'm going to merge into `test/providers` so we can get some official runs against it.

I have some concerns about retry loop in `tryGetFileByBlobID` - is this needed at all? I removed it and all the tests are passing - this retry loop appears to be slowing down tests, so if it's not needed I'd love to remove it (separate PR is fine, we can also add docs).

jkowalski:
Looks like the test passed on GitHub: https://github.com/kopia/kopia/runs/5227924251?check_suite_focus=true

I'm going to merge this into `master` now and we can iterate over there.

xkxx:
Woohoo! Thanks for guiding me through this PR!

RE tryGetFileByBlobID: It depends on how important it is to have concurrent access by multiple clients. Due to Drive indexing being eventually consistent, looking up a blob right after it's created may not work. For single client usage, this issue is resolved with `fileIdCache`. For multi clients, the retry loop mitigates the issue by essentially adding more latency to the lookup.

In the majority of cases, I don't think concurrent mutations of the same blobs (save for the repository file) is a concern. If that's the case, we can remove the retry loop and get some performance back.

PR #1741, opened 2022-02-10
docs(general): update compilation from source for Go 1.17

### LINKED ISSUE #1740: Install using go fails
Using
`go get github.com/kopia/kopia`
as per build instructions fails with
```
go get: installing executables with 'go get' in module mode is deprecated.
	Use 'go install pkg@version' instead.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```
Trying with new go syntax 
`go install github.com/kopia/kopia@latest`
also fails doing nothing with no error message just exits back to console prompt.
Trying
`go install github.com/kopia/kopia@Latest`
also fails with:
`go install: github.com/kopia/kopia@Latest: git init --bare in /root/go/pkg/mod/cache/vcs/7f85dd982a45216de1c7884420740bb11f6185a4fb413a2f8b9cf7082448e0e9: exec: "git": executable file not found in $PATH`

julio-lopez:
Fixes #1740

### DIFF
--- a/site/content/docs/Installation/_index.md
+++ b/site/content/docs/Installation/_index.md
@@ -269,7 +269,7 @@ to a repository. The names will be persisted in a configuration file and used af
 If you have [Go 1.16](https://golang.org/) or newer, you may download and build Kopia yourself. No special setup is necessary, other than the Go compiler and [git](https://git-scm.com/). You can simply run:
 
 ```shell
-$ go install github.com/kopia/kopia
+$ go install github.com/kopia/kopia@latest
 ```
 
 The resulting binary will be available in `$HOME/go/bin`. Note that this will produce basic binary that has all the features except support for HTML-based UI. To build full binary, download the source from GitHub and run:
--- a/site/content/docs/Installation/_index.md
+++ b/site/content/docs/Installation/_index.md
@@ -266,10 +266,10 @@ to a repository. The names will be persisted in a configuration file and used af
 
 ### Compilation From Source
 
-If you have [Go 1.16](https://golang.org/) or newer, you may download and build Kopia yourself. No special setup is necessary, other than the Go compiler. You can simply run:
+If you have [Go 1.16](https://golang.org/) or newer, you may download and build Kopia yourself. No special setup is necessary, other than the Go compiler and [git](https://git-scm.com/). You can simply run:
 
 ```shell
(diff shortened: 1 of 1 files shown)


site/content/docs/Installation/_index.md
```diff
@@ -266,10 +266,10 @@ to a repository. The names will be persisted in a configuration file and used af
 
 ### Compilation From Source
 
-If you have [Go 1.16](https://golang.org/) or newer, you may download and build Kopia yourself. No special setup is necessary, other than the Go compiler. You can simply run:
+If you have [Go 1.16](https://golang.org/) or newer, you may download and build Kopia yourself. No special setup is necessary, other than the Go compiler and [git](https://git-scm.com/). You can simply run:
 
 ```shell
-$ go get github.com/kopia/kopia
+$ go install github.com/kopia/kopia
```
miquella, line 272: Pretty sure this should be:
```console
$ go install github.com/kopia/kopia@latest
```

Shrekster: Hmm..

```
If the arguments don't have version suffixes, "go install" may run in module-aware mode or GOPATH mode, depending on the GO111MODULE environment variable and the presence of a go.mod file. See 'go help modules' for details. If module-aware mode is enabled, "go install" runs in the context of the main module.
```

miquella: @Shrekster: Correct!

It used to be that `go get` would download and install a package. It was then upgraded to also update `go.mod`.

Go 1.18 is going to continue that set of changes such that `go get` will no longer install anything and will only be used to update `go.mod` (same as using the `-d` flag for `go get`).

Conversely, `go install` has been upgraded to fill the gap left by `go get` no longer installing downloaded packages. The version suffix was added as a way to ignore GOPATH and any local `go.mod` and instead download the specified module and install a package from it into `$GOBIN`.

The `go install` command as written (no version suffix) will install the executable into `$GOBIN`, but it will source the version from a local `go.mod` file. If no `go.mod` file exists or if Kopia isn't present in that `go.mod` file, the installation will fail.

julio-lopez:
```suggestion
$ go install github.com/kopia/kopia@latest
```

PR #1742, opened 2022-02-11
feat(repository): repository writer upgrade lock apis

Shrekster:
Repository Writer APIs to set/update/revoke a lock intent as was described in this PR: https://github.com/kopia/kopia/pull/1727. This PR lets the writers set the upgrade lock intent on the format blob itself.

Design REF: https://docs.google.com/document/d/1zIghQc4i4vJSANjSDBOvjzMiqnZSEp1cCasxSUkltSk/edit#heading=h.khston2oxmxy

### DIFF
--- a/repo/open.go
+++ b/repo/open.go
@@ -34,9 +34,9 @@ const CacheDirMarkerFile = "CACHEDIR.TAG"
 // CacheDirMarkerHeader is the header signature for cache dir marker files.
 const CacheDirMarkerHeader = "Signature: 8a477f597d28d172789f06886806bc55"
 
-// defaultRepositoryBlobCacheDuration is the duration for which we treat cached kopia.repository
+// DefaultRepositoryBlobCacheDuration is the duration for which we treat cached kopia.repository
 // as valid.
-const defaultRepositoryBlobCacheDuration = 15 * time.Minute
+const DefaultRepositoryBlobCacheDuration = 15 * time.Minute
 
 // throttlingWindow is the duration window during which the throttling token bucket fully replenishes.
 // the maximum number of tokens in the bucket is multiplied by the number of seconds.
@@ -73,6 +73,10 @@ type Options struct {
 
 // ErrInvalidPassword is returned when repository password is invalid.
 var ErrInvalidPassword = errors.Errorf("invalid repository password")
+
+// ErrRepositoryUnavailableDueToUpgrageInProgress is returned when repository
+// is undergoing upgrade that requires exclusive access.
+var ErrRepositoryUnavailableDueToUpgrageInProgress = errors.Errorf("repository upgrade in progress")
 
 // Open opens a Repository specified in the configuration file.
 func Open(ctx context.Context, configFile, password string, options *Options) (rep Repository, err error) {
@@ -405,7 +409,7 @@ func readAndCacheRepositoryBlobBytes(ctx context.Context, st blob.Storage, cache
 	cachedFile := filepath.Join(cacheDirectory, blobID)
 
 	if validDuration == 0 {
-		validDuration = defaultRepositoryBlobCacheDuration
+		validDuration = DefaultRepositoryBlobCacheDuration
 	}
 
 	if cacheDirectory != "" {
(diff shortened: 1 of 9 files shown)


repo/open.go
```diff
@@ -34,9 +34,9 @@ const CacheDirMarkerFile = "CACHEDIR.TAG"
 // CacheDirMarkerHeader is the header signature for cache dir marker files.
 const CacheDirMarkerHeader = "Signature: 8a477f597d28d172789f06886806bc55"
 
-// defaultRepositoryBlobCacheDuration is the duration for which we treat cached kopia.repository
+// DefaultRepositoryBlobCacheDuration is the duration for which we treat cached kopia.repository
 // as valid.
-const defaultRepositoryBlobCacheDuration = 15 * time.Minute
+const DefaultRepositoryBlobCacheDuration = 15 * time.Minute
```
redgoat650, line 39: Wondering if this change to export this const is actually needed. Seems to only be consumed by repo tests, but may be more correct to pull it out of the test environment repository directly?

repo/upgrade_lock_test.go
```diff
@@ -0,0 +1,92 @@
...
+	ctx, env := repotesting.NewEnvironment(t, content.FormatVersion1)
+
+	l := &content.UpgradeLock{
+		OwnerID:                "upgrade-owner",
+		CreationTime:           env.Repository.Time(),
+		AdvanceNoticeDuration:  0,
+		IODrainTimeout:         repo.DefaultRepositoryBlobCacheDuration * 2,
+		StatusPollInterval:     repo.DefaultRepositoryBlobCacheDuration,
+		Message:                "upgrading from format version 2 -> 3",
+		MaxPermittedClockDrift: repo.DefaultRepositoryBlobCacheDuration / 3,
```
redgoat650, line 24: Would it be more correct to use
```go
fmtBlobCacheDuration := env.Repository.ClientOptions().FormatBlobCacheDuration
```
instead of exporting `DefaultRepositoryBlobCacheDuration`?

repo/upgrade_lock.go
```diff
@@ -0,0 +1,123 @@
...
+	repoConfig, err := r.updateRepoConfig(ctx, func(repoConfig *repositoryObjectFormat) error {
+		if err := l.Validate(); err != nil {
+			return errors.Wrap(err, "invalid upgrade lock intent")
+		}
+
+		if repoConfig.FormattingOptions.UpgradeLock == nil {
+			// when we are putting a new lock then ensure that we can upgrade
+			// to that version
+			if repoConfig.FormattingOptions.Version >= content.MaxFormatVersion {
+				return errors.Errorf("repository is already upgraded to %d, we can only upgrade till %d",
```
redgoat650, line 60: Something like this maybe?
```suggestion
				return errors.Errorf("repository is using version %d, and version %d is the maximum",
```

redgoat650: you'd need to update the unit test if you choose to reword this

repo/upgrade_lock_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	require.EqualError(t, env.RepositoryWriter.CommitUpgrade(ctx), "no upgrade in progress")
+
+	_, err := env.RepositoryWriter.SetUpgradeLockIntent(ctx, *l)
+	require.NoError(t, err)
+
+	require.NoError(t, env.RepositoryWriter.CommitUpgrade(ctx))
+
+	// verify that rollback after commit fails
+	require.EqualError(t, env.RepositoryWriter.RollbackUpgrade(ctx), "no upgrade in progress")
```
redgoat650, line 91: Any plans to test the rollback path?

Shrekster: Actually I should. Will fix.

repo/open.go
Shrekster: You are right. Actually, this PR is a spin off from the whole thing here:

https://github.com/kopia/kopia/pull/1622

And there we use the const outside of the module. But you are right, lets expose the const when we get there.

repo/upgrade_lock.go
Shrekster: Will do.

repo/upgrade_lock_test.go
Shrekster: Good point. I am not sure whether client `FormatBlobCacheDuration` is set in this test env. Will check.

PR #1752, opened 2022-02-16
feat(snapshots): big upload performance improvements by parallelizing directory traversal

jkowalski:
Previously directories were walked strictly sequentially which means
we could never be uploading data from multiple directories in parallel,
even if they had just a few files each.

This change switches to using the new `workshare` utility which improves
parallelism. It also reduces memory allocations, goroutine creations
and overall memory usage when taking large snapshots, while increasing
CPU utilization.

Tests on realistic directory structures show huge speed-ups during cold
snapshots (without any metadata caching:)

Photo library - 160GB, files:41717 dirs:1350

    Before: 3m11s
    After: 1m50s
    Total time reduction: 43%

Working code directory - 30.7 GB files:194560 dirs:42455

    Before: 55s
    After: 25s
    Total time reduction: 55%

### DIFF
--- a/internal/workshare/workshare_pool.go
+++ b/internal/workshare/workshare_pool.go
@@ -6,14 +6,14 @@ import (
 	"sync/atomic"
 )
 
-// ProcessFunc processes the provided input, the result must be written within the input itself.
-// To avoid allocations, the function should not be a local closure but a named function (either global or
-// a method).
-type ProcessFunc func(c *Pool, input interface{})
+// ProcessFunc processes the provided request, which is typically a pointer to a structure.
+// The result will typically be written within the request structure. To avoid allocations,
+// the function should not be a local closure but a named function (either global or a method).
+type ProcessFunc func(c *Pool, request interface{})
 
 type workItem struct {
 	process ProcessFunc     // function to call
-	input   interface{}     // parameter to the function
+	request interface{}     // parameter to the function
 	wg      *sync.WaitGroup // signal the provided WaitGroup after work is done
 }
 
@@ -41,7 +41,9 @@ func NewPool(numWorkers int) *Pool {
 	}
 
 	w := &Pool{
-		work:      make(chan workItem), // channel must be unbuffered
+		// channel must be unbuffered so that it has exactly as many slots as there are goroutines capable of reading from it
+		// this way by pushing to the channel we can be sure that a pre-spun goroutine will pick it up soon.
+		work:      make(chan workItem),
 		closed:    make(chan struct{}),
 		semaphore: make(chan struct{}, numWorkers),
 	}
@@ -56,7 +58,7 @@ func NewPool(numWorkers int) *Pool {
 				select {
 				case it := <-w.work:
 					atomic.AddInt32(&w.activeWorkers, 1)
-					it.process(w, it.input)
+					it.process(w, it.request)
 					atomic.AddInt32(&w.activeWorkers, -1)
 					<-w.semaphore
 					it.wg.Done()
@@ -74,4 +76,5 @@ func NewPool(numWorkers int) *Pool {
 // Close closes the worker pool.
 func (w *Pool) Close() {
 	close(w.closed)
+	w.wg.Wait()
 }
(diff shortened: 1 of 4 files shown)


internal/workshare/workshare_pool.go
```diff
@@ -0,0 +1,77 @@
...
+		work:      make(chan workItem), // channel must be unbuffered
+		closed:    make(chan struct{}),
+		semaphore: make(chan struct{}, numWorkers),
+	}
+
+	for i := 0; i < numWorkers; i++ {
+		w.wg.Add(1)
+
+		go func() {
+			defer w.wg.Done()
```
julio-lopez, line 53: Is there a corresponding `w.wg.Wait()` ? What's the purpose of using this wait group here?

internal/workshare/workshare_waitgroup.go
```diff
@@ -0,0 +1,54 @@
...
+	w.work <- workItem{
+		process: process,
+		input:   input,
+		wg:      g.wg,
+	}
+}
+
+// CanShareWork determines if the provided worker pool has capacity to share work.
+// If the function returns true, the use MUST call RunAsync() exactly once. This pattern avoids
+// allocations required to create asynchronous input if the worker pool is full.
```
julio-lopez, line 45: Any chance the expected usage pattern could be encapsulated in a helper function to simplify usage / the API?

It seems that otherwise it'd be error prone as it is fairly easy to misuse the API by calling `CanShareWork()` once and then `RunAsync()` multiple times. Or calling `CanShareWork` multiple times, but then calling `RunAsync()` fewer times, which would reduce the effective capacity of the pool.

Also, `CanShareWork` does not seem to convey the semantics of the function.

internal/workshare/workshare_pool.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// NewPool creates a worker pool that launches a given number of goroutines that can invoke shared work.
+func NewPool(numWorkers int) *Pool {
+	if numWorkers < 0 {
+		numWorkers = 0
+	}
+
+	w := &Pool{
+		work:      make(chan workItem), // channel must be unbuffered
```
julio-lopez, line 44: Why must the channel be unbuffered?

internal/workshare/workshare_waitgroup.go
jkowalski: the big problem with trying to encapsulate this is that you want to avoid allocation typically required when calling RunAsync(). Since CanShareWork() will return false most of the time, this extra if() helps avoid many allocations.

internal/workshare/workshare_pool.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+// ProcessFunc processes the provided input, the result must be written within the input itself.
+// To avoid allocations, the function should not be a local closure but a named function (either global or
+// a method).
+type ProcessFunc func(c *Pool, input interface{})
```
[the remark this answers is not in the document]
jkowalski: renamed as `request`.

internal/workshare/workshare_pool.go
jkowalski: added comment.

internal/workshare/workshare_waitgroup.go
```diff
@@ -0,0 +1,54 @@
...
+	g.wg.Wait()
+
+	return g.requests
+}
+
+// RunAsync starts the asynchronous work to process the provided input, the user must call Wait().
+func (g *AsyncGroup) RunAsync(w *Pool, process ProcessFunc, input interface{}) {
+	if g.wg == nil {
+		g.wg = &sync.WaitGroup{}
+	}
```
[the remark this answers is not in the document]
jkowalski: intentionally not adding mutex here, added comment instead.

internal/workshare/workshare_waitgroup.go
```diff
@@ -0,0 +1,54 @@
...
+
+	g.wg.Add(1)
+
+	g.requests = append(g.requests, input)
```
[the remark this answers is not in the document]
jkowalski: same

PR #1753, opened 2022-02-17
Implement API to get Storage free space

### LINKED ISSUE #1520: Expose the notion of "free space" in blob.Storage
For storage systems where free space is limited (basically filesystem and other non-cloud), it is important to prevent filling up the underlying disk, which can result in the repository becoming hard to recover (since deleting snapshots requires temporarily using more disk space for tombstones)

Assuming we implement something like #1519, we can have a condition that prevents new snapshots from being taken if the free space in the repository drops below certain level.

adowair:
This PR introduces the `Volume` interface, which is meant to encapsulate APIs that are used to access a storage provider's volume/disk access methods. For now, this interface contains one API, `GetCapacity()`, which returns a storage provider's total and available space. This PR resolves #1520


repo/blob/filesystem/filesystem_storage.go
```diff
@@ -70,6 +70,22 @@ func (fs *fsImpl) isRetriable(err error) bool {
 	return errors.Is(err, errRetriableInvalidLength)
 }
 
+func (fs *fsStorage) GetCapacity(ctx context.Context) (blob.Capacity, error) {
+	c, err := retry.WithExponentialBackoff(ctx, "GetCapacity", func() (interface{}, error) {
+		stat, err := fs.Impl.(*fsImpl).osi.Statfs("/")
```
jkowalski, line 75: this will return the root filesystem capacity, not necessarily the capacity of the actual blob storage location, is that intended?

talios, line 75: If I'm reading this right - this is checking statfs of the root filesystem - this may not be the filesystem of the actual repository backup right?

`/` might only have 1mb free, but `/data` may have 300gb?

repo/blob/sftp/sftp_storage.go
```diff
@@ -101,6 +101,22 @@ func (s *sftpImpl) IsConnectionClosedError(err error) bool {
 	return false
 }
 
+func (s *sftpStorage) GetCapacity(ctx context.Context) (blob.Capacity, error) {
+	i, err := s.Impl.(*sftpImpl).rec.UsingConnection(ctx, "GetCapacity", func(conn connection.Connection) (interface{}, error) {
+		stat, err := sftpClientFromConnection(conn).StatVFS("/")
```
jkowalski, line 106: same thing here, should we be passing the actual path and not the root?

talios: @jkowalski (as a drive by reviewer) that makes some logic - since we'd want the free space at that path ( I suspect even bucket based backends you may, as remote quota policies may be set?).

The paranoid monkey in me has seen ops being quickly create symlinks to other file systems and move an entire tree INSIDE used paths for quick storage space fixes, which means we're really at a best case scenario here anyway).

Look forward to @adowair's follow up..

repo/blob/filesystem/filesystem_storage.go
adowair: Thanks for catching this @jkowalski , I don't think I intended it that way–I had a misconception that a repo would have all the frees space on a volume available to it, which is not true, not least because quota policies can implemented as @talios suggested. I'll change these to use the repo path.

adowair: Yes, will fix.

repo/blob/sftp/sftp_storage.go
adowair: Yes, will fix.

repo/blob/filesystem/osinterface_realos.go
```diff
@@ -58,6 +59,14 @@ func (realOS) Stat(fname string) (os.FileInfo, error) {
 	return os.Stat(fname)
 }
 
+func (realOS) Statfs(fname string) (syscall.Statfs_t, error) {
+	var stat syscall.Statfs_t
+	err := syscall.Statfs(fname, &stat)
```
adowair, line 64: It looks like `Statfs()` and `Statfs_t` are not portable to all systems. Will have to find something that works across the board.

adowair: Perhaps this library is a good candidate: https://github.com/shirou/gopsutil

jkowalski: that dependency is quite large, can we use also syscall on Windows (a different one):

https://stackoverflow.com/questions/20108520/get-amount-of-free-disk-space-using-go

This requires conditional compilation - you can put Statfs() in separate files, one for Windows, and one for non-Windows.

adowair: It looks like we'll have to define this call at least thrice: `windows`, `openbsd_arm64`, and `linux_arm64`. At the risk of writing a lot of code that I'd like not to be in our purview to maintain. What does everyone think of importing just `GitHub.com/shirou/gopsutil/disk`? That sub package contains just the definitions we need (albeit for a larger set of architectures than we ourselves test/support).

jkowalski:
I would recommend against gopls dependency unless we end up using a lot more of things it covers. Supporting openbsd directly does not seem like a huge effort right for this case. 

Or we can simply say it's not supported for now on this particular OS.

repo/blob/filesystem/filesystem_storage_capacity_openbsd.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (fs *fsStorage) GetCapacity(ctx context.Context) (blob.Capacity, error) {
+	c, err := retry.WithExponentialBackoff(ctx, "GetCapacity", func() (interface{}, error) {
+		var stat syscall.Statfs_t
+		if err := syscall.Statfs(fs.RootPath, &stat); err != nil {
+			return blob.Capacity{}, errors.Wrap(err, "GetCapacity")
+		}
+
+		return blob.Capacity{
+			SizeB: uint64(stat.F_blocks) * uint64(stat.F_bsize), // nolint:unconvert,nolintlint
```
jkowalski, line 24: i'm surprised that `nolint` works, actually, but if the tests passes, we can take it.

repo/blob/filesystem/filesystem_storage_capacity_windows.go
```diff
@@ -0,0 +1,48 @@
...
+// nolint:gosec
+func (fs *fsStorage) GetCapacity(ctx context.Context) (blob.Capacity, error) {
+	dll, err := syscall.LoadLibrary("kernel32.dll")
+	defer syscall.FreeLibrary(dll) // nolint:errcheck
+
+	if err != nil {
+		return blob.Capacity{}, errors.Wrap(err, "load kernel32 library")
+	}
+
+	proc, err := syscall.GetProcAddress(dll, "GetDiskFreeSpaceExW")
```
jkowalski, line 25: According to 

https://stackoverflow.com/questions/54664751/how-to-get-total-size-of-a-drive-in-windows

there's a simpler solution where you import  "golang.org/x/sys/windows", can we use it here?

repo/blob/filesystem/filesystem_storage_capacity_openbsd.go
adowair: Yes, we can, with a couple of caveats:
1. If any syscalls fail the program will crash. I took the more verbose approach to be able to handle such errors gracefully. What do you think is the more appropriate way to handle errors here?
2. In the linked document, `syscall.StringToUTF16Ptr()` is deprecated. I'd use `syscall.UTF16PtrFromString()` instead.

repo/blob/filesystem/filesystem_storage_capacity_windows.go
```diff
@@ -15,34 +15,23 @@ import (
 
 // nolint:gosec
 func (fs *fsStorage) GetCapacity(ctx context.Context) (blob.Capacity, error) {
-	dll, err := syscall.LoadLibrary("kernel32.dll")
-	defer syscall.FreeLibrary(dll) // nolint:errcheck
+	dll := syscall.MustLoadDLL("kernel32.dll")
+	getDiskFreeSpaceExW := dll.MustFindProc("GetDiskFreeSpaceExW")
```
jkowalski, line 19: it can be even simpler, since there's already w windows wrapper that does GetDiskFreeSpaceExW

    var free, total, avail uint64

    path = "c:\\"
    pathPtr, err := windows.UTF16PtrFromString(path)
    if err != nil{
        panic(err)
    }
    err = windows.GetDiskFreeSpaceEx(pathPtr, &free, &total, &avail)

julio-lopez (commented):
+1: LG 👍

PR #1758, opened 2022-02-18
feat(general): Upgrade lock retries & monitoring during open & write sessions

Shrekster:
Upgrade lock monitor for Kopia does two things:

- Monitors the upgrade lock during runtime before each operation. This is protected by a cache-refresh timeout so that we do not do a hard pull on ever Get/Put etc or impact performance.
- The repository open() operation will wait poll and block on the upgrade lock until timeout.

Monitor Design: https://docs.google.com/document/d/1zIghQc4i4vJSANjSDBOvjzMiqnZSEp1cCasxSUkltSk/edit#heading=h.khston2oxmxy

### DIFF
--- a/repo/open.go
+++ b/repo/open.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -16,6 +17,7 @@ import (
 	"github.com/kopia/kopia/internal/clock"
 	"github.com/kopia/kopia/internal/epoch"
 	"github.com/kopia/kopia/internal/gather"
+	"github.com/kopia/kopia/internal/retry"
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/blob/beforeop"
 	loggingwrapper "github.com/kopia/kopia/repo/blob/logging"
@@ -69,6 +71,7 @@ type Options struct {
 	TraceStorage       bool             // Logs all storage access using provided Printf-style function
 	TimeNowFunc        func() time.Time // Time provider
 	DisableInternalLog bool             // Disable internal log
+	UpgradeOwnerID     string           // Owner-ID of any upgrade in progress, when this is not set the access may be restricted
 }
 
 // ErrInvalidPassword is returned when repository password is invalid.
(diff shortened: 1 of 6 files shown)


repo/open.go
```diff
@@ -387,25 +510,31 @@ func formatBytesCachingEnabled(cacheDirectory string, validDuration time.Duratio
 	return validDuration > 0
 }
 
-func readRepositoryBlobBytesFromCache(ctx context.Context, cachedFile string, validDuration time.Duration) ([]byte, error) {
+func readRepositoryBlobBytesFromCache(ctx context.Context, cachedFile string, validDuration time.Duration) ([]byte, time.Time, error) {
```
jkowalski, line 513: nit: can we add names to the output parameters, that time result is not obvious.

repo/open.go
```diff
@@ -420,11 +549,11 @@ func readAndCacheRepositoryBlobBytes(ctx context.Context, st blob.Storage, cache
 
 	cacheEnabled := formatBytesCachingEnabled(cacheDirectory, validDuration)
 	if cacheEnabled {
-		b, err := readRepositoryBlobBytesFromCache(ctx, cachedFile, validDuration)
+		b, cacheTime, err := readRepositoryBlobBytesFromCache(ctx, cachedFile, validDuration)
```
jkowalski, line 552: `cacheTime` is ambiguous, can we have a better/longer name that explains it better?

repo/open.go
```diff
@@ -183,51 +186,125 @@ func openDirect(ctx context.Context, configFile string, lc *LocalConfig, passwor
...
-	// Read repo blobs, potentially from cache.
-	fb, rb, err := readAndCacheRepositoryBlobs(ctx, st, caching.CacheDirectory, lc.FormatBlobCacheDuration)
+// nolint:gocritic
+func readAndCacheRepoConfig(
+	ctx context.Context,
+	st blob.Storage,
+	password string,
+	cacheOpts *content.CachingOptions,
+	validDuration time.Duration,
+) (
```
jkowalski, line 196: can this return a proper structure for readability?

repo/open.go
```diff
@@ -183,51 +186,125 @@ func openDirect(ctx context.Context, configFile string, lc *LocalConfig, passwor
...
+	repoConfig, err = f.decryptFormatBytes(formatEncryptionKey)
 	if err != nil {
-		return nil, ErrInvalidPassword
+		return nil, time.Time{}, nil, nil, nil, ErrInvalidPassword
+	}
+
+	return fb, cacheTime, f, repoConfig, formatEncryptionKey, nil
+}
+
+// ReadAndCacheRepoConfig loads the config from cache and returns the cache
```
jkowalski, line 237: again please elaborate on "cache time" here.

repo/open.go
```diff
@@ -333,6 +408,54 @@ func addThrottler(ctx context.Context, st blob.Storage) (blob.Storage, throttlin
...
+		// prevent backward jumps on lastSync
+		if lastSync.Before(cacheTime) {
+			lastSync = cacheTime
+		}
+		m.Unlock()
+
+		return nil
+	}
+
+	return beforeop.NewWrapper(st,
```
jkowalski, line 453: this is hard to understand, do we really need all these `cb` variants, can we just have one cb parameter to this wrapper and all it can do is fail?

repo/open.go
```diff
@@ -333,6 +408,54 @@ func addThrottler(ctx context.Context, st blob.Storage) (blob.Storage, throttlin
...
+		}
+
+		// only allow the upgrade owner to perform storage operations
+		if locked, _ := repoConfig.UpgradeLock.IsLocked(now()); locked && upgradeOwnerID != repoConfig.UpgradeLock.OwnerID {
+			return ErrRepositoryUnavailableDueToUpgrageInProgress
+		}
+
+		m.Lock()
+		// prevent backward jumps on lastSync
+		if lastSync.Before(cacheTime) {
```
jkowalski, line 445: nit: `cacheTime.After(lastSync)` instead?

repo/open.go
```diff
@@ -333,6 +408,54 @@ func addThrottler(ctx context.Context, st blob.Storage) (blob.Storage, throttlin
...
+		if err != nil {
+			return err
+		}
+
+		// only allow the upgrade owner to perform storage operations
+		if locked, _ := repoConfig.UpgradeLock.IsLocked(now()); locked && upgradeOwnerID != repoConfig.UpgradeLock.OwnerID {
+			return ErrRepositoryUnavailableDueToUpgrageInProgress
+		}
+
+		m.Lock()
```
jkowalski, line 443: shouldn't we move the `Lock()` right after `RUnlock()` to have full fence around caching?

repo/open.go
jkowalski, line 237: does this need to be exported?

repo/open.go
```diff
@@ -183,51 +186,125 @@ func openDirect(ctx context.Context, configFile string, lc *LocalConfig, passwor
...
+	}
+
+	var (
+		cacheTime           time.Time
+		f                   *formatBlob
+		repoConfig          *repositoryObjectFormat
+		formatEncryptionKey []byte
+	)
+
+	if err := retry.WithExponentialBackoffNoValue(ctx, "read repo config and wait for upgrade", func() error {
```
jkowalski, line 271: btw, as of now this exponential backoff is limited and very short overall duration, just a few seconds (it's a bug, I think)

you want a back-off here that lasts forever-ish, which we don't have a helper for.

repo/open.go
Shrekster: not atm, maybe on the next PR. Will unnexport for now

repo/open.go
Shrekster: will create a separate wrapper for the uniform callback case

repo/open.go
Shrekster: The desire is to avoid a mutex around a possible network bound operation to minimize the risk of locking down all goroutines. The caching mechanism uses atomic cache file operations, so this should not cause any corruption.

I understand that this might cause an inconsistent cache timestamp being tracked, but that'd be a minor compromise to favor a reduced scope for the lock.

Shrekster: Will do.

repo/open.go
Shrekster: dropped this function, I needed it on the POC PR for some reason but not anymore

repo/open.go
Shrekster: created another constructor for `beforeop` that sets up the same wrapper with a uniform callback

repo/open.go
Shrekster: done in [4dabb7d](https://github.com/kopia/kopia/pull/1758/commits/4dabb7d600c13aa983325619ac79ce3fcd271cdc)

repo/open.go
Shrekster: done in [4dabb7d](https://github.com/kopia/kopia/pull/1758/commits/4dabb7d600c13aa983325619ac79ce3fcd271cdc)

Issue #1796, opened 2022-03-03
Cannot successfully finish snapshots on KRS


budachst:
[root@kopia ~]# kopia --version
20220301.0.222516 build: 8f778ce951756f1ec50f0a858d3283f5a452f017 from: kopia/kopia

Since updating KRS to the version noted above, I am experiencing issues, where snapshots from remote Kopia clients fail at the very end, when everything seems to have been uploaded/verified. The error message from my KopiaUI on macOS looks like this:

```
Error: error flushing writer: error waiting for async writes: unable to create pending pack: unable to get session ID: unable to write session marker: unable to write session marker: sf8306bb0c991f4571df9382512156a1a-s885bc9df9ef6c2b910d: unable to read format blob: error getting kopia.repository blob: context canceled
```

I can only finish a snapshot, if I stop KRS and empty the local cache on the KRS. Afterwards, the following snapshot will succeed, but the following ones will again fail.

Some info:

- repo ist located on a standard XFS volume
- KRS gets started like this:

```
/usr/bin/kopia server start \
 --password=XXXXXXXXXX \
 --tls-cert-file /kopia/repos/ST7A.cert \
 --tls-key-file /kopia/repos/ST7A.key \
 --address 0.0.0.0:5151 \
 --config-file="/kopia/repos/ST7A.config" \
 --server-username stephan.budach@stephan \
 --log-level=error \
 --file-log-level=error \
 --log-dir="/kopia/repos" &
```

Config file looke like this:

```
{
  "storage": {
    "type": "filesystem",
    "config": {
      "path": "/mnt/kopia/ST7A",
      "dirShards": null,
      "fileMode": 384,
      "dirMode": 448
    }
  },
  "caching": {
    "cacheDirectory": "/kopia/cache/ST7A",
    "maxCacheSize": 5242880000,
    "maxMetadataCacheSize": 5242880000,
    "maxListCacheDuration": 600
  },
  "hostname": "vm",
  "username": "kopia",
  "description": "Repository in Filesystem: /mnt/kopia/ST7A"
}
```

cli-logs and content-logs are actually empty.

PR #1800, opened 2022-03-04
chore(repository): added context to beforeop callbacks

jkowalski:
This helps eliminate a class of bugs where 'ctx' used in a callbacks
is captured from a closure variable/parameter, which is almost never
correct, because it can be canceled while the beforeop wrapper is
still alive.

Follow-up for #1796.


Shrekster (approved):
LGTM

PR #1818, opened 2022-03-10
feat(cli): New Upgrade CLI / Switch to Format Version 3 (upgrade coordination)

Shrekster:
This change lets any Kopia repository owner set an upgrade intent on the repository to enable the draining protocol based upgrade coordination between multiple clients of Kopia and help gain an upgrade quorum before actually upgrading the repository.

- A dedicated upgrade CLI with the ability to set a new upgrade lock, update an existing lock and also revoke a previous orphaned lock.
- The change maintains backwards compatibility with existing `set-parameters --upgrade` based upgrade mechanism. However, it is encouraged to use the new CLI with new projects.
- Adding a new upgrade status section to the Kopia repository status CLI to allow any user to check for an ongoing upgrade status.
- Increments the max supported format version from 2 to 3. Version 3 repository format is not a data format change but has the draining protocol support which helps coordinate an upgrade quorum between all Kopia clients.

https://docs.google.com/document/d/1zIghQc4i4vJSANjSDBOvjzMiqnZSEp1cCasxSUkltSk/edit#heading=h.khston2oxmxy

PRs that introduced the upgrade-lock:

- https://github.com/kopia/kopia/pull/1742
- https://github.com/kopia/kopia/pull/1727
- https://github.com/kopia/kopia/pull/1758

Upgrading from format 1 to 3:
```sh
$ kopia repository upgrade --password 12345678 --config-file /tmp/kopia/repository.config --upgrade-owner-id owner --io-drain-timeout 1s --status-poll-interval 1s --max-clock-drift 1s
Repository upgrade lock intent has been placed.
Waiting for 2s to allow all other kopia clients to drain ...
Waiting for 1s to allow all other kopia clients to drain ...
Waiting for 0s to allow all other kopia clients to drain ...
Successfully drained all repository clients, the lock has been fully-established now.
migrating current indexes to epoch format
Repository indices have been upgraded.
Repository has been successfully upgraded
```

Upgrading from format 3 to 3:
```sh
$ kopia repository upgrade --password 12345678 --config-file /tmp/kopia/repository.config --upgrade-owner-id owner --io-drain-timeout 1s --status-poll-interval 1s --max-clock-drift 1s
ERROR error setting the upgrade lock intent: repository is already upgraded to 3, we can only upgrade till 3
```

Upgrading from format 2 to format 3:
```sh
$ kopia repository upgrade --password 12345678 --config-file /tmp/kopia/repository.config --upgrade-owner-id owner --io-drain-timeout 1s --status-poll-interval 1s --max-clock-drift 1s
Repository upgrade lock intent has been placed.
repository indices have already been migrated to the epoch format
Repository has been successfully upgraded
```

Non-owner upgraders during an upgrade:
```sh
$ kopia repository upgrade --password 12345678 --config-file /tmp/kopia/repository.config --upgrade-owner-id ownerx --io-drain-timeout 10s --status-poll-interval 5s --max-clock-drift 1s
ERROR failed to open repository: repository upgrade in progress
ERROR open repository: unable to open repository: repository upgrade in progress
```

Connections to repository during an upgrade:
```sh
$ kopia repository connect filesystem --password 12345678 --config-file /tmp/kopia/repository.config --path /tmp/kopia/repo/
ERROR failed to open repository: repository upgrade in progress
deleted repository password for /tmp/kopia/repository.config.
kopia: error: error connecting to repository: repository upgrade in progress, try --help
```

Upgrade status with advance notice:
```sh
$ kopia repository status --password 12345678 --config-file /tmp/kopia/repository.config --upgrade-owner-id owner
Config file:         /tmp/kopia/repository.config

Description:         Repository in Filesystem: /tmp/kopia/repo
Hostname:            c02fq0ttmd6r
Username:            shikhar.mall
Read-only:           false
Format blob cache:   15m0s

Storage type:        filesystem
Storage config:      {
                       "path": "/tmp/kopia/repo",
                       "fileMode": 384,
                       "dirMode": 448,
                       "dirShards": [
                         3,
                         3
                       ]
                     }

Unique ID:           83e7ec60180d56181c979213b22b9401c3709d1fcac1a2b2d9d8b921940e0c7e
Hash:                BLAKE2B-256-128
Encryption:          AES256-GCM-HMAC-SHA256
Splitter:            DYNAMIC-4M-BUZHASH
Format version:      3
Content compression: false
Password changes:    false
Max pack length:     20 MiB
Index Format:        v1
Epoch Manager:       disabled
Ongoing upgrade:     Format Version 3 -> 3
Upgrade Time:        2021-12-22 17:05:27.98409 -0800 PST
Upgrade lock:        Unlocked
Lock status:         Draining
```

Upgrade status with immediate lock:
```sh
$ kopia repository status --password 12345678 --config-file /tmp/kopia/repository.config --upgrade-owner-id owner
Config file:         /tmp/kopia/repository.config

Description:         Repository in Filesystem: /tmp/kopia/repo
Hostname:            c02fq0ttmd6r
Username:            shikhar.mall
Read-only:           false
Format blob cache:   15m0s

Storage type:        filesystem
Storage config:      {
                       "path": "/tmp/kopia/repo",
                       "fileMode": 384,
                       "dirMode": 448,
                       "dirShards": [
                         3,
                         3
                       ]                                                                                                                                                                                                                                                                             }

Unique ID:           83e7ec60180d56181c979213b22b9401c3709d1fcac1a2b2d9d8b921940e0c7e
Hash:                BLAKE2B-256-128
Encryption:          AES256-GCM-HMAC-SHA256
Splitter:            DYNAMIC-4M-BUZHASH
Format version:      3
Content compression: false
Password changes:    false
Max pack length:     20 MiB
Index Format:        v1
Epoch Manager:       disabled
Ongoing upgrade:     Format Version 1 -> 3
Upgrade Time:        2021-12-22 16:55:49.98409 -0800 PST
Upgrade lock:        Locked
Lock status:         Fully Established
```

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -245,8 +245,8 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("advanced-commands", "Enable advanced (and potentially dangerous) commands.").Hidden().Envar(c.EnvName("KOPIA_ADVANCED_COMMANDS")).StringVar(&c.AdvancedCommands)
 	app.Flag("track-releasable", "Enable tracking of releasable resources.").Hidden().Envar(c.EnvName("KOPIA_TRACK_RELEASABLE")).StringsVar(&c.trackReleasable)
 	app.Flag("dump-allocator-stats", "Dump allocator stats at the end of execution.").Hidden().Envar(c.EnvName("KOPIA_DUMP_ALLOCATOR_STATS")).BoolVar(&c.dumpAllocatorStats)
-	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar("KOPIA_REPO_UPGRADE_OWNER_ID").StringVar(&c.upgradeOwnerID)
-	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress.").Hidden().Default("false").Envar("KOPIA_REPO_UPGRADE_NO_BLOCK").BoolVar(&c.doNotWaitForUpgrade)
+	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar(c.EnvName("KOPIA_REPO_UPGRADE_OWNER_ID")).StringVar(&c.upgradeOwnerID)
+	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress.").Hidden().Default("false").Envar(c.EnvName("KOPIA_REPO_UPGRADE_NO_BLOCK")).BoolVar(&c.doNotWaitForUpgrade)
 
 	c.observability.setup(c, app)
 
--- a/cli/app.go
+++ b/cli/app.go
@@ -246,7 +246,7 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("track-releasable", "Enable tracking of releasable resources.").Hidden().Envar(c.EnvName("KOPIA_TRACK_RELEASABLE")).StringsVar(&c.trackReleasable)
 	app.Flag("dump-allocator-stats", "Dump allocator stats at the end of execution.").Hidden().Envar(c.EnvName("KOPIA_DUMP_ALLOCATOR_STATS")).BoolVar(&c.dumpAllocatorStats)
 	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar("KOPIA_REPO_UPGRADE_OWNER_ID").StringVar(&c.upgradeOwnerID)
-	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress.").Default("false").Envar("KOPIA_REPO_UPGRADE_NO_BLOCK").BoolVar(&c.doNotWaitForUpgrade)
+	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress.").Hidden().Default("false").Envar("KOPIA_REPO_UPGRADE_NO_BLOCK").BoolVar(&c.doNotWaitForUpgrade)
 
 	c.observability.setup(c, app)
 
--- a/cli/app.go
+++ b/cli/app.go
@@ -246,7 +246,7 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("track-releasable", "Enable tracking of releasable resources.").Hidden().Envar(c.EnvName("KOPIA_TRACK_RELEASABLE")).StringsVar(&c.trackReleasable)
 	app.Flag("dump-allocator-stats", "Dump allocator stats at the end of execution.").Hidden().Envar(c.EnvName("KOPIA_DUMP_ALLOCATOR_STATS")).BoolVar(&c.dumpAllocatorStats)
 	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar(c.EnvName("KOPIA_REPO_UPGRADE_OWNER_ID")).StringVar(&c.upgradeOwnerID)
-	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress.").Hidden().Default("false").Envar(c.EnvName("KOPIA_REPO_UPGRADE_NO_BLOCK")).BoolVar(&c.doNotWaitForUpgrade)
+	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress, instead exit with a message.").Hidden().Default("false").Envar(c.EnvName("KOPIA_REPO_UPGRADE_NO_BLOCK")).BoolVar(&c.doNotWaitForUpgrade)
 
 	c.observability.setup(c, app)
 
--- a/cli/app.go
+++ b/cli/app.go
@@ -126,7 +126,9 @@ type App struct {
 	cliStorageProviders           []StorageProvider
 	trackReleasable               []string
 
-	observability observabilityFlags
+	observability       observabilityFlags
+	upgradeOwnerID      string
+	doNotWaitForUpgrade bool
 
 	currentAction   string
 	onExitCallbacks []func()
@@ -243,6 +245,8 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("advanced-commands", "Enable advanced (and potentially dangerous) commands.").Hidden().Envar(c.EnvName("KOPIA_ADVANCED_COMMANDS")).StringVar(&c.AdvancedCommands)
 	app.Flag("track-releasable", "Enable tracking of releasable resources.").Hidden().Envar(c.EnvName("KOPIA_TRACK_RELEASABLE")).StringsVar(&c.trackReleasable)
 	app.Flag("dump-allocator-stats", "Dump allocator stats at the end of execution.").Hidden().Envar(c.EnvName("KOPIA_DUMP_ALLOCATOR_STATS")).BoolVar(&c.dumpAllocatorStats)
+	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar("KOPIA_REPO_UPGRADE_OWNER_ID").StringVar(&c.upgradeOwnerID)
+	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress.").Default("false").Envar("KOPIA_REPO_UPGRADE_NO_BLOCK").BoolVar(&c.doNotWaitForUpgrade)
 
 	c.observability.setup(c, app)
 
--- a/cli/command_repository_set_parameters_test.go
+++ b/cli/command_repository_set_parameters_test.go
@@ -141,13 +141,13 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
 	if s.formatVersion < content.MaxFormatVersion {
 		env.RunAndExpectSuccess(t, "repository", "upgrade",
 			"--upgrade-owner-id", "owner",
-			"--io-drain-timeout", "1s",
+			"--io-drain-timeout", "1s", "--force",
 			"--status-poll-interval", "1s",
 			"--max-clock-drift", "1s")
 	} else {
 		env.RunAndExpectFailure(t, "repository", "upgrade",
 			"--upgrade-owner-id", "owner",
-			"--io-drain-timeout", "1s",
+			"--io-drain-timeout", "1s", "--force",
 			"--status-poll-interval", "1s",
 			"--max-clock-drift", "1s")
 	}
--- a/cli/command_repository_set_parameters_test.go
+++ b/cli/command_repository_set_parameters_test.go
@@ -3,9 +3,11 @@ package cli_test
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/kopia/kopia/cli"
 	"github.com/kopia/kopia/internal/blobtesting"
 	"github.com/kopia/kopia/internal/repotesting"
 	"github.com/kopia/kopia/repo/blob"
@@ -146,8 +148,9 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
 			"--upgrade-owner-id", "owner",
 			"--io-drain-timeout", "1s", "--force",
 			"--status-poll-interval", "1s",
-			"--max-clock-drift", "1s",
 		}
+
+		cli.MaxPermittedClockDrift = func() time.Duration { return time.Second }
 
 		// You can only upgrade when you are not already upgraded
 		if s.formatVersion < content.MaxFormatVersion {
--- a/cli/command_repository_set_parameters_test.go
+++ b/cli/command_repository_set_parameters_test.go
@@ -145,7 +145,7 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
 		env.RunAndExpectSuccess(t, "index", "epoch", "list")
 	}
 
-	if s.formatVersion < content.FormatVersion3 {
+	if s.formatVersion < content.MaxFormatVersion {
 		env.RunAndExpectSuccess(t, "repository", "upgrade",
 			"--upgrade-owner-id", "owner",
 			"--io-drain-timeout", "1s",
--- a/cli/command_repository_set_parameters_test.go
+++ b/cli/command_repository_set_parameters_test.go
@@ -1,6 +1,7 @@
 package cli_test
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -31,15 +32,7 @@ func (s *formatSpecificTestSuite) TestRepositorySetParameters(t *testing.T) {
 
 	// default values
 	require.Contains(t, out, "Max pack length:     20 MiB")
-
-	switch s.formatVersion {
-	case content.FormatVersion1:
-		require.Contains(t, out, "Format version:      1")
-	case content.FormatVersion2:
-		require.Contains(t, out, "Format version:      2")
-	default:
-		require.Contains(t, out, "Format version:      3")
-	}
+	require.Contains(t, out, fmt.Sprintf("Format version:      %d", s.formatVersion))
 
 	// failure cases
 	env.RunAndExpectFailure(t, "repository", "set-parameters")
--- a/cli/command_repository_set_parameters_test.go
+++ b/cli/command_repository_set_parameters_test.go
@@ -140,18 +140,21 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
 
 	env.Environment["KOPIA_UPGRADE_LOCK_ENABLED"] = "1"
 
-	if s.formatVersion < content.MaxFormatVersion {
-		env.RunAndExpectSuccess(t, "repository", "upgrade",
+	{
+		cmd := []string{
+			"repository", "upgrade",
 			"--upgrade-owner-id", "owner",
 			"--io-drain-timeout", "1s", "--force",
 			"--status-poll-interval", "1s",
-			"--max-clock-drift", "1s")
-	} else {
-		env.RunAndExpectFailure(t, "repository", "upgrade",
-			"--upgrade-owner-id", "owner",
-			"--io-drain-timeout", "1s", "--force",
-			"--status-poll-interval", "1s",
-			"--max-clock-drift", "1s")
+			"--max-clock-drift", "1s",
+		}
+
+		// You can only upgrade when you are not already upgraded
+		if s.formatVersion < content.MaxFormatVersion {
+			env.RunAndExpectSuccess(t, cmd...)
+		} else {
+			env.RunAndExpectFailure(t, cmd...)
+		}
 	}
 
 	env.RunAndExpectSuccess(t, "repository", "set-parameters", "--upgrade")
--- a/cli/command_repository_set_parameters_test.go
+++ b/cli/command_repository_set_parameters_test.go
@@ -32,10 +32,13 @@ func (s *formatSpecificTestSuite) TestRepositorySetParameters(t *testing.T) {
 	// default values
 	require.Contains(t, out, "Max pack length:     20 MiB")
 
-	if s.formatVersion == content.FormatVersion1 {
+	switch s.formatVersion {
+	case content.FormatVersion1:
 		require.Contains(t, out, "Format version:      1")
-	} else {
+	case content.FormatVersion2:
 		require.Contains(t, out, "Format version:      2")
+	default:
+		require.Contains(t, out, "Format version:      3")
 	}
 
 	// failure cases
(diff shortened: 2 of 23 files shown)


cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+	cmd.Action(svc.directRepositoryWriteAction(c.runPhase(c.commitUpgrade)))
+
+	c.svc = svc
+}
+
+func (c *commandRepositoryUpgrade) runPhase(act func(context.Context, repo.DirectRepositoryWriter) error) func(context.Context, repo.DirectRepositoryWriter) error {
+	return func(ctx context.Context, rep repo.DirectRepositoryWriter) error {
+		if !c.skip {
+			if err := act(ctx, rep); err != nil {
+				// exc=plicitly skip all stages on error because tests do not
```
Shrekster, line 59: `exc=plicitly` -> `explicitly`

Shrekster, line 59: `exc=plicitly` -> `explicitly`

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+
+	svc advancedAppServices
+}
+
+func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent commandParent) {
+	cmd := parent.Command("upgrade", "Upgrade repository format.")
+
+	// TODO:
+	// cmd.GetFlag("upgrade-owner-id").Required()
+	// cmd.("owner-id", "The upgrade lock unique owner ID").Required().StringVar(&c.ownerId)
```
Shrekster, line 35: I should clean this up.

Shrekster, line 35: I should clean this up.

cli/command_repository_upgrade_test.go
```diff
@@ -0,0 +1,205 @@
...
+			"--max-clock-drift", "1s")
+	}
+
+	out = env.RunAndExpectSuccess(t, "repository", "status", "--upgrade-no-block")
+	require.Contains(t, out, "Epoch Manager:       enabled")
+	require.Contains(t, out, "Index Format:        v2")
+	require.Contains(t, out, "Format version:      3")
+}
+
+// TODO(small): add a test to verify that 0.8 client will choke on version-3
```
Shrekster, line 53: Let me go ahead an add this as well to this PR.

Shrekster, line 53: Let me go ahead an add this as well to this PR.

tests/end_to_end_test/suite_test.go
```diff
@@ -19,3 +19,7 @@ func TestFormatV1(t *testing.T) {
 func TestFormatV2(t *testing.T) {
 	testutil.RunAllTestsWithParam(t, &formatSpecificTestSuite{[]string{"--format-version=2"}, content.FormatVersion2})
 }
+
+func TestFormatV3(t *testing.T) {
+	testutil.RunAllTestsWithParam(t, &formatSpecificTestSuite{[]string{"--format-version=2"}, content.FormatVersion3})
```
Shrekster, line 24: crap, this needs to be `--format-version=3`

Shrekster, line 24: crap, this needs to be `--format-version=3`

Shrekster:
I will add the over-the-network lock monitor field to the intent as well as this is making the format version change.

Will update again tonight.

adowair:
Usage looks good to me. Nice Work! Shall we also the documentation on the [upgrade](https://kopia.io/docs/upgrade/) page?

cli/command_repository_set_parameters_test.go
```diff
@@ -32,10 +32,13 @@ func (s *formatSpecificTestSuite) TestRepositorySetParameters(t *testing.T) {
 	// default values
 	require.Contains(t, out, "Max pack length:     20 MiB")
 
-	if s.formatVersion == content.FormatVersion1 {
+	switch s.formatVersion {
```
adowair, line 35: Suggestion: shall we change this to:
```go
require.Contains(t, out, fmt.Sprintf("Format version:      %d", s.formatVersion))
```

cli/command_repository_set_parameters_test.go
```diff
@@ -111,14 +114,33 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
...
 		require.Contains(t, out, "Format version:      2")
 		require.Contains(t, out, "Epoch Manager:       enabled")
 		env.RunAndExpectSuccess(t, "index", "epoch", "list")
+	default:
+		require.Contains(t, out, "Format version:      3")
+		require.Contains(t, out, "Epoch Manager:       enabled")
+		env.RunAndExpectSuccess(t, "index", "epoch", "list")
+	}
+
+	if s.formatVersion < content.FormatVersion3 {
```
adowair, line 132:
```suggestion
	if s.formatVersion < content.MaxFormatVersion {
```

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/kopia/kopia/repo/content"
+)
+
+type commandRepositoryUpgrade struct {
+	forceRollback   bool
+	skip            bool
+	blockUntilDrain bool
+
+	// lock settings
+	advanceNoticeInterval  time.Duration
```
adowair, line 22: I see we assign`UpgradeLock{AdvanceNoticeDuration: ...}` to this field's value. Should we name this field accordingly? (more changes needed for suggestion).
```suggestion
	advanceNoticeDuration  time.Duration
```

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+	// cmd.GetFlag("upgrade-owner-id").Required()
+	// cmd.("owner-id", "The upgrade lock unique owner ID").Required().StringVar(&c.ownerId)
+
+	cmd.Flag("advance-notice", "Advance notice for upgrade to allow enough time for other Kopia clients to notice the lock").DurationVar(&c.advanceNoticeInterval)
+	cmd.Flag("io-drain-timeout", "Max time it should take all other Kopia clients to drop repository connections").Default(repo.DefaultRepositoryBlobCacheDuration.String()).DurationVar(&c.ioDrainTimeout)
+	cmd.Flag("status-poll-interval", "An advisory polling interval to check for the status of upgrade").Default("60s").DurationVar(&c.statusPollInterval)
+	cmd.Flag("max-clock-drift", "Maximum tolerated drift on clocks between all Kopia clients").Default("5m").DurationVar(&c.maxPermittedClockDrift)
+	cmd.Flag("block-until-drain", "Drain all the clients but do not perform the upgrade").BoolVar(&c.blockUntilDrain)
+
+	// TODO(small): make rollback a sub-command instead of a flag
```
adowair, line 43: This looks like it's going to be a small change (I may be wrong)–should we just include ti in this PR?

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+	now := rep.Time()
+	mp := rep.ContentReader().ContentFormat().MutableParameters
+	openOpts := c.svc.optionsFromFlags(ctx)
+	l := &content.UpgradeLock{
+		OwnerID:                openOpts.UpgradeOwnerID,
+		CreationTime:           now,
+		AdvanceNoticeDuration:  c.advanceNoticeInterval,
+		IODrainTimeout:         c.ioDrainTimeout,
+		StatusPollInterval:     c.statusPollInterval,
+		Message:                fmt.Sprintf("Upgrading from format version %d -> %d", mp.Version, content.FormatVersion3),
```
adowair, line 90: Is this always true?
```suggestion
		Message:                fmt.Sprintf("Upgrading from format version %d -> %d", mp.Version, content.MaxFormatVersion),
```

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+		MaxPermittedClockDrift: c.maxPermittedClockDrift,
+	}
+
+	// Update format-blob and clear the cache.
+	// This will fail if we have alread upgraded.
```
adowair, line 95:
```suggestion
	// This will fail if we have already upgraded.
```

cli/command_repository_upgrade_test.go
```diff
@@ -0,0 +1,205 @@
...
+	case content.FormatVersion2:
+		require.Contains(t, out, "Format version:      2")
+		_, stderr := env.RunAndExpectSuccessWithErrOut(t, "repository", "upgrade",
+			"--upgrade-owner-id", "owner",
+			"--io-drain-timeout", "1s",
+			"--status-poll-interval", "1s",
+			"--max-clock-drift", "1s")
+		require.Contains(t, stderr, "Repository indices have already been migrated to the epoch format, no need to drain other clients")
+		require.Contains(t, stderr, "Repository has been successfully upgraded.")
+	default:
```
adowair, line 38: Just thinking about the usage here: the "upgrade" API is not idempotent (i.e. two subsequent upgrade commands won't give the same result). I'm sure if we publish documentation it'll be okay, but I wonder if there's any potential for trouble.

adowair (commented):
Thanks for putting this up, some minor nits and questions

cli/command_repository_set_parameters_test.go
Shrekster: will do

cli/command_repository_upgrade.go
Shrekster: This ended up becoming a larger change, will look into this in a separate PR.

cli/command_repository_upgrade.go
Shrekster: Yes, that is true as of now.

cli/app.go
```diff
@@ -231,6 +233,8 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("persist-credentials", "Persist credentials").Default("true").Envar("KOPIA_PERSIST_CREDENTIALS_ON_CONNECT").BoolVar(&c.persistCredentials)
 	app.Flag("disable-internal-log", "Disable internal log").Hidden().Envar("KOPIA_DISABLE_INTERNAL_LOG").BoolVar(&c.disableInternalLog)
 	app.Flag("advanced-commands", "Enable advanced (and potentially dangerous) commands.").Hidden().Envar("KOPIA_ADVANCED_COMMANDS").StringVar(&c.AdvancedCommands)
+	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar("KOPIA_REPO_UPGRADE_OWNER_ID").StringVar(&c.upgradeOwnerID)
+	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress.").Default("false").Envar("KOPIA_REPO_UPGRADE_NO_BLOCK").BoolVar(&c.doNotWaitForUpgrade)
```
jkowalski, line 237: that should be hidden

Shrekster:
Moving the upgrade lock under maintenance schedule blob. Will update again shorty and put it back in review.

Shrekster:
@julio-lopez PTAL

cli/app.go
```diff
@@ -243,6 +245,8 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("advanced-commands", "Enable advanced (and potentially dangerous) commands.").Hidden().Envar(c.EnvName("KOPIA_ADVANCED_COMMANDS")).StringVar(&c.AdvancedCommands)
 	app.Flag("track-releasable", "Enable tracking of releasable resources.").Hidden().Envar(c.EnvName("KOPIA_TRACK_RELEASABLE")).StringsVar(&c.trackReleasable)
 	app.Flag("dump-allocator-stats", "Dump allocator stats at the end of execution.").Hidden().Envar(c.EnvName("KOPIA_DUMP_ALLOCATOR_STATS")).BoolVar(&c.dumpAllocatorStats)
+	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar(c.EnvName("KOPIA_REPO_UPGRADE_OWNER_ID")).StringVar(&c.upgradeOwnerID)
```
julio-lopez, line 248: As discussed offline: let's use the repo (maintenance) owner, instead of introducing another entity here. This means, the upgrade needs to be done from the repo owner instance. It has the advantage that 1. ensures maintenance is not running concurrently, and 2. it allows delayed execution (as part of maintenance) if we decide to do so in the future.

cli/app.go
```diff
@@ -243,6 +245,8 @@ func (c *App) setup(app *kingpin.Application) {
...
+	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress.").Hidden().Default("false").Envar(c.EnvName("KOPIA_REPO_UPGRADE_NO_BLOCK")).BoolVar(&c.doNotWaitForUpgrade)
```
julio-lopez, line 249: do these flags need to be "global"? can they be made specific to the upgrade command?

cli/app.go
```diff
@@ -126,7 +126,9 @@ type App struct {
 	cliStorageProviders           []StorageProvider
 	trackReleasable               []string
 
-	observability observabilityFlags
+	observability       observabilityFlags
+	upgradeOwnerID      string
```
julio-lopez, line 130: see comment below about _upgrade owner_

cli/command_repository_set_parameters_test.go
```diff
@@ -127,14 +123,35 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
...
+			"--upgrade-owner-id", "owner",
+			"--io-drain-timeout", "1s", "--force",
+			"--status-poll-interval", "1s",
+			"--max-clock-drift", "1s")
+	} else {
+		env.RunAndExpectFailure(t, "repository", "upgrade",
+			"--upgrade-owner-id", "owner",
+			"--io-drain-timeout", "1s", "--force",
+			"--status-poll-interval", "1s",
+			"--max-clock-drift", "1s")
```
julio-lopez, line 154: What's the difference between the `if` and `else` blocks?

julio-lopez (commented):
preliminary comments / questions

cli/app.go
Shrekster: these need to be supplied when you are connecting to the repo, so yes global

cli/command_repository_set_parameters_test.go
Shrekster: `RunAndExpectFailure` vs `RunAndExpectSuccess`

But I can still make this a bit nicer.

Shrekster: refactored the code slightly

https://github.com/kopia/kopia/pull/1818/commits/4f4917ec02093d3d222205ad5b3348477f519ca6

cli/app.go
Shrekster: Discussed with @julio-lopez , will make this change in a separate PR.

Shrekster:
@julio-lopez @adowair PTAL

- dropped --max-clock-drift flag, converting that into a const (still allowing tests to override so they do not have to block for the long default wait)
- dropped --block-until-drain flag, this was used for advance-notice feature. We can add this later as needed.
- updated documentation on --upgrade-no-block flag as it was used by status cli.
- Added subcommands to upgrade cli: begin & rollback.

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,291 @@
...
+
+	// Set the upgrade lock intent.
+	beginCmd.Action(svc.directRepositoryWriteAction(c.runPhase(c.setLockIntent)))
+	// If requested then drain all the clients otherwise stop here.
+	beginCmd.Action(svc.directRepositoryWriteAction(c.runPhase(c.drainOrCommit)))
+	// If the lock is fully established then perform the upgrade.
+	beginCmd.Action(svc.directRepositoryWriteAction(c.runPhase(c.upgrade)))
+	// Commit the upgrade and revoke the lock, this will also cleanup any
+	// backups used for rollback.
+	beginCmd.Action(svc.directRepositoryWriteAction(c.runPhase(c.commitUpgrade)))
```
julio-lopez, line 73: Why are these separate command `Action`s, more specifically, why are thy separate repo sessions (`directRepositoryWriteAction`)? Do they need to be?

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,291 @@
...
+	// lock settings
+	advanceNoticeDuration time.Duration
+	ioDrainTimeout        time.Duration
+	statusPollInterval    time.Duration
+
+	svc advancedAppServices
+}
+
+const (
+	experimentalWarning = `WARNING: The upgrade command is an EXPERIMENTAL feature. Please use with caution.
```
julio-lopez, line 32:
```suggestion
	experimentalWarning = `WARNING: The upgrade command is an EXPERIMENTAL feature. Please DO NOT use it, it may corrupt your repository and cause data loss.
```

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,291 @@
...
+	cf := rep.ContentReader().ContentFormat()
+	if cf.MutableParameters.EpochParameters.Enabled {
+		log(ctx).Infof("Repository indices have already been migrated to the epoch format, no need to drain other clients")
+
+		l, err := rep.GetUpgradeLockIntent(ctx)
+		if err != nil {
+			return errors.Wrap(err, "failed to get upgrade lock intent")
+		}
+
+		if l.AdvanceNoticeDuration == 0 {
```
julio-lopez, line 172: I do not understand the behavior here. If no (advance) notice was set, then no client draining is performed? that seems incorrect, unless I'm missing something here.

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,291 @@
...
+		if locked {
+			if writersDrained {
+				// we have the lock now
+				break
+			}
+		} else {
+			return errors.Wrap(err, "upgrade lock got revoked after the intent was placed, giving up")
+		}
+
+		// TODO: this can get stuck
```
julio-lopez, line 240: why? how?

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,291 @@
...
+				// skip/exit on error. Tests override os.Exit() that prevents
+				// running rest of the phases until we set the skip flag here.
+				// This flag is designed for testability and also to support
+				// rollback.
+				c.skip = true
+				return err
+			}
+		}
+
+		return nil
```
julio-lopez, line 111: Invert the check here to simplify these blocks?

```suggestion
		if c.skip {
			return nil
		}
		if err := act(ctx, rep); err != nil {
			// Explicitly skip all stages on error because tests do not
			// skip/exit on error. Tests override os.Exit() that prevents
			// running rest of the phases until we set the skip flag here.
			// This flag is designed for testability and also to support
			// rollback.
			c.skip = true
			return err
		}

```

cli/command_repository_upgrade.go
Shrekster: Good question, each of these functions are individual idempotent components which are retryable. Also, it is easier to read it in phases of intent -> drain -> upgrade -> commit.

We do not need this to be separate write sessions though, I can fix this in a follow-up PR.

cli/command_repository_upgrade.go
Shrekster: the if conditions wrapping this and the the advance notice block here is saying this:

if we have already upgraded (to version 2) and no advance notice was set then let the upgrade continue

cli/command_repository_upgrade.go
Shrekster: on Macs and systems which have resume from sleep capability, certain time jumps could cause a longer than expected wait time here

cli/command_repository_upgrade.go
Shrekster: thank you! committed

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,291 @@
...
+			if v := os.Getenv(c.svc.EnvName(upgradeLockFeatureEnv)); v == "" {
+				return errors.Errorf("please set %q env variable to use this feature", upgradeLockFeatureEnv)
+			}
+			return nil
+		})
+
+	beginCmd := parent.Command("begin", "Begin upgrade.").Default()
+	beginCmd.Flag("advance-notice", "Advance notice for upgrade to allow enough time for other Kopia clients to notice the lock").DurationVar(&c.advanceNoticeDuration)
+	beginCmd.Flag("io-drain-timeout", "Max time it should take all other Kopia clients to drop repository connections").Default(repo.DefaultRepositoryBlobCacheDuration.String()).DurationVar(&c.ioDrainTimeout)
+	beginCmd.Flag("force", "Force using an unsafe io-drain-timeout").Default("false").Hidden().BoolVar(&c.force)
```
julio-lopez, line 60:
```suggestion
	beginCmd.Flag("allow-unsafe-upgrade", "Force using an unsafe io-drain-timeout for the upgrade").Default("false").Hidden().BoolVar(&c.force)
```

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,293 @@
...
+
+	return nil
+}
+
+// drainOrCommit is the upgrade CLI phase that will actually wait for all the
+// clients to be drained out of the upgrade quorum. This phase will cause all other phases to be
+// skipped until the lock is fully established.
+func (c *commandRepositoryUpgrade) drainOrCommit(ctx context.Context, rep repo.DirectRepositoryWriter) error {
+	cf := rep.ContentReader().ContentFormat()
+	if cf.MutableParameters.EpochParameters.Enabled {
```
julio-lopez, line 166: Is it possible to have a repository in version 2 with `EpochParameters.Enabled == false`?

cli/command_repository_upgrade.go
```diff
@@ -0,0 +1,293 @@
...
+
+		locked, writersDrained := l.IsLocked(now)
+		if locked {
+			if writersDrained {
+				// we have the lock now
+				break
+			}
+		} else {
+			return errors.Wrap(err, "upgrade lock got revoked after the intent was placed, giving up")
+		}
```
julio-lopez, line 240:
```suggestion
		if !locked {
			return errors.Wrap(err, "upgrade lock got revoked after the intent was placed, giving up")
		}

		if writersDrained {
			// we have the lock now
			break
		}
```

Issue #540, opened 2020-08-16
Sparse file support


davidandreoletti:
Large files (eg virtual machine disk) with holes (aka sparse file) have different sizes on disk depending on the tools/options used.

```du -h --apparent-size file.img```: Shows a file' size on the host system as the max size of the image.
```du -h file.img```: Shows a file' size on the host system as the current actual size of the image (less than max size of the image).

Does kopia support sparse files for backup and restore (assuming identical host/filesystem as when it was backed up) ?

PS: +1 on creating kopia.

julio-lopez:
Kopia does not perform "sparse-region detection" per se. If the sparse regions are large enough (order ~4MB), then they may be captured by de-duplication. Also, Kopia supports compression, so regions with 0s are trivially compressible.

SarenT:
I want to quickly add here that in my case snapshot  "estimate" was way off due to a map file created by `ddrescue`. Actualy size: 100MB, `ls -lah` was showing 8TB

PR #1823, opened 2022-03-11
feat(snapshots): support restoring sparse files

adowair:
Implements support for restoring sparse files from snapshots.

When specifying `--mode=sparse` in a `snapshot restore`
command, Kopia will make a best effort to make sure the underlying
filesystem allocates the minimum amount of blocks needed to persist
restored files. In other words, enabling this feature will "force"
all restored files to be sparse-blocks of zero bytes in the source
file should not be allocated. This feature will only be enabled if:

- the OS is not Windows
- `--write-files-atomically` is not also selected

Otherwise, the flag will be silently ignored. Partially addresses the
concern in #540.

### DIFF
--- a/internal/sparsefile/sparsefile.go
+++ b/internal/sparsefile/sparsefile.go
@@ -1,0 +1,107 @@
+// Package sparsefile provides wrappers for handling the writing of sparse files (files with holes).
+package sparsefile
+
+import (
+	"io"
+	"os"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/internal/iocopy"
+	"github.com/kopia/kopia/internal/stat"
+)
+
+// Write writes the contents of src to the given targetPath, omitting any holes.
+func Write(targetPath string, src io.Reader, size int64) error {
+	dst, err := os.OpenFile(targetPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o600) //nolint:gosec,gomnd
+	if err != nil {
+		return err //nolint:wrapcheck
+	}
+
+	// ensure we always close f. Note that this does not conflict with the
+	// close below, as close is idempotent.
+	defer dst.Close() //nolint:errcheck,gosec
+
+	if err = dst.Truncate(size); err != nil {
+		return errors.Wrap(err, "error writing sparse file")
+	}
+
+	s, err := stat.GetBlockSize(targetPath)
+	if err != nil {
+		return errors.Wrap(err, "error writing sparse file")
+	}
+
+	buf := iocopy.GetBuffer()
+	defer iocopy.ReleaseBuffer(buf)
+
+	w, err := copySparse(dst, src, buf[0:s])
+	if err != nil {
+		return errors.Wrap(err, "error writing sparse file")
+	}
+
+	if w != size {
+		return errors.Errorf("")
+	}
+
+	if err := dst.Close(); err != nil {
+		return err //nolint:wrapcheck
+	}
+
+	return nil
+}
+
+func copySparse(dst io.WriteSeeker, src io.Reader, buf []byte) (written int64, err error) {
+	for {
+		nr, er := src.Read(buf)
+		if nr > 0 { // nolint:nestif
+			// If non-zero data is read, write it. Otherwise, skip forwards.
+			if isAllZero(buf) {
+				dst.Seek(int64(nr), os.SEEK_CUR) // nolint:errcheck
+				written += int64(nr)
+
+				continue
+			}
+
+			nw, ew := dst.Write(buf[0:nr])
+			if nw < 0 || nr < nw {
+				nw = 0
+
+				if ew == nil {
+					ew = errors.New("invalid write result")
+				}
+			}
+
+			written += int64(nw)
+
+			if ew != nil {
+				err = ew
+				break
+			}
+
+			if nr != nw {
+				err = io.ErrShortWrite
+				break
+			}
+		}
+
+		if er != nil {
+			if er != io.EOF {
+				err = er
+			}
+
+			break
+		}
+	}
+
+	return written, err
+}
+
+func isAllZero(buf []byte) bool {
+	for _, b := range buf {
+		if b != 0 {
+			return false
+		}
+	}
+
+	return true
+}
--- a/snapshot/restore/local_fs_output.go
+++ b/snapshot/restore/local_fs_output.go
@@ -15,6 +15,7 @@ import (
 	"github.com/kopia/kopia/fs/localfs"
 	"github.com/kopia/kopia/internal/atomicfile"
 	"github.com/kopia/kopia/internal/iocopy"
+	"github.com/kopia/kopia/internal/stat"
 	"github.com/kopia/kopia/snapshot"
 )
 
@@ -56,6 +57,9 @@ type FilesystemOutput struct {
 
 	// SkipTimes when set to true causes restore to skip restoring modification times.
 	SkipTimes bool `json:"skipTimes"`
+
+	// Sparse when set to true causes restore to allocate the minimum amount of disk space to write the file.
+	Sparse bool `json:"sparse"`
 }
 
 // Parallelizable implements restore.Output interface.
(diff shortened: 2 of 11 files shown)


snapshot/restore/local_fs_output.go
```diff
@@ -326,6 +336,81 @@ func write(targetPath string, r fs.Reader) error {
...
 
+type sparseWriter struct {
+	f *os.File
+}
+
+// nolint:wrapcheck
+func (sw *sparseWriter) Write(b []byte) (int, error) {
+	w := 0
+
+	bSize, err := stat.GetFileAllocSize(sw.f.Name())
```
jkowalski, line 347: this will be invoked tons of times (every 64KB), perhaps do it outside and pass as a field?

snapshot/restore/local_fs_output.go
```diff
@@ -315,7 +319,13 @@ func write(targetPath string, r fs.Reader) error {
 
 	name := f.Name()
 
-	if err := iocopy.JustCopy(f, r); err != nil {
+	if sparse {
+		err = iocopy.JustCopy(&sparseWriter{f: f}, r)
```
jkowalski, line 323: perhaps move this to `iocopy.CopySparse(w, r, blockSize)` ?

w can be *os.File or a smaller interface.

snapshot/restore/local_fs_output.go
```diff
@@ -326,6 +336,81 @@ func write(targetPath string, r fs.Reader) error {
...
+// nolint:wrapcheck
+func (sw *sparseWriter) Write(b []byte) (int, error) {
+	w := 0
+
+	bSize, err := stat.GetFileAllocSize(sw.f.Name())
+	if err != nil {
+		return w, err // nolint:wrapcheck
+	}
+
+	for i, j := 0, bSize; i < len(b); i, j = int(j), j+bSize {
```
jkowalski, line 352: i think this makes a hidden assumption that `Write()` is always called for multiples `bSize`. Instead of complicating writer we should just make a dedicated iocopy.CopySparse() .

snapshot/restore/local_fs_output.go
```diff
@@ -326,6 +336,81 @@ func write(targetPath string, r fs.Reader) error {
...
+			fi, err := sw.f.Stat()
+			if err != nil {
+				return w, err
+			}
+
+			// If at the end of file, extend it
+			if head == fi.Size() {
+				head += int64(len(block))
+
+				err = sw.f.Truncate(head)
```
jkowalski, line 382: Can we call Truncate() once at the very beginning with the final size - this will enable us to parallelize the work and speed id up, which for sparse files (which can be up to hundreds of gigabytes) is going to be critical.

We can:

1. Create result file.
2. Call Truncate on it to the requested size (which we know up front).
3. Divide the work into large segments, say 1GiB each (power of two is likely important here).
   [0..1GiB), [1GiB..2GiB), [2GiB..3GiB), [3GiB..4GiB) .... 
4. Create a gorotuine that will be returning these ranges in a channel.
5. Start N goroutines that will be reading from the channel, and:
       - seek to the start offset in the input and output
       - processing N bytes up to the provided count
6. Wait for all goroutines to finish work.
7. Close the result file.

jkowalski (changes requested):

tests/end_to_end_test/restore_test.go
```diff
@@ -488,6 +493,266 @@ func TestRestoreSnapshotOfSingleFile(t *testing.T) {
 	verifyFileMode(t, filepath.Join(restoreDir, "restored-5"), defaultRestoredFilePermission)
 }
 
+func TestSnapshotSparseRestore(t *testing.T) {
```
julio-lopez, line 496: Currently failing in CI on Ubuntu

snapshot/restore/local_fs_output.go
adowair: Ah right, thanks for pointing that out! We can make that part of the `sparseWriter` struct.

snapshot/restore/local_fs_output.go
adowair: I imagine we'd want CopySparse to take `io.Writer` as well (so that the interfaces align). At that point, shouldn't we just push the write logic down to the writer?

How about we introduce `SparseFile` in a new package, which implements `File` and custom `Writer`?

snapshot/restore/local_fs_output.go
adowair: That sounds really good actually–thanks for listing out the steps. I can follow up this PR with another one.

tests/end_to_end_test/restore_test.go
adowair: The following (and similar) lines are failing:
```go
```
with output:
```log
snapshot_create_test.go:74: finished in 337: 'kopia --config-file /tmp/kopia-test1214146990/fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff/.kopia.config --no-use-keyring snapshot list -l --manifest-id'
    snapshot_create_test.go:74: can't parse source info: "\x1b[m  2022-03-18 19:08:39 UTC k75283678450059522bca4aaa551375ac 54.8 KB drwx------ manifest:3810e2da6d7d38ee3e28f778addfa32c files:1151 dirs:211 (latest-2)"
```
This issue also coming up on latest master on other snapshot related tests (i.e `TestSnapshotCreate`), not just the ones introduced in this PR. The culprit is the ANSI reset code `\x1b[m` at the beginning of the line–the function `MustParseSnapshots()` mistakes it for a valid source string and tries to parse it. We should raise a separate issue for this.

snapshot/restore/local_fs_output.go
adowair: Alright, the context needed to perform all the ops the way we want to is all available in `FileSystemOutput.copyFileContent()`, but not any lower in the call chain. I've introduced a new package `sparsefile`, much like `atomicfile`, that has `Write()` and internal `copySparse()` functions. For now, we call only one of `atomicfile.Write(...)` and `sparsefile.Write(...)`. I'd like to come back and refactor this code to support both features together.

tests/end_to_end_test/restore_test.go
adowair: Fixed in the latest push.

internal/sparsefile/sparsefile.go
```diff
@@ -0,0 +1,104 @@
...
+	dst, err := os.OpenFile(targetPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o600) //nolint:gosec,gomnd
+	if err != nil {
+		return err //nolint:wrapcheck
+	}
+
+	// ensure we always close f. Note that this does not conflict with the
+	// close below, as close is idempotent.
+	defer dst.Close() //nolint:errcheck,gosec
+
+	dst.Truncate(size) // nolint:errcheck
```
jkowalski, line 25: why not check for error?

internal/sparsefile/sparsefile.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return nil
+}
+
+func copySparse(dst io.WriteSeeker, src io.Reader, buf []byte) (written int64, err error) {
+	for {
+		nr, er := src.Read(buf)
+		if nr > 0 { // nolint:nestif
+			// If non-zero data is read, write it. Otherwise, skip forwards.
+			isZeros := true
```
jkowalski, line 56: nit - extract a function `isAllZero()` ?

tests/end_to_end_test/restore_test.go
```diff
@@ -488,6 +493,271 @@ func TestRestoreSnapshotOfSingleFile(t *testing.T) {
...
+	t.Parallel()
+
+	if !(runtime.GOOS == linuxOSName && runtime.GOARCH == amdArchName) {
+		t.Skip("Sparse file behavior not guaranteed for this OS/Arch pair")
```
jkowalski, line 500: can you add more comments here?

jkowalski:
LGTM

internal/sparsefile/sparsefile.go
adowair: Good catch let me add a check here.

tests/end_to_end_test/restore_test.go
adowair: Done. Also moved this to `testutil.TestSkipUnlessLinux()`

jkowalski (approved):

PR #1839, opened 2022-03-21
feat(repository): speed up large file restores

jkowalski:
This introduces pipelining of blob reads and replaces many individual
content fetches with larger blob fetches for objects composed of
multiple contents (usually 10MB+).

For example - restoring ~13 GB file from Wasabi with clean cache:

Before: time 841 seconds, 15.5 MB/s

After:
    10MB read-ahead: time 462 seconds, 28.1 MB/s
    30MB read-ahead: time 373 seconds, 34.8 MB/s
    60MB read-ahead: time 344 seconds, 37.8 MB/s
    100MB read-ahead: time 280 seconds, 46.4 MB/s

This was on a 500Mbps download link.

The default read-ahead value is chosen based on the object size.


repo/object/object_reader.go
```diff
@@ -102,7 +106,38 @@ func (r *objectReader) Read(buffer []byte) (int, error) {
...
+
+	for i := r.currentChunkIndex + 1; i < len(r.seekTable); i++ {
+		if cid, _, ok := r.seekTable[i].Object.ContentID(); ok {
+			if i > r.maxReadAheadChunkIndex {
+				contentIDs = append(contentIDs, cid)
+				r.maxReadAheadChunkIndex = i
+			}
+
+			totalLength += r.seekTable[i].Length
+			if totalLength > r.opts.ReadAheadBytes {
```
Shrekster, line 127: nit;

```suggestion
			if totalLength >= r.opts.ReadAheadBytes {
```

repo/object/object_reader.go
```diff
@@ -102,7 +106,38 @@ func (r *objectReader) Read(buffer []byte) (int, error) {
...
+		}
+	}
+
+	if len(contentIDs) > 0 {
+		go r.cr.PrefetchContents(ctx, contentIDs, "blobs")
+	}
+}
+
 func (r *objectReader) openCurrentChunk() error {
+	r.readAhead(r.ctx)
```
Shrekster, line 139: Would `openCurrentChunk()` ever be called while there is a previous `PrefetchContents()` still running ? Does this have a worst case where this parallelism could explode on the total number of active fetch threads ?

repo/object/object_reader.go
```diff
@@ -102,7 +106,38 @@ func (r *objectReader) Read(buffer []byte) (int, error) {
...
+
+			totalLength += r.seekTable[i].Length
+			if totalLength > r.opts.ReadAheadBytes {
+				break
+			}
+		}
+	}
+
+	if len(contentIDs) > 0 {
+		go r.cr.PrefetchContents(ctx, contentIDs, "blobs")
```
Shrekster, line 134: Each call to `readAhead()` will lead to creation of five goroutines for parallel fetch with their own buffers.  We should share the prefetch pool of workers and a cyclic buffer with each call to `PrefetchContents()`.

repo/object/object_reader.go
Shrekster, line 139: One down side of this approach is that we won't pull the next content until we read the first byte. Also, if you are reading large content blobs then you are constantly hitting your prefetch boundary on each `openCurrentChunk()`, hence this will not prefetch until read if avg object size is close to `ReadAheadBytes`.

How about starting long running prefetch workers at the time of creation of the object-reader ? That way you are not continuously jump-starting your prefetch routines every time you hit a boundary.

tests/end_to_end_test/snapshot_migrate_test.go
```diff
@@ -52,7 +52,7 @@ func (s *formatSpecificTestSuite) TestSnapshotMigrate(t *testing.T) {
 
 	dstenv.RunAndExpectSuccess(t, "repo", "create", "filesystem", "--path", dstenv.RepoDir)
 
-	dstenv.RunAndExpectSuccess(t, "snapshot", "migrate", "--source-config", filepath.Join(e.ConfigDir, ".kopia.config"), "--all", "--parallel=5", "--overwrite-policies")
+	dstenv.RunAndExpectSuccess(t, "snapshot", "migrate", "--read-ahead-mb=30", "--source-config", filepath.Join(e.ConfigDir, ".kopia.config"), "--all", "--parallel=5", "--overwrite-policies")
```
Shrekster, line 55: It'd be great to add a test that verifies the max go-routine explosion of prefetches in the worst case.

repo/object/object_reader.go
jkowalski: Subsequent reads will generally block on completion of previous blob prefetches and multiple blob prefetches execute only once, so I don't think this will explode - it may create some relatively cheap goroutines though.

The only case I think where it may not behave great is if somebody opens a bunch of large readers, read just one byte and never read the rest, but it's pretty unlikely I'd say?

Having said that we should still have a hard concurrency limit at a lower layer (inside `content.WriteManager`). I'm going to add it.

repo/object/object_reader.go
jkowalski: Sharing buffers is already done using gather buffer, i will add a semaphore around goroutines.

repo/object/object_reader.go
jkowalski: The time between reading first byte and opening the reader is usually negligible. The upside of not reading immediately is that it lets you check the length and seek in the reader before you start.

tests/end_to_end_test/snapshot_migrate_test.go
jkowalski: will do.

repo/object/object_reader.go
Shrekster: Sounds good.

Also probably add a comment to this path around where the subsequent read will block. This might help readers to understand how that works.

cli/command_restore.go
```diff
@@ -96,6 +98,34 @@ type restoreSourceTarget struct {
 	isplaceholder bool
 }
 
+type objectReaderOptionsFlags struct {
+	disableReadAhead bool
+	readAheadMB      int64
```
adowair, line 103: What is the behavior of the system when `readAheadMB == 0` and disableReadAhead == false`? Is it necessary to have the flag variable?

NickIAm:
Any chance of this getting reviewed/merged soon-ish? I think it would be a good base to use for improving the full maintenance repacking speed, as that currently appears to fetches content in a similar way to restore.

PR #1844, opened 2022-03-21
fix(repository): fixed 'unable to get content data and info'

### LINKED ISSUE #1843: Content cache keys should be format-specific (including all format information from the index)
Currently any given content is cached using only content ID as the cache key.

This leads to the following problem:

1. Two different Kopia clients (A & B) can create a snapshot of the same file in 2 different directories with separate policies, one which uses compression and one which does not (or two different compression methods), at roughly the same time

2. There will be 2 index entries written - one which indicates compression, and one which does not, pointing at two separate blobs.

3. The merged index will pick either content from A or B based on time so will either be compressed or not.

4. One of the clients (A or B, depending on who loses the race) will not be able to decompress/decrypt the content because its cached format does not match the format indicated by the index.

5. Clearing the cache fixes the issue.

The issue typically manifests itself as: `error decompressing: invalid compression header, expected 00001000 but got 2320426c` but could also be regular decryption error.

The fix is to add format-specific suffix to content cache keys.

jkowalski:
We were incorrectly assigning the same content cache key regardless of
the storage format of the content (compression format, etc.)

This means that if a content has both compressed and non-compressed
entry in the index, it would sometimes get cached incorrectly
and the subsequent reads would fail.

Clearing the cache would fix the issue - this change appends
format-specific suffix to cache keys, so that clearing the cache is not
needed.

Huge thanks to Mark Derricutt who helped get to the bottom of this
on Slack.

Fix #1843

### DIFF
--- a/internal/cache/content_cache_data.go
+++ b/internal/cache/content_cache_data.go
@@ -16,9 +16,8 @@ type contentCacheForData struct {
 
 // ContentIDCacheKey computes the cache key for the provided content ID.
 func ContentIDCacheKey(contentID string) string {
-	// content IDs with odd length have a single-byte prefix.
 	// move the prefix to the end of cache key to make sure the top level shard is spread 256 ways.
-	if len(contentID)%2 == 1 {
+	if contentID[0] >= 'g' && contentID[0] <= 'z' {
 		return contentID[1:] + contentID[0:1]
 	}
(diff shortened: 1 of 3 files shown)


talios (commented):
+2 LGTM.

PR #1867, opened 2022-03-29
feat(cli): show storage capacity in repo status

adowair:
The connected repository's backing storage capacity and available
space can be now retrieved from `kopia repository status`. In text
format, these fields are printed in a human friendly form (MB, GB).
In JSON mode (`--json`), they are output as bytes.

### DIFF
--- a/cli/command_repository_status.go
+++ b/cli/command_repository_status.go
@@ -34,6 +34,7 @@ type RepositoryStatus struct {
 
 	ClientOptions repo.ClientOptions        `json:"clientOptions"`
 	Storage       blob.ConnectionInfo       `json:"storage"`
+	Capacity      blob.Capacity             `json:"volume"`
 	ContentFormat content.FormattingOptions `json:"contentFormat"`
 	ObjectFormat  object.Format             `json:"objectFormat"`
 	BlobRetention content.BlobCfgBlob       `json:"blobRetention"`
@@ -50,7 +51,7 @@ func (c *commandRepositoryStatus) setup(svc advancedAppServices, parent commandP
 	c.jo.setup(svc, cmd)
 }
 
-func (c *commandRepositoryStatus) outputJSON(r repo.Repository) error {
+func (c *commandRepositoryStatus) outputJSON(ctx context.Context, r repo.Repository) error {
 	s := RepositoryStatus{
 		ConfigFile:    c.svc.repositoryConfigFileName(),
 		ClientOptions: r.ClientOptions(),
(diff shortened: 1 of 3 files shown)


cli/command_repository_status.go
```diff
@@ -99,6 +110,16 @@ func (c *commandRepositoryStatus) run(ctx context.Context, rep repo.Repository)
 	ci := dr.BlobReader().ConnectionInfo()
 	c.out.printStdout("Storage type:        %v\n", ci.Type)
 
+	switch cp, err := dr.BlobVolume().GetCapacity(ctx); {
+	case err == nil:
+		c.out.printStdout("Storage capacity:    %v\n", units.BytesStringBase2(int64(cp.SizeB)))
```
jkowalski, line 115: use base10 (GiB) here. Base2 (GB) is only historically used for RAM sizes.

jkowalski (approved):
minor comment, otherwise LGTM.

cli/command_repository_status.go
adowair: Done

repo/blob/storage.go
```diff
@@ -56,9 +56,9 @@ type OutputBuffer interface {
 // Capacity describes the storage capacity and usage of a Volume.
 type Capacity struct {
 	// Size of volume in bytes.
-	SizeB uint64
+	SizeB uint64 `json:"capacity"`
```
julio-lopez, line 59: Printing out `0` size does not seem to make a lot of sense.
Also, what's a good way to handle or convey "unlimited" capacity.
In the unlimited case, does it make sense to have both of these fields have the `omitempty` tag?

cli/command_repository_status.go
```diff
@@ -34,6 +34,7 @@ type RepositoryStatus struct {
 
 	ClientOptions repo.ClientOptions        `json:"clientOptions"`
 	Storage       blob.ConnectionInfo       `json:"storage"`
+	Capacity      blob.Capacity             `json:"volume"`
```
julio-lopez, line 37: Should this be a pointer and have `omitempty` so it is not included when the capacity is unbounded? Or how should unlimited capacity be conveyed?

repo/blob/storage.go
adowair: That's right, thanks for catching that. I think `omitempty` makes sense here–we could argue that if the storage backend "isn't a volume" (doesn't have a limited capacity), then these fields are irrelevant and shouldn't show up.

cli/command_repository_status.go
adowair: Done

repo/blob/storage.go
```diff
@@ -56,9 +56,9 @@ type OutputBuffer interface {
 // Capacity describes the storage capacity and usage of a Volume.
 type Capacity struct {
 	// Size of volume in bytes.
-	SizeB uint64
+	SizeB uint64 `json:"capacity,omitempty"`
 	// Available (writeable) space in bytes.
-	FreeB uint64
+	FreeB uint64 `json:"available,omitempty"`
```
Shrekster, line 61: nit; Wondering if it makes sense to set `omitempty` on `FreeB` ? What happens when the volume is full ?

Shrekster (approved):
only nits; LGTM

repo/blob/storage.go
julio-lopez: Yes, a 0 value for `FreeB` should be included.
```suggestion
	FreeB uint64 `json:"available"`
```

adowair: That's a good point: we wouldn't be able to distinguish that from when FreeB is unset (unbounded). I think it should be enough for just `*Capacity` to be omitted.

PR #1890, opened 2022-04-14
feat(cli): added KOPIA_IGNORE_MALFORMED_MANIFEST_CONTENTS

jkowalski:
This can usually help open corrupted repository and partially restore data.


adowair:
@jkowalski I assume this is meant to be set by users as a bandaid to get some stuff out. Should we warn that the repo is inoperable and shouldn't be manipulated further? Are there more things Kopia can do to heal/stabilize when this happens?

PR #1899, opened 2022-04-23
feat(snapshots): Add min-retention-days feature

akash-rp:
This feature adds an option "min-retention-days" to policy. This option when set greater than 0 keeps all the snapshots within provided days. 

For example:
If we set "min-retention-days" to 14. All the snapshots within 14 days will be kept. The snapshots behind 14 days will be expired


cli/command_policy_set_retention.go
```diff
@@ -24,6 +25,7 @@ func (c *policyRetentionFlags) setup(cmd *kingpin.CmdClause) {
 	cmd.Flag("keep-weekly", "Number of most-recent weekly backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepWeekly)
 	cmd.Flag("keep-monthly", "Number of most-recent monthly backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepMonthly)
 	cmd.Flag("keep-annual", "Number of most-recent annual backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepAnnual)
+	cmd.Flag("keep-within", "Keep Backup upto time").PlaceHolder("N").StringVar(&c.policySetKeepWithin)
```
jkowalski, line 28: can we call this `min-retention-days` everywhere and rename . The word "within" is not obvious.

akash-rp: Okay

snapshot/policy/retention_policy.go
```diff
@@ -145,25 +148,35 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
 		{cutoff.weekly, fmt.Sprintf("%04v-%02v", yyyy, wk), "weekly", r.KeepWeekly},
 		{cutoff.daily, s.StartTime.Format("2006-01-02"), "daily", r.KeepDaily},
 		{cutoff.hourly, s.StartTime.Format("2006-01-02 15"), "hourly", r.KeepHourly},
+		{cutoff.within, s.StartTime.Format("2006-01-02 15"), "within", r.KeepWithin},
```
jkowalski, line 151: i don't think this logic belongs here, you should have a separate for loop, this will avoid a bunch of hacks in the current loop.

akash-rp: What changes do I need to make

akash-rp:
Please approve pending workflows

cli/command_policy_set_retention.go
```diff
@@ -24,6 +25,7 @@ func (c *policyRetentionFlags) setup(cmd *kingpin.CmdClause) {
 	cmd.Flag("keep-weekly", "Number of most-recent weekly backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepWeekly)
 	cmd.Flag("keep-monthly", "Number of most-recent monthly backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepMonthly)
 	cmd.Flag("keep-annual", "Number of most-recent annual backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepAnnual)
+	cmd.Flag("min-retention-days", "Minimun number of days to keep per source(or 'inherit)").PlaceHolder("N").StringVar(&c.policySetKeepMinDays)
```
adowair, line 28:
```suggestion
	cmd.Flag("min-retention-days", "Minimum number of days to keep per source (or 'inherit)").PlaceHolder("N").StringVar(&c.policySetKeepMinDays)
```

snapshot/policy/retention_policy.go
```diff
@@ -160,6 +168,14 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
 			continue
 		}
 
+		if *r.KeepMinDays != *newOptionalInt(0) {
+			if !s.StartTime.Before(cases[6].cutoffTime) {
+				keepReasons = []string{"Inside max retention days"}
```
adowair, line 173: Should this say "min"?
```suggestion
				keepReasons = []string{"Inside min retention days"}
```

adowair (commented):
Hey @akash-rp, welcome to the Kopia community! I'm still not sure exactly what the feature entails (the verbiage in the PR isn't consistent). Will you please update the PR definition with a summary of the change/usage/effects? I have a couple of nits/questions below.  Thanks for putting this PR up!

snapshot/policy/retention_policy.go
akash-rp: Yes, that should be min

snapshot/policy/retention_policy.go
```diff
@@ -160,6 +168,14 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
 			continue
 		}
 
+		if *r.KeepMinDays != *newOptionalInt(0) {
+			if !s.StartTime.Before(cases[6].cutoffTime) {
+				keepReasons = []string{"Inside min retention days"}
```
adowair, line 173: Should we append the reason here instead? There may be an overlap in the reasons for which a snapshot qualifies to be retained. Since `within` is the last member of `cases` struct, this won't get clobbered right now, but what if the order of members in `cases` struct changes in the future?

snapshot/policy/retention_policy.go
```diff
@@ -160,6 +168,14 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
 			continue
 		}
 
+		if *r.KeepMinDays != *newOptionalInt(0) {
```
adowair, line 171: Nit: is it necessary to compare to zero here? What happens if we remove lines `154-155` and do this:
```suggestion
		if r.KeepMinDays != nil {
```

adowair (changes requested):
Thanks @akash-rp, I've made some suggestions below

snapshot/policy/retention_policy.go
adowair: I've left some suggestions below. Alternatively, (and probably more correctly), this logic should be pulled out of this loop/function completely. See #660 for a similar change

snapshot/policy/retention_policy.go
```diff
@@ -160,6 +168,14 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
...
+			if !s.StartTime.Before(cases[6].cutoffTime) {
+				keepReasons = []string{"Inside min retention days"}
+			}
+
+			continue
+		}
```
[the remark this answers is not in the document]
akash-rp: I used `c.max = r.KeepMinDays` as the variable `max` already exists, Or else I think we need to add a new variable to the `case `struct. Can you suggest any changes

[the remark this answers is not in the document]
adowair: I believe that cutoff time here will serve your purpose (do double check that though), it will retain all snapshots within the calculated period, and if `c.max` is `MAX_INT`, it'll do so for any number of such snapshots.

snapshot/policy/retention_policy.go
akash-rp: I have removed this logic from loop.

snapshot/policy/retention_policy.go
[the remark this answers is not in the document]
akash-rp: I have moved this logic out of `for` loop

akash-rp:
@jkowalski Hello, can you review changes

cli/command_policy_set_retention.go
```diff
@@ -38,6 +40,7 @@ func (c *policyRetentionFlags) setRetentionPolicyFromFlags(ctx context.Context,
 		{"number of daily backups to keep", &rp.KeepDaily, c.policySetKeepDaily},
 		{"number of hourly backups to keep", &rp.KeepHourly, c.policySetKeepHourly},
 		{"number of latest backups to keep", &rp.KeepLatest, c.policySetKeepLatest},
+		{"minimun number of days to keep", &rp.KeepMinDays, c.policySetKeepMinDays},
```
jkowalski, line 43: please fix the typo "minimun", also rename the fields to match the flags.

the description of the flag is too similar to "number of daily backups to keep". 
How about "minimum number of days to retain the snapshot after creation" or something to that effect

snapshot/policy/retention_policy.go
```diff
@@ -147,7 +150,17 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
...
 	}
 
+	if r.KeepMinDays != nil {
+		if !s.StartTime.Before(cutoff.minDays) && *r.KeepMinDays > *newOptionalInt(0) {
+			keepReasons = []string{"Inside min retention days"}
+		}
+	}
+
 	for _, c := range cases {
+		if r.KeepMinDays != nil && *r.KeepMinDays > 0 {
```
jkowalski, line 160: you wouldn't need this if you simply remove entry from `cases` and handle it separately.

snapshot/policy/policy_tree.go
```diff
@@ -46,6 +46,7 @@ var (
 		KeepWeekly:  newOptionalInt(defaultKeepWeekly),
 		KeepMonthly: newOptionalInt(defaultKeepMonthly),
 		KeepAnnual:  newOptionalInt(defaultKeepAnnual),
+		KeepMinDays: newOptionalInt(defaultKeepMinDays),
```
jkowalski, line 49: nil is a valid option here, you can leave this uninitialized.

snapshot/policy/retention_policy.go
```diff
@@ -147,7 +150,17 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
 		{cutoff.hourly, s.StartTime.Format("2006-01-02 15"), "hourly", r.KeepHourly},
 	}
 
+	if r.KeepMinDays != nil {
+		if !s.StartTime.Before(cutoff.minDays) && *r.KeepMinDays > *newOptionalInt(0) {
+			keepReasons = []string{"Inside min retention days"}
```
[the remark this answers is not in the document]
akash-rp: Here we need to pass the variable `maxCompleteStartTime` to this function which is not available now

akash-rp:
@jkowalski please look into this

jkowalski:
this definitely needs tests before we can merge it

akash-rp:
Okay. Is their any ETA

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

adowair:
@akash-rp for any new feature, I think we'd normally expect tests for the feature to be included in the same pull request. If you take a look at `snapshots/policy/retention_policy_test.go`, there are some tests there that look like they could be a good starting point for you. Let us know if you have questions 😄

adowair:
Thanks for the tests @akash-rp, it looks like this code won't compile though:
```shell
ali.dowair -> ~/dev/kopia @ (retention-improvements) $ make ci-tests
...
# github.com/kopia/kopia/cli [github.com/kopia/kopia/cli.test]
cli/command_policy_show.go:146:164: syntax error: unexpected newline, expecting comma or )
cli/command_policy_show.go:147:3: syntax error: unexpected newline, expecting comma or }
cli/command_policy_show.go:148:2: syntax error: unexpected newline, expecting comma or )
cli/command_policy_show.go:150:27: syntax error: unexpected string at end of statement
# github.com/kopia/kopia/cli
cli/command_policy_show.go:146:164: syntax error: unexpected newline, expecting comma or )
cli/command_policy_show.go:147:3: syntax error: unexpected newline, expecting comma or }
cli/command_policy_show.go:148:2: syntax error: unexpected newline, expecting comma or )
cli/command_policy_show.go:150:27: syntax error: unexpected string at end of statement
make: *** [check-locks] Error 2
```

Please ensure that the code compiles (`make kopia`) and that the test runs successfully (`make ci-tests`). There's a short [contribution guideline page](https://kopia.io/docs/contribution-guidelines/) that'll be helpful.

akash-rp:
Fixed Error.

akash-rp:
@adowair I have cleared errors.

cli/command_policy_show.go
```diff
@@ -143,6 +143,7 @@ func appendRetentionPolicyRows(rows []policyTableRow, p *policy.Policy, def *pol
 		policyTableRow{"  Hourly snapshots:", valueOrNotSet(p.RetentionPolicy.KeepHourly), definitionPointToString(p.Target(), def.RetentionPolicy.KeepHourly)},
 		policyTableRow{"  Latest snapshots:", valueOrNotSet(p.RetentionPolicy.KeepLatest), definitionPointToString(p.Target(), def.RetentionPolicy.KeepLatest)},
 		policyTableRow{"  Ignore identical snapshots:", boolToString(p.RetentionPolicy.IgnoreIdenticalSnapshots.OrDefault(false)), definitionPointToString(p.Target(), def.RetentionPolicy.IgnoreIdenticalSnapshots)},
+		policyTableRow{"  Minimun Retention Days:", valueOrNotSet(p.RetentionPolicy.MinRetentionDays), definitionPointToString(p.Target(), def.RetentionPolicy.MinRetentionDays)},
```
jkowalski, line 146: You have a typo in the name: perhaps use `Min. Retention Days`

snapshot/policy/retention_policy.go
```diff
@@ -28,6 +28,7 @@ type RetentionPolicy struct {
 	KeepMonthly              *OptionalInt  `json:"keepMonthly,omitempty"`
 	KeepAnnual               *OptionalInt  `json:"keepAnnual,omitempty"`
 	IgnoreIdenticalSnapshots *OptionalBool `json:"ignoreIdenticalSnapshots,omitempty"`
+	MinRetentionDays         *OptionalInt  `json:"MinRetentionDays,omitempty"`
```
jkowalski, line 31: json fields should be lowercase

snapshot/policy/retention_policy.go
```diff
@@ -39,6 +40,7 @@ type RetentionPolicyDefinition struct {
 	KeepMonthly              snapshot.SourceInfo `json:"keepMonthly,omitempty"`
 	KeepAnnual               snapshot.SourceInfo `json:"keepAnnual,omitempty"`
 	IgnoreIdenticalSnapshots snapshot.SourceInfo `json:"ignoreIdenticalSnapshots,omitempty"`
+	MinRetentionDays         snapshot.SourceInfo `json:"MinRetentionDays,omitempty"`
```
jkowalski, line 43: same here

snapshot/policy/retention_policy.go
```diff
@@ -115,7 +118,7 @@ func (r *RetentionPolicy) ComputeRetentionReasons(manifests []*snapshot.Manifest
 }
 
 func (r *RetentionPolicy) effectiveKeepLatest() *OptionalInt {
-	if r.KeepLatest.OrDefault(0)+r.KeepHourly.OrDefault(0)+r.KeepDaily.OrDefault(0)+r.KeepWeekly.OrDefault(0)+r.KeepMonthly.OrDefault(0)+r.KeepAnnual.OrDefault(0) == 0 {
+	if r.KeepLatest.OrDefault(0)+r.KeepHourly.OrDefault(0)+r.KeepDaily.OrDefault(0)+r.KeepWeekly.OrDefault(0)+r.KeepMonthly.OrDefault(0)+r.KeepAnnual.OrDefault(0)+r.MinRetentionDays.OrDefault(0) == 0 {
```
jkowalski, line 121: this is not correct, the value returns the "count of snapshots" not the "number of snapshot + number of days" which does not really make sense. I think reverting this line is fine.

snapshot/policy/retention_policy.go
```diff
@@ -149,6 +152,12 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
 		{cutoff.hourly, s.StartTime.Format("2006-01-02 15"), "hourly", r.KeepHourly},
 	}
 
+	if r.MinRetentionDays != nil {
+		if !s.StartTime.Before(cutoff.minDays) && *r.MinRetentionDays > *newOptionalInt(0) {
```
jkowalski, line 156: something like:

```
if r.MinRetentionDays.OrDefault(0) > 0 && !s.StartTime.Before(cutoff.minDays) {
  keepReasons = append(keepReasons, "min-retention-days")
}
```

snapshot/policy/retention_policy.go
```diff
@@ -149,6 +152,12 @@ func (r *RetentionPolicy) getRetentionReasons(i int, s *snapshot.Manifest, cutof
...
+			keepReasons = []string{"Inside min retention days"}
```
jkowalski, line 157: let's not capitalize the string and should not include spaces.

README.md
```diff
@@ -1,82 +1,81 @@
-Kopia
```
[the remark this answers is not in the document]
jkowalski: please revert the entire file, this only has invisible changes

snapshot/policy/retention_policy.go
akash-rp: The problem is, if we don't add `r.MinRetentionDays.OrDefault(0)` and only `Min. Retention days` is set to some value and  other polices is set to zero then effectiveKeepLatest() function returns `keep-latest = maxInt`, so then every snapshot will have retention reasons as `[min-retention-days, latest-n ]` where `n = number of snapshots`

akash-rp: @jkowalski Can you suggest something

PR #1904, opened 2022-04-26
test(general): Ignore directory size check before and after restore

chaitalisg:
The directory size is affected by the underlying filesystem. Filter directory size check to make the tests more robust.

### DIFF
--- a/tests/tools/fswalker/fswalker.go
+++ b/tests/tools/fswalker/fswalker.go
@@ -35,7 +35,7 @@ func NewWalkCompare() *WalkCompare {
 		GlobalFilterFuncs: []func(string, fswalker.ActionData) bool{
 			filterFileTimeDiffs,
 			isRootDirectoryRename,
-			dirSizeMightBeOffByBlockSizeMultiple,
+			filterDirSizeCheck,
 			ignoreGIDIfZero,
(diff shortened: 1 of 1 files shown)


redgoat650 (approved):
LGTM; the only thing I'd suggest is a comment explaining the justification for this filter (different file systems may have different dir size), and maybe an explanation of its use (filter this "diff" item out of the list if the element is a directory and the problem was a difference in size).

tests/tools/fswalker/fswalker.go
```diff
@@ -186,18 +186,8 @@ func isRootDirectoryRename(diffItem string, mod fswalker.ActionData) bool {
...
-	}
-
-	if !strings.Contains(str, "size: ") {
-		return false
-	}
-
-	const blockSize = 4096
-
-	return (mod.Before.Stat.Size-mod.After.Stat.Size)%blockSize == 0
+func filterDirSizeCheck(str string, mod fswalker.ActionData) bool {
```
julio-lopez, line 189: As Nick suggested, it'd be good to add an inline comment explaining that the check is too restrictive across different file system setups.

julio-lopez (approved):

adowair (approved):

PR #1929, opened 2022-05-02
feat(cli): Allow customizing S3 upload part size (#1909)

### LINKED ISSUE #1909: [Feature Request] Allow customizing S3 PartSize
When uploading to Storj DCS I want to be able to specify a bigger part size, it seems that default one is 16 MiB

https://docs.storj.io/dcs/api-reference/s3-compatible-gateway/multipart-upload/multipart-part-size/

nazar554:
Add a new UploadPartSize JSON uint64 field for S3 storage config and "upload-part-size-bytes" cli flag.
Part size is specified in bytes, 0 means using the default minio client value (currently 16MiB).

Closes #1909.

### DIFF
--- a/cli/storage_s3.go
+++ b/cli/storage_s3.go
@@ -25,6 +25,7 @@ func (c *storageS3Flags) Setup(_ StorageProviderServices, cmd *kingpin.CmdClause
 	cmd.Flag("prefix", "Prefix to use for objects in the bucket").StringVar(&c.s3options.Prefix)
 	cmd.Flag("disable-tls", "Disable TLS security (HTTPS)").BoolVar(&c.s3options.DoNotUseTLS)
 	cmd.Flag("disable-tls-verification", "Disable TLS (HTTPS) certificate verification").BoolVar(&c.s3options.DoNotVerifyTLS)
+	cmd.Flag("upload-part-size-bytes", "Custom part size used for uploading the object").Default("0").Uint64Var(&c.s3options.UploadPartSize)
 
 	commonThrottlingFlags(cmd, &c.s3options.Limits)
 
--- a/repo/blob/s3/s3_storage.go
+++ b/repo/blob/s3/s3_storage.go
@@ -181,6 +181,7 @@ func (s *s3Storage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes, opt
 		StorageClass:    storageClass,
 		RetainUntilDate: retainUntilDate,
 		Mode:            retentionMode,
+		PartSize:        s.Options.UploadPartSize,
 	})
 
 	if isInvalidCredentials(err) {
--- a/repo/blob/s3/s3_options.go
+++ b/repo/blob/s3/s3_options.go
@@ -29,4 +29,7 @@ type Options struct {
 
 	// PointInTime specifies a view of the (versioned) store at that time
 	PointInTime *time.Time `json:"pointInTime,omitempty"`
+
+	// UploadPartSize specifies a custom part size used for uploading the object
+	UploadPartSize uint64 `json:"uploadPartSize,omitempty"`
 }


nazar554:
Not sure if it's better to use `-bytes` or `-mb` for this flag.

Should I:
- Write some tests?
- Create a form field for this flag inside S3 UI in [`kopia/htmlui`](https://github.com/kopia/htmlui)
- Validate the values? I think the [minio range](https://docs.min.io/docs/minio-server-limits-per-tenant.html) is 5 MiB to 5 GiB. Last part can be 0 B to 5 GiB

cli/storage_s3.go
```diff
@@ -25,6 +25,7 @@ func (c *storageS3Flags) Setup(_ StorageProviderServices, cmd *kingpin.CmdClause
 	cmd.Flag("prefix", "Prefix to use for objects in the bucket").StringVar(&c.s3options.Prefix)
 	cmd.Flag("disable-tls", "Disable TLS security (HTTPS)").BoolVar(&c.s3options.DoNotUseTLS)
 	cmd.Flag("disable-tls-verification", "Disable TLS (HTTPS) certificate verification").BoolVar(&c.s3options.DoNotVerifyTLS)
+	cmd.Flag("upload-part-size-bytes", "Custom part size used for uploading the object").Default("0").Uint64Var(&c.s3options.UploadPartSize)
```
adowair, line 28: Can we add some more context in this description? For example:
```suggestion
	cmd.Flag("upload-part-size-bytes", "Size of parts objects are split into for multipart uploading").Default("0").Uint64Var(&c.s3options.UploadPartSize)
```

repo/blob/s3/s3_storage.go
```diff
@@ -181,6 +181,7 @@ func (s *s3Storage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes, opt
 		StorageClass:    storageClass,
 		RetainUntilDate: retainUntilDate,
 		Mode:            retentionMode,
+		PartSize:		 s.Options.UploadPartSize,
```
adowair, line 184: Does it make sense to have this configurable per `putBlob()` as well? If y'all think that'd be useful, we can follow up with a PR in the future.

adowair (commented):
Hey @nazar554 ! Thanks for contributing. Some thoughts:
- I'm partial to `-mb`, but we should follow precedent and take @jkowalski's input as well
- Looking at [the page you linked](https://docs.storj.io/dcs/api-reference/s3-compatible-gateway/multipart-upload/multipart-part-size/) it looks like this option is called `multipart_threshold` in AWS cli. Let's follow suit and call it something like `multipart-upload-threshold-<bytes|mb>`?
- Let's definitely validate these values if there are published bounds. I imagine it's better for the user to see an error when configuring the repo rather than when uploading
- Yes, let's add this to the UI. @shrekster and @jkowalski might have input on whether to split that into its own PR.

repo/blob/s3/s3_storage.go
jkowalski: I wouldn't do it per blob, since it's hard to plumb through all the places we'd need this. Provider-level config is much easier to reason about.

cli/storage_s3.go
jkowalski: Yes, let's do the UI (https://github.com/kopia/htmlui/blob/main/src/SetupS3.js). This needs to be in separate PR anyway, since the code now lives in a different repository.

jkowalski:
Let's merge this into `testing/providers` branch first as this will run S3 tests after merge. Then we can merge this into `master`.

jkowalski:
Also, can we add some test that actually uses custom upload part sizes?

repo/blob/s3/s3_storage.go
jkowalski, line 184: this should be auto-formatted to align with the line above.

cli/storage_s3.go
nazar554: @jkowalski 

I think the Storj docs I linked inside the issue have a mistake
They reference it as `aws configure set default.s3.multipart_threshold 64MB`

But [AWS cli docs](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-threshold) say:

> When uploading, downloading, or copying a file, the S3 commands will switch to multipart operations if the file reaches a given size threshold. The multipart_threshold controls this value. You can specify this value in one of two ways:

It should actually be [`multipart_chunksize`](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-chunksize)

> Once the S3 commands have decided to use multipart operations, the file is divided into chunks. This configuration option specifies what the chunk size (also referred to as the **part size**) should be. This value can specified using the same semantics as multipart_threshold, that is either as the number of bytes as an integer, or using a size suffix.

nazar554: This is also mentioned as **part size** with [minio docs](https://docs.min.io/docs/golang-client-api-reference.html)
> opts.PartSize	uint64	Specify a custom part size used for uploading the object

repo/blob/s3/s3_storage.go
nazar554: I ran `go fmt` on that file and did a force-push.

nazar554:
> Also, can we add some test that actually uses custom upload part sizes?

I'm not sure I how to do that, I don't actually know Go 😅

cli/storage_s3.go
nazar554: I did a PR for docs there  https://github.com/storj/docs/pull/192

julio-lopez: Can anything be done on the client side to **NOT USE** multi-part uploads?

In general, multi-part uploads should be avoided altogether, and kopia is implemented to not require multi-part uploads.

nazar554: @julio-lopez That's the whole idea of this PR.

See it inside [minio client source](https://github.com/minio/minio-go/blob/1a65a8381319e069eee85921b5ffe11289e59a72/api-put-object.go#L257).

If a file (a pack) is smaller than provided `PartSize` or `opts.DisableMultipart` is true then it's uploaded with a single request.
If no `PartSize` is set then `minPartSize` is used which is equal to 16MiB right now.
Note that the default repsository pack size Kopia uses is 20 MiB (`max-pack-size-mb`) so it'll result in 2 parts.

Otherwise see calls to [`OptimalPartInfo`](https://github.com/minio/minio-go/blob/1a65a8381319e069eee85921b5ffe11289e59a72/api-put-object-streaming.go#L256)
It will try using `configuredPartSize` or switch to `minPartSize`. In case of later it'll split it up with this formula:

```go
configuredPartSize = minPartSize // 16MiB = 16777216
partSizeFlt = float64(objectSize / maxPartsCount)  // maxPartsCount is 10000
partSizeFlt = math.Ceil(partSizeFlt/float64(configuredPartSize)) * float64(configuredPartSize)
```

If you calculate it for `objectSize = 20MiB` you'll get `partSizeFlt = 16MiB`.
Note that there are performance benefits to using multiple parts if uploading concurrently, but I'm not sure if thats relevant if individual packs are uploaded in parallel.

julio-lopez: I see, thanks for the explanation.

What about always setting `opts.DisableMultipart=true` ? Would that work?

nazar554: @julio-lopez that works for me

Issue #1906, opened 2022-04-28
Recover from a corruption


oderwat:
I recently wanted to restore a large snapshot but I got a checksum error while doing that. So I took some time to fix the problem with the original data not using the backup.

Then I thought I may be able to get rid of that content and take a new full hashed snapshot to not lose about 250 GB of data which seems to be fine besides that single file error. But I can't figure out how to remove this object (or file) and create a new snapshot that verifies the file correctly.

Snapshot verify gives me the follow error for one of the files:

> error reading object Ix1912b8720256c1cb950878a429e53ed5: unable to read data: unexpected content error: invalid checksum at p2eb2ff505b6fd69113eed6a3ad93107d-s44e8a6f26712de1b108 offset 6118325 length 8316482/8316482: decrypt: unable to decrypt content: cipher: message authentication failed

So I tried to clean the cache and "content delete x1912b8720256c1cb950878a429e53ed5" and then back up the directory where this file is in and then snapshot verify with sources using this directory. But it still does not verify.

What can I do to not lose all the blobs in the repository and still get a correct new snapshot of all the data? I do not even care about the old snapshots, I just don't want to create a new repository and copy everything over again.

jkowalski:
If you still have the original file, deleting the content and re-uploading the snapshot with `--force-hash=100` should do the trick.

jkowalski:
I'll try to make a small change which adds ability to fully drop (as opposed to delete) contents from the index.

jkowalski:
@oderwat can you build  #1932 and then try:

```
$ kopia content delete --forget x1912b8720256c1cb950878a429e53ed5
$ kopia snapshot create ... --force-hash=100
```

Let me know if this fixes the issue.

oderwat:
Sorry @jkowalski, but I had to recreate the repository because of its size and the need to be able to really be able to restore it. I also switched this repository from backing up images of VM drives to actually back up the content after mounting (which is much slower) but having one file of the split VM drive corrupted brings everything down.

PR #1930, opened 2022-05-03
feat(cli): implementation for 'kopia snapshot fix'

### LINKED ISSUE #799: feature: implement ability to remove objects and their references after snapshot
There are some scenarios where you may want to get rid of objects (and their references) after they have been snapshotted in the repository:

- you accidentally snapshot a very large file or directory that should have been excluded
- you accidentally snapshot a sensitive file (say containing passwords, etc.)
- repository blobs got damaged and you want to remove references to them from manifests and directory listings.

I propose we add a feature to purge objects, say:

```
$ kopia purge <object-id>
```

This would:

- scan all snapshot roots
- recursively rewrite all directory listings, eliminating entries that had references to the objects to be removed to exclude them
- rewrite snapshots manifests whose object IDs have changed

The contents containing old object will become orphaned and subject to next GC.

Assuming `q` blobs are cached, this actually should be reasonably quick.

More ideas:

```
$ kopia drop-ignored
```

This would re-run old snapshots but apply current ignore rules (similar to migration in-place)

```
$ kopia snapshot verify --fix-unreadable
```

This would automatically rewrite directories to eliminate references to any objects for which the blobs are missing during snapshot verification.

jkowalski:
feat(cli): implementation for 'kopia snapshot fix'

This allows modifications and fixes to the snapshots after they have
been taken.

Supported are:

* `kopia snapshot fix remove-invalid-files [--verify-files-percent=X]`

Removes all directory entries where the underlying files cannot be
read based on index analysis (this does not read the files, only index
structures so is reasonably quick).

`--verify-files-percent=100` can be used to trigger full read for
all files.

* `kopia snapshot fix remove-files --object-id=<object-id>`

Removes the object with a given ID from the entire snapshot tree.
Useful when you accidentally snapshot a sensitive file.

* `kopia snapshot fix remove-files --filename=<wildcard>`

Removes the files with a given name from the entire snapshot tree.
Useful when you accidentally snapshot a sensitive file.

By default all snapshots are analyzed and rewritten. To limit the scope
use:

```
--source=user@host:/path
--manifest-id=manifestID
```

By default the rewrite operation writes new directory entries but
does not replace the manifests. To do that pass `--commit`.

Related #1906
Fixes #799


jkowalski:
This is not complete yet, needs testing, sending out early to get feedback on the CLI.

cli/command_snapshot_fix.go
```diff
@@ -0,0 +1,188 @@
...
+	verifier *snapshotfs.Verifier
+}
+
+func (c *commandSnapshotFix) setup(svc appServices, parent commandParent) {
+	cmd := parent.Command("fix", "Fix snapshot consistency issues")
+
+	// action to perform
+	cmd.Flag("remove-invalid-files", "Delete invalid files").BoolVar(&c.removeInvalidFiles)
+	cmd.Flag("remove-object", "Remove objects with specified ID").StringsVar(&c.removeObjectIDs)
+	cmd.Flag("remove-file", "Remove file with specified name (unqualified)").StringsVar(&c.removeFilesByName)
```
adowair, line 36: Since these are mutually exclusive in usage (or I imagine so), can we make them subcommands rather than flags?

```shell
kopia snapshot fix remove-invalid-files
kopia snapshot fix remove-object
kopia snapshot fix remove-file
```

What do you think?

cli/command_snapshot_fix.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// scope of rewrite
+	cmd.Flag("manifest-id", "Manifest IDs").StringsVar(&c.manifestIDs)
+	cmd.Flag("source", "Source to target (username@hostname:/path)").StringsVar(&c.sources)
+
+	// verification options
+	cmd.Flag("full-read", "Perform full read when verifying files").BoolVar(&c.fullRead)
+
+	cmd.Flag("commit", "Update snapshot manifests").BoolVar(&c.commit)
```
adowair, line 45: Forgive my ignorance: what is the point of having this default to off?

cli/command_snapshot_fix.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	for _, n := range c.removeFilesByName {
+		if ent.Name == n {
+			return nil, nil
+		}
+	}
+
+	if c.verifier != nil {
+		fname := dirRelativePath + "/" + ent.Name
```
adowair, line 64: tiny nit:
```suggestion
		fname := filepath.Join(dirRelativePath, ent.Name)
```

adowair (commented):
Nice! initial thoughts

- If we run this commands once without `--commit`, then once with, will there be duplicated work?

cli/command_snapshot_fix.go
jkowalski: FWIW this is actually subtly different (not sure if that matters), on Windows it will use `\` even for Unix snapshots. and `/` and `\` are both supported on Windows, although slash looks a bit out of place.

cli/command_snapshot_fix.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+
+	return ent, nil
+}
+
+func (c *commandSnapshotFix) run(ctx context.Context, rep repo.RepositoryWriter) error {
+	rw := snapshotfs.NewDirRewriter(rep, c.rewriteEntries)
+
+	if c.removeInvalidFiles {
```
Shrekster, line 79: Might be good idea to also specify a dr-run flag to not actually perform repo changes, instead just log them.

cli/command_snapshot_fix.go
jkowalski: That's a great idea. I implemented something very similar, PTAL.

cli/command_snapshot_fix.go
jkowalski: Safety, mostly. If the `fix` ends up deleting the entire repository (because of some broad deletion rule), that's not really helpful.

cli/command_snapshot_fix.go
jkowalski: That's the idea behind `--commit` flag.

adowair (commented):
This is great @jkowalski , I've left some cosmetic suggestions

cli/command_snapshot_fix_invalid_files.go
```diff
@@ -0,0 +1,59 @@
...
+func (c *commandSnapshotFixInvalidFiles) setup(svc appServices, parent commandParent) {
+	cmd := parent.Command("remove-invalid-files", "Remove references to invalid (unreadable) files in snapshots.")
+	c.common.setup(svc, cmd)
+
+	cmd.Flag("verify-files-percent", "Verify a percentage of files by fully downloading them [0.0 .. 100.0]").Default("0").Float64Var(&c.verifyFilesPercent)
+
+	cmd.Action(svc.repositoryWriterAction(c.run))
+}
+
+func (c *commandSnapshotFixInvalidFiles) rewriteEntries(ctx context.Context, dirRelativePath string, ent *snapshot.DirEntry) (*snapshot.DirEntry, error) {
```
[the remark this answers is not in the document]
jkowalski: yes, renamed, the function has gotten more advanced and is not really about skipping anymore. PTAL

cli/command_snapshot_fix.go
```diff
@@ -0,0 +1,133 @@
...
+				man.ID,
+				old.RootEntry.ObjectID,
+				man.RootEntry.ObjectID)
+
+			fixed = true
+		}
+	}
+
+	if fixed && !c.commit {
+		return errors.Errorf("fixes made, but not committed, pass --commit to update snapshots")
```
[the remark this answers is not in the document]
jkowalski: done

cli/command_snapshot_fix_remove_files.go
```diff
@@ -0,0 +1,57 @@
...
+	cmd := parent.Command("remove-files", "Remove references to files from snapshots.")
+	c.common.setup(svc, cmd)
+
+	cmd.Flag("object-id", "Remove files by their object ID").StringsVar(&c.removeObjectIDs)
+	cmd.Flag("filename", "Remove files by filename (wildcards are supported)").StringsVar(&c.removeFilesByName)
+
+	cmd.Action(svc.repositoryWriterAction(c.run))
+}
+
+func (c *commandSnapshotFixRemoveFiles) rewriteEntries(ctx context.Context, dirRelativePath string, ent *snapshot.DirEntry) (*snapshot.DirEntry, error) {
```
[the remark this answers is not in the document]
jkowalski: done

cli/command_snapshot_fix_invalid_files.go
```diff
@@ -0,0 +1,59 @@
...
+
+type commandSnapshotFixInvalidFiles struct {
+	common commonRewriteSnapshots
+
+	verifyFilesPercent float64
+	verifier           *snapshotfs.Verifier
+}
+
+func (c *commandSnapshotFixInvalidFiles) setup(svc appServices, parent commandParent) {
+	cmd := parent.Command("remove-invalid-files", "Remove references to invalid (unreadable) files in snapshots.")
```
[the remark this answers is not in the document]
jkowalski: done

adowair (commented):
Thank you Jarek, looks good to me, with a couple of notes:

Shrekster (approved):
Looks good. Thank you@

cli/command_snapshot_fix_remove_files.go
```diff
@@ -0,0 +1,59 @@
...
+	cmd.Flag("object-id", "Remove files by their object ID").StringsVar(&c.removeObjectIDs)
+	cmd.Flag("filename", "Remove files by filename (wildcards are supported)").StringsVar(&c.removeFilesByName)
+
+	cmd.Action(svc.repositoryWriterAction(c.run))
+}
+
+func (c *commandSnapshotFixRemoveFiles) rewriteEntry(ctx context.Context, dirRelativePath string, ent *snapshot.DirEntry) (*snapshot.DirEntry, error) {
+	for _, id := range c.removeObjectIDs {
+		if string(ent.ObjectID) == id {
+			return nil, nil
```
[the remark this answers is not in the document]
jkowalski: done

PR #1931, opened 2022-05-03
feat(repository): gdrive supports shared drives

ImSingee:
The current implementation of gdrive repository doesn't support [Shared drives](https://support.google.com/a/answer/7212025). The PR adds support for it


jkowalski:
I verified locally that current test is not affected.

jkowalski:
Thanks for your contribution!

PR #1932, opened 2022-05-05
feat(cli): added 'content delete --forget' flag

jkowalski:
This allows low-level hiding of entries in the index, which makes them completely invisible instead of just marking for deletion.

For #1906

### DIFF
--- a/repo/content/content_manager.go
+++ b/repo/content/content_manager.go
@@ -193,10 +193,6 @@ func (bm *WriteManager) ForgetContent(ctx context.Context, contentID ID) error {
 
 	bm.log.Debugf("forget-content %v", contentID)
 
-	if err := bm.maybeRefreshIndexes(ctx); err != nil {
-		return err
-	}
-
 	bi, err := bm.committedContents.getContent(contentID)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


repo/content/content_manager.go
```diff
@@ -174,7 +176,33 @@ func (bm *WriteManager) DeleteContent(ctx context.Context, contentID ID) error {
...
+// corruption.
+func (bm *WriteManager) ForgetContent(ctx context.Context, contentID ID) error {
+	bm.lock()
+	defer bm.unlock()
+
+	atomic.AddInt64(&bm.revision, 1)
+
+	bm.log.Debugf("forget-content %v", contentID)
+
+	if err := bm.maybeRefreshIndexes(ctx); err != nil {
```
julio-lopez, line 196: Why is this refresh needed here?

jkowalski: you're right, it's not, copy/paste from delete

adowair (commented):
👍

julio-lopez:
@jkowalski Given that directly deleting contents is dangerous since it'd lead to repo corruption, I am still trying to understand what the use case is, and how it'd be difference from a _"simple"_ content delete?

PR #1937, opened 2022-05-09
fix(repository): fix deletion immediately after creation

jkowalski:
In a very rare case, when content is created then deleted or forgotten
and immediately recreated in the same second, the newly recreated
content may be ignored due to how indices are merged.

This change ensures that on each {write,delete,forget} we always move
the time forward relative to latest index entry, even if the local clock
did not advance at all or moves backwards.

### DIFF
--- a/repo/content/content_manager.go
+++ b/repo/content/content_manager.go
@@ -199,9 +199,20 @@ func (bm *WriteManager) deletePreexistingContent(ctx context.Context, ci Info) e
 		return errors.Wrap(err, "unable to create pack")
 	}
 
-	pp.currentPackItems[ci.GetContentID()] = &deletedInfo{ci, bm.timeNow().Unix()}
+	pp.currentPackItems[ci.GetContentID()] = &deletedInfo{ci, bm.contentWriteTime(ci.GetTimestampSeconds())}
 
 	return nil
+}
+
+// contentWriteTime returns content write time for new content
+// by computing max(timeNow().Unix(), previousUnixTimeSeconds + 1).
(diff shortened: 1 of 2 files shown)


repo/content/content_manager.go
```diff
@@ -852,9 +887,13 @@ func (bm *WriteManager) getOverlayContentInfoReadLocked(contentID ID) (*pendingP
 }
 
 // +checklocksread:bm.mu
-func (bm *WriteManager) getContentInfoReadLocked(ctx context.Context, contentID ID) (*pendingPackInfo, Info, error) {
+// NOTE: previousWriteTime is written even if error is ErrContentNotFound
+// in cases where the content was previously forgotten.
+func (bm *WriteManager) getContentInfoReadLocked(ctx context.Context, contentID ID, previousWriteTime *int64) (*pendingPackInfo, Info, error) {
```
julio-lopez, line 892: Given that this function is already returning 3 values, why not return a fourth, instead of mixing _output_ parameters and return values?

It seems that it would also make it simpler to ignore where it needs to be ignored. Thoughts?

julio-lopez (commented):
What is the implication of:
- potentially changing the "logical" ordering of content writes wrt their time stamps.
- potentially having a content timestamp that is after the blob time stamp? Although, that possibility already exists in the current implementation.

LG overall, just an inline question / nit.

repo/content/content_manager.go
jkowalski: done

jkowalski:
The PR got super simplified after reverting the forget feature. PTAL @julio-lopez

julio-lopez (approved):

PR #1941, opened 2022-05-10
fix(providers): ensure multi-part upload is disabled for S3 backends

julio-lopez:


julio-lopez:
Partially addresses #1909, at least for the desire objective.

repo/blob/s3/s3_storage.go
```diff
@@ -172,7 +172,8 @@ func (s *s3Storage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes, opt
 	}
 
 	uploadInfo, err := s.cli.PutObject(ctx, s.BucketName, s.getObjectNameString(b), data.Reader(), int64(data.Length()), minio.PutObjectOptions{
-		ContentType: "application/x-kopia",
+		ContentType:      "application/x-kopia",
+		DisableMultipart: true,
```
adowair, line 176: Can we add a comment explaining why we do this? Along the lines of: `"Kopia already splits snapshot contents into small blobs for uploading, so there is no need for further splitting"`.

adowair (commented):
Thanks for putting this up @julio-lopez

repo/blob/s3/s3_storage.go
julio-lopez: @adowair Good idea. Please submit a PR with the change. 🙏🏼

Issue #1955, opened 2022-05-16
FR: Stream directory entries to uploader for a dynamic set of directory entries


ashmrtn:
I can spin up a PR for this, but thought it'd be helpful to get some feedback on the idea. I can have a draft PR out soon that has the skeleton of the idea but does not address the stuff in the "possible issues" section if that would help in thinking about things

#### Use case
an application is streaming data from an external source and would like to use kopia to create a backup of the data in a hierarchical manner. The application does not want to write the data to the local file system prior to adding it to kopia. In this instance, kopia would be used as a library instead of a CLI tool (i.e. application code would interface directly with kopia code)

The current directory implementation assumes that all directory entries are known when `Readdir` is called. This does not allow one discover/stream directory entries on the fly and send them to the uploader.

#### Proposal
Create a new abstraction `StreamedDirectory` in virtualfs that returns a `<- chan (fs.Entry, error)` and can be used to iterate over directory entries as they are discovered. Extend `uploader` implementation to iterate over entries from a `StreamedDirectory` using a range for-loop on the channel

`StreamedDirectory` interface
```go
// Caller provides/controls `incoming` channel and sends entries over it
func NewStreamedDirectory(name string, incoming <-chan (fs.Entry, error)) fs.StreamedDirectory
// Used by uploader
func (s *StreamedDirectory) Stream(ctx context.Context) <-chan (fs.Entry, error)
```

#### Affected components
* uploader
* ignorefs
* virtualfs

#### Possible Issues
**Ignoring selected files**
`StreamedDirectories` don't fit with the ignorefs interface as they cannot return a specific file in the directory. A simple solution would be to have ignorefs use the parent directory's ignorefs context to filter out ignored files

**Resource cleanup on cancellation**
In the API sketch above, `StreamedDirectory` has no way to notify the owner of the `incoming` channel that an upload has been canceled so that it can close the channel. This may fall outside the scope of kopia because presumably whatever controls the channel would know the kopia uploader returned an error, but is worth bringing up for discussion

**Playing nice with the uploader worker pool**
`StreamedDirectories` may block uploader worker pool goroutines while waiting for new entries. This could reduce the effective parallelism of the system.

A solution for this would be to hand all the `StreamDirectory` channels off to a single goroutine (a "watcher"), have it monitor them, and pass the work back to the worker pool when new entries come in. This would require a minimal amount of context for each channel the watcher has.

[...]

PR #1957, opened 2022-05-18
feat(snapshots): Add callback-based iteration function to Directory interface

ashmrtn:
As the first part of #1955 add an interface method that uses callbacks to process directory entries. Right now it defaults to calling the underlying `Readdir` implementation and then iterating over that result internally

### DIFF
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -79,6 +79,21 @@ type ErrorEntry interface {
 // ErrEntryNotFound is returned when an entry is not found.
 var ErrEntryNotFound = errors.New("entry not found")
 
+// Adapter for a naive Readdir -> IterateEntries implementation.
+func ReaddirToIterate(ctx context.Context, d Directory, cb func(context.Context, Entry) error) error {
+	entries, err := d.Readdir(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, e := range entries {
+		if err := cb(ctx, e); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // ReadDirAndFindChild reads all entries from a directory and returns one by name.
 // This is a convenience function that may be helpful in implementations of Directory.Child().
 func ReadDirAndFindChild(ctx context.Context, d Directory, name string) (Entry, error) {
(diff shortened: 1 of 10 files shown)


fs/localfs/local_fs.go
```diff
@@ -141,6 +141,20 @@ func toDirEntryOrNil(dirEntry os.DirEntry, prefix string) (fs.Entry, error) {
 	return entryFromDirEntry(fi, prefix), nil
 }
 
+func (fsd *filesystemDirectory) IterateEntries(ctx context.Context, cb func(context.Context, fs.Entry) error) error {
+	entries, err := fsd.Readdir(ctx)
```
jkowalski, line 145: all those implementations are the same, can you introduce a helper and call everywhere?

fs/localfs/local_fs.go
```diff
@@ -141,6 +141,14 @@ func toDirEntryOrNil(dirEntry os.DirEntry, prefix string) (fs.Entry, error) {
 	return entryFromDirEntry(fi, prefix), nil
 }
 
+func (fsd *filesystemDirectory) IterateEntries(ctx context.Context, cb func(context.Context, fs.Entry) error) error {
+	if err := fs.ReaddirToIterate(ctx, fsd, cb); err != nil {
```
jkowalski, line 145: nit: in wrapping is not that convenient in this particular case, it would also be ok to do `// nolint:wrapcheck` or even add a rule to exclude this in `.golangci.yml`

fs/entry.go
```diff
@@ -78,6 +79,22 @@ type ErrorEntry interface {
 // ErrEntryNotFound is returned when an entry is not found.
 var ErrEntryNotFound = errors.New("entry not found")
 
+// ReaddirToIterate is an adapter for a naive Readdir -> IterateEntries implementation.
+func ReaddirToIterate(ctx context.Context, d Directory, cb func(context.Context, Entry) error) error {
```
jkowalski, line 83: can we add a simple unit test here? other than that LGTM.

ashmrtn: is there another package that this function would be a good fit for or a good `Directory` implementation in which to place the test? Unfortunately even a simple test in `entries_test.go` would require minimal mocks for the `Directory`, `Entry`, and `os.FileInfo` interfaces as I cannot import `internal/mockfs` or any other concrete implementations of `Directory` due to import cycles

jkowalski:
Thanks for the contribution!

PR #1959, opened 2022-05-20
chore(general): minor cleanups

julio-lopez:


internal/cache/content_cache.go
```diff
@@ -97,8 +97,7 @@ func (c *contentCacheImpl) fetchBlobInternal(ctx context.Context, blobID blob.ID
 	if err := c.st.GetBlob(ctx, blobID, 0, -1, blobData); err != nil {
 		reportMissError()
 
-		// nolint:wrapcheck
-		return err
+		return errors.Wrapf(err, "failed to get blob: %s", blobID)
```
julio-lopez, line 100: @jkowalski are there any issues if the blob ID is included here?

jkowalski: only failing tests, but no other issue

PR #1963, opened 2022-05-22
feat(snapshots): streaming upload support

jkowalski:
This removes dependency on `fs.Directory.Readdir()` from the Uploader, ignorefs and repofs so instead of reading entire directory ahead of time, the contents can be streamed using `IterateEntries()`.

This both reduces memory consumption by quite a lot and speeds up snapshot creation by a tiny bit:

Backing up Kopia git enlistment (2.8 GB files:74722 dirs:12352):

* before (680e3f29bba59e76d118d877afea8c70fd782dcf) : 7.4s     420.422 MB RAM
* after:    7.2s     358.328 MB RAM

Note that this temporarily disables asynchronous uploads of singular large files, since it wasn't compatible with asynchronous iteration model. This will be restored and further improved by #1984 following this one.

For #1955

### DIFF
--- a/fs/ignorefs/ignorefs.go
+++ b/fs/ignorefs/ignorefs.go
@@ -77,19 +77,19 @@ type ignoreDirectory struct {
 	fs.Directory
 }
 
-func isCorrectCacheDirSignature(ctx context.Context, f fs.File) (bool, error) {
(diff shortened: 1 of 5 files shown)


julio-lopez (commented):
+1: LG 👍

Issue #1969, opened 2022-05-23
Hard-coded cmd.exe is causing error


bkmeneguello:
When I try to run my backup from Windows Task Scheduler, even using my own user, I get this error when Kopia tries to run the `beforeFolder` action:
```
ERROR kopia/cli upload error: error executing before-folder action: error running 'before-folder' action: essential action failed: exec: "C:\\Users\\me\\AppData\\Local\\Temp\\kopia-action3861831628\\cmd.exe": file does not exist
```
I'm not very sure why, but I found [some discussion](https://github.com/golang/go/issues/43724) about Windows PATH handling with `exec.Command` and I thought it could be related.

By replacing the `cmd.exe` with the fully qualified path [here](https://github.com/kopia/kopia/blob/2b38b4b99d69d95c7c4d3fd939a2736613e5ef61/snapshot/snapshotfs/upload_actions.go#L115) it worked like a charm.

So, I suggest replacing this hard-coded `cmd.exe` with `os.Getenv("COMSPEC")` which points to the full path of the CMD executable.

PR #1970, opened 2022-05-23
fix(snapshots): replace hard-coded cmd.exe with COMSPEC env var on Windows (#1969)

bkmeneguello:


julio-lopez:
Thanks @bkmeneguello

PR #1971, opened 2022-05-24
refactor(general): minor cleanups in robustness framework 

julio-lopez:


julio-lopez:
@chaitalisg

tests/robustness/robustness_test/main_test.go
```diff
@@ -118,7 +118,7 @@ func (th *kopiaRobustnessTestHarness) init(ctx context.Context, dataRepoPath, me
 }
 
 func (th *kopiaRobustnessTestHarness) makeBaseDir() bool {
-	baseDir, err := os.MkdirTemp("", "engine-data-")
+	baseDir, err := os.MkdirTemp("", "robustness-")
```
adowair, line 121: I am wary of this change for two reasons:

1. Could it cause confusion with the other robustness directories at this same (or lower) levels, if any exist?
2. Could it muck up some existing dependencies in codefresh (or other) scripts/configs.

adowair (commented):
Looks good overall, thanks @julio-lopez !

tests/robustness/robustness_test/main_test.go
julio-lopez: This should be internal to a robustness job execution. The objective is to make it explicit and separate the data vs. the metadata directories, vs. the one used for other "temporary" data.

Nevertheless, I'll revert this change and we can merge it separately.

chaitalisg (commented):
LGTM

PR #1973, opened 2022-05-25
refactor(cli): deprecate `snapshot gc` command

julio-lopez:


cli/command_snapshot_gc.go
```diff
@@ -17,13 +17,19 @@ type commandSnapshotGC struct {
 }
 
 func (c *commandSnapshotGC) setup(svc appServices, parent commandParent) {
-	cmd := parent.Command("gc", "Mark contents as deleted which are not used by any snapshot").Hidden()
+	cmd := parent.Command("gc", "DEPRECATED: This command will be removed. Snapshot GC is now automatically done as part of repository maintence. Snapshot GC marks as deleted all the contents that are not used by any snapshot").Hidden()
```
jkowalski, line 20: typo: 'maintence'

jkowalski: also, perhaps we should immediately delete it (and replace with a stub explaining what happened here), this is not helping anyone and is in fact dangerous when running in parallel to maintenance.

julio-lopez: done

PR #1976, opened 2022-05-26
refactor(general): acquire maintenance lock before updating schedule

julio-lopez:
Avoid adding an entry to the maintenance schedule for a maintenance run that may not execute at all.
Also, avoids the potential race of overwriting the schedule of an already running maintenance task.


Shrekster (commented):
LGTM

PR #1980, opened 2022-05-26
feat(repository): switched to using go-mmap for indexes

jkowalski:
Previous mmap implementation only allowed io.ReaderAt API
this one allows direct access to the underlying bytes which helps
remove a bunch of buffers, copying and allocation in index parsing.

### DIFF
--- a/repo/content/committed_content_index_disk_cache.go
+++ b/repo/content/committed_content_index_disk_cache.go
@@ -74,6 +74,8 @@ func (c *diskCommittedContentIndexCache) mmapOpenWithRetry(path string) (mmap.MM
 
 	mm, err := mmap.Map(f, mmap.RDONLY, 0)
 	if err != nil {
+		f.Close() // nolint:errcheck
+
 		return nil, nil, errors.Wrap(err, "mmap error")
 	}
 
--- a/repo/content/committed_content_index_disk_cache.go
+++ b/repo/content/committed_content_index_disk_cache.go
@@ -7,8 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/edsrzf/mmap-go"
 	"github.com/pkg/errors"
-	"golang.org/x/exp/mmap"
 
 	"github.com/kopia/kopia/internal/cache"
 	"github.com/kopia/kopia/internal/gather"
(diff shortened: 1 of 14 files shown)


repo/content/index/index.go
```diff
@@ -26,33 +25,40 @@ type Index interface {
...
 		return nil, errors.Errorf("invalid header format: %v", h.version)
 	}
 }
 
-func readAtAll(ra io.ReaderAt, p []byte, offset int64) error {
-	n, err := ra.ReadAt(p, offset)
-	if n != len(p) {
-		return errors.Errorf("incomplete read at offset %v, got %v bytes, expected %v", offset, n, len(p))
-	}
+func safeSlice(data []byte, offset int64, length int) (v []byte, err error) {
```
julio-lopez, line 46: what are the semantics of the `data` slice returned by the mmap library? Do either `cap(data)` or `len(data)` reflect the addressable region?

Are there errors other than out-of-bounds errors that need to be considered?
If that's the case, then the error used for recovery may be misleading.

repo/content/index/index.go
```diff
@@ -26,33 +25,40 @@ type Index interface {
...
+	defer func() {
+		if recover() != nil {
+			err = errors.Errorf("invalid slice offset=%v, length=%v, data=%v bytes", offset, length, len(data))
```
julio-lopez, line 49: Explicitly return a `nil []byte`  ?

```suggestion
			v = nil
			err = errors.Errorf("invalid slice offset=%v, length=%v, data=%v bytes", offset, length, len(data))
```

julio-lopez (approved):

repo/content/index/index.go
jkowalski: the slice is sized to the length of memory-mapped segment, so the bounds checker than runs when reslicing will do the right thing.

repo/content/committed_content_index_disk_cache.go
```diff
@@ -68,10 +68,26 @@ func (c *diskCommittedContentIndexCache) mmapOpenWithRetry(path string) (*mmap.R
...
 		time.Sleep(nextDelay)
 		nextDelay *= 2
-		f, err = mmap.Open(path)
+
+		f, err = os.Open(path) // nolint:gosec
 	}
 
-	return f, errors.Wrap(err, "mmap() error")
+	mm, err := mmap.Map(f, mmap.RDONLY, 0)
+	if err != nil {
```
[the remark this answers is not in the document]
jkowalski: done.

repo/content/index/index.go
julio-lopez: In that case, would checking the bounds work instead of trying to `panic` and `recover` ?

PR #1981, opened 2022-05-26
chore(snapshots): unify sparse and normal FS output paths

adowair:
This commit refactors the code paths that exercise normal and sparse
writing of restored content to the local filesystem. The goal is to
expose sparsefile.Copy() and iocopy.Copy() to be interchangeable,
thereby allowing us to wrap or transform them more easily in the future.


snapshot/restore/local_fs_output.go
```diff
@@ -359,15 +366,19 @@ func (o *FilesystemOutput) copyFileContent(ctx context.Context, targetPath strin
...
-			return sparsefile.Write(targetPath, r, f.Size())
+		if !isWindows() {
+			return write(targetPath, r, f.Size(), sparsefile.Copy)
 		}
+
+		log(ctx).Debugf("Sparse file support is not available on Windows, writing file normally")
+	}
+
+	// Wrap io.Copy to conform with the required interface
+	copier := func(w io.WriteSeeker, r io.Reader) (int64, error) {
```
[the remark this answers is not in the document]
adowair: That's a good point, there's no need for allocation. I'll make this an internal `ioCopy()` instead.

snapshot/restore/local_fs_output.go
```diff
@@ -25,6 +25,10 @@ const (
 	maxTimeDeltaToConsiderFileTheSame = 2 * time.Second
 )
 
+// Copier is a generic function type to perform the actual copying of data bits
+// from a source stream to a destination stream.
+type Copier func(io.WriteSeeker, io.Reader) (int64, error)
```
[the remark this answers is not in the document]
adowair: Done: changed to `streamCopier` to avoid conflict with existing `copier` in package

snapshot/restore/local_fs_output.go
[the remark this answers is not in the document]
jkowalski: one last nit: perhaps introduce `getStreamCopier()`, otherwise lgtm.

jkowalski (approved):

internal/sparsefile/sparsefile.go
```diff
@@ -8,49 +8,20 @@ import (
...
-	if w != size {
-		return errors.Errorf("")
-	}
-
-	if err := dst.Close(); err != nil {
-		return err //nolint:wrapcheck
-	}
-
-	return nil
+	return copyBuffer(dst, src, buf)
```
adowair, line 19: I realize now one thing missing here is we pass the entire buffer to `copyBuffer` (64K), not the intended disk block size (usually `512B`). This makes the current implementation of E2E test fail.

The next effect of this is that zero block detection/seeking will happen at that higher granularity, which is less than ideal. On the other hand, we can now enable sparse restores for Windows without much work, as we will no longer need to implement `GetDiskBlockSize()`–something that proved to be tricky for Windows.

@Shrekster we had a discussion earlier about this and I think we agreed this is an okay tradeoff. @jkowalski What do you think?

adowair, line 19: I realize now one thing missing here is we pass the entire buffer to `copyBuffer` (64K), not the intended disk block size (usually `512B`). This makes the current implementation of E2E test fail.

The next effect of this is that zero block detection/seeking will happen at that higher granularity, which is less than ideal. On the other hand, we can now enable sparse restores for Windows without much work, as we will no longer need to implement `GetDiskBlockSize()`–something that proved to be tricky for Windows.

@Shrekster we had a discussion earlier about this and I think we agreed this is an okay tradeoff. @jkowalski What do you think?

adowair: We can also just hardcode `512B` in...

adowair: Resolved. Got around this by simply using a closure. Windows still not enabled, I'd like to work on that in a future PR =)

adowair:
Ran the HTML E2E job locally (MacOS) and on a fork, and it seems to pass. The run fails here due to a timeout, which makes me think it is a flake.

snapshot/restore/local_fs_output.go
```diff
@@ -281,6 +286,28 @@ func (o *FilesystemOutput) shouldUpdateTimes(local, remote fs.Entry) bool {
 	return !local.ModTime().Equal(remote.ModTime())
 }
 
+func (o *FilesystemOutput) getStreamCopier(ctx context.Context) streamCopier {
+	if o.Sparse {
+		if !isWindows() {
+			return func(w io.WriteSeeker, r io.Reader) (int64, error) {
+				s, err := stat.GetBlockSize(o.TargetPath)
```
jkowalski, line 293: nit: we can do this once per FilesystemOutput since it's not going to change. Perhaps call it right at the very beginning?

jkowalski:
The flakes are known, we should fix them, but you don't worry about them for now. I have only minor tweaks, still LGTM, feel free to merge after those are addressed.

snapshot/restore/local_fs_output.go
adowair: Done: moved `getStreamCopier()` to be public and called only once when first constructing `FilesystemOutput`.

snapshot/restore/local_fs_output.go
```diff
@@ -63,8 +86,8 @@ type FilesystemOutput struct {
 	// SkipTimes when set to true causes restore to skip restoring modification times.
 	SkipTimes bool `json:"skipTimes"`
 
-	// Sparse when set to true causes restore files sparsely-not writing any holes (zero regions) to disk.
-	Sparse bool `json:"sparse"`
+	// Copier is the StreamCopier to use for copying the actual bit stream to output.
+	Copier StreamCopier `json:"copier"`
```
jkowalski, line 92: this won't work, unfortunately. when serialized to JSON.

Since FilesystemOutput is passed verbatim from the UI, we need to pass `sparse` as boolean and resolve it later.

adowair: Ah good point, I haven't thought of implementing this in UI, and should. Will do

adowair: Reintroducing `Sparse` as an exported field, and made `Copier` private, to be assigned on the first call to `o.WriteFile()`.

adowair: Correction, it looks like initializing copier upon the first call to `Writefile()` is not threadsafe, and has caused the race detector test to fail. Introducing `Init()` to replace `NewFilesystemOutput()` to do this assignment. Will call `Init()` for both server and CLI paths.

PR #1986, opened 2022-05-27
feat(snapshots): Remove remaining internal uses of Readdir

ashmrtn:
Part of changes proposed in #1955 for streaming directory entries
Dependent on #1963 for ignorefs changes

Does not attempt to standardize how many callbacks may be executed by `IterateEntries` when an error occurs while reading directory entries (i.e. some implementations will fail without executing any callbacks while others will execute callbacks on entries until they reach the failure)

### DIFF
--- a/cli/command_ls.go
+++ b/cli/command_ls.go
@@ -55,11 +55,7 @@ func (c *commandList) run(ctx context.Context, rep repo.Repository) error {
 
 func (c *commandList) listDirectory(ctx context.Context, d fs.Directory, prefix, indent string) error {
 	if err := d.IterateEntries(ctx, func(innerCtx context.Context, e fs.Entry) error {
-		if err2 := c.printDirectoryEntry(innerCtx, e, prefix, indent); err2 != nil {
-			return err2
-		}
-
-		return nil
+		return c.printDirectoryEntry(innerCtx, e, prefix, indent)
 	}); err != nil {
 		return err // nolint:wrapcheck
 	}
--- a/cli/command_ls.go
+++ b/cli/command_ls.go
@@ -54,15 +54,14 @@ func (c *commandList) run(ctx context.Context, rep repo.Repository) error {
 }
 
 func (c *commandList) listDirectory(ctx context.Context, d fs.Directory, prefix, indent string) error {
-	entries, err := d.Readdir(ctx)
-	if err != nil {
-		return errors.Wrap(err, "error reading directory")
-	}
-
-	for _, e := range entries {
-		if err := c.printDirectoryEntry(ctx, e, prefix, indent); err != nil {
-			return errors.Wrap(err, "unable to print directory entry")
+	if err := d.IterateEntries(ctx, func(innerCtx context.Context, e fs.Entry) error {
+		if err2 := c.printDirectoryEntry(innerCtx, e, prefix, indent); err2 != nil {
+			return err2
 		}
+
+		return nil
+	}); err != nil {
+		return err // nolint:wrapcheck
 	}
 
 	if dws, ok := d.(fs.DirectoryWithSummary); ok && c.errorSummary {
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -75,20 +75,40 @@ type ErrorEntry interface {
 	ErrorInfo() error
 }
 
+// GetAllEntries uses IterateEntries to return all entries in a Directory.
+func GetAllEntries(ctx context.Context, d Directory) (fs.Entries, error) {
+  var entries fs.Entries(nil)
+
+	err := d.IterateEntries(ctx, func(ctx context.Context, e Entry) error {
+		entries = append(entries, e)
+		return nil
+	})
+
+	return entries, err // nolint:wrapcheck
+}
+
 // ErrEntryNotFound is returned when an entry is not found.
 var ErrEntryNotFound = errors.New("entry not found")
 
 // IterateEntriesAndFindChild iterates through entries from a directory and returns one by name.
 // This is a convenience function that may be helpful in implementations of Directory.Child().
 func IterateEntriesAndFindChild(ctx context.Context, d Directory, name string) (Entry, error) {
+	type errStop struct {
+		error
+	}
+
 	var result Entry
 
-	if err := d.IterateEntries(ctx, func(c context.Context, e Entry) error {
+	err := d.IterateEntries(ctx, func(c context.Context, e Entry) error {
 		if result == nil && e.Name() == name {
 			result = e
+			return errStop{errors.New("")}
 		}
 		return nil
-	}); err != nil {
+	})
+
+	var stopped errStop
+	if err != nil && !errors.As(err, &stopped) {
 		return nil, errors.Wrap(err, "error reading directory")
 	}
 
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -60,7 +60,6 @@ type StreamingFile interface {
 type Directory interface {
 	Entry
 	Child(ctx context.Context, name string) (Entry, error)
-	Readdir(ctx context.Context) (Entries, error)
 	IterateEntries(ctx context.Context, cb func(context.Context, Entry) error) error
 }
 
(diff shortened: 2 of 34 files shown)


cli/command_ls.go
```diff
@@ -54,15 +54,14 @@ func (c *commandList) run(ctx context.Context, rep repo.Repository) error {
...
-	if err != nil {
-		return errors.Wrap(err, "error reading directory")
-	}
-
-	for _, e := range entries {
-		if err := c.printDirectoryEntry(ctx, e, prefix, indent); err != nil {
-			return errors.Wrap(err, "unable to print directory entry")
+	if err := d.IterateEntries(ctx, func(innerCtx context.Context, e fs.Entry) error {
+		if err2 := c.printDirectoryEntry(innerCtx, e, prefix, indent); err2 != nil {
+			return errors.Wrap(err2, "unable to print directory entry")
```
jkowalski, line 59: nit: you can probably ignore wrapping in this case

jkowalski:
#1955 is now merged.

ashmrtn:
`ReadDirAndFindChild` is simple enough to remove in this patch. It may be best to discuss removing the other things as I suspect some stuff is simpler with the helper functions and I'm not sure what the expected behavior for some other things is.

Things that are simpler with `IterateEntriesToReaddir`:
* `internal/diff/diff.go` - comparison of items across snapshots
* `snapshot/restore/restore.go` - completion tracking as iteration does not provide number of entries upfront

unclear expectations for behavior:
* `fs/loggingfs/loggingfs.go` - [timing of reading a directory](https://github.com/kopia/kopia/blob/9167c081830e47cc0591236dafd57423e50747e2/fs/loggingfs/loggingfs.go#L37) if we do iteration only as I believe we don't want to include callback durations
* `fs/cachefs/cache.go` - [getting entries from the cache](https://github.com/kopia/kopia/blob/9167c081830e47cc0591236dafd57423e50747e2/fs/cachefs/cache.go#L124), is the assumption that it's either all cached or all uncached, with no directories partly cached?

Additionally, there are four places where implementations of `IterateEntries` would be near duplicates of each other (similar to the original patch in #1957) because they use `ReaddirToIterate` internally and removing `fs.Directory.Readdir` means the helper is no longer usable.

If it's alright to leave `IterateEntriesToReaddir` and `fs.Entries` (though perhaps without the stipulation that it be sorted) and the behavior for `loggingfs` and `cachefs` is clarified then I can clean up the remaining stuff

ashmrtn:
looks like github was having issues running the tests last night around. May need bumped to restart them?

reported errors look like `Warning: Failed to download action 'https://api.github.com/repos/actions/setup-go/zipball/b22fbbc2921299758641fab08929b4ac52b32923'. Error: Response status code does not indicate success: 503 (Service Unavailable).`

ntolia:
> looks like github was having issues running the tests last night around. May need bumped to restart them?
> 
> reported errors look like `Warning: Failed to download action 'https://api.github.com/repos/actions/setup-go/zipball/b22fbbc2921299758641fab08929b4ac52b32923'. Error: Response status code does not indicate success: 503 (Service Unavailable).`

@ashmrtn I just restarted the failed jobs

cli/command_ls.go
```diff
@@ -54,15 +54,14 @@ func (c *commandList) run(ctx context.Context, rep repo.Repository) error {
...
-	entries, err := d.Readdir(ctx)
-	if err != nil {
-		return errors.Wrap(err, "error reading directory")
-	}
-
-	for _, e := range entries {
-		if err := c.printDirectoryEntry(ctx, e, prefix, indent); err != nil {
-			return errors.Wrap(err, "unable to print directory entry")
+	if err := d.IterateEntries(ctx, func(innerCtx context.Context, e fs.Entry) error {
+		if err2 := c.printDirectoryEntry(innerCtx, e, prefix, indent); err2 != nil {
```
jkowalski, line 58: nit: in such cases simple return c.printDirectoryEntry() is usually enough, you may need to silence linter, which is fine.

jkowalski:
this is great, i need a bit more time to review it since it's touching areas where we don't have too good testing.

fs/entry.go
```diff
@@ -79,50 +78,25 @@ type ErrorEntry interface {
...
+		if result == nil && e.Name() == name {
+			result = e
 		}
-	}
-
-	return nil
-}
-
-// IterateEntriesToReaddir is an adapter for a naive IterateEntries -> Readdir implementation.
-func IterateEntriesToReaddir(ctx context.Context, d Directory) (Entries, error) {
```
jkowalski, line 99: it might make sense to keep this helper since there are already a few copies, but I would remove `Entries` and replace with `[]Entry` wherever needed. Perhaps call it `GetAllEntries(ctx, dir) ([]Entry, error)`

fs/localfs/local_fs_test.go
```diff
@@ -17,6 +17,25 @@ import (
 	"github.com/kopia/kopia/internal/testutil"
 )
 
+func iterateEntriesToReaddir(ctx context.Context, dir fs.Directory) (fs.Entries, error) {
```
jkowalski, line 20: we should be able to remove this and restore the previous helper.

fs/entry.go
jkowalski: idea: would it help if this helper returned `map[string]Entry`?

internal/fusemount/fusefs.go
```diff
@@ -114,7 +114,7 @@ func (dir *fuseDirectoryNode) directory() fs.Directory {
 }
 
 func (dir *fuseDirectoryNode) Lookup(ctx context.Context, fileName string, out *fuse.EntryOut) (*gofusefs.Inode, syscall.Errno) {
-	entries, err := dir.directory().Readdir(ctx)
+	e, err := fs.IterateEntriesAndFindChild(ctx, dir.directory(), fileName)
```
jkowalski, line 117: can we call `Child()` here ?

internal/fusemount/fusefs.go
```diff
@@ -147,18 +146,18 @@ func (dir *fuseDirectoryNode) Lookup(ctx context.Context, fileName string, out *
...
 			Mode: entryToFuseMode(e),
 		})
+		return nil
+	})
+	if err != nil {
+		log(ctx).Errorf("error reading directory %v: %v", dir.entry.Name(), err)
+		return nil, syscall.EIO
 	}
 
 	return gofusefs.NewListDirStream(result), gofusefs.OK
```
jkowalski, line 163: This is effectively loading all into a slice and turning back into an iterator:

```
// DirStream lists directory entries.
type DirStream interface {
	// HasNext indicates if there are further entries. HasNext
	// might be called on already closed streams.
	HasNext() bool

	// Next retrieves the next entry. It is only called if HasNext
	// has previously returned true.  The Errno return may be used to
	// indicate I/O errors
	Next() (fuse.DirEntry, syscall.Errno)

	// Close releases resources related to this directory
	// stream.
	Close()
}
```

can you put a TODO to avoid slice here?

internal/webdavmount/webdavmount.go
```diff
@@ -110,27 +110,30 @@ type webdavDir struct {
...
 
-	if n > 0 && n < len(entries) {
-		entries = entries[0:n]
-	}
+	foundEntries := 0
 
-	var fis []os.FileInfo
+	err := d.entry.IterateEntries(d.ctx, func(innerCtx context.Context, e fs.Entry) error {
+		if n > 0 && n <= foundEntries {
+			return nil
```
jkowalski, line 119: nit: returning nil does not stop iteration, perhaps return some marker error `errStop` and treat it as success.

internal/webdavmount/webdavmount.go
```diff
@@ -110,27 +110,30 @@ type webdavDir struct {
 var symlinksAreUnsupportedLogged = new(int32)
 
 func (d *webdavDir) Readdir(n int) ([]os.FileInfo, error) {
```
jkowalski, line 112: btw, now that I'm looking at this function I think it's completely incorrect - it will truncate directory and never allow pagination, no need to fix it now, but can you put a TODO/file a bug to fix it?

internal/webdavmount/webdavmount.go
```diff
@@ -213,12 +216,13 @@ func (w *webdavFS) findEntry(ctx context.Context, path string) (fs.Entry, error)
 			return nil, errors.Errorf("%q not found in %q (not a directory)", p, strings.Join(parts[0:i], "/"))
 		}
 
-		entries, err := d.Readdir(ctx)
+		var err error
+
+		e, err = fs.IterateEntriesAndFindChild(ctx, d, p)
```
jkowalski, line 221: call Child() here?

snapshot/snapshotfs/objref.go
```diff
@@ -45,12 +45,11 @@ func GetNestedEntry(ctx context.Context, startingDir fs.Entry, pathElements []st
 			return nil, errors.Errorf("entry not found %q: parent is not a directory", part)
 		}
 
-		entries, err := dir.Readdir(ctx)
+		e, err := fs.IterateEntriesAndFindChild(ctx, dir, part)
```
jkowalski, line 48: call Child() here

internal/fusemount/fusefs.go
jkowalski: in general IterateEntriesAndFindChild() should only ever be used as an implementation helper for `Child()`

snapshot/snapshotfs/snapshot_tree_walker.go
```diff
@@ -106,21 +108,9 @@ func (w *TreeWalker) processDirEntry(ctx context.Context, dir fs.Directory, entr
...
-
-		if w.TooManyErrors() {
-			break
-		}
-
-		if w.alreadyProcessed(ent) {
-			continue
+	err := dir.IterateEntries(ctx, func(c context.Context, ent fs.Entry) error {
+		if w.TooManyErrors() || w.alreadyProcessed(ent) {
+			return nil
```
jkowalski, line 113: nit: may be worth doing the same trick with returning dummy error and treating it as success to stop the iteration early in case of TooManyErrors(). Probably does not matter much.

jkowalski:
Actually, approving right now, but please fix the suggestions and we can merge it.

jkowalski (approved):

fs/entry.go
jkowalski: yeah, let's get rid of `Entries` type and replace with `[]Entry` everywhere

jkowalski:
Thank you so much for doing this!

PR #1989, opened 2022-05-29
chore(ci): restrict GH actions for provider and stress tests to upstream repository.

jeremyje:
There's a lot of CI spam in my email because some of these tests occasionally fail. This limits these flows to run only on the upstream repository since they have a schedule. This also potentially saves some GH action cycles on everyone's respective accounts.


jeremyje:
https://github.com/kopia/kopia/issues/1990 is the created issue for this.

PR #1992, opened 2022-05-30
fix(cli): add retention to JSON output

### LINKED ISSUE #1987: Lack of retention tag in json output
First of all, I like to express my gratitude for your work on kopia, great piece of code!

I would like to take advantage of this message to ask you if the following is a bug or intended design:

kopia version: 0.10.7

kopia command: `kopia snapshot list --incomplete --show-identical --retention`

Output (truncated for brevety)

```
2022-05-22 06:08:12 CEST k49485a96e3a2a88940d8821c126dbeca 806.6 MB drwxr-xr-x files:11676 dirs:1000 (daily-7)
2022-05-23 05:28:53 CEST kbaf2ee2d824af8547f31c6c8e8df1566 804.4 MB drwxr-xr-x files:11703 dirs:1001 (daily-6)
(...)
```

However, the same command with --json flag added, does not show the retention tag (i.e., daily-x)

kopia command: `kopia snapshot list --incomplete --show-identical --retention --json`

Output (truncated for brevety)

```
[
  {
    "id": "d1876aac51503cad8076163c2d9627b9",
    "source": {
      "host": "nas",
      "userName": "root",
      "path": "/root"
    },
    "description": "",
    "startTime": "2022-05-22T06:08:12.756383737+02:00",
    "endTime": "2022-05-22T06:08:14.104913436+02:00",
    "rootEntry": {
      "name": "root",
      "type": "d",
      "mode": "0755",
      "mtime": "2022-05-22T06:08:10.784062142+02:00",
      "obj": "k49485a96e3a2a88940d8821c126dbeca",
      "summ": {
        "size": 806572104,
        "files": 11676,
        "symlinks": 0,
        "dirs": 1000,
        "maxTime": "2022-05-22T06:08:12.380056999+02:00",
        "numFailed": 0
      }
    }
  },

(...)
```

Thus, the lack of retention tag in json output is intentional?

In any case, thank you very much for your work in kopia.

julio-lopez:
Fixes #1987

### DIFF
--- a/cli/command_snapshot_list.go
+++ b/cli/command_snapshot_list.go
@@ -112,11 +112,6 @@ func findManifestIDs(ctx context.Context, rep repo.Repository, source string, ta
 }
 
 func (c *commandSnapshotList) run(ctx context.Context, rep repo.Repository) error {
-	var jl jsonList
-
-	jl.begin(&c.jo)
-	defer jl.end()
-
 	tags, err := getTags(c.snapshotListTags)
 	if err != nil {
 		return err
@@ -133,6 +128,16 @@ func (c *commandSnapshotList) run(ctx context.Context, rep repo.Repository) erro
 	}
 
 	if c.jo.jsonOutput {
+		type wrapManifest struct {
+			*snapshot.Manifest
+			RetentionReasons []string `json:"retentionReason,omitempty"`
+		}
+
+		var jl jsonList
+
+		jl.begin(&c.jo)
+		defer jl.end()
+
 		for _, snapshotGroup := range snapshot.GroupBySource(manifests) {
 			snapshotGroup = snapshot.SortByTime(snapshotGroup, c.reverseSort)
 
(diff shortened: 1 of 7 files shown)


julio-lopez:
@adowair PTAL

cli/command_snapshot_list.go
```diff
@@ -133,25 +128,52 @@ func (c *commandSnapshotList) run(ctx context.Context, rep repo.Repository) erro
...
-			if c.maxResultsPerPath > 0 && len(snapshotGroup) > c.maxResultsPerPath {
-				snapshotGroup = snapshotGroup[len(snapshotGroup)-c.maxResultsPerPath:]
-			}
+	return c.outputManifestGroups(ctx, rep, manifests, strings.Split(relPath, "/"))
+}
+
+// SnapshotManifest defines the JSON output for the CLI snapshot commands.
+type SnapshotManifest struct {
+	*snapshot.Manifest
+	RetentionReasons []string `json:"retentionReason,omitempty"`
```
adowair, line 140: I'm not sure I follow: why do we need an additional `RetentionReasons` here? It looks like the code below updates `Manifest.RetentionReasons`, and then copies it into `SnapshotManifest.RetentionReasons`. Could we just use the former directly? 
```diff
type Manifest struct {
  ...
-  RetentionReasons string[] `json:"-"`
+  RetentionReasons string[] `json:"retentionReason,omitempty"`

}
```
Is there anywhere else in the code that a `Manifest`'s JSON representation is used? Would such code break if we make the above change?

adowair (commented):
Thank you @julio-lopez, I have a question and a comment below

cli/command_snapshot_list.go
adowair: The answer is here #1987: https://github.com/kopia/kopia/issues/1987#issuecomment-1140706429

> where the retention is NOT stored on purpose, since it needs to be dynamically computed.

cli/command_snapshot_list_test.go
```diff
@@ -0,0 +1,50 @@
...
+	var snapshots []*cli.SnapshotManifest
+
+	testutil.MustParseJSONLines(t, e.RunAndExpectSuccess(t, "snapshot", "list",
+		"--json"), &snapshots)
+
+	require.Len(t, snapshots, 4)
+
+	for _, s := range snapshots {
+		require.NotEmpty(t, s.RetentionReasons, "expecting retention reason to be set")
+	}
```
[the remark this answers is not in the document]
julio-lopez: Punting to keep the change self-contained

cli/command_snapshot_list.go
julio-lopez: `snapshot.Manifest` is stored in the repo, that's how snapshots are tracked.
`snapshot.Manifest.RetentionReason` is explicitly excluded when the manifest is stored in the repo. The retention is dynamically computed based on policies, since the retention reason changes as time passes and new snapshots are added.

adowair: Aha that makes sense–manifests are stored in JSON format

adowair (approved):
Looks good to me 🏄

PR #1994, opened 2022-05-31
fix(general): use a fixed time for protecting newly created content

julio-lopez:
Previously, the reference time used to determine whether a content had been recently created would change throughout a snapshot GC execution. For long-running GC tasks, this non-deterministically shrank the safety window specified in `MinContentAgeSubjectToGC`.

Now, the maintenance starting time is used as a fix reference for the safety check.

### DIFF
--- a/snapshot/snapshotmaintenance/snapshotmaintenance_test.go
+++ b/snapshot/snapshotmaintenance/snapshotmaintenance_test.go
@@ -2,6 +2,7 @@ package snapshotmaintenance_test
 
 import (
 	"context"
+	"encoding/binary"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


julio-lopez:
Review commit-wise

snapshot/snapshotmaintenance/snapshotmaintenance_test.go
```diff
@@ -168,6 +169,73 @@ func (s *formatSpecificTestSuite) TestMaintenanceReuseDirManifest(t *testing.T)
 	t.Log("root info:", pretty.Sprint(info))
 }
 
+func TestSnapshotGCMinContentAgeSafety(t *testing.T) {
+	s := formatSpecificTestSuite{formatVersion: content.FormatVersion2}
+
+	s.TestSnapshotGCMinContentAgeSafety(t)
+}
+
```
julio-lopez, line 177: Only used to run this test in isolation, removing it.

```suggestion
```

PR #1996, opened 2022-06-01
feat(snapshots): Allow users to dynamically create entries in a directory during an upload

ashmrtn:
part of #1955 

Create `virtualfs.StreamingDirectory` that allows users to dynamically materialize entries in a directory. Combined with previous changes, this allows users to begin an upload without knowing ahead of time what all the entries in the file hierarchy will be

### DIFF
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -58,6 +58,8 @@ type Directory interface {
 	Entry
 	Child(ctx context.Context, name string) (Entry, error)
 	IterateEntries(ctx context.Context, cb func(context.Context, Entry) error) error
+	// MultipleIterations returns true if the Directory supports iterating through
+	// the entries multiple times. Otherwise it returns false.
 	MultipleIterations() bool
 }
 
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -58,10 +58,6 @@ type Directory interface {
 	Entry
 	Child(ctx context.Context, name string) (Entry, error)
 	IterateEntries(ctx context.Context, cb func(context.Context, Entry) error) error
-}
-
-// DirectoryWithIterationRestrictions allows callers to check if IterateEntries is safe to call multiple times.
-type DirectoryWithIterationRestrictions interface {
 	MultipleIterations() bool
 }
 
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -58,9 +58,9 @@ type Directory interface {
 	Entry
 	Child(ctx context.Context, name string) (Entry, error)
 	IterateEntries(ctx context.Context, cb func(context.Context, Entry) error) error
-	// MultipleIterations returns true if the Directory supports iterating through
-	// the entries multiple times. Otherwise it returns false.
-	MultipleIterations() bool
+	// SupportsMultipleIterations returns true if the Directory supports iterating
+	// through the entries multiple times. Otherwise it returns false.
+	SupportsMultipleIterations() bool
 }
 
 // DirectoryWithSummary is optionally implemented by Directory that provide summary.
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -60,6 +60,11 @@ type Directory interface {
 	IterateEntries(ctx context.Context, cb func(context.Context, Entry) error) error
 }
 
+// DirectoryWithIterationRestrictions allows callers to check if IterateEntries is safe to call multiple times.
+type DirectoryWithIterationRestrictions interface {
+	MultipleIterations() bool
+}
+
 // DirectoryWithSummary is optionally implemented by Directory that provide summary.
 type DirectoryWithSummary interface {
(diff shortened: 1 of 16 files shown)


ashmrtn:
I've switched to expanding the `fs.Directory` interface as this is the second time I've forgotten that `ignorefs` wraps `fs.Directory` implementations, which makes it seem error prone. If there's a better solution, I'd be happy to hear it and work towards implementing it

ashmrtn:
seems likely that previous test failures on arm were due to #2019

fs/entry.go
```diff
@@ -58,6 +58,7 @@ type Directory interface {
 	Entry
 	Child(ctx context.Context, name string) (Entry, error)
 	IterateEntries(ctx context.Context, cb func(context.Context, Entry) error) error
+	MultipleIterations() bool
```
josephaug26, line 61: A comment on what this means would be helpful

fs/virtualfs/virtualfs.go
```diff
@@ -94,6 +99,73 @@ func NewStaticDirectory(name string, entries []fs.Entry) fs.Directory {
...
+	defer sd.mu.Unlock()
+
+	if sd.callback == nil {
+		return nil, errIteratorAlreadyUsed
+	}
+
+	cb := sd.callback
+
+	if !sd.MultipleIterations() {
+		sd.callback = nil
```
josephaug26, line 130: why provide this flexibility i.e single or multiple pass of reading entries of a directory ? Since it is a Stream directory by not make it single pass read ? Just to avoid complexity. Even if we provide this flexibility here, can we provide the same for stream files and how can we enforce that multiple pass directory have only multiple pass file and visa versa

snapshot/snapshotfs/upload_test.go
```diff
@@ -692,6 +692,136 @@ func TestUpload_VirtualDirectoryWithStreamingFile(t *testing.T) {
...
+								return err
+							}
+						}
+
+						return nil
+					},
+					iterations,
+				),
+			})
+
```
josephaug26, line 733: In this test, where are we testing the multiple or single pass ? Looks like we are testing dynamic adding of a directory to root dir and adding a file to the sub dir. Do we need to have the for loop for multiple and single ReadDir/iternation ?

josephaug26 (commented):
This PR is about dynamically adding files and directory to virtual directory. what is the need/use-case to have flexibility of single readDir or multiple readDir for a virtual directory. IMHO It's a virtual/stream directory, and just like a virtual/stream file , the read (readDir) should be single pass.

fs/entry.go
ashmrtn: done

snapshot/snapshotfs/upload_test.go
ashmrtn: The test was intended to check both when the `StreamingDirectory` supports iterating multiple times and when it does not. The estimator in `uploader` also reads through all entries of a directory, so this was meant to check that it won't break. However, that test case has a race condition, so I've also added a separate test in `estimate_test.go`

fs/entry.go
```diff
@@ -58,6 +58,9 @@ type Directory interface {
 	Entry
 	Child(ctx context.Context, name string) (Entry, error)
 	IterateEntries(ctx context.Context, cb func(context.Context, Entry) error) error
+	// MultipleIterations returns true if the Directory supports iterating through
+	// the entries multiple times. Otherwise it returns false.
+	MultipleIterations() bool
```
sfc-gh-jkowalski, line 63: Can we call it SupportsMultipleIterations?

fs/virtualfs/virtualfs_test.go
```diff
@@ -109,3 +112,131 @@ func TestStreamingFileGetReader(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("error getting dir entries %v", err)
+	}
+
+	if len(entries) != 1 {
+		t.Errorf("expected directory with 1 entry, got %v", entries)
+	}
+
+	e := entries[0]
+	if e.Name() != testFileName {
```
sfc-gh-jkowalski, line 143: Try using require helpers. They generally produce very helpful failure outputs.

fs/virtualfs/virtualfs_test.go
```diff
@@ -109,3 +112,131 @@ func TestStreamingFileGetReader(t *testing.T) {
...
+		t.Fatalf("did not get expected filename: (actual) %v != %v (expected)", e.Name(), testFileName)
+	}
+
+	// Read and compare data
+	reader, err := f.GetReader(context.TODO())
```
sfc-gh-jkowalski, line 148: Instead of context.TODO you may use testlogging.Context(t)

jkowalski (approved):

PR #1960, opened 2022-05-21
refactor(repository): ensure we always parse content.ID and object.ID

jkowalski:
This changes the types to be incompatible with string to prevent direct
conversion to and from string.

This has the additional benefit of reducing number of memory allocations
and bytes for all IDs.

content.ID went from 2 allocations to 1:
   typical case 32 characters + 16 bytes per-string overhead
   worst-case 65 characters + 16 bytes per-string overhead
   now: 34 bytes

object.ID went from 2 allocations to 1:
   typical case 32 characters + 16 bytes per-string overhead
   worst-case 65 characters + 16 bytes per-string overhead
   now: 36 bytes

PR #2024, opened 2022-06-09
chore(repository): reduced memory allocations when logging content.ID

jkowalski:
This fixes regression introduced in #1960.

Tested by backing up Linux 5.14.8 source on M1 Mac (average of 15 runs):

Before: duration=6.7s avg_heap_objects=7411657 avg_heap_bytes=871794888
After: duration=5.6s (17% faster) avg_heap_objects=5947800 (20% less) avg_heap_bytes=795762120 (9% less)

### DIFF
--- a/repo/content/index/id.go
+++ b/repo/content/index/id.go
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/kopia/kopia/repo/hashing"
+	"github.com/kopia/kopia/repo/logging"
 )
 
 // IDPrefix represents a content ID prefix (empty string or single character between 'g' and 'z').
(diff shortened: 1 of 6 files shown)


repo/content/index/id.go
```diff
@@ -94,6 +96,45 @@ func (i ID) less(other ID) bool {
...
+	shdr := (*reflect.StringHeader)(unsafe.Pointer(&s)) //nolint:gosec
+	shdr.Data = uintptr(unsafe.Pointer(&l))             //nolint:gosec
+	shdr.Len = length
+
+	return
+}
+
+// ForLogging returns short-term LoggableID suitable for logging that avoids memory allocation.
+func (i ID) ForLogging() LoggableID {
+	var l LoggableID
```
julio-lopez, line 123: While this does not allocate a string, wouldn't this end up allocating a `LoggableID` since it is escaping the context. Is the compiler able to optimize it?

jkowalski: rewrote the PR to use explicit buffer pool and allocation free `Append()` pattern.

PR #2036, opened 2022-06-11
feat(snapshots): add option to ignore empty snapshots being saved

basldfalksjdf:


jkowalski:
I think this should be driven by a policy setting instead of flags

Also not sure if zero-byte snapshots are what folks in the forum were asking for. IIUC it  was about not saving snapshots that were identical to previous ones, not necessarily zero bytes. Identical snapshots can be detected by comparing root object IDs.

I suggest adding a new field to `RetentionPolicy`, something like `IgnoreIdenticalSnapshots *OptionalBool`.

It should be easy to plumb this through to make it settable in the CLI: simply search for all places where `KeepAnnual` is being used, there are 3 of them:

* `command_policy_set_retention.go` - needs a new flag to set the policy
* `command_polich_show.go` - needs to be printed
* `policy_tree.go` - needs to have a default value (false)

Once this is done, you need to add some test - I suggest adding couple lines to `TestSnapshotCreate()`

To make UI changes you would need to edit this file in another repository: to include new field:

https://github.com/kopia/htmlui/blob/main/src/PolicyEditor.js

basldfalksjdf:
> Also not sure if zero-byte snapshots are what folks in the forum were asking for. IIUC it was about not saving snapshots that were identical to previous ones, not necessarily zero bytes. Identical snapshots can be detected by comparing root object IDs.

The code is checking `TotalFileCount==0` which means no new file has changed.
 root object IDs are not identical even for identical snapshots.

jkowalski:
TotalFileCount is the number of uploaded files, it does not talk about other changes such as symlinks, new directories or directories moving around. It's most certainly not the right thing to check when deciding if a snapshot should be saved.

Looking at stats is not really a reliable way to compare snapshots, since they are not guaranteed to be content-addressable. The only reliable thing to use is object ID:

```
$ kopia snap ls -l -m
jarek@jareks-mbp:/Users/jarek/Projects/Kopia/snapshot
  2022-02-28 22:30:01 PST kb00ba64cdec7f017679994abbee092bd 0 B drwxr-xr-x manifest:abb91b4b8faa0fd61cddb6d659dba755 files:65 dirs:6 (monthly-4)
  2022-03-31 23:51:21 PDT k432e6b1855a7ada3d3a32550b373cc39 0 B drwxr-xr-x manifest:5a798abffbf239cd79f0c17b6caed9e8 files:66 dirs:6 (monthly-3)
  2022-04-11 22:20:00 PDT k432e6b1855a7ada3d3a32550b373cc39 0 B drwxr-xr-x manifest:98d9bd901646d14cf26b19510e7709dc files:66 dirs:6 (monthly-2)
  2022-05-13 23:50:01 PDT kcb6eb8ea3f356922d0476e250e3db8d7 0 B drwxr-xr-x manifest:2e0929ec1676434d8090314a3d846ab7 files:71 dirs:6 (daily-7)
  2022-05-14 21:50:01 PDT ka8ca672fb180237224232ab36f07fa3f 0 B drwxr-xr-x manifest:97099bb7909f2bd731b44b62d722cf38 files:70 dirs:6 (daily-6)
  2022-05-15 23:50:02 PDT k1d85c27d12377e659afa9c358fe3ade9 0 B drwxr-xr-x manifest:2b4cddded829d337f7a452e83da673e3 files:70 dirs:6 (daily-5,weekly-2)
  2022-05-16 23:50:03 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:ab40b6ec875da0353607804289fd375f files:70 dirs:6 (daily-4)
  2022-05-17 23:20:15 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:e503fc487c88e498b77c99ce91c2f5bd files:70 dirs:6 (daily-3)
  2022-05-18 21:40:12 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:8aab64fbcc9770aff2cea9e8e9c0d3ff files:70 dirs:6 (daily-2)
  2022-05-19 17:28:37 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:40b0be4b5aabd436c59e91391b9cb9a7 files:70 dirs:6 (latest-10)
  2022-05-19 17:30:00 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:2147174f68b62058a01474b886dd2af5 files:70 dirs:6 (latest-9)
  2022-05-19 17:40:02 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:635d26ebe82dbb04250f4cece76264bd files:70 dirs:6 (latest-8)
  2022-05-19 17:50:00 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:28f86da578a8e70e1628d8941ae9055b files:70 dirs:6 (latest-7,hourly-2)
  2022-05-19 18:00:16 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:0b23b04564ad0b09d9b5610f14b3168e files:70 dirs:6 (latest-6)
  2022-05-19 18:10:00 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:64b3f07f205abe99b20cf33d202f36b3 files:70 dirs:6 (latest-5)
  2022-05-19 18:20:02 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:18d397035e00593452906752bd37f40f files:70 dirs:6 (latest-4)
  2022-05-19 18:30:00 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:c5cea99c8f69c7ffcfae946bccc3d2eb files:70 dirs:6 (latest-3)
  2022-05-19 18:40:02 PDT k4fe12b9ba053fb9b413e51caed226c4a 0 B drwxr-xr-x manifest:094dd9d9f976847255ad70e9e42216f1 files:70 dirs:6 (latest-2)
```

As you can see, all the snapshots with root ID `k4fe12b9ba053fb9b413e51caed226c4a` are identical and that's guaranteed by the property of the [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree) used here.

The field to look at is `snapshot.Manifest.RootEntry.ObjectID` or in JSON: `rootEntry.obj`.

cli/command_policy_set_retention.go
```diff
@@ -24,10 +25,11 @@ func (c *policyRetentionFlags) setup(cmd *kingpin.CmdClause) {
 	cmd.Flag("keep-weekly", "Number of most-recent weekly backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepWeekly)
 	cmd.Flag("keep-monthly", "Number of most-recent monthly backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepMonthly)
 	cmd.Flag("keep-annual", "Number of most-recent annual backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepAnnual)
+	cmd.Flag("ignore-identical-snapshot", "Ignore identical snapshot with the last (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetIgnoreIdenticalSnapshots)
```
jkowalski, line 28: nit: remove Placeholder

cli/command_policy_set_retention.go
```diff
@@ -40,11 +42,25 @@ func (c *policyRetentionFlags) setRetentionPolicyFromFlags(ctx context.Context,
...
 	}
 
-	for _, c := range cases {
+	for _, c := range intCases {
 		if err := applyOptionalInt(ctx, c.desc, c.max, c.flagValue, changeCount); err != nil {
 			return err
 		}
 	}
 
+	boolCases := []struct {
```
jkowalski, line 51: the loop is not needed for now since it runs once, I would just call applyPolicyBoolPtr

cli/command_policy_show.go
```diff
@@ -142,6 +142,7 @@ func appendRetentionPolicyRows(rows []policyTableRow, p *policy.Policy, def *pol
 		policyTableRow{"  Daily snapshots:", valueOrNotSet(p.RetentionPolicy.KeepDaily), definitionPointToString(p.Target(), def.RetentionPolicy.KeepDaily)},
 		policyTableRow{"  Hourly snapshots:", valueOrNotSet(p.RetentionPolicy.KeepHourly), definitionPointToString(p.Target(), def.RetentionPolicy.KeepHourly)},
 		policyTableRow{"  Latest snapshots:", valueOrNotSet(p.RetentionPolicy.KeepLatest), definitionPointToString(p.Target(), def.RetentionPolicy.KeepLatest)},
+		policyTableRow{"  Ignore identical snapshot:", boolToString(p.RetentionPolicy.IgnoreIdenticalSnapshots.OrDefault(false)), definitionPointToString(p.Target(), def.RetentionPolicy.IgnoreIdenticalSnapshots)},
```
jkowalski, line 145: snapshot**s** perhaps?

cli/command_snapshot_create.go
```diff
@@ -329,6 +329,16 @@ func (c *commandSnapshotCreate) snapshotSingleSource(ctx context.Context, rep re
 		manifest.EndTime = endTimeOverride
 	}
 
+	ignoreIdenticalSnapshot := policyTree.EffectivePolicy().RetentionPolicy.IgnoreIdenticalSnapshots.OrDefault(false)
+	if ignoreIdenticalSnapshot {
+		for _, prev := range previous {
+			if prev.RootObjectID().String() == manifest.RootObjectID().String() {
+				log(ctx).Errorf("\n Not saving snapshot because no files have been changed")
+				return errors.Wrap(err, "Not saving snapshot because no files have been changed")
```
jkowalski, line 337: unclear if you want to return an error. I would simply log.Infof() and return success.

cli/command_snapshot_create.go
```diff
@@ -329,6 +329,16 @@ func (c *commandSnapshotCreate) snapshotSingleSource(ctx context.Context, rep re
 		manifest.EndTime = endTimeOverride
 	}
 
+	ignoreIdenticalSnapshot := policyTree.EffectivePolicy().RetentionPolicy.IgnoreIdenticalSnapshots.OrDefault(false)
+	if ignoreIdenticalSnapshot {
+		for _, prev := range previous {
+			if prev.RootObjectID().String() == manifest.RootObjectID().String() {
```
jkowalski, line 335: nit: you don't need to convert to strings, object IDs are comparable

internal/server/source_manager.go
```diff
@@ -400,6 +400,15 @@ func (s *sourceManager) snapshotInternal(ctx context.Context, ctrl uitask.Contro
 			return errors.Wrap(err, "upload error")
 		}
 
+		ignoreIdenticalSnapshot := policyTree.EffectivePolicy().RetentionPolicy.IgnoreIdenticalSnapshots.OrDefault(false)
+		if ignoreIdenticalSnapshot {
+			for _, prev := range manifestsSinceLastCompleteSnapshot {
+				if prev.RootObjectID().String() == manifest.RootObjectID().String() {
```
jkowalski, line 406: nit: you don't need to convert to strings, object IDs are comparable

internal/server/source_manager.go
```diff
@@ -400,6 +400,15 @@ func (s *sourceManager) snapshotInternal(ctx context.Context, ctrl uitask.Contro
...
+					return errors.Wrap(err, "Not saving snapshot because no files have been changed")
```
jkowalski, line 407: I think it's success

tests/end_to_end_test/snapshot_create_test.go
```diff
@@ -74,6 +74,18 @@ func TestSnapshotCreate(t *testing.T) {
 
 	sources := clitestutil.ListSnapshotsAndExpectSuccess(t, e)
 	require.Len(t, sources, 3)
+
+	// test ignore-identical-snapshot
+	e.RunAndExpectSuccess(t, "policy", "set", "--global", "--ignore-identical-snapshot", "true")
+	_, outErr := e.RunAndExpectSuccessWithErrOut(t, "snapshot", "create", sharedTestDataDir2)
```
jkowalski, line 80: i would expect this to succeed and compare list of snapshots before and after.

jkowalski: compare the list of blobs as well (`kopia blob ls` before and after), since we're building this feature to avoid creating new blobs when snapshots are identical.

jkowalski:
This is looking very good so far, i only have minor comments and we can merge it!

cli/command_snapshot_create.go
basldfalksjdf: Beyond the logging issue you mention, what do you think about changing this code so that the root ID of the current snapshot is only compared to the root ID of the most recent snapshot rather than the root ID of all previous snapshots? Not only would this increase performance, but logically it makes sense.

cli/command_policy_set_retention.go
```diff
@@ -40,11 +42,15 @@ func (c *policyRetentionFlags) setRetentionPolicyFromFlags(ctx context.Context,
...
 	}
 
-	for _, c := range cases {
+	for _, c := range intCases {
 		if err := applyOptionalInt(ctx, c.desc, c.max, c.flagValue, changeCount); err != nil {
 			return err
 		}
 	}
 
+	if err := applyPolicyBoolPtr(ctx, "ignore identical backups", &rp.IgnoreIdenticalSnapshots, c.policySetIgnoreIdenticalSnapshots, changeCount); err != nil {
```
jkowalski, line 51: replace `backups` with `snapshots`

cli/command_policy_set_retention.go
```diff
@@ -24,10 +25,11 @@ func (c *policyRetentionFlags) setup(cmd *kingpin.CmdClause) {
 	cmd.Flag("keep-weekly", "Number of most-recent weekly backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepWeekly)
 	cmd.Flag("keep-monthly", "Number of most-recent monthly backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepMonthly)
 	cmd.Flag("keep-annual", "Number of most-recent annual backups to keep per source (or 'inherit')").PlaceHolder("N").StringVar(&c.policySetKeepAnnual)
+	cmd.Flag("ignore-identical-snapshot", "Ignore identical snapshot with the last (or 'inherit')").StringVar(&c.policySetIgnoreIdenticalSnapshots)
```
jkowalski, line 28: one last nit: `ignore-identical-snapshots` and update help to something like "Do not save identical snapshots"

jkowalski (approved):

jkowalski:
LGTM, just last 2 minor string corrections.

basldfalksjdf:
I will update the documentation to reflect this new feature soon.

PR #2095, opened 2022-06-28
style(cli): rename restore `--sparse` to `--write-sparse-files`

adowair:
This breaking commit renames the sparse restore flag (used in
`kopia restore` and `kopia snapshot restore`) to conform more
with the naming precedents in the Kopia code.

The original motivation can be found here:
https://github.com/kopia/htmlui/pull/61#discussion_r899155054

### DIFF
--- a/snapshot/restore/local_fs_output.go
+++ b/snapshot/restore/local_fs_output.go
@@ -88,8 +88,8 @@ type FilesystemOutput struct {
 	// SkipTimes when set to true causes restore to skip restoring modification times.
 	SkipTimes bool `json:"skipTimes"`
 
-	// Sparse when set to true causes the restored files to be sparse.
-	Sparse bool `json:"sparse"`
+	// WriteSparseFiles when set to true, write contents as sparse files, minimizing allocated disk space.
+	WriteSparseFiles bool `json:"writeSparseFiles"`
 
 	// copier is the StreamCopier to use for copying the actual bit stream to output.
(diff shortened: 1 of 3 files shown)


snapshot/restore/local_fs_output.go
```diff
@@ -88,8 +88,8 @@ type FilesystemOutput struct {
 	// SkipTimes when set to true causes restore to skip restoring modification times.
 	SkipTimes bool `json:"skipTimes"`
 
-	// Sparse when set to true causes the restored files to be sparse.
-	Sparse bool `json:"sparse"`
```
jkowalski, line 92: so this will temporarily break the UI, you'll need to update the field name in separate PR to `htmlui` repo.

adowair: Gotcha: (https://github.com/kopia/htmlui/pull/76). I wonder if there is a foolproof way to sync these PRs–is `htmlui` imported as a submodule into `kopia`?

jkowalski: I already merged HTMLUI PR. Now if you do `go get github.com/kopia/htmluibuild@bf5411c` it should be possible to make an atomic commit that changes both UI and the server.

adowair: Gotcha, thanks for explaining that. I see you have this commit here https://github.com/kopia/kopia/commit/b8f68df2cf72f3b5249306e988c65fea56752adb doing that. Thanks!

PR #2128, opened 2022-07-03
docs(site): Cleaning up Release Notes

basldfalksjdf:
I think it will be more efficient if we just direct people to the GitHub releases page for anyone that wants to read release notes. I don't see the need to maintain the same information in two different places, plus this section of the site has not been updated for a while.

If this change is okay, I will delete all the individual pages for 

[0.9](https://kopia.io/docs/release-notes/v0.9/)
[0.8](https://kopia.io/docs/release-notes/v0.8/)
[0.7](https://kopia.io/docs/release-notes/v0.7/)
[0.6](https://kopia.io/docs/release-notes/v0.6/)
[0.5](https://kopia.io/docs/release-notes/v0.5/)
[0.4](https://kopia.io/docs/release-notes/v0.4/)
[0.3](https://kopia.io/docs/release-notes/v0.3/)


jkowalski:
Thank you, it's always been a pain to keep in sync!

jkowalski:
BTW, perhaps delete the entire section?

basldfalksjdf:
> BTW, perhaps delete the entire section?

I am completely fine with deleting the whole section. I actually originally wanted to delete the section as well and keep a link in the sidebar that automatically redirects to the GitHub releases page, but I did not know if adding a `index.html` with a refresh tag would do the trick or not. Maintaining a website on GitHub is a pain :-(

basldfalksjdf:
> > BTW, perhaps delete the entire section?
> 
> I am completely fine with deleting the whole section. I actually originally wanted to delete the section as well and keep a link in the sidebar that automatically redirects to the GitHub releases page, but I did not know if adding a `index.html` with a refresh tag would do the trick or not. Maintaining a website on GitHub is a pain :-(

Although, from a user experience perspective, it actually may not be a good idea to hard redirect from kopia.io to a github page, so maybe this current approach is the best of both worlds IF we want to keep the "Release Notes" link in the sidebar of the docs website. If we want to get rid of the link altogether, then that's fine too.

PR #2131, opened 2022-07-04
docs(site): Revised main homepage

basldfalksjdf:


site/content/_index.html
```diff
@@ -20,13 +20,17 @@
 	<a class="btn btn-lg btn-success mr-3 mb-4" href="https://slack.kopia.io">
 		Developer Discussion <i class="fab fa-slack ml-2 "></i>
 	</a>
-	<p class="font-weight-bold h3 mt-5">Encrypted, Deduplicated, and Compressed Data Backups Using Your Own Cloud Storage</p>
+	<p class="font-weight-bold h3 mt-5">End-to-End Zero-Knowledge Encryption</p>
```
jkowalski, line 23: i would cut down on that, that's too loud, otherwise looks good.

jkowalski: also, where is that background from?

basldfalksjdf: We do not have to use the background image. But if we do, we should find some place to credit https://unsplash.com/photos/oyXis2kALVg. The image is free for us to use and we don't need to credit per their license -- but it is the right thing to do.

Hit me up on Slack, I have a few images I am looking at, would be great to have a second opinion on which to use.

basldfalksjdf:
This is the credit of the current background image: https://unsplash.com/photos/pREq0ns_p_E

Here are other background images we could consider:

-https://unsplash.com/photos/xuTJZ7uD7PI
-https://unsplash.com/photos/oyXis2kALVg
-https://unsplash.com/photos/VT4rx775FT4
-https://unsplash.com/photos/bZZp1PmHI0E
-https://unsplash.com/photos/Q1p7bh3SHj8
-https://unsplash.com/photos/Ud4GcZW3rOY

site/content/_index.html
```diff
@@ -20,13 +20,17 @@
 	<a class="btn btn-lg btn-success mr-3 mb-4" href="https://slack.kopia.io">
 		Developer Discussion <i class="fab fa-slack ml-2 "></i>
 	</a>
-	<p class="font-weight-bold h3 mt-5">Encrypted, Deduplicated, and Compressed Data Backups Using Your Own Cloud Storage</p>
+	<p class="font-weight-bold h3 mt-5">Encryption</p>
```
jkowalski, line 23: can you revert this part, 5 bold lines look out of place, IMO

basldfalksjdf:
This is ready to merge, as long as you are okay with all the changes.

PR #2137, opened 2022-07-05
docs(site): Minor changes to Features page

basldfalksjdf:


site/content/docs/Features/_index.md
```diff
@@ -16,9 +16,9 @@ Kopia uses [content-addressable storage](https://en.wikipedia.org/wiki/Content-a
 
 * After moving or renaming even large files, Kopia can recognize that they have the same content and won't need to upload them again.
 
-* Multiple users or computers can share the same repository: if users share the same files, they are also uploaded only once.
+* Multiple users or computers can share the same repository: if different users have the same files, the files are uploaded only once to the repository thanks to deduplication. In other words, Kopia supports repository-level deduplication whereas many other software deduplicate files only within the same backup/snapshot.
```
adowair, line 19:
```suggestion
* Multiple users or computers can share the same repository: if different users have the same files, the files are uploaded only once as Kopia deduplicates content across the entire repository.
```

site/content/docs/Features/_index.md
```diff
@@ -6,7 +6,7 @@ weight: 1
 
 ### Backup Files and Directories using Snapshots
 
-Kopia creates snapshots of the files and directories you designate and uploads theses snapshots to cloud/network/local storage called a [repository](../repositories/). Snapshots are maintained as a set of historical point-in-time records based on policies that you define.
+Kopia creates snapshots of the files and directories you designate and uploads these [encrypted](#end-to-end-zero-knowledge-encryption) snapshots to cloud/network/local storage called a [repository](../repositories/). Snapshots are maintained as a set of historical point-in-time records based on policies that you define.
```
adowair, line 9:
```suggestion
Kopia creates snapshots of the files and directories you designate, then  [encrypts](#end-to-end-zero-knowledge-encryption) and uploads these snapshots to cloud/network/local storage called a [repository](../repositories/). Snapshots are maintained as a set of historical point-in-time records based on policies that you define.
```

site/content/docs/Features/_index.md
```diff
@@ -16,9 +16,9 @@ Kopia uses [content-addressable storage](https://en.wikipedia.org/wiki/Content-a
...
+* Multiple users or computers can share the same repository: if different users have the same files, the files are uploaded only once to the repository thanks to deduplication. In other words, Kopia supports repository-level deduplication whereas many other software deduplicate files only within the same backup/snapshot.
 
-> NOTE: There is currently no access control mechanism within repository: everybody with access to the repository can see everyone's data, so be sure you trust the other users if you share a repository with someone else.
+> NOTE: There is currently no access control mechanism when users/computers share the same repository -- each user/computer can see the backed up data of all the other users/computers, so be sure you trust the other users/computers if you share a repository with someone else. Note that this only applies when multiple users/computers know the repository password and are backing up to the same repository. People who do not know the password are not able to see your backed up data because Kopia uses [end-to-end 'zero knowledge' encryption](#end-to-end-zero-knowledge-encryption).
```
adowair, line 21:
```suggestion
> NOTE: There is currently no access control mechanism between users/computers that share a repository -- every user/computer with access to the repository can see the data of all the other users/computers, so be sure you trust the other users/computers if you share a repository with someone else.
```

adowair (approved):
Looks good @basldfalksjdf, some nitpicking below

site/content/docs/Features/_index.md
basldfalksjdf: So, I disagree a bit with this suggested change. 

I know the extra language about "Note that this only applies when multiple users/computers know the repository password..." is very clunky. I don't particularly like it either. But, the reason I added it is to avoid confusion for new users of Kopia who don't understand repositories or how Kopia works yet; this blurb is, after all, one of the first things a new users reads under when looking at the Features page. Someone new reads this blurb about other people having access to their data and they may misunderstand that as your data is accessible by anyone. It isn't. It is only accessible by people who share the repo because they know the repo/encryption password.

Perhaps a compromise is to remove this note from where it is currently, and put it somewhere else so that it isn't right there in new users' faces, most of whom this will never apply to. It is worth asking, do we really need the note? Isn't it obvious that two people who share a repository will have access to each other's data because they share the same password? 

Or maybe we should change the note to emphasis that each repo only allows one password, so when you share a repository with another user, you must also share your password with them, so make sure to trust the user? See my change and LMK what you think.

PR #2147, opened 2022-07-06
docs(site): Kopia compared to other tools

basldfalksjdf:


jkowalski:
I don't think comparisons like that belong on kopia.io if there's an external blog post that summarizes, we can certainly link to that, ideally where full comparison methodology is explained and can be critiqued.

basldfalksjdf:
> I don't think comparisons like that belong on kopia.io if there's an external blog post that summarizes it, we can certainly link to that, ideally where full comparison methodology is explained and can be critiqued.

Out of curiosity, may I ask why you feel that way? There are so many similar software out there, and people regularly ask about the difference between them all. Having a page on kopia.io addresses that problem really easily.

As for comparison methodology, most all of it is just reporting facts. The only two line items that could be debated are `Backup & Restore Speed` and `Easy to Download, Install, and Use for Home & Business Users`. These two line items can be removed.

Still, if you don't want to post this on kopia.io, that is fine. LMK.

Shrekster:
This probably does not belong on kopia.io for two reasons:

1. Preparing benchmarking toolchain and hosting the infrastructure required to run such tests is beyond the development and maintenance scope of kopia.io. This involves licensing, cost overhead and dedication to "fairness" of comparison with equivalent feature-set which is best done by a third party.
2. It is far more valuable to run latest kopia perf comparisons against a prior version of a kopia release itself to isolate perf regressions. Plus you want to run these perf benchmarking "workflows" (if you may) on a fixed cadence against Kopia itself to make it really valuable. While comparing Kopia against other similar tools can help motivate certain design choices, but spending resources for the same, as of now, will not be the best use of community resources.

basldfalksjdf:
> This probably does not belong on kopia.io for two reasons:
> 
> 1. Preparing benchmarking toolchain and hosting the infrastructure required to run such tests is beyond the development and maintenance scope of kopia.io. This involves licensing, cost overhead and dedication to "fairness" of comparison with equivalent feature-set which is best done by a third party.
> 2. It is far more valuable to run latest kopia perf comparisons against a prior version of a kopia release itself to isolate perf regressions. Plus you want to run these perf benchmarking "workflows" (if you may) on a fixed cadence against Kopia itself to make it really valuable. While comparing Kopia against other similar tools can help motivate certain design choices, but spending resources for the same, as of now, will not be the best use of community resources.
> 
> It is desired to post comparative analysis which could be reverified in the near future or after a "perf impacting" change with some sort of automation to help justify the need for that change. A "kopia vs others" comparison does not help with that in the current state, hence my reasoning above.

This (proposed) page is comparing the features between software. It is not comparing performance.

But, if people don't want this comparison here, that's fine.

I just don't see why we shouldn't try to market Kopia as much as possible. Highlighting Kopia features vs other software does just that. For example, Duplicacy has a comparison on their Github page. Restic does not do a comparison per se but rather lists out the features of other software on their Github page. This type of information helps users make informed decisions.

julio-lopez (changes requested):
Keep the comparison outside kopia.io.

Yes, other tools and projects have the comparison in their own documentation, that does not mean the Kopia project has to follow the same pattern, at least not now. Our primary goal is to make the project as good, robust and useful as it can be.

Putting the comparison together and writing down the corresponding documentation took considerable effort. We appreciate it. It would be good to have the comparison in a blog post, in the users forum or other location. And, in the end, that may have a larger impact, both for you and the project.

It would be valuable to provide guidance about the features kopia has and the use cases.  Provide information about when it makes sense to use it, and when it may not make sense, and instead use other tools.

(This mainly echoes some of the comments already presented above).

Shrekster:
> This (proposed) page is comparing the features between software. It is not comparing performance.

Sorry about the confusion, I misinterpreted the PR in terms of another request around perf. However, similar reasoning apply here too. Simply saying a feature exists in a software doesn't necessarily make it a better choice. Design choices are highly influenced by the use-case, resources available and the software stack a product is meant to be used on. There might be a common ground where we can bring all these tools together for the best apples-to-apples comparison, but then someone has to curate these outcomes on a regular basis. Thinking about the same at scale and the comparison matrix where you can repeat your last comparison will require constant effort (and investment).

adowair:
Thanks @basldfalksjdf for the effort

> I just don't see why we shouldn't try to market Kopia as much as possible.

I totally agree here, a blog post will likely have good impact here (inbound links, Google search and all). Consider that posting this comparison here will necessitate ongoing maintenance (keep tabs on Kopia features _and_ all other alternatives)–this is a cost to consider.

Issue #2162, opened 2022-07-08
[Regression] KopiaUI throws "Network Error" when restoring snapshot


basldfalksjdf:
![Kopia Network Error](https://user-images.githubusercontent.com/106879934/178014589-231a23af-6482-48b9-aeea-fd891f90803b.PNG)

![Network Error2](https://user-images.githubusercontent.com/106879934/178014594-cec41cf6-331e-448c-b529-c0aa6cba5a2f.PNG)

I tried restoring from multiple different repos. All throw the same error. I also tried closing/reopening KopiaUI. Same problem. (I actually am forced to close/reopen KopiaUI after the error, as the UI won't load for that particular repo anymore.) 

I am able to restore via CLI just fine. This issue did not exist previously in v0.10.7.

jkowalski:
Looks like something crashed. Can you try restoring from a command line and see if you get any more error details?

jkowalski:
Ok, I found an issue, will make a fix later tonight and will cut 0.11.3 over the weekend since it affects all filesystem restores in the UI (zips and tars are fine). We need to seriously step up our UI testing automation game, this should have been caught in presubmits.

basldfalksjdf:
> Looks like something crashed. Can you try restoring from a command line and see if you get any more error details?

To clarify, whatever is crashing is crashing immediately. Zero files or folders are restored when doing the full snapshot restore via the GUI. I am, however, able to selectively restore individual files and mount snapshots through the GUI just fine.

Restoring via the CLI throws no errors and is able to restore files/folders just fine. I should clarify that I stopped the restore from completing fully because I dont have the hard drive space right now, but it seems to work.

jkowalski:
this particular case is totally possible to automate, we simply did not have a test case for triggering restore from UI (we had individual file downloads, but not full restore).

PR #2163, opened 2022-07-08
fix(ui): fixed filesysystem restores triggered from UI

jkowalski:
Added comprehensive test for restore API which was previously completely uncovered.

Fixes https://github.com/kopia/kopia/issues/2162


internal/server/api_restore.go
```diff
@@ -51,11 +51,13 @@ func handleRestore(ctx context.Context, rc requestContext) (interface{}, *apiErr
 
 	switch {
 	case req.Filesystem != nil:
-		out := req.Filesystem
```
Shrekster, line 54: hmm, the shadow linter did not detect this. ?

jkowalski: yeah, this is weird. those linters are finicky.

PR #2164, opened 2022-07-08
feat(repository): upgrade lock coordinator URL

Shrekster:
This PR is a spin off from original PR https://github.com/kopia/kopia/pull/1818

This allows anyone who is placing an upgrade lock intent to also specify an optional coordinator URL. If the URL could be any resource on the network serving as a helper which can:

- block if trying to establish upgrade quorum
- return 200 status if we are ok to proceed with the upgrade
- return 409 status if someone else has locked for upgrade
- return an error if it is not OK to perform the upgrade


repo/content/content_upgrade_lock.go
```diff
@@ -122,11 +127,42 @@ func (l *UpgradeLock) totalDrainInterval() time.Duration {
 	return l.MaxPermittedClockDrift + 2*l.IODrainTimeout
 }
 
+func (l *UpgradeLock) checkCoordinatorLock(ctx context.Context) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", l.CoordinatorURL, http.NoBody)
```
jkowalski, line 131: would it make sense to send some identifying information here (say hash of username@hostname) so that the server knows who's connecting?

repo/content/content_upgrade_lock.go
```diff
@@ -122,11 +127,42 @@ func (l *UpgradeLock) totalDrainInterval() time.Duration {
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to check for coordinator lock with %q", l.CoordinatorURL)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return true, nil
+	case http.StatusLocked:
```
jkowalski, line 145: not sure here, 409 feels more appropriate, as 423 belongs to WebDAV.

repo/content/content_upgrade_lock.go
Shrekster: will do

repo/content/content_upgrade_lock.go
Shrekster: will change to 409

repo/content/content_upgrade_lock.go
Shrekster: used owner-id and process-id for now, see latest commit

repo/content/content_upgrade_lock.go
```diff
@@ -122,11 +130,59 @@ func (l *UpgradeLock) totalDrainInterval() time.Duration {
 	return l.MaxPermittedClockDrift + 2*l.IODrainTimeout
 }
 
+// CoordinatorLockOwnerInfo is the json boduy for coordinator-lock URL
```
adowair, line 133:
```suggestion
// CoordinatorLockOwnerInfo is the json body for coordinator-lock URL
```

Shrekster:
Moving to draft, we may not need the external coordinator as of now. Will keep this PR around for a little while in-case the need arises.

PR #2170, opened 2022-07-09
docs(site): Updating documentation

basldfalksjdf:
-Reworked repositories page
-Minor edits to Getting Started Guide
-Minor edits to Features page
-Minor edits to Docs homepage
-Minor edits to Github homepage
-Modified side/top navigations to make better use of both navigations

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Kopia supports saving your [encrypted](https://kopia.io/docs/features/#end-to-en
 * Google Cloud Storage
 * Any remote server or cloud storage that supports WebDAV
 * Any remote server or cloud storage that supports SFTP
-* Dropbox, OneDrive, Google Drive, and all cloud storage supported by Rclone
+* The cloud storages supported by Rclone
   * Requires you to download and setup Rclone in addition to Kopia, but after that Kopia manages/runs Rclone for you
   * Rclone support is experimental and all the cloud storage supported by Rclone has not been tested to work with Kopia; Kopia has been tested to work with Dropbox, OneDrive, and Google Drive through Rclone
 * Your local machine and any network-attached storage or server
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Kopia supports saving your [encrypted](https://kopia.io/docs/features/#end-to-en
 * Any remote server or cloud storage that supports **SFTP**
 * Some of the cloud storages supported by **Rclone**
   * Requires you to download and setup Rclone in addition to Kopia, but after that Kopia manages/runs Rclone for you
-  * Rclone support is experimental: all the cloud storages supported by Rclone have not been tested to work with Kopia, and some may not work with Kopia; Kopia has been tested to work with **Dropbox**, **OneDrive**, and **Google Drive** through Rclone
+  * Rclone support is experimental: not all the cloud storages supported by Rclone have been tested to work with Kopia, and some may not work with Kopia; Kopia has been tested to work with **Dropbox**, **OneDrive**, and **Google Drive** through Rclone
 * Your local machine and any network-attached storage or server
 * Your own server by setting up a [Kopia Repository Server](https://kopia.io/docs/repository-server/)
 
--- a/README.md
+++ b/README.md
@@ -27,15 +27,14 @@ Pick the Cloud Storage Provider You Want
 Kopia supports saving your [encrypted](https://kopia.io/docs/features/#end-to-end-zero-knowledge-encryption) and [compressed](https://kopia.io/docs/features/#compression) snapshots to all of the following [storage locations](https://kopia.io/docs/features/#save-snapshots-to-cloud-network-or-local-storage):
 
 * Amazon S3 and any cloud storage that is compatible with S3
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -24,19 +24,22 @@ When ready, head to the [installation](https://kopia.io/docs/installation/) page
 Pick the Cloud Storage Provider You Want
 ---
 
-Kopia supports saving your [encrypted](https://kopia.io/docs/features/#end-to-end-zero-knowledge-encryption) and [compressed](https://kopia.io/docs/features/#compression) snapshots to all of the following [cloud storage](https://kopia.io/docs/features/#save-snapshots-to-cloud-network-or-local-storage):
+Kopia supports saving your [encrypted](https://kopia.io/docs/features/#end-to-end-zero-knowledge-encryption) and [compressed](https://kopia.io/docs/features/#compression) snapshots to all of the following [storage locations](https://kopia.io/docs/features/#save-snapshots-to-cloud-network-or-local-storage):
 
-* Amazon S3 and any cloud storage that is compatible with S3 (e.g., Wasabi, IDrive E2, DigitalOcean Spaces, and more)
+* Amazon S3 and any cloud storage that is compatible with S3
+  * Including Alibaba Cloud, Amazon Lightsail, Backblaze B2, China Mobile Cloud, Cloudflare R2, Contabo, DigitalOcean Spaces, Dreamhost, Google Cloud Storage, IBM Cloud, IDrive E2, Linode, Mail.ru Cloud, MEGA.io S4, MinIO, Oracle Cloud Infrastructure, OVH, Scaleway, Storj, Synology C2, Tencent Cloud, Vultr, Wasabi, Yandex Cloud, and many more!
```
jkowalski, line 30: BTW some of them don't work, Storj for example. I would revert this change.

README.md
```diff
@@ -24,19 +24,22 @@ When ready, head to the [installation](https://kopia.io/docs/installation/) page
...
 * Azure Blob Storage
 * Backblaze B2
 * Google Cloud Storage
 * Any remote server or cloud storage that supports WebDAV
 * Any remote server or cloud storage that supports SFTP
-* All cloud storage supported by Rclone (requires you to download and setup Rclone in addition to Kopia)
+* Dropbox, OneDrive, Google Drive, and all cloud storage supported by Rclone
```
jkowalski, line 36: I would say "Other storage solutions supported by Rclone", since examples are already listed below.

jkowalski:
This is great. thank you, just couple suggestions for README.

README.md
basldfalksjdf: Do we know which of these are known not to work? All of these claim to be fully S3 compliant, so all should work with Kopia. I also cross-referenced this list with the Rclone list of S3 compliant storages that work through Rclone's S3 remote. Of the list, either I or someone else have confirmed Kopia's S3 repo to work with: Amazon Lightsail, Backblaze B2, Cloudflare R2, DigitalOcean Spaces, Google Cloud Storage, IDrive E2, Linode, MinIO, Oracle Cloud Infrastructure, OVH, Scaleway, Storj, Vultr, and Wasabi.

Do you happen to know when it was that Storj was not working? Perhaps they did not fully implement S3 compatibility in the past and now have, as Ive used Storj before and just tested it now as well. I am able to successfully create and restore a snapshot and also run quick and full maintenance. They do the whole distributed storage thing but they claim to give instant access to files, so it is not like Amazon Deep Glacier.

If you feel strongly about it, I will remove this list. However, the reason I included the list is (1) SEO and (2) there is an incremental benefit of telling new users that their favorite cloud storage (e.g., Wasabi) works with Kopia rather than just saying all S3-compatible storages work. Perhaps a compromise is only to list the ones that have been confirmed to work. 

LMK.

jkowalski: I prefer not to list any particular ones, as that implies we've tested this. 

The compatibility is about more than S3 protocol support, which they all tend to do to a certain extent, but given how different those backends are there's usually more to it so those providers tend to "bend" the protocol a bit to their needs. It's usually about subtle things like read-after-write and list-after-write consistency, server-assigned timestamps, prefix listing support (IIRC this is where Storj failed).

I think this belongs in the community section (i.e. forum). We can start a thread, pin it and folks can post their configuration for various provifders that have been tested.

README.md
```diff
@@ -33,7 +33,7 @@ Kopia supports saving your [encrypted](https://kopia.io/docs/features/#end-to-en
 * Google Cloud Storage
 * Any remote server or cloud storage that supports WebDAV
 * Any remote server or cloud storage that supports SFTP
-* Dropbox, OneDrive, Google Drive, and all cloud storage supported by Rclone
+* The cloud storages supported by Rclone
```
jkowalski, line 36: I would specifically mention that not all providers are supported. Some for example don't guarantee unique filenames, don't have server-assigned timestamps or have no basic consistency.

site/content/docs/Repositories/_index.md
```diff
@@ -24,7 +23,7 @@ Kopia allows you to save your [encrypted](../features/#end-to-end-zero-knowledge
 * Dropbox, OneDrive, Google Drive, and all cloud storage supported by [Rclone](#rclone) 
   * Rclone is a (free and open-source) third-party program that you must download and setup seperately before you can use it with Kopia
   * Once you setup Rclone, Kopia automatically manages and runs Rclone for you, so you do not need to do much beyond the initial setup, aside from enabling Rclone's self-update feature so that it stays up-to-date
-  * Kopia's Rclone support is experimental and all the cloud storage supported by Rclone has not been tested to work with Kopia; Kopia has been tested to work with Dropbox, OneDrive, and Google Drive through Rclone
+  * Kopia's Rclone support is experimental: all the cloud storage supported by Rclone has not been tested to work with Kopia, and some may not work with Kopia; Kopia has been tested to work with Dropbox, OneDrive, and Google Drive through Rclone
```
jkowalski, line 26: nit: "not all cloud storage ... has been tested" instead of "all the cloud storage ... has not been tested"

Issue #1067, opened 2021-05-13
Feature request: ransomware protection using S3-compatible Object Locking (immutability)


bbccdd:
"I'm more concerned about something crawling into the machine and finding the credentials to the repository somewhere in Kopia's config and using that to mass delete files." - @Neurrone

[A shared document has been created to facilitate a transparent discussion logging design choicese, tweaking to changes in Kopia and for drafting documentation.](https://docs.google.com/document/d/1GCymFXrP53fvVAQsEOnFzE584kgv1o2ZIvtveHQ0VHQ/edit?usp=sharing) Feel free support and **comment**. If you want **editing rights** to this document, please send me your gmail address on [slack](https://kopia.slack.com).

This issue is [based on a conversation on slack that took place on slack from May 8th 2021 onwards](https://kopia.slack.com/archives/CH7R8L2KD/p1620469738289000).

Related to #711, #743 and #913

User story (B2 specific, but please read the question S3-generic):
**As a** Kopia user weary of [randsomware ](https://www.forbes.com/sites/daveywinder/2021/05/02/ransomware-reality-shock-92-who-pay-dont-get-their-data-back/?sh=28a9c854e0c7)attacks on backups, 
**I would like to** use S3-compatible object locking 
**so that** files cannot be deleted by randsomware or other malicious actors (that do not have access to my (MFA-protected) S3-storage account). 

How would or how could Kopia best relate to a) default [B2 bucket retention policies](https://help.backblaze.com/hc/en-us/articles/1260804786409-Default-Bucket-Retention-Immutability-) (that lock a file for a set number of days) and b) to [lifecycle rules ](https://www.backblaze.com/b2/docs/lifecycle_rules.html)(that allow for deletion after a set number of days to save money on the storage bill)? 

The challenge I imagine is in cleaning up unreferenced data at some point. Perhaps a maintainence flag could be set with the amount of days after which the S3-storage provider will delete the file. Repackaging content that is still being referenced in to be deleted files would be trivial I imagine.

**## Without  using Object Locking /  immutability the following should be possible**
Note that it should already be possible to use one pair of application keys for backing up and restoring (without delete capability) and another for maintenance (that has delete abilities).

Question by Dickson Tan  7:04 AM May 9 2021
Thanks for confirming my hypothesis about permissions.
    I see the documentation mentions that maintenance is done occasionally. Does that mean that if its been more than an hour since Kopia was last invoked, it would try to run quick maintenance the next time I invoke Kopia?
[...]

PR #2160, opened 2022-07-08
docs(site): Add documentation about Ransomware Protection

PhracturedBlue:
This is a very preliminary take at documenting Kopia's Ransomware Protection capabilities

PR #2179, opened 2022-07-11
feat(repository): Implement retention time extension on S3 buckets using Object Locks

PhracturedBlue:
If enabled and the storage implementation supports it, automatically extend the retention time of all relevant (non deleted/hidden/log/...) blobs during a full maintenance run.

A new repository parameter is added to enable/disable this functionality (which will only work if Retention is supported and enabled in the repository)

This implements functionality as discussed in #1067


cli/command_repository_set_parameters.go
```diff
@@ -39,6 +40,7 @@ func (c *commandRepositorySetParameters) setup(svc appServices, parent commandPa
 	cmd.Flag("index-version", "Set version of index format used for writing").IntVar(&c.indexFormatVersion)
 	cmd.Flag("retention-mode", "Set the blob retention-mode for supported storage backends.").EnumVar(&c.retentionMode, "none", blob.Governance.String(), blob.Compliance.String())
 	cmd.Flag("retention-period", "Set the blob retention-period for supported storage backends.").DurationVar(&c.retentionPeriod)
+	cmd.Flag("retention-maintenance-extend", "Extend retention period as part of full maintenance.").BoolVar(&c.retentionExtend)
```
jkowalski, line 43: could this be stored in the `maintenance.Params` instead?

internal/blobtesting/map.go
```diff
@@ -192,6 +192,10 @@ func (s *mapStorage) FlushCaches(ctx context.Context) error {
 	return nil
 }
 
+func (s *mapStorage) ExtendBlobRetention(ctx context.Context, b blob.ID, opts blob.ExtendOptions) error {
+	return blob.ErrUnsupportedObjectLock
```
jkowalski, line 196: nit: perhaps introduce `type UnsupportedBlobRetention struct {}` with default implementation and embed everywhere instead of repeating the same method.

repo/maintenance/blob_retain.go
```diff
@@ -0,0 +1,97 @@
...
+
+	var prefixes []blob.ID
+
+	extendOpts := blob.ExtendOptions{
+		RetentionMode:   blobCfg.RetentionMode,
+		RetentionPeriod: blobCfg.RetentionPeriod,
+	}
+
+	if !opt.DryRun {
+		// start goroutines to delete blobs as they come.
```
jkowalski, line 48: nit: fix stale copy/paste comment

repo/maintenance/blob_retain.go
```diff
@@ -0,0 +1,97 @@
...
+
+	if opt.Parallel == 0 {
+		opt.Parallel = 16
+	}
+
+	const extendQueueSize = 100
+
+	extend := make(chan blob.Metadata, extendQueueSize)
+
+	var cnt, toExtend uint64
```
jkowalski, line 36: nit: to make linter happy, this can be all in one var block:

```
var (
  ctn, toExtend uint64
  eg errgroup.Group
  prefixes []blob.ID
)
```

repo/maintenance/blob_retain.go
```diff
@@ -0,0 +1,97 @@
...
+				return nil
+			})
+		}
+	}
+
+	for _, pfx := range repo.GetLockingStoragePrefixes() {
+		prefixes = append(prefixes, blob.ID(pfx))
+	}
+
+	// iterate unreferenced blobs and count them + optionally send to the channel to be deleted
```
jkowalski, line 69: nit: fix comment

repo/maintenance/blob_retain.go
```diff
@@ -0,0 +1,97 @@
...
+	log(ctx).Infof("Extending retention time for blobs...")
+
+	if err := blob.IterateAllPrefixesInParallel(ctx, opt.Parallel, rep.BlobStorage(), prefixes, func(bm blob.Metadata) error {
```
jkowalski, line 72: i think this logic needs to be slightly more nuanced - we want to extend retention on blobs that are alive, not the ones that happen to be in the repository. Wouldn't this prevent blobs from ever getting dedleted because of delayed blob gc?

repo/maintenance/blob_retain.go
```diff
@@ -0,0 +1,97 @@
...
+		if !opt.DryRun {
+			extend <- bm
```
jkowalski, line 74: If all goroutines finish because of errors, this will hang forever. I think delete has the same bug, btw.

jkowalski:
I did first pass, looks really good, but I would defer to @julio-lopez for final approval, since we spoke about this in the past and he was interested in getting this feature in.

repo/maintenance/blob_retain.go
PhracturedBlue: Unless I'm missing something, I think it is ok as is.  Since garbage collection only 'hides' blobs (which can be done while objects are still under retention), it should be safe to extend the retention on all blobs, even inactive ones.  We do retention extension after gabage-collection to skip extending just-collected blobs, but doing it the other way would only add one additional period to the blob's retention time.

repo/maintenance/blob_retain.go
PhracturedBlue: I hope I have resolved this concern now.  If not, please clarify the concern

cli/command_maintenance_set_test.go
```diff
@@ -0,0 +1,42 @@
...
+	e := testenv.NewCLITest(t, testenv.RepoFormatNotImportant, testenv.NewInProcRunner(t))
+	defer e.RunAndExpectSuccess(t, "repo", "disconnect")
+
+	var mi cli.MaintenanceInfo
+
+	e.RunAndExpectSuccess(t, "repo", "create", "filesystem", "--path", e.RepoDir)
+
+	testutil.MustParseJSONLines(t, e.RunAndExpectSuccess(t, "maintenance", "info", "--json"), &mi)
+
+	if mi.ExtendObjectLocks {
```
jkowalski, line 23: nit: require.False()

repo/maintenance/maintenance_params.go
```diff
@@ -23,6 +23,8 @@ type Params struct {
 	FullCycle  CycleParams `json:"full"`
 
 	LogRetention LogRetentionOptions `json:"logRetention"`
+
+	ExtendObjectLocks bool `json:"extendObjectLocks"`
```
jkowalski, line 27: nit: explicitly set this to `false` in DefaultParams and add a comment for documentation purposes.

repo/maintenance/blob_retain.go
```diff
@@ -19,72 +19,80 @@ type ExtendBlobRetentionTimeOptions struct {
...
 		}
 	}
 
+	// Convert prefixes from strong to BlobID.
 	for _, pfx := range repo.GetLockingStoragePrefixes() {
 		prefixes = append(prefixes, blob.ID(pfx))
 	}
 
-	// iterate unreferenced blobs and count them + optionally send to the channel to be deleted
+	// iterate all relevant (active, extendabe) blobs and count them + optionally send to the channel to be extended
```
jkowalski, line 75: nit: typo "extendabe", missing "l"

repo/maintenance/blob_retain_test.go
```diff
@@ -35,7 +35,6 @@ func (s *formatSpecificTestSuite) TestExtendBlobRetentionTime(t *testing.T) {
 			nro.BlockFormat.HMACSecret = testHMACSecret
 			nro.RetentionMode = blob.Governance
 			nro.RetentionPeriod = time.Hour * 24
-			nro.RetentionExtend = true
```
jkowalski: why this change?

repo/maintenance/blob_retain_test.go
```diff
@@ -85,8 +84,6 @@ func (s *formatSpecificTestSuite) TestExtendBlobRetentionTimeDisabled(t *testing
 			nro.BlockFormat.MasterKey = testMasterKey
 			nro.BlockFormat.Hash = blockFormatHash
 			nro.BlockFormat.HMACSecret = testHMACSecret
-			nro.RetentionMode = blob.Governance
```
jkowalski: why this change?

repo/maintenance/blob_retain.go
```diff
@@ -19,72 +19,80 @@ type ExtendBlobRetentionTimeOptions struct {
...
 	if !opt.DryRun {
-		// start goroutines to delete blobs as they come.
+		// start goroutines to extend blob retention as they come.
 		for i := 0; i < opt.Parallel; i++ {
 			eg.Go(func() error {
+				var failed []blob.ID
 				for bm := range extend {
 					if err := rep.BlobStorage().ExtendBlobRetention(ctx, bm.BlobID, extendOpts); err != nil {
-						return errors.Wrapf(err, "unable to extend blob %q", bm.BlobID)
+						failed = append(failed, bm.BlobID)
```
jkowalski, line 54: nit: i would just log the error as things happend and increment a counter, no need to buffer all those blob IDs.

repo/maintenance/blob_retain.go
```diff
@@ -19,72 +19,80 @@ type ExtendBlobRetentionTimeOptions struct {
...
 					}
 					curCnt := atomic.AddUint64(&cnt, 1)
 					if curCnt%100 == 0 {
 						log(ctx).Infof("  extended %v blobs", curCnt)
 					}
 				}
+				if len(failed) > 0 {
+					return errors.Errorf("Failed to extend blobs: %v", failed)
+				}
 				return nil
```
jkowalski, line 65: would it make sense to replace errgroup with regular sync.WaitGroup, since it does not return an error?

repo/maintenance/blob_retain.go
```diff
@@ -19,72 +19,80 @@ type ExtendBlobRetentionTimeOptions struct {
...
 func ExtendBlobRetentionTime(ctx context.Context, rep repo.DirectRepositoryWriter, opt ExtendBlobRetentionTimeOptions, safety SafetyParameters) (int, error) {
-	blobCfg := rep.BlobCfg()
-	if !blobCfg.IsRetentionEnabled() || !blobCfg.RetentionExtend {
-		// Blob retention is disabled
-		return 0, nil
-	}
+	const extendQueueSize = 100
+
+	var (
+		cnt, toExtend uint64
```
jkowalski, line 25: btw -`atomic` will panic on 32-bit platforms (which we sadly are not testing on CI/CD), since they require 64-bit alignment of 64-bit values, which this does not guarantee.

One way to do it is to allocate them, which guarantees at least 64-bit alignment:

```
cnt := new(uint64)
toExtend := new(uint64)
```

Alternatively `uint32` is usually sufficient for all "counting" applications, I would still allocate that just to be safe.

jkowalski (commented):
i have just minor comments, LGTM from my part, but please wait for Julio's approval.

repo/maintenance/blob_retain_test.go
PhracturedBlue: Governance without retention time is not really a useful configuration, and irrelevant to what is being tested here.  The main purpose of this test is to verify that things work ok if retention is disabled (even if the maintenance parameter is enabled).  The important thing is that `blobCfg.IsRetentionEnabled()` returns false, and this test is not intended to validate the functionality of that method.

repo/maintenance/maintenance_params.go
PhracturedBlue: I'm not sure what comment you are looking for?  The assignment seems to be self-evident to me.

PhracturedBlue:
@jkowalski Thank you for taking the time to review this, I appreciate the feedback.

jocutajar:
Hi folks, gentle reminder, let's not have this important improvement rot over the summer. Is @julio-lopez unto it @jkowalski ?

jan-glx:
Would an inexperienced  tester (me) help here?

jkowalski:
@PhracturedBlue can you rebase the PR against latest master branch

@julio-lopez anything remaining before this can be merged?

PhracturedBlue:
I have rebased and fixed the tests to be clean on the latest HEAD

bbccdd:
@julio-lopez do you have time to review this please? That action would be an much so appreciated step towards ransomware protection!

Raboo:
Perhaps someone else is qualified to review the code?

ntolia:
Note that I unblocked tests a few days ago and we did have a race-related test failure here (the failed Windows test might just be a flake). This needs to be investigated before merging/approval too.

julio-lopez:
Tagging @redgoat650 to review this. I'm removing myself from the critical path here, I won't be able to look at it in the upcoming weeks.

cli/command_maintenance_set.go
```diff
@@ -121,6 +124,22 @@ func (c *commandMaintenanceSet) setMaintenanceEnabledAndIntervalFromFlags(ctx co
...
+		p.ExtendObjectLocks = lastVal
+		*changed = true
+
+		if lastVal {
+			log(ctx).Info("Object Lock extension maintenance enabled.")
+		} else {
+			log(ctx).Info("Object Lock extension maintenance disabled.")
+		}
+	}
+}
```
redgoat650, line 141: There's a problematic situation here: If the full maintenance window interval (`FullCycle.Interval`) is longer than, equal to, or even slightly less than (it takes time to issue the extensions), the retention period (`blobcfg.RetentionPeriod`), the blobs will NOT be protected, and the user might not be aware of this.

One possible solution could be:
- Calculate an acceptable relationship between full maint interval and retention period (a trivial example would be retention period must be, say, twice as long as the maintenance interval).
- Refuse to set `ExtendObjectLocks` here the current interval and period settings do not satisfy the above relationship. Throw a noisy error instead.
- Similarly, if `ExtendObjectLocks` is enabled, refuse any changes to the maintenance interval that would result in a violation of the above relationship. Noisy error.
- And same for retention period changes - if `ExtendObjectLocks` is enabled, throw a noisy error if the change would violate it.

Basically prevent the system from getting into that unsafe state altogether by policing changes to these three settings. Best if we can prevent users from unwittingly shooting themselves in the foot.

redgoat650 (commented):
Thanks for the change! It mostly looks great, a few minor suggestions/questions here and there.

The one issue I'm worried about is the relationship between full maintenance interval and retention period. See comment below for details.

repo/maintenance/blob_retain_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if got, want := len(blobsBefore), 4; got != want {
+		t.Fatalf("unexpected number of blobs after writing: %v", blobsBefore)
+	}
+
+	st := env.RootStorage().(cache.Storage)
+
+	ta.Advance(7 * 24 * time.Hour)
+
+	if err = st.TouchBlob(ctx, blobsBefore[3].BlobID, time.Hour); err != nil {
```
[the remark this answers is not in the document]
PhracturedBlue: it is just the blob we just added (the 'hello world' one).  if more blobs are added, then the got != want would fail, but I agree that just using blobsBefore-1 would be better.

repo/maintenance/blob_retain.go
```diff
@@ -0,0 +1,115 @@
...
+
+			go func() {
+				defer wg.Done()
+
+				for bm := range extend {
+					if err = rep.BlobStorage().ExtendBlobRetention(ctx, bm.BlobID, extendOpts); err != nil {
+						log(ctx).Errorf("Failed to extend blob %v: %v", bm.BlobID, err)
+						atomic.AddUint32(failedCnt, 1)
+
+						continue
```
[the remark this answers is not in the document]
PhracturedBlue: This should not be possible  attempting to add RetentiopPeriod or RetentionMode to an unsupported BlobStorage will result in an error:
```
error <initializing|updating> repository: unable to write blobcfg blob: ... blob-retention: unsupported put-blob option
```

repo/locking_storage.go
```diff
@@ -0,0 +1,22 @@
...
+	// collect prefixes that need to be locked on put
+	for _, prefix := range content.PackBlobIDPrefixes {
+		prefixes = append(prefixes, string(prefix))
+	}
+
+	prefixes = append(prefixes, indexblob.V0IndexBlobPrefix, epoch.EpochManagerIndexUberPrefix, format.KopiaRepositoryBlobID,
+		format.KopiaBlobCfgBlobID)
+
+	return prefixes
+}
```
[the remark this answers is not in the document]
PhracturedBlue: I'm not sure how to get a list of all valid prefixes to build an exclude list.  I'm also concerned that it would cause user breakage.  I am not sure that a new blob would be properly tested with extension, resulting in creating a retained blob that must be removed (per kopia policy) but cannot  (due to retention rules) and not catching this until after the code has been deployed to production.
I'm open to ideas though.

PhracturedBlue:
I could use some help debugging the Race Detector results.  Clearly there is an issue here, but I don't understand what the logs are telling me.

PhracturedBlue:
> I could use some help debugging the Race Detector results. Clearly there is an issue here, but I don't understand what the logs are telling me.

I believe I have found and fixed the race condition now.  My local run of:
```go test -race ./repo/maintenance -run TestFormatV3/TestExtendBlobRetentionTime```
now passes

File: repo/locking_storage.go (hunk 19 above)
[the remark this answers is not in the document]
Comment (redgoat650) [reply repo/locking_storage.go:22]: I think this is a task for another day; I recognize this change here is just a decomposition of existing code for reuse. I think it would be safer, but we can revisit it later.

File: cli/command_maintenance_set_test.go [hunk 20]
```diff
@@ -0,0 +1,83 @@
...
+
+	require.True(t, mi.ExtendObjectLocks, "ExtendOjectLocks should be enabled.")
+	require.True(t, mi.FullCycle.Interval == 86340000000000, "maintenance-interval should be unchanged.")
+
+	// Cannot change retention_period when retention_perion-full_maintenance_interval < 24h
+	e.RunAndExpectFailure(t, "repository", "set-parameters", "--retention-period", "47h")
+	testutil.MustParseJSONLines(t, e.RunAndExpectSuccess(t, "repo", "status", "--json"), &rs)
+	require.True(t, rs.BlobRetention.RetentionPeriod == 172800000000000, "retention-interval should be unchanged.")
+
+	// Can change retention_period when retention_perion-full_maintenance_interval > 24h
```
redgoat650, line 79: copy paste typos here `perion` --> `period`

redgoat650 (approved):
This looks good to me, thanks for fixing the config issue + adding unit test! I'll mark approval ✅

redgoat650:
Left a review for docs in #2160 ☝️  It'd be great if we can coordinate getting that PR merged alongside this one.

ballestr:
I have been testing this patch for a few weeks now and everything seems to be working fine, the object locks are extended and no side effect observed.
Is there anything else I could do to help getting this PR merged? It would be really helpful for our Disaster Recovery plan 🙂

ntolia:
@redgoat650 What's needed to merge this? Anything we can help with? It's currently unclear to me whether the race detector test is related to this PR (I took a quick glance and see connection refused errors).

ashmrtn:
I'm also curious to know what's required to finish off this PR and get it in

ashmrtn (approved):
hi @PhracturedBlue, I have permissions in the kopia repository to get this merged so I'm going to do a few fixups since main has changed since it was opened and get this in. I've chatted with jkowalski and redgoat650 out of band and they're alright with me shepherding this. I've also tried this patch locally a bit and everything seems in order

PhracturedBlue:
Thank you.  please let me know if you need me to do anything

repo/blob/storage.go
```diff
@@ -41,7 +41,7 @@ var ErrUnsupportedPutBlobOption = errors.New("unsupported put-blob option")
 var ErrNotAVolume = errors.New("unsupported method, storage is not a volume")
 
 // ErrUnsupportedObjectLock is returned when attempting to use an Object Lock specific
-// function on a storage implementation that does not have the inteded functionality.
+// function on a storage implementation that does not have the intedend functionality.
```
PhracturedBlue, line 44: Still a typo.  should be 'intended'

jan-glx:
💙Can't wait for the release! Quick question: 

I have set up my repo using object lock with infinite retention time (backblaze S3). If I add/delete files now and change the default retention time to something reasonable after this change was merged, will it just work as desired or do i need to do something else as well?

PhracturedBlue:
I wasn't aware backblaze offered an infinite retention period.  I believe it is not possible to reduce the retention period on a file without closing your account (this is the whole point of object lock), so file with an infinite retention period will be retained indefinitely.  Of course in kopia, you can delete/add files regardless, and new files will have the new retention period, but you'll be paying for those old files disk space until you close your account.

jan-glx:
Oh yes, I misunderstood backblaze's instructions (lucky I doesn't need to delete my account to get rid of my to-be-deleted files now).  

So reading https://github.com/kopia/kopia/blob/f866e61e9be62b9739ddd9998d636d55ca8eb3bb/site/content/docs/Advanced/Ransomware%20Protection/_index.md left me wondering:
Will I be able to modify a existing kopia repo with `--retention-mode COMPLIANCE --retention-period <retention time>` ? 
Should `<retention time>` match Backblaze S3 bucket setting "Lifecycle settings"->"Keep prior versions for this number of days: ",?
Is there a way to `kopia maintenance set --extend-object-locks true` from the GUI?

PhracturedBlue:
I believe you should be able to use `kopia repository set-parameters --retention-mode COMPLIANCE --retention-period <retention time>` on an existing repository (if object-locks are already enabled)
Lifecycle settings determine how much time you have after a file is marked deleted to when it is actually deleted (i.e. how much time you get to recover from an attack).
The retention period is set and refreshed each time a file is changed or when running full-maintenance.  Since only undeleted files have their retention-time extended, once a file is marked as deleted, the retention-time must elapse before a file is irrevocably deleted.  The Lifecycle determines the minimum time before backblaze automatically deletes a file, but it does not prevent manual deletion of data.  so generally you need a Lifecycle set, but from a Ransomware perspective, it is less critical than the retention time, and can be shorter than retention-time as well.

You could also make Lifecycle longer than retention time to give yourself more ability to rollback a change, but without locking yourself into a long hold-time (and also losing the assurances that a bad-actor wouldn't delete it during that time).

The most common solution is probably to just set them the same and not think about it too much

bbccdd:
> The most common solution is probably to just set them the same and not think about it too much

Not sure what I love more about your reply, the extraordinary well wordedness or the TL;DR. 

Your reply can be placed in the documentation verbatim.

PhracturedBlue:
I added similar (though not verbatim) text to the documentation patch https://github.com/kopia/kopia/pull/2160

PR #2203, opened 2022-07-17
refactor(repository): moved upgrade lock intent from content to repo

jkowalski:
There are no real `content` dependencies here and JSON at the repo level can store the upgrade lock in exactly the same way.


Shrekster:
LGTM

Shrekster (approved):

julio-lopez (commented):
Makes sense

PR #2209, opened 2022-07-20
fix(cli): content sizes are already suffixed, so 'bytes' in the string is redundant.

stpr-dev:
Noticed that running maintenance produces logs like:
```text
...

GC found 32773 unused contents (12.7 GiB bytes)

...
```
Seems like the units are already suffixed appropriately, so the 'bytes' in the string is not required.


jkowalski (approved):
LGTM, Thanks!

PR #2210, opened 2022-07-20
docs(site): Updating storage class information

basldfalksjdf:


basldfalksjdf:
The order of the pages in the "Advanced Topics" menu is temporary. I plan on slowly working on these pages and reordering accordingly.

basldfalksjdf:
@jkowalski or anyone else knowledgeable with the underlying code: can you please confirm these two details are accurate:

-"Because Kopia’s full maintenance will usually compact p blobs by reading and re-writing them, you will not be able to run full maintenance when you store p blobs on a storage class that has delayed access."

*Will full maintenance just not run at all or is Kopia smart enough to skip the compacting of `p` blobs and run the rest of full maintenance?

-"Another approach that some providers allow is to set a default storage class for a bucket that is other than their standard. This approach, too, is independent from and is not managed by Kopia: you control this from within your cloud provider account. If you follow this approach, Kopia will automatically use the default storage class that you set for your bucket."

*Some software pass the `STANDARD` storage class when uploading to S3 or S3-compatible storage. So, even if you have a bucket with a different default storage class, `STANDARD` is basically forced. Does Kopia work this way? Or does Kopia allow the bucket to decide whatever is the standard default storage?*

jkowalski:
Sorry about stalling the review, as I've been thinking what to do about it.

I'm not sure we're ready to prominently advertise this feature in the current state given all the limitations - I consider it incomplete until we have a solution for Glacier restore and maintenance, otherwise currently casual users will try it and be disappointed.

In fact I would say for now the opposite: put a big, fat disclaimer: don't use non-standard storage classes unless you have really specific need and know what you're doing. I would also refrain from explaining step-by-step how to do it, since this will lead to copy/paste cargo culting - this feature in its current state is currently like exposed live wire - can be powerful, but not for everyone.

basldfalksjdf:
A few thoughts.

First, I did not add materially new information to kopia.io. I combined existing pages into one, and rewrote it for clarity. Information about storage classes has been on kopia.io for a while now. The pages can be deleted if you want to, but questions about storage classes are common, and in my opinion it is better to have a page that informs users rather than not.

Second, this revised page is not intended to, nor does it, advocate for use of archive/glacier classes. The intention is to inform users about Kopia's support for different storage classes, inform the user what those storage classes are and which ones Kopia supports, and how to utilize the storage classes. 

The revised page already warns users away from the archive/glacier classes, but I kept the warning inline with the warning that was there in prior versions of the page. Perhaps the warning about archive/glacier classes should be more prominent and people simply told to stay away (especially given now that Kopia does regular full maintenance automatically). I made some changes along those lines. LMK is that satisfies your concern.

I do not see why we should tell users to keep standard storage classes though -- Kopia should work just fine with cold storage that gives instant access, so I think people should take advantage of cold storage if they want to.

PR #2213, opened 2022-07-21
feat(providers): add AWS IAM as a credential provider for s3 storage

wlan0:
When kopia is used as a SDK for applications running inside AWS, IAM credentials cannot be directly leveraged. This PR addresses that need. Now, when running inside AWS environment, if instance role allows access to s3, then kopia will by default have s3 access (only via SDK). 

In the future, this can be done via CLI too.

### DIFF
--- a/repo/blob/s3/s3_storage.go
+++ b/repo/blob/s3/s3_storage.go
@@ -316,7 +316,25 @@ func New(ctx context.Context, opt *Options) (blob.Storage, error) {
 }
 
 func newStorage(ctx context.Context, opt *Options) (*s3Storage, error) {
-	return newStorageWithCredentials(ctx, credentials.NewStaticV4(opt.AccessKeyID, opt.SecretAccessKey, opt.SessionToken), opt)
+	creds := credentials.NewChainCredentials(
+		[]credentials.Provider{
+			&credentials.Static{
+				Value: credentials.Value{
+					AccessKeyID:     opt.AccessKeyID,
+					SecretAccessKey: opt.SecretAccessKey,
+					SessionToken:    opt.SessionToken,
+					SignerType:      credentials.SignatureV4,
+				},
+			},
+			&credentials.EnvAWS{},
+			&credentials.IAM{
+				Client: &http.Client{
+					Transport: http.DefaultTransport,
+				},
+			},
+		},
+	)
+	return newStorageWithCredentials(ctx, creds, opt)
 }
 
 func newStorageWithCredentials(ctx context.Context, creds *credentials.Credentials, opt *Options) (*s3Storage, error) {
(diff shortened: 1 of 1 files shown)


wlan0:
ping @jkowalski

jkowalski:
looks good, let's first merge this into `test/providers` so our test against real S3 can run there.

jkowalski:
Hey @wlan0, looks like there's been some regression, that reproduces on my macOS laptop. 

Can you see what's going on with this test:

```
$ go test ./tests/robustness/snapmeta/... -timeout 20s -v -count=1 -run TestS3Connect
```

It passes at `31c5bd94bc6975b74e9de26ea39bb40afd4db6a2` but fails at `0e62d64576d58c653bc79b709c4a8ad071e701bb` which is the next commit - basically your PR merged into main.

Interestingly enough the tests on Github Actions are not affected.

wlan0:
Hi @jkowalski, I'll reproduce it on my laptop and diagnose the issue. Gimme an hour or so

wlan0:
Hi @jkowalski! I was able to reproduce and fix the issue. Here is the problem - there is a bug in minio/minio-go/v7 where an error is ignored:

https://github.com/minio/minio-go/blob/a2b545423f4d967f656288e5c0a741041be4ded7/pkg/credentials/iam_aws.go#L292

When I clone this repo locally, and return err in that line, tests on my Mac test succeeds. 

prev:

```go
func getCredentials(client *http.Client, endpoint string) (ec2RoleCredRespBody, error) {
	if endpoint == "" {
		endpoint = defaultIAMRoleEndpoint
	}

	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
	token, _ := fetchIMDSToken(client, endpoint)

	// http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
	u, err := getIAMRoleURL(endpoint)
```

new:
```go
func getCredentials(client *http.Client, endpoint string) (ec2RoleCredRespBody, error) {
	if endpoint == "" {
		endpoint = defaultIAMRoleEndpoint
	}

	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
	token, err := fetchIMDSToken(client, endpoint)
	if err != nil {
	    return ec2RoleCredRespBody{}, err 
        }

	// http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
	u, err := getIAMRoleURL(endpoint)
```

Now, to answer the question of why only Mac and why not GitHub Actions. In Github Actions, the code does not reach IAM auth retrieval.

It could be because of some environment factors.  It is likely because `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are set when running tests in CI, but not when calling `go test ./...`. 

Is this env disparity likely?

PR #2220, opened 2022-07-23
feat(repository): added `required features` to the repository

jkowalski:
This is intended for future compatibility to be able to reliably
stop old kopia client from being able to open a repository when
the old code does not understand new `required feature`.

Required features are checked on startup and periodically using the
same method as upgrade lock, where they will return errors during blob
operations.

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -88,6 +88,8 @@ type appServices interface {
 	Stderr() io.Writer
 	stdin() io.Reader
 	onCtrlC(callback func())
+	onRepositoryFatalError(callback func(err error))
+	enableTestOnlyFlags() bool
 	EnvName(s string) string
 }
 
@@ -130,8 +132,9 @@ type App struct {
 	upgradeOwnerID      string
 	doNotWaitForUpgrade bool
 
-	currentAction   string
-	onExitCallbacks []func()
+	currentAction         string
+	onExitCallbacks       []func()
+	onFatalErrorCallbacks []func(err error)
 
 	// subcommands
 	blob        commandBlob
@@ -154,14 +157,21 @@ type App struct {
 	logs        commandLogs
 
 	// testability hooks
-	osExit         func(int) // allows replacing os.Exit() with custom code
-	stdinReader    io.Reader
-	stdoutWriter   io.Writer
-	stderrWriter   io.Writer
-	rootctx        context.Context // nolint:containedctx
-	loggerFactory  logging.LoggerFactory
-	simulatedCtrlC chan bool
-	envNamePrefix  string
+	testonlyIgnoreMissingRequiredFeatures bool
+
+	isInProcessTest bool
+	exitWithError   func(err error) // os.Exit() with 1 or 0 based on err
+	stdinReader     io.Reader
+	stdoutWriter    io.Writer
+	stderrWriter    io.Writer
+	rootctx         context.Context // nolint:containedctx
+	loggerFactory   logging.LoggerFactory
+	simulatedCtrlC  chan bool
+	envNamePrefix   string
+}
+
+func (c *App) enableTestOnlyFlags() bool {
+	return c.isInProcessTest || os.Getenv("KOPIA_TESTONLY_FLAGS") != ""
 }
 
 func (c *App) getProgress() *cliProgress {
@@ -226,7 +236,7 @@ func (c *App) setup(app *kingpin.Application) {
 
 	_ = app.Flag("help-full", "Show help for all commands, including hidden").Action(func(pc *kingpin.ParseContext) error {
 		_ = app.UsageForContextWithTemplate(pc, 0, kingpin.DefaultUsageTemplate)
-		os.Exit(0)
+		c.exitWithError(nil)
 		return nil
 	}).Bool()
 
@@ -247,6 +257,8 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("dump-allocator-stats", "Dump allocator stats at the end of execution.").Hidden().Envar(c.EnvName("KOPIA_DUMP_ALLOCATOR_STATS")).BoolVar(&c.dumpAllocatorStats)
 	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar(c.EnvName("KOPIA_REPO_UPGRADE_OWNER_ID")).StringVar(&c.upgradeOwnerID)
 	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress, instead exit with a message.").Hidden().Default("false").Envar(c.EnvName("KOPIA_REPO_UPGRADE_NO_BLOCK")).BoolVar(&c.doNotWaitForUpgrade)
+
+	app.Flag("ignore-missing-required-features", "Open repository despite missing features (VERY DANGEROUS, ONLY FOR TESTING)").Hidden().BoolVar(&c.testonlyIgnoreMissingRequiredFeatures)
 
 	c.observability.setup(c, app)
 
@@ -308,7 +320,13 @@ func NewApp() *App {
 		},
 
 		// testability hooks
-		osExit:       os.Exit,
+		exitWithError: func(err error) {
+			if err != nil {
+				os.Exit(1)
+			} else {
+				os.Exit(0)
+			}
+		},
 		stdoutWriter: colorable.NewColorableStdout(),
 		stderrWriter: colorable.NewColorableStderr(),
 		stdinReader:  os.Stdin,
@@ -474,13 +492,13 @@ func (c *App) runAppWithContext(command *kingpin.CmdClause, cb func(ctx context.
 	if err != nil {
 		// print error in red
 		log(ctx).Errorf("%v", err.Error())
-		c.osExit(1)
+		c.exitWithError(err)
 	}
 
 	if len(c.trackReleasable) > 0 {
 		if err := releasable.Verify(); err != nil {
 			log(ctx).Warnf("%v", err.Error())
-			c.osExit(1)
+			c.exitWithError(err)
 		}
 	}
 
@@ -573,7 +591,7 @@ To run this command despite the warning, set --advanced-commands=enabled
 
 `)
 
-		c.osExit(1)
+		c.exitWithError(errors.Errorf("advanced commands are disabled"))
 	}
 }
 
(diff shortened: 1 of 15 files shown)


adowair (commented):
Thank you @jkowalski , I'm catching up, so here are my initial comments

cli/app.go
```diff
@@ -244,6 +248,8 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("track-releasable", "Enable tracking of releasable resources.").Hidden().Envar(c.EnvName("KOPIA_TRACK_RELEASABLE")).StringsVar(&c.trackReleasable)
 	app.Flag("dump-allocator-stats", "Dump allocator stats at the end of execution.").Hidden().Envar(c.EnvName("KOPIA_DUMP_ALLOCATOR_STATS")).BoolVar(&c.dumpAllocatorStats)
 
+	app.Flag("testonly-ignore-missing-required-features", "Open repository despite missing features (VERY DANGEROUS, ONLY FOR TESTING)").Hidden().BoolVar(&c.testonlyIgnoreMissingRequiredFeatures)
```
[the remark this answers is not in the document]
adowair: Ah, I see it is "hidden" already with `.Hidden()`.

cli/app.go
```diff
@@ -152,7 +154,9 @@ type App struct {
 	logs        commandLogs
 
 	// testability hooks
-	osExit         func(int) // allows replacing os.Exit() with custom code
+	testonlyIgnoreMissingRequiredFeatures bool
+
+	exitWithError  func(err error) // os.Exit() with 1 or 0 based on err
```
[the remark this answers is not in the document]
jkowalski: some testability aspects were missing, just that

cli/command_repository_set_parameters.go
```diff
@@ -50,6 +55,10 @@ func (c *commandRepositorySetParameters) setup(svc appServices, parent commandPa
 	cmd.Flag("epoch-delete-parallelism", "Epoch delete parallelism").IntVar(&c.epochDeleteParallelism)
 	cmd.Flag("epoch-checkpoint-frequency", "Checkpoint frequency").IntVar(&c.epochCheckpointFrequency)
 
+	cmd.Flag("add-required-feature", "Add required feature which must be present to open the repository").Hidden().StringVar(&c.addRequiredFeature)
+	cmd.Flag("remove-required-feature", "Remove required feature").Hidden().StringVar(&c.removeRequiredFeature)
+	cmd.Flag("warn-on-missing-required-feature", "Only warn (instead of failing) when the required feature is missing").Hidden().BoolVar(&c.warnOnMissingRequiredFeature)
+
```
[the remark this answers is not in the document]
jkowalski: this is currently for testing only, we will add required features for real as the system evolves

cli/command_repository_set_parameters.go
```diff
@@ -156,6 +170,8 @@ func (c *commandRepositorySetParameters) run(ctx context.Context, rep repo.Direc
 	c.setIntParameter(ctx, c.epochDeleteParallelism, "epoch delete parallelism", &mp.EpochParameters.DeleteParallelism, &anyChange)
 	c.setIntParameter(ctx, c.epochCheckpointFrequency, "epoch checkpoint frequency", &mp.EpochParameters.FullCheckpointFrequency, &anyChange)
 
+	requiredFeatures = c.addRemoveUpdateRequiredFeatures(requiredFeatures, &anyChange)
```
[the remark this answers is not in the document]
jkowalski: this is the section where we change things, the one at the top is initialization of existing values, so makes sense to keep them separate

cli/app.go
```diff
@@ -248,6 +258,8 @@ func (c *App) setup(app *kingpin.Application) {
 	app.Flag("upgrade-owner-id", "Repository format upgrade owner-id.").Hidden().Envar(c.EnvName("KOPIA_REPO_UPGRADE_OWNER_ID")).StringVar(&c.upgradeOwnerID)
 	app.Flag("upgrade-no-block", "Do not block when repository format upgrade is in progress, instead exit with a message.").Hidden().Default("false").Envar(c.EnvName("KOPIA_REPO_UPGRADE_NO_BLOCK")).BoolVar(&c.doNotWaitForUpgrade)
 
+	app.Flag("ignore-missing-required-features", "Open repository despite missing features (VERY DANGEROUS, ONLY FOR TESTING)").Hidden().BoolVar(&c.testonlyIgnoreMissingRequiredFeatures)
```
Shrekster, line 261: Would anyone ever need to use this flag ? If this is test-only then how about removing it from CLI and setting up the test with a mutable global ?

cli/command_repository_set_parameters.go
```diff
@@ -50,6 +55,12 @@ func (c *commandRepositorySetParameters) setup(svc appServices, parent commandPa
 	cmd.Flag("epoch-delete-parallelism", "Epoch delete parallelism").IntVar(&c.epochDeleteParallelism)
 	cmd.Flag("epoch-checkpoint-frequency", "Checkpoint frequency").IntVar(&c.epochCheckpointFrequency)
 
+	if svc.enableTestOnlyFlags() {
+		cmd.Flag("add-required-feature", "Add required feature which must be present to open the repository").Hidden().StringVar(&c.addRequiredFeature)
+		cmd.Flag("remove-required-feature", "Remove required feature").Hidden().StringVar(&c.removeRequiredFeature)
+		cmd.Flag("warn-on-missing-required-feature", "Only warn (instead of failing) when the required feature is missing").Hidden().BoolVar(&c.warnOnMissingRequiredFeature)
```
Shrekster, line 61: What is the motivation to add the `warn-on-missing-required-feature` flag ?

cli/command_repository_set_parameters.go
```diff
@@ -50,6 +55,12 @@ func (c *commandRepositorySetParameters) setup(svc appServices, parent commandPa
 	cmd.Flag("epoch-delete-parallelism", "Epoch delete parallelism").IntVar(&c.epochDeleteParallelism)
 	cmd.Flag("epoch-checkpoint-frequency", "Checkpoint frequency").IntVar(&c.epochCheckpointFrequency)
 
+	if svc.enableTestOnlyFlags() {
+		cmd.Flag("add-required-feature", "Add required feature which must be present to open the repository").Hidden().StringVar(&c.addRequiredFeature)
+		cmd.Flag("remove-required-feature", "Remove required feature").Hidden().StringVar(&c.removeRequiredFeature)
```
Shrekster, line 60: How about `set-features <CSV-OF-FEATURES` versus `add/remove` ? `set` could be a simpler and less confusing CLI.

cli/command_repository_set_parameters_test.go
```diff
@@ -183,3 +184,48 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
 
 	env.RunAndExpectSuccess(t, "index", "epoch", "list")
 }
+
+func (s *formatSpecificTestSuite) TestRepositorySetParametersRequiredFeatures(t *testing.T) {
+	env := s.setupInMemoryRepo(t)
+
+	env.RunAndExpectSuccess(t, "repository", "status")
+	env.RunAndExpectSuccess(t, "repository", "set-parameters", "--add-required-feature", "no-such-feature")
+	env.RunAndExpectFailure(t, "repository", "status")
```
Shrekster, line 193: nit; Can we also verify the failure message to ensure it is not failing for unexpected reasons ?

cli/command_repository_set_parameters_test.go
```diff
@@ -183,3 +184,48 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
...
+	env.RunAndExpectSuccess(t, "repository", "set-parameters", "--remove-required-feature", "no-such-feature", "--ignore-missing-required-features")
+	env.RunAndExpectSuccess(t, "repository", "status")
+
+	// now require a feature but with a warning
+	env.RunAndExpectSuccess(t, "repository", "set-parameters", "--add-required-feature", "no-such-feature", "--warn-on-missing-required-feature")
+	env.RunAndExpectSuccess(t, "repository", "status")
+	env.RunAndExpectSuccess(t, "repository", "set-parameters", "--remove-required-feature", "no-such-feature")
+}
+
+func (s *formatSpecificTestSuite) TestRepositorySetParametersRequiredFeatures_ServerMode(t *testing.T) {
```
Shrekster, line 204: nice test, thanks!

repo/open.go
```diff
@@ -450,6 +501,11 @@ func upgradeLockMonitor(
 			return err
 		}
 
+		if err := handleMissingRequiredFeatures(ctx, ufb.repoConfig, false); err != nil {
```
Shrekster, line 504: Shouldn't `options.TestOnlyIgnoreMissingRequiredFeatures` be respected here as well ?

cli/command_repository_set_parameters.go
jkowalski: I was thinking this would produce some warning to existing clients before they get broken the hard way. Happy to discuss.

cli/command_repository_set_parameters.go
jkowalski: having a file introduces all kinds of problems, add/remove is a pattern we have everywhere else.

cli/command_repository_set_parameters_test.go
jkowalski: there's currently no mechanism for that, I want to refactor RunAndExpectFailure() to always require validation in separate PR.

repo/open.go
jkowalski: done, although not sure who's gonna see the warnings this way.

cli/app.go
jkowalski: added test-only condition

cli/config.go
```diff
@@ -73,6 +77,18 @@ func (c *App) optionsFromFlags(ctx context.Context) *repo.Options {
 		DisableInternalLog:  c.disableInternalLog,
 		UpgradeOwnerID:      c.upgradeOwnerID,
 		DoNotWaitForUpgrade: c.doNotWaitForUpgrade,
+
+		// when a fatal error is encountered in the repository, run all registered callbacks
+		// and exit the program.
+		OnFatalError: func(err error) {
+			for _, cb := range c.onFatalErrorCallbacks {
+				cb(err)
```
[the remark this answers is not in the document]
jkowalski: done

cli/command_repository_set_parameters.go
Shrekster: Was proposing comma separated value/string, doesn't have to be a file though. But this is not a deal breaker.

Shrekster:
LGTM, merging ...

PR #2223, opened 2022-07-23
feat(cli): improved safety of v1->v2 index format upgrade

jkowalski:
When upgrading from legacy to epoch manager-based index, we will write
an intentionally-corrupted index blob, such that old clients won't be
able to understand it when they read the repository index using legacy
format.

The error message emitted by very old clients is not great, but it's
safer to do that rather than corrupt the repository.

Note that this additional safety has a delay of up to 15 minutes
which is the time required for old clients to stop relying on index list
cache in case of very long-running snapshots, server or KopiaUI.

### DIFF
--- a/cli/command_repository_upgrade.go
+++ b/cli/command_repository_upgrade.go
@@ -269,6 +269,12 @@ func (c *commandRepositoryUpgrade) upgrade(ctx context.Context, rep repo.DirectR
 	if err := rep.SetParameters(ctx, mp, rep.BlobCfg()); err != nil {
 		return errors.Wrap(err, "error setting parameters")
 	}
+
+	// poison V0 index so that old readers won't be able to open it.
+	if err := content.WriteLegacyIndexPoisonBlob(ctx, rep.BlobStorage()); err != nil {
+		log(ctx).Errorf("unable to write legacy index poison blob: %v", err)
+	}
+
 	// we need to reopen the repository after this point
 
 	log(ctx).Infof("Repository indices have been upgraded.")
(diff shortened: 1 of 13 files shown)


cli/command_repository_upgrade.go
```diff
@@ -269,6 +269,12 @@ func (c *commandRepositoryUpgrade) upgrade(ctx context.Context, rep repo.DirectR
 	if err := rep.SetParameters(ctx, mp, rep.BlobCfg()); err != nil {
 		return errors.Wrap(err, "error setting parameters")
 	}
+
+	// poison V0 index so that old readers won't be able to open it.
+	if err := content.WriteLegacyIndexPoisonBlob(ctx, rep.BlobStorage()); err != nil {
```
Shrekster, line 274: @jkowalski planning to write compat tests to hit this and verify with older binaries ?

repo/content/index_blob_manager_v0.go
```diff
@@ -14,7 +14,12 @@ import (
 	"github.com/kopia/kopia/repo/logging"
 )
 
+// LegacyIndexBlobPrefix is the prefix for all legacy (v0) index blobs.
+const LegacyIndexBlobPrefix = "n"
+
 const (
+	legacyIndexPoisonBlobID = "n00000000000000000000000000000000-repository_unreadable_by_this_kopia_version_upgrade_required"
```
Shrekster, line 21: nit; hardcode "n" vs using `LegacyIndexBlobPrefix`

cli/command_repository_upgrade.go
Shrekster: looks like repo-tests have some coverage, LGTM

Shrekster:
no major concerns, taking in

PR #2235, opened 2022-07-27
refactor(repository): introduce interface for reading FormattingOptions

jkowalski:
Instead of passing static content.FormattingOptions (and caching it)
we now introduce an interface to provide its values.

This will allow the values to dynamically change at runtime in the
future to support cases like live migration.


julio-lopez (commented):
+1: LG 👍

PR #2252, opened 2022-07-31
Dockerfile based on Ubuntu

Yanik39:
i guess compatible with docker-publish.sh

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -1,24 +1,29 @@
-FROM gcr.io/distroless/static:nonroot
-ARG TARGETARCH
+FROM		ubuntu:latest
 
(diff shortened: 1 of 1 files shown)


jkowalski:
Awesome. I verified it all builds locally for all 3 architectures, unfortunately we don't have pre-commit check for that (we probably should, one that skips `--push`)

When this is merged we should see ubuntu-based unstable.

jkowalski:
The first image is out:

https://hub.docker.com/layers/kopia/kopia/kopia/20220731.0.100346/images/sha256-bcf9ff5fb2de28dbf1d9581039a6efeb58a517494cd299b6244708f49af8d356?context=explore

Thank you again!

Yanik39:
it is as unstable as you said. but example docker-compose.yml is fetching kopia:latest.
What to do?

PR #2253, opened 2022-07-31
feat(general): Added example docker-compose.yml

Yanik39:
Nginx reverse proxy
```
location / {
                grpc_pass grpcs://container_ip:51515;
}
```
Fusermount

A volume should be mounted with property :shared to the /tmp folder of the kopia container.
```
-v /mnt/kopia:/tmp:shared
```
So KopiaUI mounts would be browsable at /mnt/kopia/

Above is required to add some readme to docker-compose.yml works.

### DIFF
--- a/tools/docker/docker-compose.yml
+++ b/tools/docker/docker-compose.yml
@@ -1,0 +1,52 @@
+services:
+  kopia:
+      image: kopia/kopia:latest
+      container_name: Kopia
+      user: "0:0"
+      restart: "unless-stopped"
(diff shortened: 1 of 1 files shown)


jkowalski:
Thanks, could you add some explanation for how to use it in `site/content/docs/Installation/_index.md` ?

Yanik39:
> Thanks, could you add some explanation for how to use it in site/content/docs/Installation/_index.md ?

Done

PR #2270, opened 2022-08-03
feat(repository): error correction for contents implemented as encryption.Encryptor

pescuma:
This is an initial implementation of error correction codes for blobs. It is still far from complete, but I'm posting it here to gather feedback and see if you are interested in this feature.

Currently it is based in the [infectious lib](https://github.com/vivint/infectious) that uses [Reed-Solomon forward error correction](https://en.wikipedia.org/wiki/Reed%E2%80%93Solomon_error_correction) and [Berlekamp-Welch error correction algorithm](https://en.wikipedia.org/wiki/Berlekamp%E2%80%93Welch_algorithm) to detect errors. 

It is an initial implementation, but already works with the filesystem repository and protects all the blobs in the repository. It also has a few options on how to do its job.

What I still intend to implement:
1. Reed-Solomon + CRC32 as a different approach
2. The cli/gui
3. A benchmark command
4. Logging
5. Support for splitting the files as PAR does (not sure about this one)

Things that I don't intend to fix:
1. The config must be made on repository creation and can't change
2. You can't really use less than 1% of file duplication space. This is a limitation of the Reed-Solomon lib. It also gets slower when you go bellow 2%.
3. It only protects against bit flips inside the files. It does not protect against file size changes.

Questions that I have:
1. Currently it is based on the _sharded.Impl_ interface and writes files ending in _.ec_, so it does not work with all repository types. Would it be better to base it on the _blob.Storage_ interface and, in this case, name the blobs as something like _eXXXXX_?
4. Do you have interest in this feature? (I'll keep working on it if you think it is worth)

PS: This is the first code that I write in go, so probably there are a lot of mistakes. Any pointers would be highly apreciated.


basldfalksjdf:
Not a programmer, so take this with a grain of salt... but wouldn't it be better to leave such error correction to the cloud storage providers? My understanding is that implementing error correction is not a very straightforward procedure. For example, see the [discussion](https://github.com/restic/restic/issues/804) the good folks over at Restic had on this topic.

Perhaps local filesystem repos can benefit from error correction built into Kopia, but this feature may be adding too much complexity to Kopia -- and not necessarily the good type of complexity.

jkowalski:
I think this is great and we could really use this as an option for filesystem and possibly all blob storage. The same approach could benefit sftp and webdav (which are usually backed by filesystem). 

Many folks don't realize how common bitrot is in particular for large repositories and not-so-modern filesystems like zfs/btrfs.  I agree with @basldfalksjdf that we probably don't need to do this for cloud-based repositories.

I wonder if a wrapper approach would be feasible, the same way we treat throttling, sharding, readonly aspects. There are many more types of wrappers I can think of, so if we can figure out good user experience here, we might allow adding these individual aspects on top of any blob storage.

Your code looks pretty good, I have some ideas for how to structure it and perhaps tips for how to make it faster by avoiding some memory allocations. Before that, let's sync up on Slack .

pescuma:
@jkowalski I've updated the code to use the _Encryptor_ interface as we've talked about. I've also implemented the benchmark command. 

So that you have an idea about the velocity and file size groth:

- For a 10MB file:
```
     ECC                      Throughput     Throughput       Growth
                                Encoding       Decoding
-----------------------------------------------------------------------------
  0. RS-CRC32-2%               1.4 GiB/s      1.7 GiB/s        2% [10.2 MiB]
  1. RS-CRC32-1%               1.4 GiB/s      1.8 GiB/s        1% [10.1 MiB]
  2. RS-CRC32-5%               853 MiB/s      1.7 GiB/s        6% [10.6 MiB]
  3. RS-BW-FB-2%               2.3 GiB/s    850.3 MiB/s        2% [10.2 MiB]
  4. RS-BW-FB-5%               1.8 GiB/s      758 MiB/s        5% [10.5 MiB]
  5. RS-BW-FB-10%              1.2 GiB/s    611.5 MiB/s        9% [10.9 MiB]
  6. RS-CRC32-10%            442.5 MiB/s      1.2 GiB/s       11% [11.1 MiB]
  7. RS-BW-FB-1%               1.6 GiB/s      364 MiB/s        1% [10.1 MiB]
  8. RS-BW-FS-10%              1.2 GiB/s    142.7 MiB/s        9% [10.9 MiB]
  9. RS-BW-FS-1%               2.4 GiB/s    128.9 MiB/s        1% [10.1 MiB]
 10. RS-BW-FS-2%               1.3 GiB/s    123.5 MiB/s        2% [10.2 MiB]
 11. RS-BW-FS-5%               898 MiB/s    115.7 MiB/s        5% [10.5 MiB]
-----------------------------------------------------------------------------
```

- For a 1B file:
```
     ECC                      Throughput     Throughput       Growth
                                Encoding       Decoding
-----------------------------------------------------------------------------
  0. RS-CRC32-1%              35.4 KiB/s     61.1 KiB/s   206000% [2 KiB]
  1. RS-CRC32-2%              34.9 KiB/s    122.2 KiB/s   206000% [2 KiB]
  2. RS-CRC32-5%                24 KiB/s     91.7 KiB/s   310000% [3 KiB]
  3. RS-CRC32-10%             14.8 KiB/s    243.1 KiB/s   312400% [3.1 KiB]
  4. RS-BW-FS-2%              48.9 KiB/s      1.1 KiB/s   204800% [2 KiB]
  5. RS-BW-FS-5%              12.9 KiB/s        1 KiB/s   614400% [6 KiB]
  6. RS-BW-FS-10%              9.3 KiB/s        782 B/s   1228800% [12 KiB]
  7. RS-BW-FS-1%              21.2 KiB/s        373 B/s   204800% [2 KiB]
  8. RS-BW-FB-2%              26.1 KiB/s        124 B/s     3200% [33 B]
  9. RS-BW-FB-10%              5.1 KiB/s        110 B/s    19200% [193 B]
 10. RS-BW-FB-5%               8.5 KiB/s         72 B/s     9600% [97 B]
 11. RS-BW-FB-1%              19.4 KiB/s         44 B/s     3200% [33 B]
-----------------------------------------------------------------------------
```

- For a 1KB file:
```
     ECC                      Throughput     Throughput       Growth
                                Encoding       Decoding
-----------------------------------------------------------------------------
  0. RS-CRC32-2%             120.6 MiB/s    233.5 MiB/s      201% [3 KiB]
  1. RS-CRC32-1%              61.1 MiB/s    244.3 MiB/s      201% [3 KiB]
  2. RS-CRC32-5%              59.9 MiB/s    213.6 MiB/s      303% [4 KiB]
  3. RS-BW-FS-2%              80.2 MiB/s        2 MiB/s      200% [3 KiB]
  4. RS-BW-FS-5%              34.5 MiB/s      1.8 MiB/s      600% [7 KiB]
  5. RS-BW-FS-10%             16.7 MiB/s      1.5 MiB/s     1200% [13 KiB]
  6. RS-BW-FS-1%              70.4 MiB/s      672 KiB/s      200% [3 KiB]
  7. RS-BW-FB-5%               4.9 MiB/s     91.6 KiB/s        9% [1.1 KiB]
  8. RS-BW-FB-2%              12.7 MiB/s     67.1 KiB/s        3% [1 KiB]
  9. RS-BW-FB-10%                2 MiB/s     54.9 KiB/s       19% [1.2 KiB]
 10. RS-BW-FB-1%              17.8 MiB/s     26.1 KiB/s        3% [1 KiB]
 11. RS-CRC32-10%             45.5 MiB/s          0 B/s      306% [4.1 KiB]
-----------------------------------------------------------------------------
```

Let me know what you think. Please keep in mind that this is not ready yet to merge, because I've hard-coded the ecc config when creating a new Encryptor.

jkowalski:
This looks awesome. I would focus on having good unit tests and benchmarks for the `ecc` package and then we can merge it (without `format_provider.go`, so this can only be benchmarked but not used in a repository).

I need to spend some time understanding the options here. I would like to avoid throwing all kinds of questions at people where bad answers come with penalties (in terms of overhead and throughput) and good answers are hard to choose.

I'm wondering if there's a "good" or "mostly good" choice for everyone that balances speed and overhead. We can perhaps look at a histogram of file sizes across people's repositories and pick a good default that optimizes most common case.

For example my repo which holds a mix of all kinds of data (source code, software, photos, music and videos) has this distribution:

```
        0 between 0 B and 10 B (total 0 B)
    88653 between 10 B and 100 B (total 4.9 MB)
   490837 between 100 B and 1 KB (total 225 MB)
   348274 between 1 KB and 10 KB (total 1.2 GB)
   145886 between 10 KB and 100 KB (total 5 GB)
    84591 between 100 KB and 1 MB (total 31.9 GB)
   182328 between 1 MB and 10 MB (total 557.1 GB)
        0 between 10 MB and 100 MB (total 0 B)
```

Could we have a dynamic selection of algorithm that prefers throughput for very large contents, and switches to algorithm with little overhead for smaller ones or perhaps something in the middle?

Could we perhaps have a single per-repository flag to specify how much overheard to use for ECC (0%-10%, 0==no ECC) and we figure out the rest and dynamically apply the right algorithms based on content lengths?

repo/ecc/ecc_rs_crc.go
```diff
@@ -0,0 +1,215 @@
...
+	"encoding/binary"
+	"github.com/klauspost/reedsolomon"
+	"github.com/kopia/kopia/internal/gather"
+	"github.com/kopia/kopia/repo/encryption"
+	"github.com/pkg/errors"
+	"hash/crc32"
+)
+
+const (
+	RsCrc321pEccName  = "RS-CRC32-1%"
```
jkowalski, line 13: can we spell out ReedSolomon ?

repo/ecc/ecc_rs_bw.go
```diff
@@ -0,0 +1,186 @@
...
+	RsBwFs5pEccName  = "RS-BW-FS-5%"
+	RsBwFs10pEccName = "RS-BW-FS-10%"
+	RsBwFb1pEccName  = "RS-BW-FB-1%"
+	RsBwFb2pEccName  = "RS-BW-FB-2%"
+	RsBwFb5pEccName  = "RS-BW-FB-5%"
+	RsBwFb10pEccName = "RS-BW-FB-10%"
+)
+
+// RsBwEcc implements Reed-Solomon error codes with Berlekamp-Welch error detection
+type RsBwEcc struct {
```
jkowalski, line 22: can we spell out `BerlekampWelch` instead of BW?

repo/ecc/ecc_rs_bw.go
```diff
@@ -0,0 +1,186 @@
...
+	Options
+	FixedShareSize  bool
+	RequiredShares  int
+	RedundantShares int
+	Fec             *infectious.FEC
+}
+
+func NewRsBwEcc(opts *Options, fixedShareSize bool, spaceUsedPercentage float32) (*RsBwEcc, error) {
```
jkowalski, line 30: also, Golang has slightly different convention for acronyms than say Java, so `ECC` would be all caps.

Perhaps function name would could be `NewReedSolomonBerlekampWelchECC` ? It's not going to be called in more than one or two places so the length does not matter much.

repo/ecc/ecc_rs_bw.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+
+	output.Append(dataBytes[:originalSize])
+
+	return nil
+}
+
+func (r *RsBwEcc) Overhead() int {
+	return 0
```
jkowalski, line 157: this should panic() to ensure we don't call it. We need to refactor this and fix names anyway in separate PR.

repo/ecc/ecc_rs_bw.go
```diff
@@ -0,0 +1,186 @@
...
+	RsBwFb10pEccName = "RS-BW-FB-10%"
+)
+
+// RsBwEcc implements Reed-Solomon error codes with Berlekamp-Welch error detection
+type RsBwEcc struct {
+	Options
+	FixedShareSize  bool
+	RequiredShares  int
+	RedundantShares int
+	Fec             *infectious.FEC
```
jkowalski, line 27: nit: FEC

repo/ecc/ecc_rs_bw.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	originalSize = length - r.RedundantShares*shareSize*blocks
+	return
+}
+
+func (r *RsBwEcc) Encrypt(input gather.Bytes, contentID []byte, output *gather.WriteBuffer) error {
+	blocks, shareSize, originalSize := r.ComputeSizesFromOriginal(input.Length())
+	requiredSize := r.RequiredShares * shareSize
+
+	var inputBuffer gather.WriteBuffer
```
jkowalski, line 74: perhaps add comments here to explain what's going on at each step?

repo/ecc/ecc_rs_bw.go
```diff
@@ -0,0 +1,186 @@
...
+
+	return nil
+}
+
+func (r *RsBwEcc) Decrypt(input gather.Bytes, contentID []byte, output *gather.WriteBuffer) error {
+	blocks, shareSize, originalSize := r.ComputeSizesFromStored(input.Length())
+	requiredSize := r.RequiredShares * shareSize
+	redundantSize := r.RedundantShares * shareSize
+
+	var inputBuffer gather.WriteBuffer
```
jkowalski, line 111: same here, could use some comments

repo/ecc/ecc_rs_bw.go
```diff
@@ -0,0 +1,186 @@
...
+	copied := input.AppendToSlice(inputBytes[:0])
+
+	if len(copied) < len(inputBytes) {
+		clear(inputBytes[len(copied):])
+	}
+
+	eccBytes := inputBytes[:redundantSize*blocks]
+	dataBytes := inputBytes[redundantSize*blocks:]
+
+	shares := make([]infectious.Share, r.RequiredShares+r.RedundantShares)
```
jkowalski, line 125: Sometimes (not always, though), it may make sense to use arrays on the stack (if you know max capacity), instead of `make()`. This can avoid memory allocations.

I'm guessing the reason for poor throughput for small files is memory allocations:
You can memory allocations using kopia quickly by using:

```
$ kopia <some-subcommand-and-flags> --profile-memory 1 --profile-dir /tmp/profile1
$ go tool pprof -http :9000 -alloc_objects -lines /tmp/profile1/mem.pprof
```

There are similar flags for profiling CPU, mutexes and a bunch more. Golang is awesome.

repo/ecc/ecc_rs_bw.go
pescuma: I had to implement it because `content.SharedManager.setupReadManagerCaches` calls it:
```
	sm.committedContents = newCommittedContentIndex(caching,
		uint32(sm.format.Encryptor().Overhead()),
		sm.format.WriteIndexVersion(),
		sm.enc.getEncryptedBlob,
		sm.namedLogger("committed-content-index"),
		caching.MinIndexSweepAge.DurationOrDefault(DefaultIndexCacheSweepAge))
```

repo/ecc/ecc_rs_bw.go
pescuma: That is nice to know, thanks. I had not realized that `make` allocates in the heap and just creating the array allocates in the stack. I'll play with it.

pescuma:
@jkowalski I've fixed some minor issues. I've also deleted the rs_bw variant, because it was always worse.

> This looks awesome. I would focus on having good unit tests and benchmarks for the `ecc` package and then we can merge it (without `format_provider.go`, so this can only be benchmarked but not used in a repository).
> 
> I need to spend some time understanding the options here. I would like to avoid throwing all kinds of questions at people where bad answers come with penalties (in terms of overhead and throughput) and good answers are hard to choose.
> 
> I'm wondering if there's a "good" or "mostly good" choice for everyone that balances speed and overhead. We can perhaps look at a histogram of file sizes across people's repositories and pick a good default that optimizes most common case.
> 
> For example my repo which holds a mix of all kinds of data (source code, software, photos, music and videos) has this distribution:
> 
> ```
>         0 between 0 B and 10 B (total 0 B)
>     88653 between 10 B and 100 B (total 4.9 MB)
>    490837 between 100 B and 1 KB (total 225 MB)
>    348274 between 1 KB and 10 KB (total 1.2 GB)
>    145886 between 10 KB and 100 KB (total 5 GB)
>     84591 between 100 KB and 1 MB (total 31.9 GB)
>    182328 between 1 MB and 10 MB (total 557.1 GB)
>         0 between 10 MB and 100 MB (total 0 B)
> ```
> 
> Could we have a dynamic selection of algorithm that prefers throughput for very large contents, and switches to algorithm with little overhead for smaller ones or perhaps something in the middle?

As you said, the biggest problems here are throughput and space usage. I'll explain how the algorithm works so I can put some considerations latter:

The idea of the Reed-Solomon algo is to recompute data that have missing pieces. To do that, it uses parity shards. So, you have to split your data in `D` data shards (all must have the same size) and then compute `P` parity shards (also the same size). `D` and `P` can be arbitrary numbers, but the library that's being used here requires `D+P<=256`. Also, it's recommended that `P>=2`. From this group of data shards + parity shards, you can lose up to `P` shards (of any kind). One more thing, as all shards must have the same size, usually you have to add padding to the data so it can be of the size `D * shard size`.

The second problem is how do we know that we lost some data? To compute that in this case I'm computing a CRC32 over the shard, and I have to do that to all `D+P` shards. 

(The Berlekamp-Welch allows to compute what have changed without CRC, but it only works with up to `P/2` shards lost. It is also slower.)

So, how much data can we lose? If we have 1 bit flip in a block, we lost the block. But in this case it does not matter if all the bits are flipped, we have lost the block all the same. Also, 1 bit flip inside the CRC also loses the block. 

Let's say now that we decide to use 2% more space to compute parity. Currently, the code in this case selects `D=128` and `P=2`. The thing is: if we split all our data in 130 shards (and have big shards), we allow in the worst case just 2 bit flips. But we can do better: we can divide the file in blocks, and than divide these blocks in shards. This way, we can lose 2 bits (worst case) per block. Wonderful, let's have our shard size be 1 byte! But here we hit another problem: we need to use 4 bytes per shard to store the CRC, and if the shards begin to be too small, that begins to use too much space. 

So that we can have some numbers, here is a benchmark changing the shard size to see how that affects the space used: (this uses an input that is 1mb, 128+2 shards, what should increase around 2% the size of the file)
```
     ECC                                Throughput     Throughput       Growth
                                          Encoding       Decoding         
-----------------------------------------------------------------------------
  0. REED-SOLOMON-CRC32-512b               2 GiB/s      2.9 GiB/s        2% [1 MiB]
  1. REED-SOLOMON-CRC32-1024b            1.6 GiB/s      2.8 GiB/s        2% [1 MiB]
  2. REED-SOLOMON-CRC32-256b             1.5 GiB/s      2.3 GiB/s        3% [1 MiB]
  3. REED-SOLOMON-CRC32-128b               1 GiB/s      1.6 GiB/s        5% [1 MiB]
  4. REED-SOLOMON-CRC32-64b            715.1 MiB/s        1 GiB/s        8% [1.1 MiB]
  5. REED-SOLOMON-CRC32-32b              151 MiB/s    432.6 MiB/s       14% [1.1 MiB]
-----------------------------------------------------------------------------
```

On the other side, if the shard becomes too big, for small files the parity over the padding starts using too much space. So, the next idea would be to make the shard size depend on the file size. The problem with that is that, when the shard size changes to a smaller one, the final file size becomes smaller. Let's say that if the file has up to 16kb, we use shard size 32, after that we use 64. The total size of the resulting file would be:

```
input size    shard size     blocks   resulting size 
16381          32            4        18717   
16382          32            4        18718   
16383          32            4        18719   
16384          64            2        17680   
```

And that would not work because when correcting I need to figure out the block size and the shard size based on the output file size, so it must monotonically grow. 

Having said all that, to try to fix this problem I've created 3 different approaches depending of file size:
1. For very small files just use 6 data shards and 2 parity shards - this avoid lots of padding
2. For intermediary sizes use always 1 block
3. For bigger sizes, split by blocks
The only problem is that for cases 1 and 2 the resulting file must be stored with padding, or this size will not monotonically grow.

One last optimization was to change the max shard size depending on the percentage of space overhead. This also makes it use a little less space for smaller files.

In the end, the file growth by space overhead is:
```
S.O.         1 B     %        10 B     %       100 B     %       1 KiB     %      10 KiB     %     100 KiB     %       1 MiB     %      10 MiB     %   
1  %        40 B 3900%        56 B  460%       176 B   76%     1.4 KiB   38%    11.2 KiB   13%     102 KiB    2%       1 MiB    1%    10.1 MiB    1%   
2  %        40 B 3900%        56 B  460%       176 B   76%     1.4 KiB   38%    10.9 KiB    9%   102.8 KiB    3%       1 MiB    2%    10.2 MiB    2%   
5  %        40 B 3900%        56 B  460%       176 B   76%     1.4 KiB   38%    10.8 KiB    8%   104.9 KiB    5%       1 MiB    5%    10.5 MiB    5%   
10 %        40 B 3900%        56 B  460%       176 B   76%     1.4 KiB   38%    11.2 KiB   12%   109.7 KiB   10%     1.1 MiB   10%      11 MiB   10%   
20 %        40 B 3900%        56 B  460%       176 B   76%     1.4 KiB   38%      13 KiB   30%   121.8 KiB   22%     1.2 MiB   21%    12.1 MiB   21%   
```

And the performance is:

- For 100B file:
```
     ECC                                Throughput     Throughput       Growth
                                          Encoding       Decoding         
---------------------------------------------------------------------------------------
  0. REED-SOLOMON-CRC32 - 10%           45.7 MiB/s     45.6 MiB/s       76% [176 B]
  1. REED-SOLOMON-CRC32 - 20%           38.8 MiB/s     58.6 MiB/s       76% [176 B]
  2. REED-SOLOMON-CRC32 - 2%            36.2 MiB/s     49.3 MiB/s       76% [176 B]
  3. REED-SOLOMON-CRC32 - 5%              35 MiB/s     68.5 MiB/s       76% [176 B]
  4. REED-SOLOMON-CRC32 - 1%            31.1 MiB/s     45.8 MiB/s       76% [176 B]
---------------------------------------------------------------------------------------
```

- For 1KB file:
```
     ECC                                Throughput     Throughput       Growth
                                          Encoding       Decoding
---------------------------------------------------------------------------------------
  0. REED-SOLOMON-CRC32 - 20%            323 MiB/s    406.3 MiB/s       38% [1.4 KiB]
  1. REED-SOLOMON-CRC32 - 1%                 0 B/s     96.7 MiB/s       38% [1.4 KiB]
  2. REED-SOLOMON-CRC32 - 2%            95.2 MiB/s          0 B/s       38% [1.4 KiB]
  3. REED-SOLOMON-CRC32 - 5%           163.3 MiB/s          0 B/s       38% [1.4 KiB]
  4. REED-SOLOMON-CRC32 - 10%                0 B/s      397 MiB/s       38% [1.4 KiB]
---------------------------------------------------------------------------------------
```

- For 10KB file:
```
     ECC                                Throughput     Throughput       Growth
                                          Encoding       Decoding         
---------------------------------------------------------------------------------------
  0. REED-SOLOMON-CRC32 - 5%           538.2 MiB/s    892.1 MiB/s        8% [10.8 KiB]
  1. REED-SOLOMON-CRC32 - 10%          498.6 MiB/s    825.6 MiB/s       12% [11.2 KiB]
  2. REED-SOLOMON-CRC32 - 2%           199.7 MiB/s    630.7 MiB/s        9% [10.9 KiB]
  3. REED-SOLOMON-CRC32 - 20%            178 MiB/s    787.9 MiB/s       30% [13 KiB]
  4. REED-SOLOMON-CRC32 - 1%           174.4 MiB/s    463.9 MiB/s       13% [11.2 KiB]
---------------------------------------------------------------------------------------
```

- For 100KB file:
```
     ECC                                Throughput     Throughput       Growth
                                          Encoding       Decoding         
---------------------------------------------------------------------------------------
  0. REED-SOLOMON-CRC32 - 2%             2.9 GiB/s      2.7 GiB/s        3% [102.8 KiB]
  1. REED-SOLOMON-CRC32 - 5%             1.4 GiB/s      1.6 GiB/s        5% [104.9 KiB]
  2. REED-SOLOMON-CRC32 - 1%               1 GiB/s      2.4 GiB/s        2% [102 KiB]
  3. REED-SOLOMON-CRC32 - 10%          547.3 MiB/s      890 MiB/s       10% [109.7 KiB]
  4. REED-SOLOMON-CRC32 - 20%          239.8 MiB/s        1 GiB/s       22% [121.8 KiB]
---------------------------------------------------------------------------------------
```

- For 1MB file:
```
     ECC                                Throughput     Throughput       Growth
                                          Encoding       Decoding         
---------------------------------------------------------------------------------------
  0. REED-SOLOMON-CRC32 - 2%             1.9 GiB/s      4.4 GiB/s        2% [1 MiB]
  1. REED-SOLOMON-CRC32 - 1%             1.2 GiB/s      3.7 GiB/s        1% [1 MiB]
  2. REED-SOLOMON-CRC32 - 5%             1.1 GiB/s      1.8 GiB/s        5% [1 MiB]
  3. REED-SOLOMON-CRC32 - 10%          570.5 MiB/s    918.9 MiB/s       10% [1.1 MiB]
  4. REED-SOLOMON-CRC32 - 20%          249.3 MiB/s      1.1 GiB/s       21% [1.2 MiB]
---------------------------------------------------------------------------------------
```

I tried but still could not find out why smaller files are so much slower.

> Could we perhaps have a single per-repository flag to specify how much overheard to use for ECC (0%-10%, 0==no ECC) and we figure out the rest and dynamically apply the right algorithms based on content lengths?

I think that is a good idea. The code is organized around that idea now. There are a few other parameters, but they need not be set (there are used for tests, basically)

Also, I think now this can be considered for merging.

repo/ecc/ecc_utils.go
```diff
@@ -0,0 +1,69 @@
+package ecc
+
+import "math"
+
+func ComputeShards(spaceOverhead float32) (required, redundant int) {
+	required = 128
+	redundant = between(applyPercent(required, spaceOverhead/100), 1, 128)
+
+	if redundant == 1 {
```
jkowalski, line 9: can you put a comment around this logic?

repo/ecc/ecc_utils.go
```diff
@@ -0,0 +1,69 @@
...
+	return int(math.Floor(float64(val) * float64(percent)))
+}
+
+func clear(bytes []byte) {
+	for i := range bytes {
+		bytes[i] = 0
+	}
+}
+
+func minInt(a int, b int) int {
```
jkowalski, line 43: this seems to be used once or twice, sometimes if statement is more readable

repo/ecc/ecc_utils.go
```diff
@@ -0,0 +1,69 @@
...
+
+func maxFloat32(a float32, b float32) float32 {
+	if a >= b {
+		return a
+	} else {
+		return b
+	}
+}
+
+func CeilInt(a, b int) int {
```
jkowalski, line 67: does this need to be exported? If it's just for testability, golang provides ways of package-level testing (test package needs to be `package ecc` instead of `package ecc_test`, you can actuall have both kinds of tests in one directory)

repo/ecc/ecc.go
```diff
@@ -0,0 +1,87 @@
+// Package ecc implements common support for error correction in sharded blob providers
+package ecc
+
+import (
```
jkowalski, line 4: reorder imports here based on linter failures

repo/ecc/ecc_utils.go
pescuma: I can change it if you prefer but I really think a function called `min` is easier to understand. The `Int` I did not like, but I could not make a generic version that works for all types of numbers.

repo/ecc/ecc_utils.go
pescuma: Thanks, will fix. It was only because of the tests, and I did not know you could do that.

repo/ecc/ecc.go
pescuma: I'm sorry for the question, but how can I see the linter output?

jkowalski: `make lint` should do the trick. In general `make` before sending a PR will uncover all kinds of issues and save you time.

pescuma: Sorry. I had not noticed the Makefile there. All lint errors should be fixed now.

repo/ecc/ecc_rs_crc_test.go
```diff
@@ -0,0 +1,252 @@
+package ecc
+
+import (
+	"testing"
+
+	"github.com/kopia/kopia/repo/encryption"
+	"github.com/stretchr/testify/require"
+)
+
+/*
```
jkowalski, line 10: should we uncomment this?

repo/ecc/ecc_options.go
```diff
@@ -0,0 +1,21 @@
+package ecc
+
+// Options must be anonymously embedded in sharded provider options.
```
jkowalski, line 3: the comment is stale

repo/ecc/ecc_options.go
```diff
@@ -0,0 +1,21 @@
...
+type Options struct {
+	// Algorithm name to be used. Leave empty to disable error correction.
+	Algorithm string `json:"algorithm,omitempty"`
+
+	// SpaceOverhead is how much more space can be used for ECC, in percentage.
+	// Between 0 and 100. 0 means disable ECC.
+	SpaceOverhead uint8 `json:"spaceOverhead,omitempty"`
```
jkowalski, line 10: `int` is usually preferred, byte seems like over-optimization.

repo/ecc/ecc_rs_crc.go
```diff
@@ -0,0 +1,430 @@
...
+
+	originalSize = int(binary.BigEndian.Uint32(lengthBuffer[:]))
+
+	//nolint:nakedret
+	return
+}
+
+// Overhead is not used. It's just implemented to avoid errors when running.
+func (r *ReedSolomonCrcECC) Overhead() int {
+	return 0
```
jkowalski, line 395: we should panic here, so we remember to never call this.

repo/ecc/ecc_rs_crc_test.go
pescuma: The two commented methods are not really tests, they are just the code that I used to generate some of the tables above. I was not sure if it would be better to leave them there or just remove, but un-commenting does not really help. I think I'll just remove them.

repo/ecc/ecc_rs_crc.go
pescuma: Currently this method is called by `content.SharedManager.setupReadManagerCaches`:
```
	sm.committedContents = newCommittedContentIndex(caching,
		uint32(sm.format.Encryptor().Overhead()),
		sm.format.WriteIndexVersion(),
		sm.enc.getEncryptedBlob,
		sm.namedLogger("committed-content-index"),
		caching.MinIndexSweepAge.DurationOrDefault(DefaultIndexCacheSweepAge))
```

I can panic it now, because it is not being called, but when integrating this will be a problem.

jkowalski: that's ok, when we integrate this we will need to fix this logic.

PR #2272, opened 2022-08-03
docs(site): Updating repository testing and recovery page

basldfalksjdf:


basldfalksjdf:
@jkowalski Please double check the information in the corruption repair section. I think it is correct, but you are the expert.

Also, I do not sure how you feel about the new information I put in that section. Do we want to have such directions in the docs? Or is it better to just tell people to post on the forums asking for help? Up to you.

site/content/docs/Advanced/Consistency/_index.md
```diff
@@ -1,69 +1,72 @@
...
 
->NOTE: Since all data corruption cases are unique, it's generally not recommended to attempt fixes recommended to other users even for possibly similar issues, since the particular fix method may not be applicable. 
+    * If the name(s) start with `q`, then the corruption is likely with metadata. Run `kopia snapshot fix invalid-files --parallel=10 --commit` and see if that fixes your issue.
 
+    * If the name(s) start with `p`, then the corruption may be in your indexes, metadata, or in the actual blob that stores your backed up files. First, try `kopia index recover --parallel=10 --commit`. If your issue has not been solved, try `kopia index recover --parallel=10 --commit --delete-indexes`. If you are still having issues, try `kopia snapshot fix invalid-files --parallel=10 --commit`. Finally, if the corrupt is still not fixed, try `kopia snapshot fix invalid-files --parallel=10 --commit --verify-files-percent=100` but note that this will download all your files, so it will take a while.
```
jkowalski, line 66: Given current v2 index structures, `index recover` is almost never the solution worth trying. We should remove this and probably deprecate the command.

jkowalski: In general most of the suggestions (except `kopia snapshot fix invalid-files`) are unsafe when used improperly and they can lead to further corruption, that's why they were not mentioned.

Definitely do not suggest deleting any blobs, since this will only help in very few cases if the root cause is properly understood.

I would say if you have timestamps of 1970 or 2038, the filesystem is likely incompatible with Kopia and deleting will not do much.

jkowalski: Also, let's not recommend folks run commands with `--commit` until the amount of damage/deletion is understood. Without this switch, kopia will print what it's going to do without actually rewriting snasphots.

basldfalksjdf: I just reverted the repair corruption part to what it was before my proposed edits.

PR #2299, opened 2022-08-10
refactor(repository): Rename Encryptor to Transformer (and related classes)

pescuma:
This is an initial implementation, so we can start talking about it.

### DIFF
--- a/cli/command_benchmark_crypto.go
+++ b/cli/command_benchmark_crypto.go
@@ -100,7 +100,7 @@ func (c *commandBenchmarkCrypto) runBenchmark(ctx context.Context) []cryptoBench
 				for i := 0; i < hashCount; i++ {
 					contentID := hf(hashOutput[:0], input)
 
-					if encerr := enc.Encrypt(input, contentID, &encryptOutput); encerr != nil {
+					if encerr := enc.ToRepository(input, contentID, &encryptOutput); encerr != nil {
 						log(ctx).Errorf("encryption failed: %v", encerr)
 						break
 					}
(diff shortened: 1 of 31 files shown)


jkowalski:
I feel that the notion of `Transformer` only belongs in the `content` package so I would not rename the `encryption` package and keep `encryption.Encryptor`, but instead have 2 implementations of Transformer. If i'm not mistaken this should let us avoid having `Overhead()` method in the transformer.

* encryptingTransformer (which forwards to encryption.Encrypter)
* encryptingTransformerWithErrorCorrection (which chains encryption.Encrypter and ecc.SomeOtherName)

This would allow us to give meaningful name to ecc implementation (insted of `Encryptor`, which it isn't)

BTW if we stretch our imagination, there's a perfectly valid nullTransformer or errorCorrectionTransformer which don't do any encryption.

pescuma:
I agree with you. My plan is that, in the end, you can create the transform with something like `transform.Pipeline(compression.New(), encryption.New(), ecc.New())` and if any of those are null, the pipeline just ignores it.

About the `Encryptor` interface, I agree that conceptually it makes sense by itself. But, looking at the code, I see that only 2 places create one: `NewFormattingOptionsProvider` and `AuthenticatedEncryptionProtection`. The former will also (maybe) use ecc. I don't know exactly about the latter, seems like it has something to do with the cache. For it, would ecc make sense? Or, asking it another way: there is some place that needs encryption and that we would not want ecc at all?

About the `Overhead` method, at lest currently, it is used from `format.Provider`, which would need ecc and it would require that method in the `Transform` interface anyway. Or did I get it wrong?

(I'm just throwing ideas around, I have no problem in keeping the `Encryptor` interface)

pescuma:
Closing this until we have a better idea of how to do this change.

PR #2308, opened 2022-08-11
feat(cli): Added ECC related options to repository create cli command

pescuma:
As the next step in ECC integration, this PR allows to create repositories with ECC enabled, but only from the cli for now.


repo/format/format_provider.go
```diff
@@ -99,11 +100,18 @@ func NewFormattingOptionsProvider(f *ContentFormat, formatBytes []byte) (Provide
 		return nil, errors.Wrap(err, "unable to create hash")
 	}
 
-	e, err := encryption.CreateEncryptor(f)
+	encryptionEncryptor, err := encryption.CreateEncryptor(f)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to create encryptor")
 	}
 
+	eccEncryptor, err := ecc.CreateEncryptor(f)
```
jkowalski, line 108: add a comment saying this can and in most cases will be nil, since it's not quite obvious.

repo/encryption/pipeline.go
```diff
@@ -0,0 +1,60 @@
+package encryption
+
+import "github.com/kopia/kopia/internal/gather"
+
+// Pipeline creates a pipeline encryptor, where the output of one becomes the input of the next.
+func Pipeline(es ...Encryptor) Encryptor {
```
jkowalski, line 6: that seems unnecessarily generic, do we need that here and/or that generic? The only use is in `format` package.

repo/encryption/pipeline.go
```diff
@@ -0,0 +1,60 @@
...
+		//nolint:wrapcheck
+		return err
+	}
+
+	//nolint:wrapcheck
+	return p.impl.Decrypt(tmp.Bytes(), contentID, output)
+}
+
+func (p *pipelineStep) Overhead() int {
+	return p.impl.Overhead() + p.next.Overhead()
```
jkowalski, line 59: again: panic here, Overhead() should not really be called except for legacy cases.

cli/command_repository_create.go
```diff
@@ -41,6 +44,8 @@ func (c *commandRepositoryCreate) setup(svc advancedAppServices, parent commandP
 
 	cmd.Flag("block-hash", "Content hash algorithm.").PlaceHolder("ALGO").Default(hashing.DefaultAlgorithm).EnumVar(&c.createBlockHashFormat, hashing.SupportedAlgorithms()...)
 	cmd.Flag("encryption", "Content encryption algorithm.").PlaceHolder("ALGO").Default(encryption.DefaultAlgorithm).EnumVar(&c.createBlockEncryptionFormat, encryption.SupportedAlgorithms(false)...)
+	cmd.Flag("ecc", "Error correction algorithm.").PlaceHolder("ALGO").Default(ecc.DefaultAlgorithm).EnumVar(&c.createBlockECCFormat, ecc.SupportedAlgorithms()...)
+	cmd.Flag("ecc-overhead", "How much space overhead can be used for error correction, in percentage. Use 0 to disable ECC.").Default("0").IntVar(&c.createBlockECCOverhead)
```
jkowalski, line 48: can you add `Hidden()` to both flags for now, until we're ready to advertise this functionality?

repo/encryption/pipeline.go
pescuma: The idea is to maybe support compression in the future. Also, I though it would be better if forwarding to next one and ecc would be done in different classes. I can move this to the ecc package, and put the code as internal, if you prefer.

repo/format/format_provider.go
pescuma: Added. For some reason it does not show here.

repo/encryption/pipeline.go
jkowalski: I'm sceptical we'll be able to do compression like this, since it's conditional and needs to return additional state (whether it was applied or not). 

So we'll have either 1- or 2-stage pipeline which is only used in `format`, so that's why I think we should move it there (on a tactical basis).1-stage pipeline is simply encryptor and we had before was quite reasonable as 2-stage pipeline.

repo/format/format_provider.go
```diff
@@ -99,11 +100,19 @@ func NewFormattingOptionsProvider(f *ContentFormat, formatBytes []byte) (Provide
 		return nil, errors.Wrap(err, "unable to create hash")
 	}
 
-	e, err := encryption.CreateEncryptor(f)
+	encryptionEncryptor, err := encryption.CreateEncryptor(f)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to create encryptor")
 	}
 
+	// This will return nil if the current options disable ECC
```
jkowalski, line 108: i would add if statement around here and rename things slightly for readability:

if f.GetECCAlgorithm() != "" {
  ecc, err := ecc.New(f)
  if err ... 
  e = encryptorWrapper{e, ecc}
}

I would move `encryptorWrapper` to this package.

cli/command_repository_create.go
```diff
@@ -41,6 +44,8 @@ func (c *commandRepositoryCreate) setup(svc advancedAppServices, parent commandP
 
 	cmd.Flag("block-hash", "Content hash algorithm.").PlaceHolder("ALGO").Default(hashing.DefaultAlgorithm).EnumVar(&c.createBlockHashFormat, hashing.SupportedAlgorithms()...)
 	cmd.Flag("encryption", "Content encryption algorithm.").PlaceHolder("ALGO").Default(encryption.DefaultAlgorithm).EnumVar(&c.createBlockEncryptionFormat, encryption.SupportedAlgorithms(false)...)
+	cmd.Flag("ecc", "Error correction algorithm.").Hidden().PlaceHolder("ALGO").Default(ecc.DefaultAlgorithm).EnumVar(&c.createBlockECCFormat, ecc.SupportedAlgorithms()...)
+	cmd.Flag("ecc-overhead", "How much space overhead can be used for error correction, in percentage. Use 0 to disable ECC.").Hidden().Default("0").IntVar(&c.createBlockECCOverhead)
```
jkowalski, line 48: can you rename the flag `ecc-overhead-percent` and all fields along with it? withtout the suffix it's not immediately obvious what the units are.

repo/format/format_provider.go
pescuma: Done. A side comment: i understand why you want to hide encryptorWrapper in format package, but for me it fells better to keep it together with Encryptor, because they seem very closely related.

repo/initialize.go
```diff
@@ -163,6 +166,10 @@ func repositoryObjectFormatFromOptions(opt *NewRepositoryOptions) (*format.Repos
 		f.HMACSecret = nil
 	}
 
+	if opt.BlockFormat.ECCOverheadPercent == 0 {
+		f.ContentFormat.ECC = ""
+	}
+
```
jkowalski, line 172: we should disallow ECC when `IndexVersion == 1`

Also, can you add a test or two using something like:

```
	e := testenv.NewCLITest(t, testenv.RepoFormatNotImportant, runner)

	defer e.RunAndExpectSuccess(t, "repo", "disconnect")
	e.RunAndExpectSuccess(t, "repo", "create", "filesystem", "--path", e.RepoDir, /* ECC parameters */)
	output := e.RunAndExpectSuccess(t, "repo", "status")
        // verify output contains ECC status line
```

I'm wondering if it makes sense to add few quick e2e tests for the ECC:

- create repo with 50% overhead, backup 1MB of random data, ensure the repo size is >= 1.5MB
- as a control, backup the same file with no ECC and assert that repository size is <=1.1MB.
- after backup with ECC, go through every single file in the repo (except `kopia.repository`) and flip one random bit in each file - the repo should still be readable

jkowalski:
Thank you!

it's close to LGTM, can you add some tests using framework we already have and then we can merge it!

repo/ecc/ecc_options.go
```diff
@@ -5,11 +5,11 @@ type Options struct {
 	// Algorithm name to be used. Leave empty to disable error correction.
 	Algorithm string `json:"algorithm,omitempty"`
 
-	// SpaceOverhead is how much more space can be used for ECC, in percentage.
+	// OverheadPercent is how much more space can be used for ECC, in percentage.
 	// Between 0 and 100. 0 means disable ECC.
-	SpaceOverhead int `json:"spaceOverhead,omitempty"`
+	OverheadPercent int `json:"spaceOverhead,omitempty"`
```
jkowalski, line 10: update JSON name as well

repo/initialize.go
pescuma: Done

jkowalski:
This is awesome. LGTM.

PR #2314, opened 2022-08-13
feat(server): Added ECC to server api

pescuma:
This allows the GUI to show/configure ECC.

### DIFF
--- a/internal/server/api_repo.go
+++ b/internal/server/api_repo.go
@@ -70,9 +70,15 @@ func handleRepoStatus(ctx context.Context, rc requestContext) (interface{}, *api
 			return nil, internalServerError(err)
 		}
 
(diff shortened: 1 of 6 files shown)


jkowalski:
Now that the UI part has been built, you can add it to this PR:

```
go get github.com/kopia/htmluibuild@d7803267445c9387823fb079071a63325f897bc8
```

internal/server/api_repo.go
```diff
@@ -70,9 +70,15 @@ func handleRepoStatus(ctx context.Context, rc requestContext) (interface{}, *api
 			return nil, internalServerError(err)
 		}
 
+		pars, err := dr.ContentReader().ContentFormat().GetMutableParameters()
```
jkowalski, line 73: nit: use `mp` instead of `pars` since this is what we do everywhere else.

PR #2317, opened 2022-08-16
feat(snapshots): Allow uploader to find cached `StreamingFile`s

ashmrtn:
Main changes in this PR:
* allow user to set mod time on a `StreamingFile` if they want
* set `StreamingFile` mod time when the struct is created, not in the uploader when the directory manifest entry is added
* relax metadata equality checks for `StreamingFile` in the uploader when looking for cached files

Together these allows the uploader to skip reuploading `StreamingFile` content if the file has the same mod time, mode, and owner. Size is no longer checked as that is only populated after the `StreamingFile` has been completely uploaded

This patch is not expected to change the behavior of backups that don't explicitly set the mod time as `StreamingFile` will default to using the current time. This means that different backups should have different mod times for `StreamingFile`s, resulting in reuploading the content just like it does now. The generated times for the `StreamingFile`s may be slightly earlier than previously observed though because it is set on struct initialization closer to the start of the backup, not when the data upload is completed during the backup

### DIFF
--- a/fs/virtualfs/virtualfs.go
+++ b/fs/virtualfs/virtualfs.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/kopia/kopia/fs"
+	"github.com/kopia/kopia/internal/clock"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


jkowalski:
LGTM!

PR #2327, opened 2022-08-19
feat(repository): implemented format.Manager

jkowalski:
This will provide a single place for reading, caching and all
manipulations for `kopia.repository` and `kopia.blobcfg`.

Both format blobs are automatically refreshed every 15 minutes
to support safe format changes in the futures with guaranteed maximum
latency.


repo/format/format_manager.go
```diff
@@ -0,0 +1,466 @@
...
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/encryption"
+	"github.com/kopia/kopia/repo/hashing"
+	"github.com/kopia/kopia/repo/logging"
+)
+
+var log = logging.Module("kopia/repo/format")
+
+// UniqueIDLength is the length of random unique ID of each repository.
+const UniqueIDLength = 32
```
aaron-kasten, line 23: Need units.  Length in what?  Bytes, bits, words?

repo/format/format_blob_cache.go
```diff
@@ -12,133 +14,161 @@ import (
...
-	ufb.KopiaRepositoryBytes, err = addFormatBlobChecksumAndLength(ufb.KopiaRepositoryBytes)
-	if err != nil {
-		return nil, errors.Errorf("unable to add checksum")
+// NewDiskCache returns on-disk blob cache.
+func NewDiskCache(cacheDir string) BlobCache {
+	return &onDiskCache{cacheDir}
+}
+
+// NewMemoryBlobCache returns in-memory blob cache.
+func NewMemoryBlobCache(timeNow func() time.Time) BlobCache {
```
aaron-kasten, line 149: Is it necessary to include the BlobCache in this PR?  If so, why?

repo/format/format_blob_cache.go
```diff
@@ -12,133 +14,161 @@ import (
...
-	"github.com/kopia/kopia/repo/logging"
 )
 
 // DefaultRepositoryBlobCacheDuration is the duration for which we treat cached kopia.repository
 // as valid.
 const DefaultRepositoryBlobCacheDuration = 15 * time.Minute
 
-var log = logging.Module("kopia/repo/format")
+// BlobCache encapsulates cache for format blobs.
+type BlobCache interface {
```
aaron-kasten, line 25: Is it necessary to include the BlobCache in this PR?  If so, why?

repo/format/crypto_key_derivation.go
```diff
@@ -5,16 +5,17 @@ import (
...
 // DeriveKeyFromMasterKey computes a key for a specific purpose and length using HKDF based on the master key.
 func DeriveKeyFromMasterKey(masterKey, uniqueID, purpose []byte, length int) []byte {
 	key := make([]byte, length)
 	k := hkdf.New(sha256.New, masterKey, uniqueID, purpose)
 
-	if _, err := io.ReadFull(k, key); err != nil {
-		panic("unable to derive key from master key, this should never happen")
-	}
+	_, err := io.ReadFull(k, key)
+	impossible.PanicOnError(err)
```
aaron-kasten, line 18: Can this be moved to another PR?

repo/format/format_manager.go
```diff
@@ -0,0 +1,466 @@
...
+	if err := m.maybeRefreshLocked(); err != nil {
+		return nil, err
+	}
+
+	return m.current, nil
+}
+
+// +checklocks:m.mu
+func (m *Manager) maybeRefreshLocked() error {
+	if n := m.timeNow(); !n.Before(m.validUntil) {
```
aaron-kasten, line 76: Nit: early return is possible.  Return `nil` when `n := m.timeNow(); n.Before(m.validUntil)`

repo/format/format_blob_cache.go
```diff
@@ -12,133 +14,161 @@ import (
...
-	if err != nil {
-		return nil, errors.Errorf("unable to add checksum")
+// NewDiskCache returns on-disk blob cache.
+func NewDiskCache(cacheDir string) BlobCache {
+	return &onDiskCache{cacheDir}
+}
+
+// NewMemoryBlobCache returns in-memory blob cache.
+func NewMemoryBlobCache(timeNow func() time.Time) BlobCache {
+	return &inMemoryCache{
```
aaron-kasten, line 150: This cache looks like its unbounded.

repo/format/format_manager.go
```diff
@@ -0,0 +1,466 @@
...
+
+	return m.refreshCounter
+}
+
+// refreshLocked reads `kopia.repository` blob, potentially from cache and decodes it.
+// +checklocks:m.mu
+func (m *Manager) refreshLocked(ctx context.Context) error {
+	log(ctx).Infow("refreshLocked", "now", m.timeNow())
+
+	b, cacheMTime, err := m.readAndCacheRepositoryBlobBytes(ctx, KopiaRepositoryBlobID)
```
aaron-kasten, line 124: Can remove the cache and instead keep (cache) the parsed forms (result of `ParseKopiaRepositoryJSON`, `addFormatBlobChecksumAndLength` and `blobCfg` in `Manager`).  `refreshLocked` compares current time with what was cache reload duration.
A session approach might be better for use of the `blobCfg` and the `repoConfig` - maintain values for an instance of a callpath instead of wallclock-time.

aaron-kasten (commented):
See comments.

repo/format/format_blob_cache.go
jkowalski: It's a micro-abstraction just for caching the two blobs  - we have 3 strategies, in-memory, null and on-disk. That's part of the refactoring, the previous approach assumed disk and was generally fragile.

repo/format/format_blob_cache.go
jkowalski: same, it's only for 2 blobs.

repo/format/format_blob_cache.go
jkowalski: it's only 2 items.

repo/format/format_manager.go
jkowalski: the on-disk cache allows us to not fetch the blob over and over again when kopia is invoked from the CLI.

Happy to chat more about rationales here, ping me on slack.

repo/format/format_blob_cache.go
jkowalski: added comment and made unexported.

repo/format/format_blob_cache.go
jkowalski: added comment, made this unexported.

repo/format/crypto_key_derivation.go
jkowalski: reverted

repo/format/format_manager.go
jkowalski: done

aaron-kasten (commented):
@jkowalski Do you see a way that this can be broken up into a PR train?  There are enough substantial changes that its difficult to get through as a single review.

jkowalski:
I could not find a way to meaningfully break this into PR train since there are many interdependent moves and renames, but I opened #2349 which is sliced into individual commits by area. I'm going to mark this one as draft.

PR #2329, opened 2022-08-19
feat(cli): flag to allow placing the upgrade lock without actually upgrading

Shrekster:
This adds a hidden flag to the upgrade CLI to allow tests to lock a repository without actually performing the upgrade draining mechanism and the index upgrade itself. The current mechanism to simulate a locked repo is by done by killing the CLI during draining but that approach is pretty flaky. This should help avoid failures like this:

https://github.com/kopia/kopia/runs/7920935223?check_suite_focus=true


aaron-kasten (approved):
Looks good!

PR #2331, opened 2022-08-19
feat(snapshots): Callback for when uploader finishes processing a file

ashmrtn:
Add a callback that gives implementations of `UploadProgress` insight into the fate of uploaded files. This callback provides a slightly different surface area compared to listening on `FinishedHashingFile` and `Error`. Notably, it combines the completion of a file with whether or not there was an error processing the file. By combining this, listeners do not have to worry about seeing a `FinishedHashingFile` call and then maybe seeing an `Error` call. Instead, they know the outcome right away. This can save listeners from caching all files reported by `FinishedHashingFile` until the end of the upload (at which point `Error` will definitely not be called on them)

### DIFF
--- a/snapshot/snapshotfs/upload.go
+++ b/snapshot/snapshotfs/upload.go
@@ -136,8 +136,12 @@ func (u *Uploader) incompleteReason() string {
 	return ""
 }
 
-func (u *Uploader) uploadFileInternal(ctx context.Context, parentCheckpointRegistry *checkpointRegistry, relativePath string, f fs.File, pol *policy.Policy) (*snapshot.DirEntry, error) {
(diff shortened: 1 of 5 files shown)


snapshot/snapshotfs/upload.go
```diff
@@ -136,8 +136,12 @@ func (u *Uploader) incompleteReason() string {
 	return ""
 }
 
-func (u *Uploader) uploadFileInternal(ctx context.Context, parentCheckpointRegistry *checkpointRegistry, relativePath string, f fs.File, pol *policy.Policy) (*snapshot.DirEntry, error) {
+func (u *Uploader) uploadFileInternal(ctx context.Context, parentCheckpointRegistry *checkpointRegistry, relativePath string, f fs.File, pol *policy.Policy) (dirEntry *snapshot.DirEntry, ret error) {
 	u.Progress.HashingFile(relativePath)
+
+	defer func() {
+		u.Progress.FinishedFile(relativePath, ret != nil)
```
jkowalski, line 143: why not pass the error itself instead of boolean?

jkowalski (approved):

PR #2334, opened 2022-08-22
feat(repository): Filesystem snapshots for windows

pescuma:
This is an initial implementation of windows' VSS snapshots. I still need to implement some configuration method, probably through policies. Also, this only works if the user running kopia is admin, but I could not find a way around that.

Please let me know if you would like me to keep working in this one. And, if so, if the current implementation seems ok.


jkowalski:
This is very interesting, do you have plans to maintain the library and enhance it to support other types of snapshots (ZFS, LVM)?

pescuma:
I plan to maintain it as far as my needs go, so only windows for now. I'll probably need to use a Mac later this year - if that happens, I'll see if there is something similar there. But my idea was to allow other implementations in the same library, or at least try to support them. But I don't have the time to implement the ones I won't use.

basldfalksjdf:
Out of curiosity, what does this library get Kopia that the before/after action script at https://kopia.io/docs/advanced/actions/ is unable to accomplish in terms of creating a VSS snapshot and having Kopia backup from that snapshot?

Would it be better to keep it simple and add to Kopia/KopiaUI an option to enable VSS which, then, in the backend enables the before/after action script? The same methodology could be implemented for ZFS, etc. This way users could easily enable VSS/other snapshotting at their leisure rather than having to go through actions for it.

pescuma:
Nothing major, but it has some advantages:

1. You don't have to configure actions
2. The snapshot path appears as the correct path (looking at the vss script, seems like the path will always have some other path as a prefix - if I'm not mistaken)
3. It allows to disable vss writers. I may make it work in some cases. Not sure if it's useful, though.
4. It handles mount points. I didn't even know windows had mount points before, so, again, not sure if this is useful 
5. If we implement it was a policy, it could allow snapshot of only some folders. Again, not sure if anyone would want this.

For me, 1 and 2 are enough to make this worth.

basldfalksjdf:
> Nothing major, but it has some advantages:
> 
> 1. You don't have to configure actions

I absolutely agree that enabling or disabling VSS within policies with Kopia is a fantastic feature to have. My point was simply, what is the better way to do it: by adding the library you propose in this PR or to submit a PR that embeds the action scripts as a Kopia policy? In effect, under both scenarios there would be an option in policies to enable/disable this feature. The difference would be under-the-hood in what happens when that policy is enabled. The former adds a lot of complexity to Kopia for a feature that can arguably be implemented by leveraging the actions framework that is already built into Kopia.

fs/fssnapshotfs/snapshot_fs.go
```diff
@@ -0,0 +1,268 @@
...
+	fs.Entry
+}
+
+// NewEntry returns fs.Entry for the specified path, the result will be one of supported entry types: fs.File, fs.Directory, fs.Symlink
+// or fs.UnsupportedEntry.
+func NewEntry(originalPath string) (fs.Entry, error) {
+	snapshoter, err := fs_snapshot.CreateSnapshoter()
+	if err != nil {
+		// If we can't use snapshots, just use the filesystem directly
+		fmt.Printf("Error creatting filesystem snapshot: %v\n", err)
```
phidevz, line 29: typo: double t

pescuma:
@jkowalski I think this is now ready to merge, if you want. Current state is:

Windows:
 - Works if kopia is run as administrator
 - Works without administrator if fs_snapshot.exe is copied somewhere and `fs_snapshot enable for current-user` is run as Administrator. This needs to be done only once. It registers a scheduled task that runs a server as administrator to take snapshots.

MacOS:
 - Works if computer has one drive formatted as APFS. I did not yet understand how to list current drives/mounting points/relationship between data and mount point.

fs/fssnapshotfs/snapshot_fs.go
```diff
@@ -0,0 +1,305 @@
...
+			if level == fs_snapshot.OutputLevel {
+				fmt.Printf(msg+"\n", a...)
+			}
+		},
+	}
+
+	snapshoter, err := fs_snapshot.NewSnapshoter(cfg)
+	if err != nil {
+		// If we can't use snapshots, just use the filesystem directly
+		fmt.Printf("Error creating filesystem snapshot: %v\n", err)
```
jkowalski, line 48: log an error here instead of Printf - you'll need to pass `ctx`

fs/fssnapshotfs/snapshot_fs.go
```diff
@@ -0,0 +1,305 @@
...
+
+		return entry, nil
+	}
+
+	backuper, err := snapshoter.StartBackup(nil)
+	if err != nil {
+		// If we can't use snapshots, just use the filesystem directly
+		fmt.Printf("Error creating filesystem snapshot: %v\n", err)
```
jkowalski, line 56: same

cli/command_policy_set_files.go
```diff
@@ -57,5 +61,9 @@ func (c *policyFilesFlags) setFilesPolicyFromFlags(ctx context.Context, fp *poli
 		return err
 	}
 
+	if err := applyPolicyBoolPtr(ctx, "use fs snapshots", &fp.UseFsSnapshots, c.policyUseFsSnapshots, changeCount); err != nil {
```
jkowalski, line 64: perhaps `"use fileystem snapshots where available"` ?

cli/command_policy_set_files.go
```diff
@@ -43,6 +45,8 @@ func (c *policyFilesFlags) setup(cmd *kingpin.CmdClause) {
 	cmd.Flag("one-file-system", "Stay in parent filesystem when finding files ('true', 'false', 'inherit')").EnumVar(&c.policyOneFileSystem, booleanEnumValues...)
 
 	cmd.Flag("ignore-cache-dirs", "Ignore cache directories ('true', 'false', 'inherit')").EnumVar(&c.policyIgnoreCacheDirs, booleanEnumValues...)
+
+	cmd.Flag("use-fs-snapshots", "Use filesystem snapshots ('true', 'false', 'inherit')").EnumVar(&c.policyUseFsSnapshots, booleanEnumValues...)
```
jkowalski, line 49: nit `use-filesystem-snapshots`

fs/fssnapshotfs/snapshot_fs.go
```diff
@@ -0,0 +1,305 @@
...
+	fs_snapshot "github.com/pescuma/go-fs-snapshot/lib"
+
+	"github.com/kopia/kopia/fs"
+	"github.com/kopia/kopia/fs/localfs"
+	"github.com/kopia/kopia/snapshot"
+	"github.com/kopia/kopia/snapshot/policy"
+)
+
+type snapshotEntry struct {
+	policyTree      *policy.Tree
```
jkowalski, line 19: I think this is unnecessarily coupling this package with policies, while you only need one bit of information.

Can we remove this dependency on policies and lift the decision to use snapshots to the caller?

jkowalski: One idea - (which may or may not work) - can we merge this package with `localfs` and explicitly add snapshotting support there (while still keeping it loosely coupled and not dependent on `fs_snapshot` library).

I'm thinking the API could be something like:

```
package localfs

type SnapshotFunc func(originalDir string) (newDir string, closeFunc func(), err error)

func NoSnapshot(originalDir string) (newDir string, closeFunc func(), err error) {
  return oriignalDir, func() {}, nil
}

localfs.NewEntry(path, SnapshotFunc)
```

We would call it by:

```
localfs.NewEntry(path, localfs.NoSnapshot)
localfs.NewEntry(path, fssnapshot.SnapshotFuncFromPolicy(policyTree))
```

This decouples Kopia implementation from specifics of any snapshotting library - anything that matches this function signature can be used as long as it can create snapshots and close them.

fs/fssnapshotfs/snapshot_fs.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		// If we can't use snapshots, just use the filesystem directly
+		fmt.Printf("Error creating filesystem snapshot: %v\n", err)
+
+		return entry, nil
+	}
+
+	backuper, err := snapshoter.StartBackup(nil)
+	if err != nil {
+		// If we can't use snapshots, just use the filesystem directly
```
jkowalski, line 55: I think we should actually fail hard if snapshot is requested but could not be made.

pescuma: I like the idea of just leaving it enabled and it works where available. Maybe make it have 3 options: required, if available, no ?

fs/fssnapshotfs/snapshot_fs.go
pescuma: The reason I put the dependency on `policyTree` is that I wanted to allow the user to configure fs snapshots just for some folders. So it needs to track the policy at each `Child` and `IterateEntries` call, and these do not receive a policy as parameter. I can move it to a function, but I'd need to still store the policy inside there. And it would also always use the function with fs snapshots enabled, because the root folder may have fs snapshots disabled, but a folder inside it may change that. Or maybe I'm over complicating things and we could use only the configuration from the root folder.

About the function, I think it is a good idea to not only depend on fs_snapshots and a function (or something like it) there makes sense. But I think a wrapper around `localfs` is a better approach because it keeps the 2 concerns separated: interact with fs and map different paths.

fs/fssnapshotfs/snapshot_fs.go
jkowalski: sure, that's gonna work, as long as we don't silently fall back to inconsistent snapshots when the user thinks we're snapshotting.

fs/fssnapshotfs/snapshot_fs.go
jkowalski: i would not do snapshots within folders as it makes it much harder to reason about things - say you had N policies - consistency-wise you'd still have N snapshots that are not at the same point in time - let's just limit this to the root and keep it simple.

I really don't think we should be depending on policies in this layer.

fs/fssnapshotfs/snapshot_fs.go
jkowalski: BTW, when we fail to create FS snapshot and decide to move forward with it, this should be recorded somewhere.

In fact, we should probably record details of snapshot on `*snapshot.Manifest` in all cases, something like:

```
"rootEntry": {
   ...
},
"filesystemSnapshot": {
  "id": "..."
  "timestamp": "...",
  "method": "VSS",
}
```

or

```
"rootEntry": {
   ...
},
"filesystemSnapshot": {
  "method": "VSS",
  "error": {
    "message": "Unable to create VSS snapshot because ..."
   }
}
```

fs/fssnapshotfs/snapshot_fs.go
pescuma: The idea is not to have one snapshot per policy, but to re-use snapshots when possible. The problem with only considering the root snapshot is:
Suppose you ask kopia to snapshot /a and you have in /a/b another drive mounted and in /a/c a different one. The options at that point is (1) to snapshot all mounted drives inside /a or (2) snapshot those as we see them. I went with (2) because /a/b can be ignored by some parameter and we may need not to snapshot it. Because of that, I though it makes sense to also allow to choose that by policies. If you prefer, I can change it to to (1). 

As a side note, this may seen not applicable to Windows, but I discovered that it allows to mount drives at random places too (don't know if anyone uses it). For MacOS (and maybe in the future Linux) that is common concern.

pescuma:
@jkowalski I tried a different approach. Please take a look and tell me if it seems better.

fs/fssnapshotfs/snapshot_fs.go
pescuma: I added logging when creating/an error occurred. But I don't have the snapshot manifest at hand to add more info to it. Is there an easy way to do that? Another thing is that it can contain multiple snapshots (when there are different drives mounted inside the snapshot root), so this have to be a list.

pescuma: Ok, after the change in approach it became easy to do. Now it adds info to the manifest too.

philiplb:
This is great. "Direct" (no script copied from the documentation which may or may not outdate) VSS support was keeping me back from actually using Kopia :)

akash07k:
@pescuma  It's really wonderful.
I really loved Kopia but lack of VSS snapshots was making myself reluctant to use it.
I hope this PR gets merged soon and we can this added in Kopia

akash07k:
@jkowalski  Any idea that by when will it ready for testing?

cristihcd:
setting another policy "use filesystem snapshots" to no.

```
kopia policy set --use-filesystem-snapshots=required c:\path
Setting policy for user@system:c:\path
 - setting "use filesystem snapshots" to required.

kopia policy set --compression=zstd-better-compression c:\path
Setting policy for user@system:c:\path
 - setting "use filesystem snapshots" to no.
 - setting compression algorithm to zstd-better-compression
 ```

pescuma:
@cristihcd Thanks for the report. Fixed.

stephenjamieson:
Would love to see this merged 👍

akash07k:
@pescuma  @jkowalski  Any ideas till when it will be merged and shipped?

philiplb:
> Closed due to inactivity. Re-open and remove "stale" label if it should remain open for an additional period of time

Oh no, I was looking forward to this!

DoTheEvo:
Sad here too. Windows is huge enough thing on its own that having VSS integrated well in to the program itself would feel better than dealing with before and after actions and all the steps one has to take to get it working.

akash07k:
@pescuma Have you stopped working on it?

pescuma:
Yes, I'm not working on this anymore.

akash07k:
Oh, sad to know this
> Yes, I'm not working on this anymore.

PR #2335, opened 2022-08-22
feat(cli): Allow restore from snapshoted path

pescuma:
When using a path, it finds the last snapshot that contains that path.

One thing that is somewhat sketchy: the function `findSnapshotsForSource` is being used from another command. But it did not seem like a good idea to duplicate the function.

This also allows to use `\` in windows after a snapshot id.

### DIFF
--- a/cli/command_restore.go
+++ b/cli/command_restore.go
@@ -435,7 +435,7 @@ func tryToConvertPathToID(ctx context.Context, rep repo.Repository, source strin
 		return "", errors.Errorf("the source must contain a path element")
 	}
 
-	manifestIDs, relPath, err := findSnapshotsForSource(ctx, rep, si, map[string]string{})
+	manifestIDs, err := findSnapshotsForSource(ctx, rep, si, map[string]string{})
 	if err != nil {
 		return "", err
 	}
@@ -449,16 +449,19 @@ func tryToConvertPathToID(ctx context.Context, rep repo.Repository, source strin
 		return "", errors.Wrap(err, "unable to load snapshots")
 	}
 
-	manifest := findLastManifestWithPath(ctx, rep, manifests, relPath)
+	manifest := findLastManifestWithPath(ctx, rep, manifests, si.Path)
 	if manifest == nil {
 		return "", errors.Errorf("no snapshots contain data for %v", source)
 	}
 
-	if relPath != "" {
-		relPath = "/" + relPath
+	relPathParts, err := findRelativePathParts(manifest, si.Path)
+	if err != nil {
+		return "", err
 	}
 
-	result := string(manifest.ID) + relPath
+	relPathParts = append([]string{string(manifest.ID)}, relPathParts...)
+
+	result := filepath.ToSlash(filepath.Join(relPathParts...))
 
 	log(ctx).Infof("Restoring from snapshot %v...", result)
 
--- a/cli/command_restore.go
+++ b/cli/command_restore.go
@@ -17,6 +17,9 @@ import (
 	"github.com/kopia/kopia/internal/timetrack"
 	"github.com/kopia/kopia/internal/units"
 	"github.com/kopia/kopia/repo"
+	"github.com/kopia/kopia/repo/content/index"
+	"github.com/kopia/kopia/repo/object"
+	"github.com/kopia/kopia/snapshot"
 	"github.com/kopia/kopia/snapshot/restore"
 	"github.com/kopia/kopia/snapshot/snapshotfs"
 )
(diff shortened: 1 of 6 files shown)


jkowalski:
I think it's a good idea, but we need to develop it further. 

1. We need syntax to select a particular time. Here are some ideas, not sure which ones we want:

We can either embed it in a single string:

```
- myfile@2022-01-01
- myfile@latest
- myfile@yesterday
- myfile@january
- myfile@tuesday
- myfile@oldest
- myfile@2021
- myfile@2022-01-01T23:00:00Z
- myfile@2days-ago
```

The logic would look for `@<valid-timespec>` to trigger this behavior, not just a path by itself.

Alternatively have separate flags:

```
--snapshot-of <path>
--snapshot-time <timespec>
```

2. We should log the snapshot selection so that the user know what is going on.

3. We definitely need unit tests for this logic since it's pretty complex and a bunch of e2e tests.

Here's a more interesting question: should a single-argument restore work, where the output replaces the exact file specified:

```
$ kopia restore somefile.go
$ kopia restore somefile.go@2022-01-22
```

@Shrekster @julio-lopez any suggestions?

pescuma:
I think it's nice to be able to set a date. About the `@`, currently it supports user@machine:path, so it would be 2 `@` in there. Not sure if this is a problem though.

pescuma:
About 2, it does this: `log(ctx).Infof("Restoring from snapshot %v...", result)`

jkowalski:
I've been thinking about it more and I'm convinced name mangling would be problematic, how about  we make `--snapshot-time <timespec>` required if a filename is provided:

So the following 6 forms would be possible:

```
1. kopia restore k<dir-id> /tmp/output
2. kopia restore <manifest-id> /tmp/output
3. kopia restore username@hostname:/some/path/to/snapshotted/file.go /tmp/output --snapshot-time latest
4. kopia restore username@hostname:some/path/to/snapshotted/file.go --snapshot-time 2022-08-23T15:34:01
5. kopia restore /some/path/to/snapshotted/file.go /tmp/output --snapshot-time latest
6. kopia restore /some/path/to/snapshotted/file.go --snapshot-time latest
```

The 6th one would restore in-place replacing the provided file and would be the only case where a single argument is allowed.

What do you think?

pescuma:
I have too say that I can't see why the time should be required given that we are giving preference to IDs, but probably I'm missing something. I agree that for 5 it'd be better if it was required, but for the other cases I think using latest as default would work fine.

About the ambiguity, I had not thought of that. But you already have this problem in `snapshot list` (from where I copied this code) - this would only be consistent. Also, current approach uses the snapshot from the path closer to the file, but it could be extended to include all paths that have the file, so it would really pick the latest one and the ambiguity would be gone.

One last note: the free time that I have to contributing to kopia is nearing its end. I don't know if I'll be able to implement the date matching code. I'd suggest to implement an initial version with only the latest snapshot (maybe without 5), and then later extend it. What do you think?

pescuma:
I've implemented handling of all possible paths for a file and some timespec patterns (not all of the above), but I've made them optional.

pescuma:
Ok, I think it now handles everything pointed out. 

A few things to notice:
1. Both `snapshot list` and `restore` were changed to handle the case where one file is in multiple snapshot paths
2. Most of the timespec patters are implemented. We can add more latter, if needed
3. For some reason some tests are failing, but it does not seem related to this change

pescuma:
Ops, of course the test fails were my fault (I should know better by now). They should be fixed now. 

BUT the ARM tests are still failing... This time I think (hope) that it is not related to this PR. The error is: 
```
go build -ldflags "-s -w -X github.com/kopia/kopia/repo.BuildVersion=20220828.0.0- -X github.com/kopia/kopia/repo.BuildInfo= -X github.com/kopia/kopia/repo.BuildGitHubRepo=kopia/kopia" -o dist/kopia_linux_arm64/kopia -tags "" github.com/kopia/kopia
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```
Seems to be a problem of the go compiler trying to get information from git: https://github.com/microsoft/CBL-Mariner/issues/3273#issuecomment-1172982438

cli/command_restore.go
```diff
@@ -404,3 +445,234 @@ func (c *commandRestore) run(ctx context.Context, rep repo.Repository) error {
...
+}
+
+var timeAgoRE = regexp.MustCompile(`^(?:(\d{1,3})(?:y|year|years)-)?(?:(\d{1,3})(?:m|mo|month|months)-)?(?:(\d{1,3})(?:d|day|days)-)?ago$`)
+
+// computeMaxTime returns the first time after the max allowed.
+func computeMaxTime(timespec string) (time.Time, error) {
+	now := clock.Now()
+
+	if timespec == "yesterday" {
+		return time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, now.Location()), nil
```
jkowalski, line 540: I would use `time.Local` since instead of `now.Location()`

jkowalski (approved):
LGTM, thanks!

PR #2337, opened 2022-08-22
test(general): Fix upgrade portion of robustness test

chaitalisg:
Moved the upgrade repository function used in the robustness test to `snapmeta` package. This gets rid of the poison blob error  the test hit earlier this month.

The root cause of the error was separate `kopia.config` used by the restore operation and upgrade operation. With this change, all operations in the test use the same `kopia.config`.


tests/robustness/snapmeta/kopia_snapshotter.go
```diff
@@ -178,3 +180,23 @@ func (ks *KopiaSnapshotter) ConnectOrCreateFilesystemWithServer(serverAddr, repo
 func (ks *KopiaSnapshotter) Cleanup() {
 	ks.snap.Cleanup()
 }
+
+// GetRepositoryStatus returns the repository status in JSON format.
+func (ks *KopiaSnapshotter) GetRepositoryStatus() (cli.RepositoryStatus, error) {
+	a1, _, _ := ks.snap.Run("repository", "status", "--json")
```
adowair, line 186: Should we check for error here?
```suggestion
	a1, _, err := ks.snap.Run("repository", "status", "--json")
	if err != nil {
	  return nil, err
	}
```

tests/tools/kopiarunner/kopia_snapshotter_upgrade_test.go
```diff
@@ -1,38 +0,0 @@
...
-	ks.UpgradeRepository(repoDir)
-
-	rs, err = ks.GetRepositoryStatus()
-	require.NoError(t, err)
-
-	got := rs.ContentFormat.MutableParameters.Version
-	require.Equal(t, got, format.Version(2), "The format version should be upgraded to 2.")
-
-	require.NotEqual(t, got, prev, "The format versions should be different.")
-}
```
adowair, line 38: Forgive my ignorance, why is this removed?

adowair (commented):
Looks good for the most part, two questions

tests/tools/kopiarunner/kopia_snapshotter_upgrade_test.go
chaitalisg: I moved the functions GetRepositoryStatus and UpgradeRepository to snapmeta package. May be I can add a test there.

chaitalisg: Actually, both the functions are exercised in the robustness test, so, do we need another test here?

adowair: Unit tests are all always a good idea, if they are applicable. At the very least we can check that `GetRepositoryStatus` does not return error on `Run()` or parse.

chaitalisg: Added test

adowair (approved):

PR #2343, opened 2022-08-24
feat(repository): introduced fs.UTCTimestamp

### LINKED ISSUE #2342: Optimize `DirEntry` by not storing timezone pointers?
A small-ish optimisation idea:

In

https://github.com/kopia/kopia/blob/dd1a6aff8d5a1e7653f402593ad9f7fb65f246e6/snapshot/manifest.go#L122-L133

`ModTime time.Time` is [this struct](https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/time/time.go;l=129-150):

```go
type Time struct {
  wall uint64
  ext  int64
  loc *Location
}
```

This might be overkill for storing file modification times; single `uint64` is usually enough for storing them on most file systems, and a translation to UTC could be done at storage time.

This would have the benefit of reducing emory overhead of `DirEntry` (less small objects allocated because the `*Location` pointer is gone), and also reduce GC traversal times: See e.g. https://segment.com/blog/allocation-efficiency-in-high-performance-go-services/:

> Now the `retryItem` doesn’t contain any pointers. This dramatically reduces the load on the garbage collector as the entire footprint of `retryItem` is knowable at compile time

jkowalski:
Fixes #2342

### DIFF
--- a/cli/command_snapshot_copy_move_history.go
+++ b/cli/command_snapshot_copy_move_history.go
@@ -108,13 +108,13 @@ func (c *commandSnapshotCopyMoveHistory) run(ctx context.Context, rep repo.Repos
 
 		if snapshotExists(dstSnapshots, dstSource, manifest) {
 			if isMoveCommand && !c.snapshotCopyOrMoveDryRun {
-				log(ctx).Infof("%v (%v) already exists - deleting source", dstSource, formatTimestamp(manifest.StartTime))
+				log(ctx).Infof("%v (%v) already exists - deleting source", dstSource, formatTimestamp(manifest.StartTime.ToTime()))
 
 				if err := rep.DeleteManifest(ctx, manifest.ID); err != nil {
 					return errors.Wrap(err, "unable to delete source manifest")
 				}
 			} else {
-				log(ctx).Infof("%v (%v) already exists", dstSource, formatTimestamp(manifest.StartTime))
+				log(ctx).Infof("%v (%v) already exists", dstSource, formatTimestamp(manifest.StartTime.ToTime()))
 			}
 
 			continue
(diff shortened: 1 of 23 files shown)


julio-lopez (commented):
High level questions and comments:

1. Are there any compatibility implications here? While it does not seem to be the case, does it change the serialized representation of the directory entries? is there anything to guarantee compatibility here?

2. In principle, this seems a good change and there does not appear to be a reason for not doing it (pardon the double negative). Having said that:
    - Is there any data showing that this is a problem / is causing memory pressure?
    - Does this change reduce the memory pressure in any way? Or what is the gain? Only based on observation, it seems that the total number of `time.Time` objects instantiated while creating a snapshot is the same, since these are the result of traversing the source FS. Perhaps I'm missing something.

jkowalski:
Your question 1 got me thinking and I double checked it - the times in JSON appear to be always serialized using local timezone (e.g. "mtime":"2022-08-13T18:19:13.932475613-07:00") even when we construct the time using `time.Unix()` like this PR does. Since we're not setting the timezone explicitly this should not be a visible change. I compared directory format before and after this change and they are the same.

Regarding question 2, @nh2 has been trying to use Kopia with huge directories (200M files+) and provided this suggestion in https://github.com/kopia/kopia/issues/2342. 

For his use case this will save 16 bytes per dir entry (time.Time was 24, UTCTimestamp is 8 bytes) for each timestamp (which multiplied by 200M is >3GB of RAM) and reduce GC traversal time, since dirEntry will have 1 pointer instead of 2. Not sure if this GC overhead is measurable, but I would imagine at this scale it may matter.

Note that this change won't make Kopia support 200M entries per directory by itself, but is a step in the right direction. There are more optimizatons we can do here that will follow.

julio-lopez:
It'd be good to consider normalizing the (serialized) timestamps to UTC. In a separate PR that is.

PR #2357, opened 2022-09-01
feat(repository): Allow clock drift to be specified on upgrade.

aaron-kasten:
When upgrading a repository, allow the user to specify a clock-drift with `--max-permitted-clock-drift`.


adowair:
Thanks for all the typo fixes, shall we split that into a separate PR?

adowair (approved):
Approved, with a nit about splitting off typos into its own PR–I believe rebase+merge is not enabled for this repository.

adowair:
> LGTM, minor nit:
> 
> Would it help to verify that the users are setting the max-allowed-drift to a minimum value unless you specify the `--force` flag ... ?

I suppose we can let it only be zero if `--allow-unsafe-upgrade` is set.

jkowalski (approved):

PR #2364, opened 2022-09-03
feat(repository): added Set and Map backed by custom on-disk hashtable

jkowalski:
This is just a new data structure, not used yet.

# Background:

We used to use `sync.Map` to keep track of various content IDs visited in various places. This is quite slow and wastes a lot of memory due to fragmentation.

For example on 2021 MacBook Pro M1 with 16 GB RAM, sync.Map can only keep ~55M keys, each of 32 bytes (SHA256 hashes) due to excessive memory fragmentation (without fragmentation it should be <2GB)

This custom implementation uses explicitly managed contiguous buffers, is faster than sync.Map and only uses 8 bytes per key in RAM and offloads all payload to memory-mapped files which are allocated in segments.

On a Core i5 machine with 32GB RAM, this supports ~100M Put() operations per minute (32-byte keys).

### DIFF
--- a/internal/bigmap/bigmap_internal.go
+++ b/internal/bigmap/bigmap_internal.go
@@ -29,6 +29,7 @@ const (
 	defaultMemorySegmentSize    = 18 * 1e6   // 18MB enough to store >1M 16-17-byte keys
 	defaultFileSegmentSize      = 1024 << 20 // 1 GiB
 	defaultInitialSizeLogarithm = 20
+	mmapFileMode                = 0o600
 
 	// grow hash table above this percentage utilization, higher values (close to 100) will be very slow,
 	// smaller values will waste memory.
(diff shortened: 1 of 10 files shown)


adowair (commented):
First pass, this is great. I wonder–would it be better to spin this off into its own repository? It looks fairly self-contained

internal/bigmap/bigmap_internal.go
```diff
@@ -0,0 +1,476 @@
...
+	1<<59 - 871,
+	1<<60 - 453,
+	1<<61 - 465,
+	1<<62 - 273,
+	1<<63 - 471,
+}
+
+type entry struct {
+	segment uint32 // 0-empty, otherwise index into internalMap.segments+1
+	offset  uint32 // offset within a segment
```
julio-lopez, line 147: @jkowalski thinking about this...

Are the entries (key + value) expected to be of fixed size?

If so, given that this map is specialized for this use case, etc.,
Would it make sense to make the key and value sizes be parameters of the map, and replace `offset` with an "`index`" into the segment? The offset for the entry would then be computed using the entry size.

Thoughts?

PR #2396, opened 2022-09-11
feat(repository): reduced mutex contention on committed index cache

jkowalski:


nh2:
This change improves CPU utilisation from 110% to 1600% during `kopia maintenance` GC on my 8-core, 16-thread machine.

PR #2437, opened 2022-09-24
feat(snapshots): Added option to ignore file by extended attributes

### LINKED ISSUE #1539: Automatically exclude Time Machine excluded files
From https://kopia.discourse.group/t/automatically-exclude-time-machine-excluded-files/368

> On macOS applications can mark transient and cache files with an extended attribute com.apple.metadata:com_apple_backup_excludeItem to inform time machine that those files and directories should not be backed up. Since the application developers have much better idea about what files need and need not be preserved this approach works better than centralized exclusion list.
> 
> Some notable third party backup tools (Duplicacy, Arq, CrashPlan) support exclusion by that attribute as well.
> 
> Kopia should too. This cuts down drastically on the need for the user to write (and maintain!) elaborate exclusion patters.

pescuma:
It allows to set any attribute, and use as default the attribute used in MacOS.

This fixes #1539.

### DIFF
--- a/cli/command_policy_set_files.go
+++ b/cli/command_policy_set_files.go
@@ -25,6 +25,11 @@ type policyFilesFlags struct {
 	policyOneFileSystem string
 
 	policyIgnoreCacheDirs string
+
+	// Ignore extended attributes rules.
+	policySetAddIgnoreExtendedAttributes    []string
+	policySetRemoveIgnoreExtendedAttributes []string
+	policySetClearIgnoreExtendedAttributes  bool
 }
 
 func (c *policyFilesFlags) setup(cmd *kingpin.CmdClause) {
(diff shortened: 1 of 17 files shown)


pescuma:
I can't find out why the windows test is failing. When I run on my machine it works. Sorry about that and I'd appreciate some help find out what is happening.

jkowalski:
This is great, I have some suggestions:

- we should ensure attribute fetching does not slow down snapshots, if it does, we should make it opt-in, via policies
- since we're scanning xattrs, it probably makes sense to store them in the repo - policy driven
- restore and mount should probably support xattr - we need to carefully think of cross-OS restores and mounting (linux-to-windows, etc.)

pescuma:
>     * we should ensure attribute fetching does not slow down snapshots, if it does, we should make it opt-in, via policies

The current implementation is using policies for that, and is opt in for all OSes, except for MacOS where it is opt out. I think it makes sense as a good default, but I agree that performance is an issue. I'll do some performance tests and report back 


>     * since we're scanning xattrs, it probably makes sense to store them in the repo - policy driven

Makes sense, but I think this could be done in a separate PR. Also, we need to think about the format. Currently I'm only looking at if the attribute is there, but they also have a []byte value. Not sure how it's used though.


>     * restore and mount should probably support xattr - we need to carefully think of cross-OS restores and mounting (linux-to-windows, etc.)

I'm not the person for this one. I don't know enough about xattrs to answer that. Also, the xattr lib used does not support windows.

pescuma:
Timing data for MacOS (Monterey) snapshoting 22GB:
- Version 0.11.3:
   - First run: 65.21s user 9.92s system 495% cpu 15.174 total
   - Second run (no changes): 0.17s user 0.05s system 116% cpu 0.189 total
- This PR with default config, ignore by attributes enabled:
   - First run: 66.20s user 10.86s system 470% cpu 16.386 total
   - Second run (no changes): 1.36s user 0.72s system 77% cpu 2.702 total
 
So it has a bit impact on run time when no data must be transferred (1.36s x 0.17s), but it is a lot less than the time used to write data to disk.

probablykasper:
> we should ensure attribute fetching does not slow down snapshots, if it does, we should make it opt-in, via policies

Also be worth considering that not checking the attributes can slow down snapshots too, by including unnecessary files

probablykasper:
What's the status on this @jkowalski?

pocketpixels:
This sounds like a great feature for Mac users, I'd really love to have this. 
It'd be great to re-open this and get it merged into a release (if it works as expected).

PR #2464, opened 2022-09-30
fix(server): Respect manual scheduling policy

pescuma:
The server was ignoring when "Manual Snapshots Only" was set to true in policy. This fixes that.


jkowalski (changes requested):
Can you add a simple test?

jkowalski (approved):

PR #2496, opened 2022-10-13
feat(repository): add more split size options: 128K, 256K, 512K

le-duane:
Hi,

I am looking into using Kopia for Postgres backups. Since Postgres makes updates in 8k blocks, I have found better efficiency in using smaller split sizing.


le-duane:
Thanks @jkowalski! 

In case it's of any interest: in our synthetic tests using a 5 GB DB, with 90 iterations at 3% data change rate, we saw a 50% space savings at 256K vs 1MB split size, while maintaining the same runtime/CPU usage.

Issue #2492, opened 2022-10-10
[Feature Request] CLI/UI: Support binary (base-2) display of storage values


sowaret:
I'm experimenting with different backup solutions, and I hit a bit of a speed bump evaluating Kopia by not realizing that storage values are displayed in decimal (base-10). Most solutions I've seen seem to settle on `1 GB = 1024 MB`. An option to display values in binary units would be useful outside of e.g. benchmarks, which is the only place I've encountered them so far.

I looked through the documentation and didn't find any indication that this was currently supported.

(Kopia v0.12.1 on FreeBSD)

PR #118, opened 2019-10-03
site: added version compatibility page

jkowalski:
Fixes #108

Preview link: https://deploy-preview-118--kopia.netlify.com/docs/compatibility/

PR #2502, opened 2022-10-14
feat(cli): Support displaying storage values in base-2 [#2492]

sowaret:
CLI functionality in support of #2492 (equivalent PR for UI: [#118](https://github.com/kopia/htmlui/pull/118)).

- Pull `KOPIA_BYTES_STRING_BASE_2` configuration from environment
- Add `BytesStringBase2` field to `UIPreferences` struct to support change in frontend config file
- Fix typo in `units.go`

Thanks for taking a look.

### DIFF
--- a/internal/units/units.go
+++ b/internal/units/units.go
@@ -46,10 +46,12 @@ func BytesStringBase2(b int64) string {
 	return toDecimalUnitString(float64(b), 1024.0, base2UnitPrefixes, "B")
 }
 
+// BytesString formats the given value as bytes with the unit provided from the environment.
 func BytesString(b int64) string {
 	if v, _ := strconv.ParseBool(os.Getenv(bytesStringBase2Envar)); v {
 		return BytesStringBase2(b)
 	}
+
 	return BytesStringBase10(b)
 }
 
--- a/internal/units/units.go
+++ b/internal/units/units.go
@@ -47,7 +47,7 @@ func BytesStringBase2(b int64) string {
 }
 
 func BytesString(b int64) string {
-	if bool, err := strconv.ParseBool(os.Getenv(bytesStringBase2Envar)); err == nil && bool {
+	if v, _ := strconv.ParseBool(os.Getenv(bytesStringBase2Envar)); v {
 		return BytesStringBase2(b)
 	}
 	return BytesStringBase10(b)
--- a/internal/units/units.go
+++ b/internal/units/units.go
@@ -14,6 +14,10 @@ var (
 	base2UnitPrefixes  = []string{"", "Ki", "Mi", "Gi", "Ti"}
 )
 
+const (
+	bytesStringBase2Envar = "KOPIA_BYTES_STRING_BASE_2"
+)
+
 func niceNumber(f float64) string {
 	return strings.TrimRight(strings.TrimRight(fmt.Sprintf("%.1f", f), "0"), ".")
 }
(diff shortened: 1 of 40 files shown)


internal/units/units.go
```diff
@@ -40,6 +42,15 @@ func BytesStringBase2(b int64) string {
 	return toDecimalUnitString(float64(b), 1024.0, base2UnitPrefixes, "B")
 }
 
+func BytesStringBaseEnv(b int64) string {
```
jkowalski, line 45: `BytesString` is probably sufficient

jkowalski: also it might make sense to make it a function pointer and statically initialize it:

```
var BytesString func(int64) string

func init() {
   if ... {
     BytesString = BytesStringBase2
   } else {
     BytesString = BytesStringBase10
   }
}
```

sowaret: Good callouts.

Re: initialization, my impression of Go on the testing front would be that the `units` package would be initialized once, without a chance to set the envvar, and at that point the base-2 case wouldn't be accessible for testing.

internal/units/units.go
```diff
@@ -40,6 +46,15 @@ func BytesStringBase2(b int64) string {
 	return toDecimalUnitString(float64(b), 1024.0, base2UnitPrefixes, "B")
 }
 
+func BytesString(b int64) string {
+	if v := os.Getenv(bytesStringBase2Envar); v != "" {
+		if bool, err := strconv.ParseBool(v); err == nil && bool {
+			return BytesStringBase2(b)
+		}
+	}
```
sowaret, line 54: Think it's extraneous to check for a non-empty value as we're parsing a bool anyways:
```suggestion
func BytesString(b int64) string {
	if bool, err := strconv.ParseBool(os.Getenv(bytesStringBase2Envar)); err == nil && bool {
		return BytesStringBase2(b)
	}
```

internal/units/units.go
```diff
@@ -40,6 +46,13 @@ func BytesStringBase2(b int64) string {
 	return toDecimalUnitString(float64(b), 1024.0, base2UnitPrefixes, "B")
 }
 
+func BytesString(b int64) string {
+	if bool, err := strconv.ParseBool(os.Getenv(bytesStringBase2Envar)); err == nil && bool {
```
jkowalski, line 50: nit: you can ignore err check, bool can't be true if error is != nil

if v, _ := strconv.ParseBool(os.Getenv(...)); v {
  
}

tests/tools/kopiaclient/kopiaclient.go
```diff
@@ -143,7 +143,7 @@ func (kc *KopiaClient) SnapshotRestore(ctx context.Context, key string) ([]byte,
 		return nil, err
 	}
 
-	log.Printf("restored %v", units.BytesStringBase10(int64(len(val))))
+	log.Printf("restored %v", units.BytesStringBaseEnv(int64(len(val))))
```
jkowalski, line 146: still to be renamed, I guess?

tests/tools/kopiaclient/kopiaclient.go
```diff
@@ -100,7 +100,7 @@ func (kc *KopiaClient) SnapshotCreate(ctx context.Context, key string, val []byt
 		return errors.Wrap(err, "cannot get manifest")
 	}
 
-	log.Printf("snapshotting %v", units.BytesStringBase10(atomic.LoadInt64(&man.Stats.TotalFileSize)))
+	log.Printf("snapshotting %v", units.BytesStringBaseEnv(atomic.LoadInt64(&man.Stats.TotalFileSize)))
```
jkowalski, line 103: still to be renamed, I guess?

sowaret:
Will continue to look through tests when I have the time

jkowalski:
those are linter failures, just run `make lint` and fix the issues until it comes back clean

PR #2526, opened 2022-10-28
chore(ci): upgrade golang to 1.19.2 and linter to 1.50.1

jkowalski:
Lack of generics support is blocking more and more dependency upgrades so this unblocks that.

Had to temporarily disable `checklocks` linter until it's fixed upstream.


julio-lopez (approved):
🥇Thanks for doing this.

PR #2527, opened 2022-10-28
refactor(general): introduced generics to reduce boilerplate code

jkowalski:
This removes tons of boilerplate code around:

- retry loop
- connection management
- storage registration


jeremyje (commented):
I think splitting out the 1.19 ci upgrade and if possible the linter enhancements would make the review simpler.

.golangci.yml
```diff
@@ -85,16 +86,24 @@ linters:
...
     - tparallel
+    - varcheck
     - varnamelen # this one may be interesting, but too much churn
+    - wastedassign
     - whitespace
 
 issues:
   exclude-use-default: false
   exclude-rules:
+    - path: \.pb\.go
```
jeremyje, line 104: Nah I would basically encourage go fmt before reach commit here. Typically having a `make presubmit` command that line and formats is a good way to avoid needing these rules

julio-lopez: That seems like a good idea. Can you push a PR with the approach you are suggesting?
OTOH, it seems OK to have exceptions for generated code.

PR #2528, opened 2022-10-28
feat(validate index): new `repository upgrade` sub-command to validate indices.

aaron-kasten:
This PR adds functionality to validate indices.

The full feature will add a validation phase to the current V0 to V1 index upgrade command.  This phase will be run before the `commitUpgrade` phase.  If validation fails, console output will be generated that indicates the failed blobs and the command exited, leaving the repository locked.  The user may then either rollback and re-run the upgrade, or continue the upgrade (by re-executing `repo upgrade begin` with the `--force` flag).  The user may also choose to change the index blob(s) using a Kopia blob-editor tool and then re-validate the index (to check the work) using `repo upgrade validate`.

Re-running `repo upgrade begin` command with the `--force` flag will commit the existing index and unlock the repository, making the V0 index blob available for reclamation.

DO NOT MERGE

This PR is a proof-of-concept, designed to promote discussion.  Maintainers may choose to approve development into a full feature (additional testing, completion of command line invocation, and finalize report output).

### DIFF
--- a/cli/command_repository_upgrade.go
+++ b/cli/command_repository_upgrade.go
@@ -81,13 +81,8 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
 }
 
 func (c *commandRepositoryUpgrade) validateAction(ctx context.Context, rep repo.DirectRepositoryWriter) error {
-
 	_, err := rep.ContentManager().SharedManager.ValidateIndexes(ctx)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func (c *commandRepositoryUpgrade) forceRollbackAction(ctx context.Context, rep repo.DirectRepositoryWriter) error {
--- a/cli/command_repository_upgrade.go
+++ b/cli/command_repository_upgrade.go
@@ -72,7 +72,22 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
 
 	rollbackCmd.Action(svc.directRepositoryWriteAction(c.forceRollbackAction))
 
+	validateCmd := parent.Command("validate", "Validate the upgraded indexes.")
+	validateCmd.Flag("force", "Force rollback the repository upgrade, this action can cause repository corruption").BoolVar(&c.forceRollback)
+
+	validateCmd.Action(svc.directRepositoryWriteAction(c.validateAction))
+
 	c.svc = svc
+}
+
+func (c *commandRepositoryUpgrade) validateAction(ctx context.Context, rep repo.DirectRepositoryWriter) error {
+
+	_, err := rep.ContentManager().SharedManager.ValidateIndexes(ctx)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func (c *commandRepositoryUpgrade) forceRollbackAction(ctx context.Context, rep repo.DirectRepositoryWriter) error {
--- a/repo/content/committed_read_manager.go
+++ b/repo/content/committed_read_manager.go
@@ -2,6 +2,7 @@ package content
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sync"
@@ -21,6 +22,7 @@ import (
 	"github.com/kopia/kopia/repo/blob/filesystem"
 	"github.com/kopia/kopia/repo/blob/sharded"
 	"github.com/kopia/kopia/repo/compression"
+	"github.com/kopia/kopia/repo/content/index"
 	"github.com/kopia/kopia/repo/format"
 	"github.com/kopia/kopia/repo/hashing"
 	"github.com/kopia/kopia/repo/logging"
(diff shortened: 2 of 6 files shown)


jkowalski:
I think direction-wise this is ok, but I would move implementation to be mostly in the CLI without modifying the content package much. Definitely revert DiffInfo() method as it is too specific and does not belong in this package.

cli/command_repository_upgrade.go
```diff
@@ -72,9 +72,24 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
 
+func (c *commandRepositoryUpgrade) validateAction(ctx context.Context, rep repo.DirectRepositoryWriter) error {
+
+	_, err := rep.ContentManager().SharedManager.ValidateIndexes(ctx)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
Shrekster, line 91: nit;

```suggestion
	_, err := rep.ContentManager().SharedManager.ValidateIndexes(ctx)
	return err
}
```

repo/content/committed_read_manager.go
```diff
@@ -115,6 +117,81 @@ type SharedManager struct {
...
+	ibis0, _, err := sm.indexBlobManagerV0.listActiveIndexBlobs(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, ibi0 := range ibis0 {
+		// skip the V0 blob poison that is used to prevent client reads.
+		if ibi0.BlobID == legacyIndexPoisonBlobID {
+			continue
+		}
```
Shrekster, line 157: Should not this check belong in the layer within `listActiveIndexBlobs()` above ? Conceptually, a poison blob is not part of the active index.

cc: @jkowalski

cli/command_repository_upgrade.go
```diff
@@ -72,9 +72,19 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
+	validateCmd.Flag("force", "Force rollback the repository upgrade, this action can cause repository corruption").BoolVar(&c.forceRollback)
+
+	validateCmd.Action(svc.directRepositoryWriteAction(c.validateAction))
+
 	c.svc = svc
 }
 
+func (c *commandRepositoryUpgrade) validateAction(ctx context.Context, rep repo.DirectRepositoryWriter) error {
+	_, err := rep.ContentManager().SharedManager.ValidateIndexes(ctx)
+	return err
```
Shrekster, line 85: Should not we check if the returned report length is more than zero and fail the upgrade if it is ?

repo/format/upgrade_lock_test.go
```diff
@@ -116,6 +117,77 @@ func TestFormatUpgradeCommit(t *testing.T) {
 	require.EqualError(t, env.RepositoryWriter.FormatManager().RollbackUpgrade(ctx), "no upgrade in progress")
 }
 
+func TestFormatUpgradeValidate(t *testing.T) {
```
Shrekster, line 120: might be a good idea to simulate a negative test-case as well to ensure upgrade fails if validate fails ..

Since all the blobs are in-memory, you can access the blob-storage and corrupt it in a way you would get a report of length more than zero.

repo/content/content_manager_indexes.go
```diff
@@ -79,8 +79,7 @@ func (sm *SharedManager) CompactIndexes(ctx context.Context, opt CompactOptions)
 	return nil
 }
 
-// ParseIndexBlob loads entries in a given index blob and returns them.
-func ParseIndexBlob(ctx context.Context, blobID blob.ID, encrypted gather.Bytes, crypter crypter) ([]Info, error) {
+func ParseIndexBlob(blobID blob.ID, encrypted gather.Bytes, crypter crypter) ([]Info, error) {
```
adowair, line 82: nit: looks like this doc-comment was removed by accident
```suggestion
// ParseIndexBlob loads entries in a given index blob and returns them.
func ParseIndexBlob(blobID blob.ID, encrypted gather.Bytes, crypter crypter) ([]Info, error) {

File: repo/content/committed_read_manager.go [hunk 6]
```diff
@@ -115,6 +117,75 @@ type SharedManager struct {
...
+	v[i] = iif
+	m[iif.GetContentID()] = v
+}
+
+func (sm *SharedManager) ValidateIndexes(ctx context.Context) ([]string, error) {
+
+	d := gather.WriteBuffer{}
+	indexEntries := map[ID][2]Info{}
+
+	ibis0, _, err := sm.indexBlobManagerV0.listActiveIndexBlobs(ctx)
```
adowair, line 148: nit: can we rename these (`ibis0, `ibi0`, `iifs`, etc.) to be more descriptive? Say, `activeIndexBlobsV0` or `indexBlobsV0`?

repo/content/index/info.go
```diff
@@ -90,3 +91,32 @@ func ToInfoStruct(i Info) *InfoStruct {
...
+		qs = append(qs, fmt.Sprintf("mixed EncryptionKeyIDs: %v %v", i0.GetEncryptionKeyID(), i1.GetEncryptionKeyID()))
+	}
+	if i0.GetDeleted() != i1.GetDeleted() {
+		qs = append(qs, fmt.Sprintf("mixed Deleted flags: %v %v", i0.GetDeleted(), i1.GetDeleted()))
+	}
+	if i0.GetTimestampSeconds() != i1.GetTimestampSeconds() {
+		qs = append(qs, fmt.Sprintf("mixed TimestampSeconds: %v %v", i0.GetTimestampSeconds(), i1.GetTimestampSeconds()))
+	}
+	return qs
+}
```
adowair, line 124: nit: how about "mismatched" instead of "mixed"? I think that makes more sense given the context

cli/command_repository_upgrade.go
```diff
@@ -72,9 +72,24 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
 
 	rollbackCmd.Action(svc.directRepositoryWriteAction(c.forceRollbackAction))
 
+	validateCmd := parent.Command("validate", "Validate the upgraded indexes.")
+	validateCmd.Flag("force", "Force rollback the repository upgrade, this action can cause repository corruption").BoolVar(&c.forceRollback)
```
adowair, line 76: I'm not sure I understand how `--force` relates to its description. How about `--rollback-on-failure` instead?

cli/command_repository_upgrade.go
Shrekster: Noticed that we have not yet hooked this up in a way that will prevent `commitUpgrade()`. I am guessing you want to do that in a separate PR ?

aaron-kasten:
> I think direction-wise this is ok, but I would move implementation to be mostly in the CLI without modifying the content package much. Definitely revert DiffInfo() method as it is too specific and does not belong in this package.

@jkowalski note that the reason why its on the shared manager is that most of the variables and methods it uses are private.  Should I make those public so they can be exported to the CLI?

cli/command_repository_upgrade.go
adowair: From our offline conversation, I understand now that when `--force` is specified, any validation errors will be ignored and the repository unlocked. The description of the command above does not match up with that and needs to be reworded.

repo/content/committed_read_manager.go
aaron-kasten: @Shrekster agreed.  Note that the method that returns these indices is private, so the intent may be to list _all_ active index blobs of that version (which would necessarily include the poison blob).

repo/content/committed_read_manager.go
```diff
@@ -115,6 +117,81 @@ type SharedManager struct {
...
+func assign(iif Info, i int, m map[ID][2]Info) {
+	v := m[iif.GetContentID()]
+	v[i] = iif
+	m[iif.GetContentID()] = v
+}
+
+// ValidateIndexes returns an array of strings (report) that describes differences between
+// the V0 index blob and V1 index blob content.  This is used to check that the upgraded index
+// (V1 index) reflects the content of the old V0 index.
+func (sm *SharedManager) ValidateIndexes(ctx context.Context) ([]string, error) {
```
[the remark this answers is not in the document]
aaron-kasten: @Shrekster that works for me.

cli/command_repository_upgrade.go
aaron-kasten: @Shrekster yes.  This is code to show how an index comparison can be done.  Reporting the results is viewed as low-risk and so was not included in this code (proof of concept of index comparison).

repo/format/upgrade_lock_test.go
aaron-kasten: @Shrekster agreed.  Unsure as of yet how that happens.  Likely to depend on the final implementation of the index checker.

repo/content/content_manager_indexes.go
aaron-kasten: @adowair yes... your suggestion was committed - thank you.

cli/command_repository_upgrade.go
aaron-kasten: @adowair the code is a proof of concept for index comparison - command line wasn't really considered for this PR.

Issue #2419, opened 2022-09-18
Suddently cannot connect to my repository; a problem with 'cancel connection'; 'unconnected' repository shows even after recreating the repository; backups did not run for a week


LinuxOnTheDesktop:
I have been using - and appreciating! - Kopia for a while (on Windows 10; Kopia version 0.11.3; using B2). 

Today when I tried to open the main Kopia window the program told me, persistently, that it is 'initializing' and I see this in the log:

> unable to open repository: get password: password prompt error: The handle is invalid.

My Internet connection is working and I have not (to my knowledge) recently messed with my cloud storage setup. Call this problem '(1)'.

In an attempt to fix things, I chose Kopia's 'cancel connection' option - which did not cancel my current attempt to connect to my repository (which is what I expected) but instead delete the repository information from Kopia. So here we have problem (2): Some sort of warning should be given, surely, before the program does that.

(3) After re-creating the repository, within Kopia, I managed to connect to it. Everything worked, except that I see this within my tray menu:

![image](https://user-images.githubusercontent.com/47056543/190913128-6710a81e-2a19-4824-a598-236cdb2477ab.png)

Should the 'unconnected' element not have disappeared from the menu?

Finally: (4) it seems that my backups did not run - they failed to run - for a week; and I received no notification of this. Should I have used Kopia to set up some sort of notification? Surely the default should be some sort of message.

LinuxOnTheDesktop:
I would welcome. for one thing, advice on how to remove that '<not connected>' entry in the tray menu. (Clicking it yields a window that contains a 'cancel' button, but clicking that button does not remove the menu entry.)

LinuxOnTheDesktop:
Bump.

lupusA:
Hi @LinuxOnTheDesktop, 

regarding point 3 - this a consequence of the current design. There is one primary repository which cannot be deleted once disconnected. It shows up at the top. If you connect to another repository (B2) and disconnect the primary repository it stays within the list. The code does not allow to delete the primary (eg. the first) repository. 

What you can do is, to disconnect from B2 and then use the primary repository to connect to B2 again. Then it will only show one entry. 


Cheers,

lupusA:
Hi @LinuxOnTheDesktop, 

please do the following:

1. Disconnect from B2 repository and close the window. 
2. Open Kopia from the system tray and check if only one repository is shown within the list.
3. If so, click on "not connected" and reconnect within the newly openend window to your B2 repository. 

[...]

PR #2543, opened 2022-11-02
fix(repository): Manifest label comparison when value is empty

ashmrtn:
Allow matching on labels with empty values. Previously, the "empty value" for `map[string]string` ("") would match the empty value of the label key, causing the system to return all manifests. This separates key existence from value equality to handle labels with empty values.


LinuxOnTheDesktop:
Hi there. Might this commit fix problem #2419? More specifically (for, unfortunately perhaps, that bug report treats of more than one problem), might the commit fix the problem of undeleteable and 'unconnected' repositories? Thanks.

PR #2548, opened 2022-11-03
feat(snapshot): Add labels to snapshot checkpoint manifests

ashmrtn:
Allow adding labels to snapshot manifests created by checkpoint. This aids in finding these checkpoints later on and in a more granular fashion because `Repository.FindManifests` can now be used instead of just `snapshot.ListSnapshots`

For backwards compatibility, none of the internal calls of `Upload` populate checkpoint labels, leaving the CLI output undisturbed

### DIFF
--- a/snapshot/snapshotfs/upload.go
+++ b/snapshot/snapshotfs/upload.go
@@ -464,10 +464,10 @@ func newDirEntry(md fs.Entry, fname string, oid object.ID) (*snapshot.DirEntry, 
 }
 
 // uploadFileWithCheckpointing uploads the specified File to the repository.
-func (u *Uploader) uploadFileWithCheckpointing(ctx context.Context, relativePath string, file fs.File, pol *policy.Policy, sourceInfo snapshot.SourceInfo) (*snapshot.DirEntry, error) {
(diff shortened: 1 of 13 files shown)


snapshot/snapshotfs/upload.go
```diff
@@ -1192,11 +1193,13 @@ func (u *Uploader) maybeOpenDirectoryFromManifest(ctx context.Context, man *snap
 
 // Upload uploads contents of the specified filesystem entry (file or directory) to the repository and returns snapshot.Manifest with statistics.
 // Old snapshot manifest, when provided can be used to speed up uploads by utilizing hash cache.
+// labels is applied to every checkpoint of a backup.
 func (u *Uploader) Upload(
 	ctx context.Context,
 	source fs.Entry,
 	policyTree *policy.Tree,
 	sourceInfo snapshot.SourceInfo,
+	labels map[string]string,
```
jkowalski, line 1202: this looks like an optional parameter with `nil` being a reasonable default, can we add a field to `Uploader` instead?

jkowalski (approved):

PR #2561, opened 2022-11-08
feat(repository): new `repository upgrade` sub-command to validate indices.

aaron-kasten:
This PR adds functionality to validate indices and replaces #2528.

The full feature will add a validation phase to the current V0 to V1 index upgrade command.  This phase will be run before the `commitUpgrade` phase.  If validation fails, the tool will fail with an error indicating the first failed blob(s), leaving the repository locked.  The user may then either rollback and re-run the upgrade, or continue the upgrade (by re-executing `repo upgrade begin` with the `--commit-mode=always` option).  The user may also choose to change the index blob(s) using a Kopia blob-editor tool and then re-validate the index (to check the work) using `repo upgrade validate`.

Re-running `repo upgrade begin` command with the `--commit-mode=always` flag will commit the existing v1 index, delete the v0 index backup, and unlock the repository.

upgrade validate examples:
```
$ KOPIA_UPGRADE_LOCK_ENABLED=1 kopia repo upgrade --upgrade-owner-id jdoe@example.com validate

upgraded index matches old index
```

commit mode "never" example:
```
$ KOPIA_UPGRADE_LOCK_ENABLED=1 kopia repo upgrade --upgrade-owner-id jdoe@example.com begin --commit-mode=never

deleting /Users/aalpar/Library/Caches/kopia/9ba004d5e92006b9/kopia.repository
Repository upgrade lock intent has been placed.
.
.
.
migrating current indices to epoch format
deleting /Users/aalpar/Library/Caches/kopia/9ba004d5e92006b9/kopia.repository
deleting /Users/aalpar/Library/Caches/kopia/9ba004d5e92006b9/kopia.blobcfg
Repository indices have been upgraded.
upgraded index matches old index
WARN Commit mode is set to 'never'.  Skipping commit.

$ KOPIA_UPGRADE_LOCK_ENABLED=1 kopia repo status --upgrade-owner-id jdoe@example.com

Config file:         /Users/jdoe/Library/Application Support/kopia/repository.config
.
.
.
Max pack length:     21 MB
Index Format:        v2

Epoch Manager:       enabled
Current Epoch: 0
.
.
.
Ongoing upgrade:     Upgrading from format version 1 -> 3
Upgrade Time:        2022-11-09 09:56:27.861252 -0800 PST
Upgrade lock:        Locked
Lock status:         Fully Established
```

failed validation and commit mode "always" example:
```
$ KOPIA_UPGRADE_LOCK_ENABLED=1 kopia repo upgrade --upgrade-owner-id jdoe@example.com begin


deleting /Users/aalpar/Library/Caches/kopia/9ba004d5e92006b9/kopia.repository
Repository upgrade lock intent has been placed.
.
.
.
ERROR failed to load index entries for v1 index entry: failed to load index blob with BlobID xn0_081213a775af0bae06d4d81f4f5f0666-sb28e5d344128a4ed-c1: unable to decrypt index blob: error decrypting BLOB xn0_081213a775af0bae06d4d81f4f5f0666-sb28e5d344128a4ed-c1: unable to decrypt content: cipher: message authentication failed

$ KOPIA_UPGRADE_LOCK_ENABLED=1 kopia repo upgrade --upgrade-owner-id jdoe@example.com validate

ERROR failed to load index entries for v1 index entry: failed to load index blob with BlobID xn0_081213a775af0bae06d4d81f4f5f0666-sb28e5d344128a4ed-c1: unable to decrypt index blob: error decrypting BLOB xn0_081213a775af0bae06d4d81f4f5f0666-sb28e5d344128a4ed-c1: unable to decrypt content: cipher: message authentication failed

```


cli/command_repository_upgrade.go
```diff
@@ -72,9 +86,123 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
+
+	err = loadIndexBlobs(ctx, indexEntries, sm, 0, indexBlobInfos)
+	if err != nil {
+		return errors.Wrapf(err, "failed to load index entries for v0 index entry")
+	}
+
+	indexBlobInfos, _, err = sm.IndexReaderV1().ListIndexBlobInfos(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "failed to list index blobs for v1 index")
+	}
```
redgoat650, line 141: Hmm don't the two blob reader versions need to be inputs to this function? Say, assuming some hypothetical future upgrade from version X to version Y, we still want to verify the before-upgrade indices are equivalent to the after-upgrade indices?

redgoat650 (commented):
Posed a few questions

cli/command_repository_upgrade.go
```diff
@@ -72,9 +86,123 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
+		for _, indexInfo := range indexInfos {
+			assign(indexInfo, index, indexEntries)
+		}
+	}
+	return nil
+}
+
+// validateAction returns an array of strings (report) that describes differences between
+// the V0 index blob and V1 index blob content.  This is used to check that the upgraded index
+// (V1 index) reflects the content of the old V0 index.
```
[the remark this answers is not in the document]
redgoat650: I see, it was probably copy/paste from #2528

cli/command_repository_upgrade.go
```diff
@@ -72,9 +86,123 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
+		log(ctx).Errorf("%v", err)
+		return nil
+	}
+	return err
+}
+
+// checkIndexInfo compare two index infos.  If a mismatch exists, return an error with diagnostic information.
+func checkIndexInfo(i0, i1 index.Info) error {
+	var err error
+	if i0.GetFormatVersion() != i1.GetFormatVersion() {
```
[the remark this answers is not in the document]
Shrekster: Hmm. @aaron-kasten does `i0.GetFormatVersion()` return us > 1 after upgrade ? Do we need to fix that ?

cli/command_repository_upgrade.go
```diff
@@ -72,9 +86,123 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
+	}
+	if i0.GetEncryptionKeyID() != i1.GetEncryptionKeyID() {
+		err = errors.Errorf("mismatched EncryptionKeyIDs: %v %v", i0.GetEncryptionKeyID(), i1.GetEncryptionKeyID())
+	}
+	if i0.GetDeleted() != i1.GetDeleted() {
+		err = errors.Errorf("mismatched Deleted flags: %v %v", i0.GetDeleted(), i1.GetDeleted())
+	}
+	if i0.GetTimestampSeconds() != i1.GetTimestampSeconds() {
+		err = errors.Errorf("mismatched TimestampSeconds: %v %v", i0.GetTimestampSeconds(), i1.GetTimestampSeconds())
+	}
```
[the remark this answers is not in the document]
aaron-kasten: @redgoat650 moving to just returning one error instead of a collection to simplify implementation.

cli/command_repository_upgrade.go
aaron-kasten: @redgoat650 in theory yes, but that would be designing before we have a use-case.  For now just dealing with v0 -> v1 index upgrade.

cli/command_repository_upgrade.go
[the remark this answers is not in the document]
aaron-kasten: @redgoat650 @Shrekster This check was inherited from the preexisting index testing.  This should be the blob format version, not the index structure version.

cli/command_repository_set_parameters_test.go
```diff
@@ -182,6 +182,71 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
...
+	env.RunAndExpectFailure(t, "repository", "set-parameters", "--epoch-cleanup-safety-margin", "10s")
+	env.RunAndExpectFailure(t, "repository", "set-parameters", "--epoch-advance-on-count", "1")
+
+	out = env.RunAndExpectSuccess(t, "repository", "status")
+	require.Contains(t, out, "Epoch Manager:       enabled")
+	require.Contains(t, out, "Index Format:        v2")
+	require.Contains(t, out, "Format version:      3")
+	require.Contains(t, out, "Epoch cleanup margin:    23h0m0s")
+	require.Contains(t, out, "Epoch advance on:        22 blobs or 80.7 MB, minimum 3h0m0s")
+	require.Contains(t, out, "Epoch checkpoint every:  9 epochs")
```
redgoat650, line 245: same suggestion ☝️  for `require.Regexp` would apply here

repo/content/committed_read_manager.go
```diff
@@ -62,6 +62,11 @@ var allIndexBlobPrefixes = []blob.ID{
 	epoch.RangeCheckpointIndexBlobPrefix,
 }
 
+// IndexBlobReader provides and API for reading index blobs.
```
redgoat650, line 65:
```suggestion
// IndexBlobReader provides an API for reading index blobs.
```

repo/content/committed_read_manager.go
```diff
@@ -113,6 +118,26 @@ type SharedManager struct {
...
+
+	return ParseIndexBlob(ibid, d.Bytes(), sm.format)
+}
+
+// IndexReaderV0 return an index reader for reading V0 indexes.
+func (sm *SharedManager) IndexReaderV0() IndexBlobReader {
+	return sm.indexBlobManagerV0
+}
+
+// IndexReaderV1 return an index reader for reading V0 indexes.
```
redgoat650, line 136:
```suggestion
// IndexReaderV1 return an index reader for reading V1 indexes.
```

tests/testenv/cli_test_env.go
```diff
@@ -106,6 +109,39 @@ func (e *CLITest) RunAndExpectSuccess(t *testing.T, args ...string) []string {
...
+	for {
+		// find a location that isn't already
+		// the value we want to write
+		_, err := f.ReadAt(bs[:], i)
+		require.NoError(t, err)
+		if bs[0] != 0x00 {
+			break
+		}
+		i = rand.Int63n(l)
+	}
```
redgoat650, line 139: Before we enter an infinite loop, maybe we should check the file we're tweaking actually has any non-zero bytes in it?

Alternatively you could change the mechanism of simulating file corruption to XOR a random byte, which works no matter the content of the file (plus you won't need to loop, cause it'll work for any value)

cli/command_repository_set_parameters_test.go
aaron-kasten: @redgoat650 `require.Contains` is part of the testify framework, so changing the implementation of `Contains` will be challenging.  Using, or creating, another assertion that would allow regular expressions is a good approach.  The rest of the code that inspects output from `require.Contains` should be moved over as well (otherwise there will be one bit of code that doesn't conform).

(I added ticket fix this for the kasten branch)

repo/format/upgrade_lock.go
```diff
@@ -86,6 +103,11 @@ func (m *Manager) CommitUpgrade(ctx context.Context) error {
 		return errors.New("no upgrade in progress")
 	}
 
+	// poison V0 index so that old readers won't be able to open it.
+	if err := WriteLegacyIndexPoisonBlob(ctx, m.blobs); err != nil {
+		log(ctx).Errorf("unable to write legacy index poison blob: %v", err)
+	}
+
 	// restore the old format version
```
Shrekster, line 111: Can upgrade succeed without `WriteLegacyIndexPoisonBlob()` succeeding ? What would be the implications for unlocking the repo when that happens ?

Shrekster: Never mind, this was also the existing behavior. So, you are not changing it.

repo/content/index/info.go
```diff
@@ -92,3 +93,32 @@ func ToInfoStruct(i Info) *InfoStruct {
...
+	if i0.GetEncryptionKeyID() != i1.GetEncryptionKeyID() {
+		qs = append(qs, fmt.Sprintf("mixed EncryptionKeyIDs: %v %v", i0.GetEncryptionKeyID(), i1.GetEncryptionKeyID()))
+	}
+	if i0.GetDeleted() != i1.GetDeleted() {
+		qs = append(qs, fmt.Sprintf("mixed Deleted flags: %v %v", i0.GetDeleted(), i1.GetDeleted()))
+	}
+	if i0.GetTimestampSeconds() != i1.GetTimestampSeconds() {
+		qs = append(qs, fmt.Sprintf("mixed TimestampSeconds: %v %v", i0.GetTimestampSeconds(), i1.GetTimestampSeconds()))
+	}
+	return qs
```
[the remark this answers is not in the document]
aaron-kasten: @adowair the above was taken out in an earlier version.  There was a bad merge that re-introduced it.

The code has been removed because its more sophisticated that what was required for the use-case.

cli/command_repository_set_parameters.go
```diff
@@ -140,32 +210,28 @@ func (c *commandRepositorySetParameters) run(ctx context.Context, rep repo.Direc
...
-			mp.EpochParameters = epoch.DefaultParameters()
-			upgradeToEpochManager = true
-			mp.IndexVersion = 2
-		}
+	c.setSizeMBParameter(ctx, c.maxPackSizeMB, "maximum pack size", &mp.MaxPackSize, &anyChange)
 
-		if mp.Version < format.FormatVersion2 {
-			mp.Version = format.FormatVersion2
+	// prevent downgrade of index format
+	if c.indexFormatVersion != 0 && c.indexFormatVersion != mp.IndexVersion {
```
Shrekster, line 223: Might be a good idea to separate this change out in its own PR for better visibility.

cc @jkowalski , I believe the only scenario anyone would ever consider downgrading index version (outside of tests) is during a corrupted repository recovery attempts. Any concerns over disallowing this anymore ?

cli/command_repository_set_parameters_test.go
```diff
@@ -182,6 +182,71 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
 	env.RunAndExpectSuccess(t, "index", "epoch", "list")
 }
 
+func (s *formatSpecificTestSuite) TestRepositorySetParametersDowngrade(t *testing.T) {
```
Shrekster, line 185: This sounds like a negative test which would attempt to perform an invalid index downgrade and successfully hit a failure. Is that correct ? If so then shouldn't we be using `--index-version` operation here somewhere ?

cli/command_repository_set_parameters_test.go
```diff
@@ -182,6 +182,71 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
...
+
+		// You can only upgrade when you are not already upgraded
+		if s.formatVersion < format.MaxFormatVersion {
+			env.RunAndExpectSuccess(t, cmd...)
+		} else {
+			env.RunAndExpectFailure(t, cmd...)
+		}
+	}
+
+	env.RunAndExpectSuccess(t, "repository", "set-parameters", "--upgrade")
```
Shrekster, line 226: nit; Might be a good idea to write a comment here explaining what is the intention of running the `set-parameters --upgrade` command after running the `upgrade` command above ? I think I know what you might be doing here, but it may not be apparent for the readers.

cli/command_repository_upgrade.go
```diff
@@ -72,9 +83,130 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
+
+		for _, indexInfo := range indexInfos {
+			assign(indexInfo, which, indexEntries)
+		}
+	}
+
+	return nil
+}
+
+// validateAction returns an error of the new V1 index blob content does not match the source V0 index blob content.
```
Shrekster, line 120:
```suggestion
// validateAction returns an error if the new V1 index blob content does not match the source V0 index blob content.
```

cli/command_repository_upgrade.go
```diff
@@ -72,9 +83,130 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
+		return errors.Wrapf(err, "failed to list index blobs for old index")
+	}
+
+	indexBlobInfos1, _, err := sm.IndexReaderV1().ListIndexBlobInfos(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "failed to list index blobs for new index")
+	}
+
+	if len(indexBlobInfos0) == 0 && len(indexBlobInfos1) > 0 {
+		log(ctx).Infof("index has already been upgraded.  Nothing to do")
```
Shrekster, line 138:
```suggestion
		log(ctx).Infof("index has already been upgraded, nothing to do")
```

cli/command_repository_upgrade.go
```diff
@@ -72,9 +83,130 @@ func (c *commandRepositoryUpgrade) setup(svc advancedAppServices, parent command
...
+	for contentID, indexEntryPairs := range indexEntries {
+		if indexEntryPairs[0] == nil || indexEntryPairs[1] == nil {
+			err = errors.Errorf("lop-sided index entries for contentID %q", contentID)
+		} else {
+			err = checkIndexInfo(indexEntryPairs[0], indexEntryPairs[1])
+		}
+
+		if err != nil {
+			break
+		}
```
Shrekster, line 161: It might be a good idea to dump a minimum number of failures before breaking out. This will help answering questions like "Did we somehow spoil just a single index or more of them ?".

You can add a CLI flag to let the user specify a non-default minimum too. Not required though.

cli/command_repository_upgrade.go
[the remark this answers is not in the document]
Shrekster: Ah! Thanks for the response.

Shrekster:
Thanks for taking in the feedbacks. I'll take a final pass on this today as soon as I get out of meetings.

cli/command_repository_set_parameters_test.go
```diff
@@ -182,6 +182,49 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
...
+			require.Contains(t, out, "Format version:      1")
+			require.Contains(t, out, "Epoch Manager:       disabled")
+			env.RunAndExpectFailure(t, "index", "epoch", "list")
+		case format.FormatVersion2:
+			require.Contains(t, out, "Format version:      2")
+			require.Contains(t, out, "Epoch Manager:       enabled")
+			env.RunAndExpectSuccess(t, "index", "epoch", "list")
+		default:
+			require.Contains(t, out, "Format version:      3")
+			require.Contains(t, out, "Epoch Manager:       enabled")
```
redgoat650, line 209: @aaron-kasten - I wasn't suggesting changing the behavior of `require.Contains`. I was suggesting you instead use  the existing `require.Regexp` instead. Both are part of the testify framework.

https://github.com/stretchr/testify/blob/1333b5d3bda8cf5aedcf3e1aaa95cac28aaab892/require/require.go#L1731

cli/command_repository_set_parameters_test.go
redgoat650: @aaron-kasten - I wasn't suggesting changing the behavior of `require.Contains`. I was suggesting you instead use  the existing `require.Regexp` instead. Both are part of the testify framework.

https://github.com/stretchr/testify/blob/1333b5d3bda8cf5aedcf3e1aaa95cac28aaab892/require/require.go#L173

Shrekster (approved):
Nice work! The negative test enhancements and JSON syntax support for failure dump is future work.

Makefile
```diff
@@ -486,3 +486,4 @@ perf-benchmark-results:
 	gcloud compute scp $(PERF_BENCHMARK_INSTANCE):psrecord-* tests/perf_benchmark --zone=$(PERF_BENCHMARK_INSTANCE_ZONE) 
 	gcloud compute scp $(PERF_BENCHMARK_INSTANCE):repo-size-* tests/perf_benchmark --zone=$(PERF_BENCHMARK_INSTANCE_ZONE)
 	(cd tests/perf_benchmark && go run process_results.go)
+
```
Shrekster, line 489: needless blank line

Shrekster, line 489:
```suggestion
```

Shrekster (approved):
just one minor comment

good attempt at untangling the codecov issues

PR #2562, opened 2022-11-09
Add native Google Drive support to Kopia GUI

JumpProg:
Native Google Drive support is currently only available through Kopia CLI.


julio-lopez:
It seems that the intent was to open an issue, not to create a PR.

PR #2597, opened 2022-11-23
fix(snapshots): Partly fix #544 by supporting setuid/setgid/sticky bits …

### LINKED ISSUE #544: Support for advanced filesystem features (setuid/setgid/xattrr/hardlinks/sockets, etc)
Version: 0.6.1

Followup to #543 

```
# ls -al /usr/bin/sudo
-rwsr-xr-x 1 root root 147560 Feb  2  2020 /usr/bin/sudo
```
After the `kopia restore`
```
# ls -al /var/tmp/restore/sudo
-rwxr-xr-x 1 root root 147560 Feb  2  2020 /var/tmp/restore/sudo
```

I assume aside from the SUID bit, SGID and sticky bit are also not (yet) properly supported.

lciti:
…when snapshotting, restoring and fuse-mounting.

Also modified an end-to-end test to also check that these extra mode flags work when snapshotting+restoring. Manually tested fuse-mount.

### DIFF
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -9,6 +9,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ModBits is a bitmask representing the mode flags supported.
+const ModBits = os.ModePerm | os.ModeSetgid | os.ModeSetuid | os.ModeSticky
+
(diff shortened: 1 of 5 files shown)


lciti:
I don't think the Windows failed test has anything to do with the PR. Is there anything left for me to do before this is ready to be merged?

PR #2628, opened 2022-12-05
fix(general): prevent storage retries on ESTALE errors

Shrekster:
Running Kopia over certain (like NFS) backends can lead to ESTALE errors when working with a stale mount or file-descriptors. Currently an ESTALE from FS calls like o.Stat() end up getting retried unnecessarily. This patch introduces a fix to prevent retries on UNIX platforms.


jkowalski:
looks like windows build is broken due to linter.

jkowalski (approved):

PR #2682, opened 2023-01-06
Improve upload rate throttling

### LINKED ISSUE #2665: Extremely high upload rate burst, despite rate limit
with 
```
"throttlingLimits": {
    "maxUploadSpeedBytesPerSecond": 500000
  }
```
Kopia still maxes out the connection for a while, and then sleeps, logging "sleeping for XmYs to refill token bucket upload-bytes"

 See Task Manager:

image

It would be better if the traffic was spread evenly over the whole time window.

leth:
Should fix #2665!
A word of caution though; it compiles and the unit tests pass but I've not run this code for real, as I don't know how to 😅


leth:
I've fixed some of the linting issues. I'm not sure fixing the others is a good idea, given what the code is doing!
I'd appreciate some guidance :)

repo/blob/throttling/throttling_reader.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Length returns the length.
+func (t ThrottledBytes) Length() int {
+	return t.data.Length()
+}
+
+// WriteTo writes to the writer while meeting the rate limit.
+func (t ThrottledBytes) WriteTo(w io.Writer) (int64, error) {
+	// Use Copy & t.Reader rather than delegate to t.data.WriteTo to ensure we rate limit
+	return io.Copy(w, t.Reader()) //nolint:wrapcheck
```
jkowalski, line 27: `io.Copy()` will allocate unnecessary memory for temporary buffers. We have `iobuf.Copy()` and `iobuf.JustCopy()` which avoids this.

jkowalski:
Were you able to test this (even manually) and are you able to get any smoother upload rate? 
Ideally we'd have some automated tests to prove this.

leth:
I've not tried yet. Do you have any docs on manual testing from source? I have kopia installed on windows so far.

leth:
You can see the throttler being called during upload in this test: https://github.com/kopia/kopia/pull/2682/files#diff-43c9a863f13dd291b01f9db79a62110305a94196c8cb84e4c7d8e18a4f666b67R95-R111

It probably won't be a perfectly even throttler; as the buffer grows (see the test, I'm not sure why the buffer grows, perhaps it's go's built in copy implementation) it'll get a bit bursty.
But since it's dealing with individual reads rather than whole uploads it should be a solid improvement.

If you want a really even upload rate I think you'd want to use a different throttler behind the upload-read-throttler.
The current tokenbucket one makes sense for existing usages, 'saving up' for bandwidth kinda works well there as you want to make sure the # of concurrent uploads started fits within rates.

To ensure a really even read rate, you need to replenish tokens on an even schedule, and limit the amount of data read to what is available, instead of reading the full amount asked but making it take longer.
I quick google found https://github.com/aybabtme/iocontrol which looks like a good implementation.

leth:
I finally found some time to give this a for-real test on my laptop. It doesn't seem to change the upload rate.
I am however running this on mac, and I originally saw the throttling issue on windows, though people are reporting the issue on linux.
I'm unsure if that's related, or if I am just not testing this properly.

PR #2689, opened 2023-01-11
Support specifying a credential container during S3 init

vkamra:
This commit allows us to specify a custom credential container when initializing an S3 storage provider.

This functionality can be leveraged to plugin custom credential providers to override the default credential
chain currently used.

### DIFF
--- a/repo/blob/s3/s3_options.go
+++ b/repo/blob/s3/s3_options.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/kopia/kopia/repo/blob/throttling"
+	"github.com/minio/minio-go/v7/pkg/credentials"
 )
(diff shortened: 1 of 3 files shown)


repo/blob/s3/s3_options.go
```diff
@@ -29,4 +30,7 @@ type Options struct {
 
 	// PointInTime specifies a view of the (versioned) store at that time
 	PointInTime *time.Time `json:"pointInTime,omitempty"`
+
+	// Creds is an optional credential container to use
+	Creds *credentials.Credentials `json:"-"`
```
vkamra, line 35: Unfortunately doesn't look like this works well because we will at some point serialize this config to a file and read from there.

PR #2690, opened 2023-01-12
fix(snapshots): cached stream file size

ashmrtn:
When checking if an entry is cached and using StreamingFile, source the file size from the cached entry. StreamingFiles don't know their size until they are read so this fixes the issue where files would have 0 size but still have their full content. Also fixes stats reporting for StreamingFiles


vkamra (approved):
LGTM. Also reviewed this last week [here](https://github.com/alcionai/kopia/pull/57/files). This was pulled into Corso and verified with the tests that caught the issue.

PR #2691, opened 2023-01-12
Enable configuration of kopia server arguments when using KopiaUI

### LINKED ISSUE #1432: High volume of B2 class C transactions, "Periodic maintenance" task?
I have a test snapshot running to B2 which contains a small folder of only 9 files (125MB) and I am seeing a large volume of B2 "Class C" transactions with only this small snapshot set to run once every 24 hours. Today I have already amassed over 2000 transactions and this is climbing steadily.
image
While these are fairly inexpensive (2500 transactions is only $0.10), this is with only **9 files** and my concern is that once I am ready and add the bulk of my other backup data to a snapshot, that these transactions and therefore the daily/monthly cost will balloon exponentially.

A list of the types of transactions is [here](https://www.backblaze.com/b2/b2-transactions-price.html). The bulk of the transactions on my account seem to fall under **b2_list_file_names** with **b2_list_file_versions** second.

The strange thing is I would expect this to be much much lower with this small amount of files as this B2 [b2_list_file_names ](https://www.backblaze.com/b2/docs/b2_list_file_names.html)API document indicates each single transaction returns a list of up to 1000 files. How is it at all possible to be using thousands of transactions for this tiny amount of files?

I am not sure what all would be contributing with this small set of files, but I noticed that KopiaUI shows a maintenance task running every 10 minutes with the description "Periodic maintenance":
image

Over a 24 hour period this runs almost 150 times and I suspect is at least partially contributing to these transactions.

I was unable to determine what this **Periodic maintenance** task is or how it is scheduled. I have not changed the maintenance intervals which appear to be **quick** maintenance every hour, and **full** every 24 hours as confirmed by **kopia maintenance info**:
image

Based on posts I've read on other forums for other backup products, this level of Class C transactions is quite abnormal especially with a snapshot of this size.

gorootde:
This PR enables to supply additional arguments to the kopia server processes that are started by KopiaUI.

Fixes #1432

### DIFF
--- a/app/public/config.js
+++ b/app/public/config.js
@@ -10,6 +10,19 @@ let configDir = "";
 let isPortable = false;
 let firstRun = false;
 
+// returns additional server args for starting Kopia server
+function additionalServerArgs(repoID) {
+    const argsFile=path.resolve(globalConfigDir(), `server_args-${repoID}.conf`)
+    log.info(`Checking for additional server args in ${argsFile}`)
+    if(fs.existsSync(argsFile)){
+        const additionalArgs=fs.readFileSync(argsFile).toString().split("\n").filter(v => v.trim().length >= 0)
+        log.info(`Found additional server args for ${repoID}: ${JSON.stringify(additionalArgs)}`)
+        return additionalArgs
+    } else {
+        return []
+    }
+}
+
 // returns the list of directories to be checked for portable configurations
 function portableConfigDirs() {
     let result = [];
@@ -107,6 +120,7 @@ function deleteConfigIfDisconnected(repoID) {
 }
 
 module.exports = {
+    additionalServerArgs,
     loadConfigs() {
         fs.mkdirSync(globalConfigDir(), { recursive: true, mode: 0700 });
         let entries = fs.readdirSync(globalConfigDir());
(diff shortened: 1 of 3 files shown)


site/content/docs/Reference/Command-Line/_index.md
```diff
@@ -97,3 +97,27 @@ The configuration file stores the connection parameters, for example:
 ```
 
 The password to the repository is stored in operating-system specific credential storage (KeyChain on macOS, Credential Manager on Windows or KeyRing on Linux).
+
```
jkowalski, line 100: can you move this to Advanced section?

app/public/config.js
```diff
@@ -10,6 +10,19 @@ let configDir = "";
 let isPortable = false;
 let firstRun = false;
 
+// returns additional server args for starting Kopia server
+function additionalServerArgs(repoID) {
+    const argsFile=path.resolve(globalConfigDir(), `server_args-${repoID}.conf`)
+    log.info(`Checking for additional server args in ${argsFile}`)
+    if(fs.existsSync(argsFile)){
```
jkowalski, line 17: I have couple comments:

1. Can this be a JSON file for future extensibility, also to match all other config files? this will make parsing easier.

```
{
   "extraArgs": [
     "--foo",
     "--bar"
   ]
}
```

2. Later we can set additional things like restart policies, back-off on failure, specify custom-built `kopia` path, pass environment variables and more. 

3. Can make the filename  be `{config-name}.server`?

Issue #1535, opened 2021-11-23
[feature request] snapshot restore: add "robocopy /MIR" (mirror) equivalent


0x53A:
I'd like to add a restore mode which, like ``--skip-existing`` works incremental, but also **deletes** extra files.

1) is this a functionality you'd like to support?
2) if yes, should I add a new flag ``--mirror``, or should it be a combination of ``--skip-existing`` and a new flag ``--delete-extra`` (exact name tbd)?

### Use case:

We use kopia snapshots to copy files between gitlab ci stages. The total size is ~13GB, about 10GB of that is static and does not change between runs.

Currently we delete everything and do a full, clean restore.

We could save a lot of time by keeping the directory _dirty_ between runs, then using an incremental kopia restore.

jkowalski:
Makes sense to me. PRs are welcome.

cristihcd:
Looks like a welcome feature. I would suggest --skip-existing and --delete-extra flags.

cristihcd:
Currently I am using `robocopy /mir` over an on-demand vpn connection to delete extra files. The restored files are over 400GB, so deleting before restore is not a good option. 

Does anyone have another solution? I am not experienced enough to modify kopia's code.

nickchomey:
I'd love to see this feature as well. Right now (I believe) I'm forced to mount the repo in either kopia or restic and use rclone sync to perform a mirror as described here. I'd love to simplify this and keep it all within rclone restore.

renardguill:
Hello all, @jkowalski, now that @0x53A has made a PR, when can this be merged? because it's a really useful feature 🤩 .

geekwave9421:
Hi @jkowalski  when this PR will be merged , this is really a important feature adds value to kopia

trompx:
Also wanted an equivalent of rclone sync (mirror) where I don't end up with incomplete files, in case the restore fail and I have to restart it multiple times. Ended up using `--skip-existing --write-files-atomically --delete-extra` but not sure if this is bullet proof (have not enough experience with huge snapshots restores).

PR #2721, opened 2023-01-31
feat(snapshots): restore: add new flag '--delete-extra'

0x53A:
closes https://github.com/kopia/kopia/issues/1535

### DIFF
--- a/snapshot/restore/restore.go
+++ b/snapshot/restore/restore.go
@@ -2,6 +2,7 @@ package restore
 
 import (
 	"context"
+	"os"
 	"path"
 	"runtime"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


snapshot/restore/restore.go
```diff
@@ -260,6 +270,54 @@ func (c *copier) copyDirectory(ctx context.Context, d fs.Directory, targetPath s
...
+
+		if e.IsDir() {
+			dirs[e.Name()] = struct{}{}
+		} else /* file */ {
+			files[e.Name()] = struct{}{}
+		}
+	}
+
+	// read existing entries on disk
+	existingEntries, err := os.ReadDir(targetPath)
```
jkowalski, line 294: you should not assume targetPath is an operating system directory, there's `output` abstraction to support ZIP and TAR outputs too.

0x53A: thanks, wouldn't have thought of that!
Is typechecking c.Output the correct way?

0x53A:
I added a test, but couldn't test the test yet, because for me (on gitpod) even main fails at every ``compareDirs``.
Will pick this up again tomorrow.

PR #2733, opened 2023-02-03
feat(repository): added streaming of FindManifests responses

### LINKED ISSUE #2660: Cannot list snapshots from remote Kopia client
Somwhere along the last 3 months, my remote Kopia client, which connects to a Kopia server, has lost the ability to list all it's snapshots on the Kopia server. The error generated looks like this:

`ERROR error listing all snapshot manifests: unable to find snapshot manifests: unable to complete invoking GRPC API despite 10 retries: should retry`

When performing the same action on the Kopia server itself, it does work, although it takes a good 132s to perform that task, due to the number of snapshots in the repo. My estimate would be 100k+ snapshots. I updated either Kopia server and client, but this doesn't help a bit. Also, accessing the repo via Kopias built-in web Ui is tantalizingly slow, when it comes to listing the Snapshots.

jkowalski:
This change adds a new streaming response to the FindManifests API. The server will deliver the response in chunks of N manifests where N is requested by the client. This allows the client to process the response in chunks and improves pipelining of responses.

For now client will hold the entire response in memory since this is what FindManifests() API currently does. This will be fixed in a follow up change.

Replaces #2713
Fixes #2660


budachst:
Works great on my installation. Pulled-in this PR and my backups have been running smoothly again. Hope to see this soon merged.

internal/grpcapi/repository_server.pb.go
```diff
@@ -1158,6 +1158,8 @@ type FindManifestsRequest struct {
 	unknownFields protoimpl.UnknownFields
 
 	Labels map[string]string `protobuf:"bytes,1,rep,name=labels,proto3" json:"labels,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
+	// if set to non-zero, the server will return up to N entries.
+	PageSize int32 `protobuf:"varint,2,opt,name=page_size,json=pageSize,proto3" json:"page_size,omitempty"`
```
Shrekster, line 1162: Better to use sint32 for efficiency. If we do not expect these to be negative values ever then we can use `uint32`.

internal/grpcapi/repository_server.pb.go
```diff
@@ -1620,6 +1630,13 @@ func (x *SessionResponse) GetRequestId() int64 {
 	return 0
 }
 
+func (x *SessionResponse) GetHasMore() bool {
+	if x != nil {
+		return x.HasMore
+	}
+	return false
```
Shrekster, line 1637:
```suggestion
	return x != nil && x.HasMore
```

internal/server/grpc_session.go
```diff
@@ -361,16 +362,32 @@ func handleFindManifestsRequest(ctx context.Context, dw repo.DirectRepositoryWri
...
+			respond(&grpcapi.SessionResponse{
+				HasMore: true,
+				Response: &grpcapi.SessionResponse_FindManifests{
+					FindManifests: &grpcapi.FindManifestsResponse{
+						Metadata: makeEntryMetadataList(filtered),
+					},
+				},
+			})
+
+			filtered = nil
```
Shrekster, line 377: better to keep the allocated memory to avoid fragmentation

```go
filtered = filtered[:0]
```

repo/grpc_repository_client.go
```diff
@@ -109,12 +114,17 @@ func (r *grpcInnerSession) readLoop(ctx context.Context) {
...
+		if !msg.HasMore {
+			delete(r.activeRequests, msg.RequestId)
+		}
 
 		r.activeRequestsMutex.Unlock()
 
 		ch <- msg
-		close(ch)
+		if !msg.HasMore {
+			close(ch)
```
Shrekster, line 126: We should close the channel when `err != nil` as well.

repo/grpc_repository_client.go
```diff
@@ -109,12 +114,17 @@ func (r *grpcInnerSession) readLoop(ctx context.Context) {
 	for ; err == nil; msg, err = r.cli.Recv() {
 		r.activeRequestsMutex.Lock()
 		ch := r.activeRequests[msg.RequestId]
-		delete(r.activeRequests, msg.RequestId)
+
+		if !msg.HasMore {
+			delete(r.activeRequests, msg.RequestId)
```
Shrekster, line 119: we should remove from active list also when `err != nil`

repo/grpc_repository_client.go
```diff
@@ -321,23 +329,34 @@ func (r *grpcInnerSession) PutManifest(ctx context.Context, labels map[string]st
...
 		case *apipb.SessionResponse_FindManifests:
-			return decodeManifestEntryMetadataList(rr.FindManifests.GetMetadata()), nil
+			entries = appendManifestEntryMetadataList(entries, rr.FindManifests.GetMetadata())
+
+			if !resp.GetHasMore() {
+				return entries, nil
+			}
 
 		default:
 			return nil, unhandledSessionResponse(resp)
```
Shrekster, line 362: nit; might be helpful for debugging later if we include the count of the entries read so far in the error message

PR #2747, opened 2023-02-09
feat(repository): added RepositoryWriter.ReplaceManifests

jkowalski:
This simplifies use cases where we intend to replace a manifest uniquely identified by a set of labels with another one as is the case for policies.

This helped fix annoying test flake on Windows where the clock is not guaranteed to move forward when read in quick succession.

This is now passing on Windows:

```
$ go test -timeout 1000s ./internal/server -run TestSourceRefreshesAfterPolicy -count=1000
```


jkowalski:
@julio-lopez PTAL, curious what you think of this helper, do you think it belongs in the `repo` package?

PR #2756, opened 2023-02-10
feat(cli): new storage config input from file when creating a repo

Shrekster:
When creating a repo we are currently required to supply storage config secrets as CLI arguments. CLI commands often get logged which can accidentally expose these creds. This PR extends the `from-config` storage flags to be supplied at the time of creation with a dedicated `--token-file` option or `--token-stdin` option that reads from stdin.

```
echo "${kopia_string}" | kopia repository create from-config --token-stdin
```

```
kopia repository create from-config --token-file /path/to/token-file
```

### DIFF
--- a/cli/command_repository_connect_from_config.go
+++ b/cli/command_repository_connect_from_config.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"io"
 	"os"
 
 	"github.com/alecthomas/kingpin/v2"
@@ -15,6 +16,7 @@ type storageFromConfigFlags struct {
 	connectFromConfigFile  string
 	connectFromConfigToken string
 	connectFromTokenFile   string
+	connectFromTokenStdin  bool
 
 	sps StorageProviderServices
 }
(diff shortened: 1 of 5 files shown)


cli/command_repository_connect_from_config.go
```diff
@@ -13,31 +14,37 @@ import (
...
 	connectFromConfigToken string
+	connectFromTokenFile   string
 
 	sps StorageProviderServices
 }
 
 func (c *storageFromConfigFlags) Setup(sps StorageProviderServices, cmd *kingpin.CmdClause) {
 	cmd.Flag("file", "Path to the configuration file").StringVar(&c.connectFromConfigFile)
 	cmd.Flag("token", "Configuration token").StringVar(&c.connectFromConfigToken)
+	cmd.Flag("token-file", "Path to the configuration token file").StringVar(&c.connectFromTokenFile)
```
jkowalski, line 25: how about (also) adding a boolean `token-stdin` or something similar ? This would make it work on Windows where /dev/stdin is not available.

Shrekster: good idea, will do

r4rajat: @jkowalski, According to my understanding, we want to introduce a flag `token-stdin`, which is basically of boolean type. If the value is false than user have to pass the token file path and if true than user have to pass the value of token thru stdin ?

Shrekster: added the `--token-stdin` changes with the unit-test

julio-lopez:
@Shrekster nit: modified the PR description to remove `--password ....` in order to better reflect the desired usage pattern.

julio-lopez (approved):

julio-lopez:
Thanks for the change!

pavannd1:
Thank you, @Shrekster! 🎉

PR #2824, opened 2023-03-14
feat(cli): Make format upgrade command idempotent

adowair:
This commit changes the behavior of the command
`kopia repo upgrade begin...` to not fail (exit code 1) when the repository is already using the latest format version. Instead, a helpful message is output and the program exits with zero code. In effect the command becomes idempotent-successive upgrades would return the same exit code. Such an idempotent api is desirable, especially in cases where we build automation around format upgrades.

Before this change, an error code `1` is returned when upgrading a repository that is already up to date:
```console
$ kopia repo status | grep "Format Version"
Format version:      3
$ kopia repo upgrade begin --upgrade-owner-id admin
[1] ERROR error setting the upgrade lock intent: repository is using version 3, and version 3 is the maximum
```

and after this change, a `0` code is returned:
```console
$ kopia repo upgrade begin --upgrade-owner-id admin
[0] Repository format is already upto date.
```

### DIFF
--- a/cli/command_repository_set_parameters_test.go
+++ b/cli/command_repository_set_parameters_test.go
@@ -154,7 +154,8 @@ func (s *formatSpecificTestSuite) TestRepositorySetParametersUpgrade(t *testing.
 		if s.formatVersion < format.MaxFormatVersion {
 			env.RunAndExpectSuccess(t, cmd...)
 		} else {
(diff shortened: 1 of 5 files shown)


aaron-kasten (approved):
Looks good.

denisvmedia (approved):
Lgtm.

Shrekster:
LGTM 👍

Issue #728, opened 2020-12-17
(Re)Compression of existing repository?


oderwat:
Is there a way to set compression after taking snapshots already and then recompress the repository?

And probably related: How does deduplication work when compression was enabled later? Will new files be re-uploaded compressed or matched with blocks that are already in the repository?

I find it hard to find any information about how compression works.

julio-lopez:
Compression is policy based. It can be enabled after having taken snapshots, and can also be set at the subdirectory granularity. However, the setting is not applied retroactively. Existing snapshots remain uncompressed.

For snapshots taken after compression has been enabled, completely unmodified files (the file paths and file times match those of the last snapshot) will not be re-uploaded. New files that end up being compressed are uploaded, even if their uncompressed contents match those of existing files.

I hope this helps.

oderwat:
I still don't understand the process, really. From what you write, it sounds like the compression is basically done per file and before deduplication.

Wouldn't this harm the possibilities of deduplication? I usually use `gzip --rsyncable` with other deduplicating storage tools. Which works very well with tar gzip piped data.

You also write they are re-uploaded even if their content matches. Let's take a big file (for example a MySQL database binary data or index file), which is modified somewhere in between or at least growing all the time. It sounds as if enabling compression will re-upload this whole file again (and again?) because it now looks different every time (because of the compression).

Is compression basically just a "preprocessing" step as if one would compress the files beforehand?

I expected that the data gets compressed after the deduplication as an additional way to save storage and bandwidth. 
Basically, the blobs of data are compressed before transmission and inside the repository, but their hashes refer to the uncompressed data.

jkowalski:
objects are written in the following stages:

splitter -> compressor -> hash/dedup -> encrypt -> package

Assuming compressors produce stable outputs (and they usually do), deduplication property is preserved.

We could consider compressing between hash and encrypt, although that would require minor index format change (we'd need to remember which encrypted blocks are also compressed and which ones are not, that is just one bit) but would potentially allow re-compression of repositories.

Not sure how compatible that would be with directory-level policies, but perhaps worth trying at some point.

oderwat:
[...]

PR #2835, opened 2023-03-16
feat(snapshots): Enable compression for StreamingFile entries

ashmrtn:
Also take fs.Entry for helper that gets compression policy instead of fs.File.

Hooks into existing logic for parsing policy trees to get the policy for the
current path. Users expecting compression for StreamingFiles do need to be
cognizant of the fact that StreamingFiles return a size of 0 when `f.Size()`
is called. This means that setting a minimum compression size may interfere
with enabling compression on StreamingFiles


Shrekster:
Thank you for posting this change! I have a few questions:

1. Does this change make compression mandatory for StreamingFile(s) which were otherwise not respecting the compression policy so far on an exiting repository ? That is, would a restore from an older repository fail ? Can we add a compatibility test for this ?
2. Would adding compression to an existing repository bloat its next incremental run when a user will upgrade Kopia to a version that has this change ?

ashmrtn:
Great questions! I've put answers to the questions inline with quotes for context

>1. Does this change make compression mandatory for StreamingFile(s) which were otherwise not respecting the compression policy so far on an exiting repository ?
>

If compression is configured for the kopia repository and there's no minimum size set for files to be compressed then yes, it will make compression mandatory for all StreamingFiles. Since StreamingFiles have return 0 when `StreamingFile.Size()` is called an easy way to disable compression for them is to make a policy that has a minimum size for compression > 0 (i.e. minimum file size = 1)

>That is, would a restore from an older repository fail ?
>

No, this reuses existing logic for compressing files. All the added code does is ensure that we allow compression to take place if the policy requests compression. The existing code for compressing data [adds a compression header](https://github.com/kopia/kopia/blob/cd5de60c87c2d497e5f6929a062d0eba96c14758/repo/compression/compressor_zstd.go#L41) to the blob when creating it. The header is then read back when fetching the data from kopia, ensuring kopia can distinguish compressed and uncompressed data as well as different compression types

>Can we add a compatibility test for this ?
>

I can add a test if desired, though I must admit I'm not sure precisely where to add it nor what existing test frameworks I may be able to reuse to help setup said test. If you have any pointers on this I'd really appreciate it.

Though it doesn't give the assurances of a regression test that's checked into the codebase, I can say I've run a manual test where I created a kopia backup without this patch, updated kopia, and restored the (uncompressed) data from kopia without issue

>2. Would adding compression to an existing repository bloat its next incremental run when a user will upgrade Kopia to a version that has this change ?
>

I've tried looking around for docs to answer this solidly and came away with conflicting answers. Because this ties into existing logic for compressing data, it should follow the regular compression/dedupe rules for kopia. Looking at the current [doc page](https://kopia.io/docs/advanced/compression/) for compression, it states

> When Kopia sees a file to backup, it first splits its data into “chunks” with the configured splitter. Then compute its hash with the configured hash algorithm, and compare it against the existing database. If a matching hash is found, this chunk is already stored, thus can be safely discarded. Otherwise, Kopia applies the compression algorithm on the chunk, encrypt it, and pack a few of such into a blob if fit.

The above seems to imply that no, changing compression shouldn't cause mass file reuploads because the (uncompressed) hash of the data will still match

On the other hand there's this [comment](https://github.com/kopia/kopia/issues/728#issuecomment-748270429) from an open question in the kopia repo which seems to imply that compression happens before hashing/dedupe, which would probably cause reuploads of data as kopia would compare uncompressed hashes to compressed hashes

>objects are written in the following stages:
>
>```text
>splitter -> compressor -> hash/dedup -> encrypt -> package
>```

Empirically testing this, the answer appears to be no, it wouldn't cause lots of files to be reuploaded on the next run due to a change in compression. For reference, the empirical test I ran was to backup some StreamingFiles with the unpatched version of kopia. The first run showed all bytes hashed and uploaded. The second run (with patched kopia) showed all bytes hashed (expected as modtime for the StreamingFiles wasn't explicitly set) and only a small amount of data uploaded (i.e. < 1MB for a backup that included 2 trivially compressible files of 512MB)

jkowalski (approved):

ashmrtn:
re: the discussion on whether this change will cause mass file reuploads. On closer examination, it appears the answer to that may actually be "it depends." I was digging around the codebase a bit more and realized that compression may actually occur in one of two different places during a snapshot: the content manager or the object manager, with [preference for compressing in the content manager](https://github.com/kopia/kopia/blob/cc00388f6d2a2b78f03e2507b4b2f25519d1f7e4/repo/object/object_writer.go#L197)

If compression is supported by and done in the content manager ([index version](https://github.com/kopia/kopia/blob/cc00388f6d2a2b78f03e2507b4b2f25519d1f7e4/repo/content/content_manager.go#L767) >= 2) then this change won't cause file reuploads because the content is hashed before being compressed

If, on the hand, compression is _not_ supported by the content manager, then the data will be compressed by the object writer. That will cause reuploads of files because the data will be compressed and then hashed, causing hash mismatches

From what I've been able to see, kopia switched to v2 index on Sept 2, 2021. For repositories that haven't upgraded yet I'd assume it would reupload any `StreamingFile`s now that they'd be compressed. For repositories that started out with v1 but then upgraded, I'm not sure what the behavior would be

Shrekster (approved):
Sorry it took me longer to respond back here.

> No, this reuses existing logic for compressing files. All the added code does is ensure that we allow compression to take place if the policy requests compression.

This resolves my first concern. You may add a test here: tests/compat_test/compat_test.go. This test works with older binaries like from 0.8 release. However, with better understanding of this now I do not see any value in adding a test that verifies the existing restore behavior of relying on the header to do the right thing, and this PR does not change it. Thanks for verifying it manually.

> If, on the hand, compression is not supported by the content manager, then the data will be compressed by the object writer. That will cause reuploads of files because the data will be compressed and then hashed, causing hash mismatches

That info helps, I was not aware. For repositories which are on v1 format, we should encourage them to upgrade to v2 using the `kopia repository upgrade` tooling.

Thanks for the responses. LGTM 👍 .

pavannd1 (approved):
LGTM 💯 
I ran some manual tests with this patch too. The results seem similar to Ashlie's observations 👍🏼

ashmrtn:
>Sorry it took me longer to respond back here.

it gave me a chance to dig into the code a bit more so it's all good. Mostly me commenting here has been me trying to make sure I tie up all the loose ends and give the best information I can on how it will affect things moving forward 😅

PR #2844, opened 2023-03-21
docs(site): added a documentation on kopiaignore

### LINKED ISSUE #1945: Document the .kopiaignore format
I see no documentation for the `.kopiaignore` format. A search in https://kopia.io/docs/ finds no matches (which looks like a bug in the docs search engine, since `.kopiaignore` is mentioned, but not described, in the documentation).

#773 suggests that `.kopiaignore` now behave similarly to `.gitignore`, but this is not mentioned in the documentation, and presumably there are some differences specific to the way kopia uses the file.

Disabling backup of some files by mistake is a potentially catastrophic error for a backup solution, so documenting the way ignoring files and directory works is really important.

lupusA:
Hi, 

this PR comes with the following changes:

- Added a new under "Advanced Topics" to document kopiaignore files and patterns
- Added an example for kopiaignore for better understanding

I have tested the patterns with kopia 0.12.1 and they should work as described. However, I think there is still a strange behavior when using "/" together with "*" in a pattern. 

This PR closes #1945 and #2847

Open for feedback.

### DIFF
--- a/site/content/docs/Advanced/Actions/_index.md
+++ b/site/content/docs/Advanced/Actions/_index.md
@@ -1,7 +1,7 @@
 ---
 title: "Actions"
 linkTitle: "Actions"
-weight: 40
+weight: 45
 ---
 
 ## Actions
--- a/site/content/docs/Advanced/Caching/_index.md
+++ b/site/content/docs/Advanced/Caching/_index.md
@@ -1,7 +1,7 @@
 ---
 title: "Caching"
 linkTitle: "Caching"
-weight: 45
+weight: 50
 ---
 
 ## Caching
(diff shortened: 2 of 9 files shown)


ntolia:
@jkowalski I took a look at this docs patch and it seems reasonable to me but I am not a `.kopiaignore` user. Would be great if you could take a look.

@lupusA Did you validate all the examples in the last table of the main docs change?

lupusA:
Hi @ntolia 

i`ve tested all the examples by setting up the same folder which is mentioned in the example. However, I can retest tomorrow and report back before we merge. 

Cheers,

ntolia:
Sounds good. LMK when you have reverified and we can then merge unless we hear any other feedback.

lupusA:
Hi @ntolia, 

i corrected the main table and corrected the number of files that are ignored. Additionally, the wording was improved. 

Cheers,

ntolia:
Thank you. This has now been merged.

PR #2850, opened 2023-03-25
fix(general): Fixed ordering in function "isAbs" in ospath

lupusA:
There is a special case when checking if a given path is absolute. 

On Windows paths like "\\hostname\share" are treated as absolute. Therefore, the function "isAbs" has to first check for this before calling "filepath.IsAbs" as this returns true for non-absolute paths (e.g. \\host).  

The test for internal\ospath now return success and was failing before on windows due to the wrong order.  

Cheers,

### DIFF
--- a/internal/ospath/ospath.go
+++ b/internal/ospath/ospath.go
@@ -26,11 +26,6 @@ func LogsDir() string {
 
 // IsAbs determines if a given path is absolute, in particular treating \\hostname\share as absolute on Windows.
 func IsAbs(s string) bool {
-	//nolint:forbidigo
-	if filepath.IsAbs(s) {
-		return true
-	}
-
 	// On Windows, treat \\hostname\share as absolute paths, which is not what filepath.IsAbs() does.
 	if runtime.GOOS == "windows" {
 		if strings.HasPrefix(s, "\\\\") {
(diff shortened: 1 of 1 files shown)


jkowalski:
Hmm, i'm not seeing ospath test failures on Windows, can you share a log?

lupusA:
Hey @jkowalski, 

here is the log:

Error Trace:	c:/Users/andso/Desktop/GoProjects/kopia/internal/ospath/ospath_test.go:69
        	Error:      	Should be false
        	Test:       	TestIsAbs
        	Messages:   	\\host

The code comments already state that. 

- "// On Windows, treat \\hostname\share as absolute paths, which is not what filepath.IsAbs() does."

So in this case, filepath.IsAbs returns true and the code below that call cannot be reached. 

image

jkowalski:
This is very strange, I'm running on Windows and am not able to reproduce this failure, neither in CLI nor VS Code.

lupusA:
Strange indeed. 

Does it has to do with how the tests are setup? I've noticed that the automatic test from the github workflow uses "make ci-tests". The test from github's workflow also succeeds. 

Never the less, I've tested it with a sample go application and the method "IsAbs" from "path\filesystem" returns indeed "true" for ""\\\\Host"". 

But according to the test cases, it should be false on Windows. 

Cheers,

lupusA:
I am using go1.20.2 windows/amd64

jkowalski (approved):

jkowalski:
Approved.

Issue #2655, opened 2022-12-21
rclone-v1.61 does not work anymore with kopia


kapitainsky:
problem as in the title - reverting to rclone-v1.60 fix the problem

fivestones:
When I try to connect with kopia to a previously created repository, `kopia repository connect rclone...`, I get `got error BLOB not found when WriteTemporaryFileAndCreateParentDirs` 10 times, and then `unable to persist sharding parameters: unable to complete WriteTemporaryFileAndCreateParentDirs despite 10 retries: BLOB not found`.
I think this is because of this bug?
I am running rclone 1.61.1, kopia 0.12.1 build: 5227d74996b6520f9f96e4203cfe00b832a60d5f

kapitainsky:
testing 13RC2 and indeed it works now again with the latest rclone.

PR #2853, opened 2023-03-27
fix(repository): fixed connect broken since rclone v1.61.0

yfdyh000:
https://github.com/rclone/rclone/commit/4444d2d102cc187f8cc4907f3e06611576c10a36#diff-3bfe93b28218b8e93a7480a5684d21ac566795a0fa8e540ec1bc6cbe7028439eR295

Fix https://github.com/kopia/kopia/issues/2655.

e.g. `rclone -v serve webdav Disk/Kopia_test --addr 127.0.0.1:0`
old: NOTICE: Local file system at .../rclone-v1.59.2-windows-amd64: WebDav Server started on http://127.0.0.1:41656/
new: NOTICE: Local file system at ...: WebDav Server started on [http://127.0.0.1:51898/]
p.s. They will be https in Kopia.

Note: 
```
if l[len(l)-1:] == "]" {
						rcloneAddressChan <- l[p:len(l)-1]
```
has been considered, but regexp permit more flexibility and reliability for future.


jkowalski:
Thanks for the fix, looking at the regexp, it should work with both old and new rclone, but would be nice to check in both to make sure we don't break people.

While at it, can you upgrade rclone used for testing to the latest (update `tools.mk`, run `make regenerate-checksums` and paste the appropriate checksums into `tools/gettool/checksums.txt`)

jkowalski (approved):

julio-lopez:
Post-merge rclone provider tests are currently failing.

PR #2858, opened 2023-03-28
fix(general): remove uses of deprecated rand.Read

julio-lopez:
Lint fixes in preparation for moving to Go 1.20

Split into self-contained commits to simplify reviewing


adowair (commented):
LGTM

Issue #2823, opened 2023-03-14
Side effect of implicit `create` in `snapshot` subcommand


Juerd:
```
# kopia snapshot hlep 
Snapshotting root@hostname:/tmp/hlep ...
ERROR unable to get local filesystem entry: resolveSymlink: stat: lstat /tmp/hlep: no such file or directory
```

I mistyped `help` and it tried to create a snapshot :)

I'd like to suggest deprecating the implicit `create` subcommand, so that mistyped subcommands can get proper error messages.

jkowalski:
I like the idea, I don't think we ever advertised the shortcut and it's actively confusing (say you had a directory called "help"), we should remove this (and possibly review other default subcommand and remove some more), mark it as a breaking change.

PR welcome.

julio-lopez:
Addressed in #2861

PR #2861, opened 2023-03-28
breaking(cli): remove default behavior for CLI command

julio-lopez:
| Command        | Previous default subcommand |
| ---------------| --------------------------- |
| `cache`        | `info`   |
| `index`        | `list`   |
| `manifest`     | `list`   |
| `maintenance`  | `run`    |
| `repo upgrade` | `begin`  |
| `snapshot`     | `create` |

Ref #2823

### DIFF
--- a/app/public/server.js
+++ b/app/public/server.js
@@ -34,7 +34,7 @@ function newServerForRepo(repoID) {
             let kopiaPath = defaultServerBinary();
             let args = [];
 
-            args.push('server', '--ui',
+            args.push('server', 'start', '--ui',
                 '--tls-print-server-cert',
                 '--tls-generate-cert-name=127.0.0.1',
                 '--random-password',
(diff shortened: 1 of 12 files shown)


jkowalski (approved):

jkowalski:
Can we remove all defaults at once? We have 7 defaults right now, it would be cleaner to have none.

Also, can we list defails of all removed defaults in this PR so we can include it in release notes.

julio-lopez:
> Can we remove all defaults at once? We have 7 defaults right now, it would be cleaner to have none.

Let me first attempt to address the test failures, and then remove the other ones in separate commits, so at least it is easier to track.

> 
> Also, can we list details of all removed defaults in this PR so we can include it in release notes.

What do you mean by listing the details? For example, that the default for `snap` was `create`? or do you mean something else?

jkowalski:
> > Can we remove all defaults at once? We have 7 defaults right now, it would be cleaner to have none.
> 
> Let me first attempt to address the test failures, and then remove the other ones in separate commits, so at least it is easier to track.
> 
> > Also, can we list details of all removed defaults in this PR so we can include it in release notes.
> 
> What do you mean by listing the details? For example, that the default for `snap` was `create`? or do you mean something else?

Yes, that.

PR #2862, opened 2023-03-28
breaking(cli): remove deprecated 'snapshot gc' command

julio-lopez:


Shrekster (approved):

Shrekster:
Just one suggestion, I know we want to remove all default commands. So, could there be a kingpin setting to disallow registering any new default commands so to prevent any future additions ? Just my 2c 🪙 . Approach looks good.

PR #2868, opened 2023-03-29
fix(cli): fixed help formatting broken with recent upgrade to kingpin/v2

jkowalski:


julio-lopez (approved):

Shrekster:
Thanks! I was wondering about this too while looking at another issue.

PR #2903, opened 2023-04-06
build(infra): checkout repo before setting up Go

julio-lopez:
This way GH caching for Go packages has a chance to work.

See https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs

A cursory test shows that restoring the Go mod cache takes about ~20s.
Downloading the dependencies without a cache takes 60+s. With a cache, that step is ~5s.
Overall the gains are between 2x - 3x for this stage.

Now, when the go mod cache is modified (e.g., during a dependency upgrade) there is an additional 60s cost at the end of the job to update the cache. Once the cache is updated, other jobs get the benefit. Notice that the cache is only updated when the job is successful. When the job fails, the cache artifact is not uploaded.

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -11,15 +11,15 @@ jobs:
     name: Code Coverage Tests
     runs-on: ubuntu-latest
     steps:
-    - name: Set up Go.
+    - name: Check out repository
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Set up Go
       uses: actions/setup-go@v4
       with:
         go-version: '1.20'
       id: go
-    - name: Check out code into the Go module directory
-      uses: actions/checkout@v3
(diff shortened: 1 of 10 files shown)


jkowalski (approved):

Shrekster (approved):

Shrekster:
Retried this failure: https://github.com/kopia/kopia/actions/runs/4633902013/jobs/8199573860?pr=2903

Haven't yet repro'd locally, but the issue is unrelated.

julio-lopez:
There's an issue on how CI is currently set up to run the kopia-ui tests on MacOS.
That's causing the run on this PR to stall, it's not inherent to the changes, but to the way the PR was submitted. I'll merge this as is, and take a stab at the issue later.
Alternatively, I could submit the PR in a different way, but that would end up running all the CI jobs again. Taking a shortcut here.

PR #2910, opened 2023-04-10
build(infra): fix minio's mc installation

julio-lopez:
MinIO's MC installation broke in minio/mc#4531
This is a temporary fix until upstream is fixed.

### DIFF
--- a/tools/tools.mk
+++ b/tools/tools.mk
@@ -198,7 +198,7 @@ $(kopia08):
 MINIO_MC_PATH=$(TOOLS_DIR)/bin/mc$(exe_suffix)
 
 $(MINIO_MC_PATH):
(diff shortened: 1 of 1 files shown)


tools/tools.mk
```diff
@@ -198,7 +198,7 @@ $(kopia08):
 MINIO_MC_PATH=$(TOOLS_DIR)/bin/mc$(exe_suffix)
 
 $(MINIO_MC_PATH):
-	GOBIN=$(TOOLS_DIR)/bin go install github.com/minio/mc@latest
+	GOBIN=$(TOOLS_DIR)/bin go install github.com/minio/mc@RELEASE.2023-04-06T16-51-10Z
```
harshavardhana, line 201: You should be really downloading the released binaries or use our docker containers. Not sure what value does it add to compile?

PR #2920, opened 2023-04-11
fix(deps-dev): fixup mutex for make check-locks

### LINKED ISSUE #2919: Kopia `make check-locks` fails with `Fault` mutex changes.
The mutex changes to `fault_set.go` did not anticipate support for `gvisor`'s `checklocks` tool.  `make check-locks` will fail once `checklocks` is updated.  `fault_set.go` and possibly `fault.go` need to be modified so `make check-locks` will pass.

aaron-kasten:
Fixes #2919 .


internal/blobtesting/faulty.go
```diff
@@ -25,13 +25,14 @@ const (
 type FaultyStorage struct {
 	base blob.Storage
 
-	fault.Set
+	*fault.Set
```
aaron-kasten, line 28: lint will complain if the `FaultSet` `mu` is copied for use in `FaultyStorage`.  While the scope of the construction of the `mu` in `FaultSet` does not extend beyond the `NewFaultyStorage` call, a pointer will be used to avoid warnings.

internal/fault/fault_set.go
```diff
@@ -16,25 +16,28 @@ type Method int
...
-func (s *Set) ensureInitialized() {
-	if s.faults == nil {
-		s.faults = map[Method][]*Fault{}
-		s.callCounter = map[Method]int{}
+func NewSet() *Set {
+	q := &Set{
+		mu:          &sync.Mutex{},
+		faults:      map[Method][]*Fault{},
+		callCounter: map[Method]int{},
 	}
```
aaron-kasten, line 31: Initialize in construction, not lazily.

internal/blobtesting/faulty.go
julio-lopez: Q: would making `mu` a pointer => `*mu` solve the issue as well?

No change requested, mainly curiosity.

julio-lopez (approved):
+1: LG 👍
Other folks should take a look as well.

PR #2933, opened 2023-04-12
test(general): added tests for tlsutil

denisvmedia:
This PR adds some tests to tlsutil.

### DIFF
--- a/internal/tlsutil/tlsutil_test.go
+++ b/internal/tlsutil/tlsutil_test.go
@@ -1,0 +1,68 @@
+package tlsutil_test
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/kopia/kopia/internal/clock"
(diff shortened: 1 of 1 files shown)


denisvmedia:
cc @adowair

internal/tlsutil/tlsutil_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/kopia/kopia/internal/tlsutil"
+)
+
+func TestGenerateServerCertificate(t *testing.T) {
+	ctx := context.Background()
+	certValid := 24 * time.Hour
+	names := []string{"127.0.0.1", "localhost"}
+
+	cert, priv, err := tlsutil.GenerateServerCertificate(ctx, 2048, certValid, names)
+	if err != nil {
```
jkowalski, line 21: we generally use require library to avoid if statements like that

require.NoError(t, err)
require.NotNil(t, cert)

and so on...

denisvmedia: Done.

jkowalski (approved):

PR #2939, opened 2023-04-14
chore(general): fixed remaining checklocks violations

jkowalski:
We can't enable checklocks on CI yet until https://github.com/google/gvisor/pull/8807 is merged upstream.

This was tested with private build of checklocks with this patch applied and the results were clean.


julio-lopez (approved):

Shrekster:
Thank you! LGTM!

PR #2942, opened 2023-04-14
test(general): added tests for parallelwork

denisvmedia:
This PR adds unit tests for parallelwork (100% coverage).


denisvmedia:
@jkowalski do you think this can be merge or is there anything I can improve here?

PR #2956, opened 2023-04-18
feat(repository): Reduce memory usage when parsing manifests

ashmrtn:
Golang's stdlib JSON decoder allocates lots of memory when decoding large objects/arrays because it needs to have the entire object/array in-memory (see a [medium post](https://medium.com/safetycultureengineering/analyzing-and-improving-memory-usage-in-go-46be8c3be0a8) by another person for some debugging of this phenomenon in an unrelated project). This can cause OOMs or increased costs when accessing manifests in a kopia repo that has lots of snapshots in it

This PR adds some manual parsing to the decoding of manifests in kopia. It manually parses the `manifest` struct and pulls off the `[` and `]` denoting the array of manifest entries. Parsing of individual entries in the array is deferred to the stdlib decoder as it behaves reasonably well for those

This does not completely solve the memory problem when there are large numbers of manifests in a repo, but it does help

Some [benchmarking](https://github.com/alcionai/corso/blob/json-debug-benchmarks/src/cmd/jsondebug/benchmark/benchmark_test.go) results for various configurations (sorry the benchmark is in an unrelated repo, we were making things on the fly). The benchmark ran against a single serialized manifest entry that contained 1000 randomly generated kopia snapshot manifests in the array. 500 runs of the benchmark were done with the input gzipped like kopia has it. `Stdlib` is the golang JSON decoder. `JsonParser` uses the JsonParser [library](https://github.com/buger/jsonparser) to extract data, though it still requires having the full JSON array in-memory. `Array` is the implementation in this PR. `ArrayFull` is an extension of this implementation that only uses golang's stdlib decoder to get the values of individual fields in each array entry
```text
Benchmark_FromFile/Stdlib/Zipped-6                   500             42829 ns/op           13125 B/op         72 allocs/op
Benchmark_FromFile/JsonParser/Zipped-6               500             38351 ns/op           22079 B/op         83 allocs/op
Benchmark_FromFile/Array/Zipped-6                    500             42462 ns/op            4764 B/op         74 allocs/op
Benchmark_FromFile/ArrayFull/Zipped-6                500             49263 ns/op            6356 B/op        168 allocs/op
```

### DIFF
--- a/repo/manifest/committed_manifest_manager.go
+++ b/repo/manifest/committed_manifest_manager.go
@@ -346,7 +346,8 @@ func loadManifestContent(ctx context.Context, b contentManager, contentID conten
 		return man, errors.Wrapf(err, "unable to unpack manifest data %q", contentID)
 	}
 
-	// Not needed?
+	// Will be GC-ed even if we don't close it?
+	//nolint:errcheck
 	defer gz.Close()
 
 	man, err = decodeManifestArray(gz)
--- a/repo/manifest/committed_manifest_manager.go
+++ b/repo/manifest/committed_manifest_manager.go
@@ -346,11 +346,12 @@ func loadManifestContent(ctx context.Context, b contentManager, contentID conten
 		return man, errors.Wrapf(err, "unable to unpack manifest data %q", contentID)
 	}
 
-	if err := json.NewDecoder(gz).Decode(&man); err != nil {
-		return man, errors.Wrapf(err, "unable to parse manifest %q", contentID)
-	}
+	// Not needed?
+	defer gz.Close()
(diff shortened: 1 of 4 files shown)


jkowalski (approved):
Very nice, thank you. I have couple nits to optionally consider.

FWIW Kopia had similar streaming code for parsing directory listings and indirect objects until it was removed long ago: https://github.com/kopia/kopia/commit/23d979f29009832cf51e4f33a043f01db2c3702e#diff-5427831a5ef9e154dea8d6be7d380c7f9e9c65e60d606b1c47d39aee5ae7f929

PR #3017, opened 2023-05-05
chore(ci): disable dependabot and replace with Snyk

jkowalski:
We are not using dependabot to automatically upgrade our dependencies, instead we will use Snyk to monitor our dependencies and alert us when they need to be upgraded.

This should hopefully reduce the noise in our PR process.


julio-lopez (approved):

julio-lopez:
It makes sense to try out different approaches.

The main motivations for keeping dependencies fairly up to date are:

- Picking up security fixes and other bug fixes.
- Facilitating upgrading dependencies when there is an urgent security issue that needs to be addressed right away.

From experience, when dependencies are not up to date, upgrading a critical dependency (say, for a security fix) ends up requiring a significant amount of effort, due to the need to upgrade other dependencies and breaking (API) changes in those dependencies.

@jkowalski I do not yet understand how this addresses the "PR noise" issue.

PR #3033, opened 2023-05-14
feat(repository): Support Azure storage class for AZ blobs

lesaugues:
As this feature was already present for S3, I want to add the support of storage classes for Azure. 
These modifications with the right `.storageconfig` file, allow the insertion of a blob type into a specific storage class within an Azure storage account.
As a beginner, I followed the same logic as the work that was done for S3.


jkowalski:
This is looking good overall, I have 2 requests:

* can you add a test specifically for this new behavior?
* we're using very old Azure client, which could not be upgraded to the latest one automatically because of breaking changes, since you're touching the code, would you mind also doing the upgrade?

vkamra:
> * we're using very old Azure client, which could not be upgraded to the latest one automatically because of breaking changes, since you're touching the code, would you mind also doing the upgrade?

@jkowalski @lesaugues - I have the upgrade in-progress (#3056 ) . Once I have it tested will update the PR.

jkowalski:
@lesaugues this has now been merged, please rebase.

Gui13:
Up!

PR #3034, opened 2023-05-15
feat(snapshots): added support for reading from named pipes

acul009:
Hi,

this is just 2 lines, which makes kopia read named pipes like files. I'd like to use that for streaming multiple database dumps in the same snapshot, so the backup is atomic.


fs/localfs/local_fs.go
```diff
@@ -356,6 +356,7 @@ func Directory(path string) (fs.Directory, error) {
 func entryFromDirEntry(fi os.FileInfo, prefix string) fs.Entry {
 	isplaceholder := strings.HasSuffix(fi.Name(), ShallowEntrySuffix)
 	maskedmode := fi.Mode() & os.ModeType
+	readNamedEnv := os.Getenv("KOPIA_SNAPSHOT_NAMED_PIPES")
```
jkowalski, line 359: this function is called in a tight loop, reading from environment has non-trivial cost and should be avoided in such a place.

acul009: Thanks for reviewing :)

I‘ll update the PR to cache the value.

acul009: I moved the variable outside the function. This should fix the performance issues.

Issue #3035, opened 2023-05-15
SFTP blob creation fails with "Operation unsupported" on initial setup


DButter:
When I try to create the initial backup blob it fails with the following error message

> Connect Error: INTERNAL: internal server error: connect error: unable to write blobcfg blob: PutBlob() failed for “kopia.blobcfg”: unable to complete PutBlob(kopia.blobcfg) despite 10 retries: unexpected error renaming file on SFTP: sftp: “Operation unsupported” (SSH_FX_OP_UNSUPPORTED)

i assume this happens because of this piece of code:

`err = sftpClientFromConnection(conn).PosixRename(tempFile, fullPath)`

[this is the file](https://github.com/kopia/kopia/blob/f9de453efc198b6e993af8922f953a7e5322dc5f/repo/blob/sftp/sftp_storage.go#L212)

my sftp provider writes in his documentation:

> Common commands to create, read, update, and delete, files and directories are supported.

> Directory rename operations, append operations, changing ownerships, permissions and timestamps, and use of symbolic and hard links are currently not supported.

tbh I'm not to sure how to work around this issue atm

DButter:
changed to satisfy linting, unfortunately i cant create a branch as i dont have the necessary permissions

```
		_, err = sftpClientFromConnection(conn).Stat(fullPath)
		if err != nil {
			if !isNotExist(err) {
				return errors.Wrap(err, "can't check if old backup folder exists")
			}
			log(ctx).Infof("no old backup folder exists: %v", err)
		} else {
			if err = sftpClientFromConnection(conn).Remove(fullPath); err != nil {
				return errors.Wrap(err, "can't remove leftover backup path file")
			}
		}

		err = sftpClientFromConnection(conn).Rename(tempFile, fullPath)
		if err != nil {
			if removeErr := sftpClientFromConnection(conn).Remove(tempFile); removeErr != nil {
				log(ctx).Warnf("can't remove temp file: %v", removeErr)
			}

			return errors.Wrap(err, "unexpected error renaming file on SFTP")
		}
```

PR #3040, opened 2023-05-22
Fix: SFTP blob creation fails with "Operation unsupported" on initial setup #3035

DButter:
This pull request is a fix to the following issue:

https://github.com/kopia/kopia/issues/3035

basically kopia fails if the ftp backend doesnt support the posix rename extention, a simple remove&&rename will fullfill the same purpose but won't cause this issue

### DIFF
--- a/repo/blob/sftp/sftp_storage.go
+++ b/repo/blob/sftp/sftp_storage.go
@@ -209,7 +209,19 @@ func (s *sftpImpl) PutBlobInPath(ctx context.Context, dirPath, fullPath string, 
 			return errors.Wrap(err, "can't close temporary file")
 		}
 
-		err = sftpClientFromConnection(conn).PosixRename(tempFile, fullPath)
+		_, err = sftpClientFromConnection(conn).Stat(fullPath)
+		if err != nil {
+			if !isNotExist(err) {
+				return errors.Wrap(err, "can't check if old backup folder exists")
+			}
+			log(ctx).Infof("no old backup folder exists: %v", err)
+		} else {
+			if err = sftpClientFromConnection(conn).Remove(fullPath); err != nil {
+				return errors.Wrap(err, "can't remove leftover backup path file")
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


repo/blob/sftp/sftp_storage.go
```diff
@@ -209,7 +209,19 @@ func (s *sftpImpl) PutBlobInPath(ctx context.Context, dirPath, fullPath string,
 			return errors.Wrap(err, "can't close temporary file")
 		}
 
-		err = sftpClientFromConnection(conn).PosixRename(tempFile, fullPath)
```
jkowalski, line 212: Can we do this as a fallback instead of a regular code path?

Basically we're changing 1 request to the server (posix rename) for 2-3 (stat/remove/rename) which will increase overall latency.  When we discover that server does not support posix rename we can execute slower path and not even attempt posix rename in the future.

BTW. This is also technically incorrect, since if the program were to fail at the worst possible moment, which is after removing the `fullPath`, nobody will execute `rename`, but it probably does not matter that much since Kopia never really rewrites data blobs in place.

jkowalski:
sorry about late response, this is very promising.

repo/blob/sftp/sftp_storage.go
DButter: > Can we do this as a fallback instead of a regular code path?

yup, makes sense to me, I'll need to research if there is a way to check the methods availability first, I'd rather not default to it on every error.

maybe this https://pkg.go.dev/github.com/pkg/sftp#Client.HasExtension would already be enough

> BTW. This is also technically incorrect, since if the program were to fail at the worst possible moment, which is after removing the fullPath, nobody will execute rename, but it probably does not matter that much since Kopia never really rewrites data blobs in place.

true, but there is no alternative standard way to emulate this exact behavior

DButter:
@jkowalski I think I've resolved the issue, it will now only use multiple operations as a fallback, is there anything else i need to do?

TheFitzZZ:
is there anything that can be done to make this happen?

PR #3041, opened 2023-05-24
feat(cli): add --override-source parameter to allow overriding source when snapshotting

### LINKED ISSUE #1530: spoof snapshot create source
As discussed in the forum thread raising a feature request issue for the same.
https://kopia.discourse.group/t/spoof-snapshot-create-source/748?u=saviodsouza

Brief info:
Kopia needs a  --set-source= parameter for kopia snapshot create to spoof snapshot create.
`kopia snapshot create /home/admin --set-source /home/user`
The above comand should backup /home/admin but record source as /home/user in the repository.

Thank you 😊

mschuwalow:
Resolves #1530


cli/command_snapshot_create.go
```diff
@@ -69,6 +70,7 @@ func (c *commandSnapshotCreate) setup(svc appServices, parent commandParent) {
 	cmd.Flag("tags", "Tags applied on the snapshot. Must be provided in the <key>:<value> format.").StringsVar(&c.snapshotCreateTags)
 	cmd.Flag("pin", "Create a pinned snapshot that's will not expire automatically").StringsVar(&c.pins)
 	cmd.Flag("flush-per-source", "Flush writes at the end of each source").Hidden().BoolVar(&c.flushPerSource)
+	pathVar(cmd.Flag("snapshot-path", "Directory or file that should be snapshot instead of source."), &c.snapshotPath)
```
mschuwalow, line 73: Decided to go this way instead of allowing to override the source that will be stored in the repo for more consistency with --stdin-file

cli/command_snapshot_create.go
```diff
@@ -91,17 +93,18 @@ func (c *commandSnapshotCreate) run(ctx context.Context, rep repo.RepositoryWrit
 		return errors.Wrap(err, "error upgrading repository")
 	}
 
-	if c.snapshotCreateAll {
+	switch {
+	case len(sources) == 0 && c.snapshotCreateAll:
```
mschuwalow, line 97: I can get rid of these new validations, just seemed like we have a few combinations of parameters that result in very unintuitive executions

cli/command_snapshot_create.go
mschuwalow: An interesting alternative would be to provide a --set-source flag as suggested in the issue originally that bypasses path validation to some degree.
One usecase I would see for this is that it would allow to create snapshots of a windows path from a unix system and the other way around. Disadvantage here is that we need to decide what exactly a valid path is for us and get rid of all the platform specific path handling for sourceinfo.

wdyt?

tests/end_to_end_test/snapshot_create_test.go
```diff
@@ -728,3 +729,39 @@ func TestSnapshotCreateAllFlushPerSource(t *testing.T) {
...
+	require.Equal(t, si[0].User, "bar")
+	require.Equal(t, si[0].Host, "bar")
+	require.Equal(t, si[0].Path, "/foo/bar")
+
+	require.Equal(t, si[1].User, "bar")
+	require.Equal(t, si[1].Host, "bar")
+	require.Equal(t, si[1].Path, "C:\\foo\\baz")
+
+	require.Equal(t, si[2].User, "foo")
+	require.Equal(t, si[2].Host, "foo")
```
jkowalski, line 760: please also make sure the snapshots are set to `manual`

cli/command_snapshot_create.go
```diff
@@ -69,6 +70,7 @@ func (c *commandSnapshotCreate) setup(svc appServices, parent commandParent) {
 	cmd.Flag("tags", "Tags applied on the snapshot. Must be provided in the <key>:<value> format.").StringsVar(&c.snapshotCreateTags)
 	cmd.Flag("pin", "Create a pinned snapshot that's will not expire automatically").StringsVar(&c.pins)
 	cmd.Flag("flush-per-source", "Flush writes at the end of each source").Hidden().BoolVar(&c.flushPerSource)
+	cmd.Flag("set-source", "Override the source of the snapshot.").StringVar(&c.sourceOverride)
```
jkowalski, line 73: Perhaps `override-source` ?

julio-lopez: This may be confusing since _"source"_ has a specific meaning in kopia, a snapshot source is the triplet `{host, user, path}`

Given that the intend here is to override the path, perhaps this should be `override-source-path` or `override-path`.

mschuwalow: Right now it allows overriding all three parts of the triplet (using the same logic to determine what you mean as list-snapshots, etc.)
https://github.com/kopia/kopia/pull/3041/files#diff-30637363ab70c6f1f34641e0e058998b680960f18a89107285161e3d479f04b6R745

cli/command_snapshot_create.go
```diff
@@ -470,3 +447,62 @@ func shouldSnapshotSource(ctx context.Context, src snapshot.SourceInfo, rep repo
...
+		sourceInfo, err = parseFullSource(c.sourceOverride, rep)
+
+		if err != nil {
+			return nil, sourceInfo, false, errors.Wrapf(err, "invalid source override %v", c.sourceOverride)
+		}
+
+		setManual = true
+	} else {
+		sourceInfo = snapshot.SourceInfo{
+			Path:     dir,
```
julio-lopez, line 475: Should this be the following ?
```suggestion
			Path:     filepath.Clean(absDir),
```

cli/command_snapshot_create.go
```diff
@@ -470,3 +447,62 @@ func shouldSnapshotSource(ctx context.Context, src snapshot.SourceInfo, rep repo
...
+		err        error
+	)
+
+	absDir, err = filepath.Abs(dir)
+	if err != nil {
+		return nil, sourceInfo, false, errors.Wrapf(err, "invalid source %v", dir)
+	}
+
+	if c.sourceOverride != "" {
+		sourceInfo, err = parseFullSource(c.sourceOverride, rep)
```
julio-lopez, line 466: nit: change the function parameter to only require either `rep.ClientOptions` or `hostname, username string`

cli/command_snapshot_create.go
julio-lopez: Got it. When I skimmed over the PR, I incorrectly assumed it was only overriding the `path`.

julio-lopez (approved):
merge & iterate

PR #3044, opened 2023-05-25
test(general): Delete snapshot when restore fails

chaitalisg:
During the set up of the robustness suite, a random snapshot is restored to local data directory. If the restore errors out, the test goes on. This PR adds the logic to delete the snapshot when `no space left on device` error is encountered on restore.


tests/robustness/robustness_test/main_test.go
```diff
@@ -48,7 +48,14 @@ func TestMain(m *testing.M) {
 	eng = th.engine
 
 	// Restore a random snapshot into the data directory
-	if _, err := eng.ExecAction(ctx, engine.RestoreIntoDataDirectoryActionKey, nil); err != nil && !errors.Is(err, robustness.ErrNoOp) {
+	if opts, err := eng.ExecAction(ctx, engine.RestoreIntoDataDirectoryActionKey, nil); err != nil && !errors.Is(err, robustness.ErrNoOp) {
+		if eng.ErrIsNotEnoughSpace(err) {
+			// Delete the snapshot
+			snapID := opts["snapshot-ID"]
+			log.Println("snap ID: ", snapID)
```
Shrekster, line 55:
```suggestion
			log.Println("deleting snap ID: ", snapID)
```

tests/robustness/engine/action.go
```diff
@@ -313,7 +317,8 @@ func pickActionWeighted(actionControlOpts map[string]string, actionList map[Acti
 	return keepKey
 }
 
-func errIsNotEnoughSpace(err error) bool {
+// ErrIsNotEnoughSpace checks if the given error is about no space left.
+func (e *Engine) ErrIsNotEnoughSpace(err error) bool {
```
julio-lopez, line 321: Why does this need to have a receiver that is, be a _member function_ of `e Engine`? The receiver `e` is not used in the function implementation.

It seems that all that is needed for for the function to be _exported_ so it is accessible from the `robustness` package. If that is the case, then simply name the function `ErrIsNotEnoughSpace` without adding a receiver

tests/robustness/robustness_test/main_test.go
```diff
@@ -48,7 +48,14 @@ func TestMain(m *testing.M) {
 	eng = th.engine
 
 	// Restore a random snapshot into the data directory
-	if _, err := eng.ExecAction(ctx, engine.RestoreIntoDataDirectoryActionKey, nil); err != nil && !errors.Is(err, robustness.ErrNoOp) {
+	if opts, err := eng.ExecAction(ctx, engine.RestoreIntoDataDirectoryActionKey, nil); err != nil && !errors.Is(err, robustness.ErrNoOp) {
+		if eng.ErrIsNotEnoughSpace(err) {
+			// Delete the snapshot
+			snapID := opts["snapshot-ID"]
+			log.Println("deleting snap ID: ", snapID)
+			_ = eng.Checker.DeleteSnapshot(ctx, snapID, opts)
```
julio-lopez, line 56: Is this deleting the snapshot in the repository?

julio-lopez (changes requested):
I may be missing something here, but I do not completely understand the goal of this change.

> If the restore errors out, the test goes on. This PR adds the logic to delete the snapshot when no space left on device error is encountered on restore.

Where is snapshot deleted from? The repository?
If so, why is this desirable and how does it address issues with lack of space in the (local) device that is the target for restoring?

chaitalisg:
> I may be missing something here, but I do not completely understand the goal of this change.
> 
> > If the restore errors out, the test goes on. This PR adds the logic to delete the snapshot when no space left on device error is encountered on restore.
> 
> Where is snapshot deleted from? The repository? If so, why is this desirable and how does it address issues with lack of space in the (local) device that is the target for restoring?

The first thing the robustness test does is to restore a snapshot from object store into local dir. If that restore fails with `no space left` error, it is an indication that the snapshot saved in object store corresponds to data that couldn't possibly be successfully restored in the test. If the test hits such snapshot again and again, the test will just keep throwing `no space left` error and won't move forward. 

I pushed another PR that tries to address the `no space left` issue at a different stage: 
https://github.com/kopia/kopia/pull/3043

I am hoping to make several incremental changes to handle the `no space left` error thrown at different stages in the test. This change is one of them. 

Having said that, I am open to more ideas on how to tackle this error.

tests/robustness/robustness_test/main_test.go
chaitalisg: Good point. The `eng.ExecAction` and `eng.Checker.xx` may be acting upon two different repositories (object store and local). Let me confirm this.

julio-lopez:
> The first thing the robustness test does is to restore a snapshot from object store into local dir. If that restore fails with `no space left` error, it is an indication that the snapshot saved in object store corresponds to data that couldn't possibly be successfully restored in the test. If the test hits such snapshot again and again, the test will just keep throwing `no space left` error and won't move forward.

Observations:
- If the snapshot was created from the same volume, it means that the volume could fit the snapshot at some point in the past. The assumption here is that total volume capacity is the same for both the source volume (the one where the snapshot was taken from) and the target volume (for the restore).  This likely means that what has changed is the available free space on the volume.
- The fact that the target volume does not have enough free space does not seem like a valid reason to delete the snapshot from the repo.

(*) Any idea of what else could be occupying data in the target volume?

What about a combination of the following approaches:
- Delete data from failed restores (which you already did).
- Removing any other extra, ephemeral data from the volume (e.g., log output, or anything resulting from answering the question above (*).
- Checking the free space on the volume, and the size of the chosen snapshot before restoring. If it is not going to fit, then skip it and randomly choose another snapshot to restore. Yes, over time some of the snapshots may become less eligible for restoring.

Also, does this fail for particular type of job setups?
- server vs. direct repos?
- file vs. object store repos?

All of them?

chaitalisg:
> > The first thing the robustness test does is to restore a snapshot from object store into local dir. If that restore fails with `no space left` error, it is an indication that the snapshot saved in object store corresponds to data that couldn't possibly be successfully restored in the test. If the test hits such snapshot again and again, the test will just keep throwing `no space left` error and won't move forward.
> 
> Observations:
> 
> * If the snapshot was created from the same volume, it means that the volume could fit the snapshot at some point in the past. The assumption here is that total volume capacity is the same for both the source volume (the one where the snapshot was taken from) and the target volume (for the restore).  This likely means that what has changed is the available free space on the volume.
> * The fact that the target volume does not have enough free space does not seem like a valid reason to delete the snapshot from the repo.
> 
> (*) Any idea of what else could be occupying data in the target volume?
> 
> What about a combination of the following approaches:
> 
> * Delete data from failed restores (which you already did).
> * Removing any other extra, ephemeral data from the volume (e.g., log output, or anything resulting from answering the question above (*).
> * Checking the free space on the volume, and the size of the chosen snapshot before restoring. If it is not going to fit, then skip it and randomly choose another snapshot to restore. Yes, over time some of the snapshots may become less eligible for restoring.
> 
> Also, does this fail for particular type of job setups?
> 
> * server vs. direct repos?
> * file vs. object store repos?
> 
> All of them?

what else could be occupying data in the target volume: may be cache or `/tmp/. I hav seen the error log messages pointing to `/tmp/` path. I can try deleting both but I am not sure of the outcome of deleting them.

The capacity check is next on my agenda. I was hoping these incremental changes will provide some relief till I can allocate bandwidth to implement the capacity check.

The `no space left` error is not deterministic. Sometimes it comes up during the first restore, sometimes it comes up during one of the tests. Yes, all of the pipelines have been failing intermittently in different stages with this error.

PR #3061, opened 2023-06-09
chore(general): output rewritten bytes to INFO

NickIAm:
Currently this data is only output into the debug log file and isn't displayed anywhere else. I think this is useful information to have and outputting it directly to INFO and is a good quick reference when running the maintenance manually as this goes to stdout. What do you think?


repo/maintenance/content_rewrite.go
```diff
@@ -115,7 +115,7 @@ func RewriteContents(ctx context.Context, rep repo.DirectRepositoryWriter, opt *
 
 	wg.Wait()
 
-	log(ctx).Debugf("Total bytes rewritten %v", units.BytesString(totalBytes))
+	log(ctx).Infof("Total bytes rewritten %v", units.BytesString(totalBytes))
```
jkowalski, line 118: please fix the whitespace

NickIAm: Whitespace removed

julio-lopez (approved):

PR #3065, opened 2023-06-13
chore(ci): exclude docsy theme from snyk

jkowalski:


jkowalski:
We're using fairly old version of Docsy which is major effort to upgrade which has a bunch of old build-time dependencies. Since the `site` build only runs on Netlify, build-time vulnerabilities are not a major concern since it has no access to secrets.

Issue #3068, opened 2023-06-14
Kopia parallelism: choose numCPU if less than max parallel value


MinyiZou:
Initial implementation hardcodes --parallel=8 in the kopia command to at least cap the parallelism for high num cpu systems. For low cpu system we may want to choose min(numCPU, maxParallel). An initial look into this functionality proved more problematic: the command is generated in a different container than where it will be executed, so it is not straightforward to find the cpu allocated.
One relatively simple option is to add an additional flag to kopia upstream that will choose min(numCPU, <flagValue>) if the existing flag `--parallel` is not set.

jkowalski:
we may want to implement additional policies that allow hard-caping the parallelism to a specific percentage of # cores or a fixed value, but I'm not sure if hard-capping to exactly numCPU is always a good idea.

PR #3069, opened 2023-06-14
fix(snapshots): choose numCPU if less than max parallel value

ChiziZou:
#3068 Kopia parallelism: choose numCPU if less than max parallel value


snapshot/snapshotfs/upload.go
```diff
@@ -744,6 +744,11 @@ func (u *Uploader) maybeIgnoreCachedEntry(ctx context.Context, ent fs.Entry) fs.
 
 func (u *Uploader) effectiveParallelFileReads(pol *policy.Policy) int {
 	p := u.ParallelUploads
+
+	if p > runtime.NumCPU() {
```
jkowalski, line 748: given that uploads are not entirely CPU bound there may be benefits to allowing p to be greater than the number of cores to allow max number of cores to do hashing while others are waiting for uploads

PR #3072, opened 2023-06-14
fix(repository): fix s3 list bucket permission error

### LINKED ISSUE #3071: AWS s3 - Repository creation fails under credentials with restrictive IAM roles
There is a use case that multiple users share one bucket, so AWS provides IAM role policies that grant access to a sub-dir(prefix) of a s3 bucket but restrict the access to the bucket itself.

During creating the repository, Kopia tries to request `s3:HeadBucket` to the bucket (by calling `minio.Clinet.BucketExists`). As a result, an `access denied` error will be returned.

Actually, in this case, the repository can be created successfully in the sub-dirs, so should not be blocked.

Lyndon-Li:
Skip the accessDenied error of `BucketExists`. As the case mentioned in the issue, the denying access to the bucket itself doesn't mean the denying access to the sub-dirs of the bucket.

Fixes #3071


jkowalski:
I think a better fix would be to read non-existent blob (UUID) and expect Not Found.

Lyndon-Li:
> I think a better fix would be to read non-existent blob (UUID) and expect Not Found similar to what `gcs_storage.go` does

@jkowalski Thanks for the suggestion, I have modified the code accordingly, please have a look.

Lyndon-Li:
@jkowalski 
There are some conflicts and Lint check errors. I have fix them, please have another look. Thanks!

repo/blob/s3/s3_storage.go
```diff
@@ -312,6 +312,40 @@ func (s *s3Storage) FlushCaches(ctx context.Context) error {
...
+	if s.Options.Prefix == "" {
+		return false, errors.Wrap(err, "no permission to access bucket itself")
+	}
+
+	// Verify the existence of bucket by getting an non-existent object from the bucket/prefix and
+	// expecting an BlobNotFound error, in case that the account has no permission to the bucket itself
+	// but has permission to the prefix of the bucket
+	nonExistentBlob := fmt.Sprintf("kopia-s3-storage-initializing-%v", clock.Now().UnixNano())
+
+	var scOutput gather.WriteBuffer
```
jkowalski, line 339: add `defer scOutput.Close()` and while at it rename it to something like `out`

Lyndon-Li: Done the changes. Thanks

Lyndon-Li:
@jkowalski 
BTW, can we include this fix into the next patch release, i.e., v0.13.1?
If so, what else should I do?

jkowalski:
This looks reasonable, but I'll also want @julio-lopez to take a look as I'm not sure if what you're doing with bucketExists() check will not break existing customers.

We definitely need to have some additional tests for this behavior - in https://github.com/kopia/kopia/blob/master/repo/blob/s3/s3_storage_test.go

Note you won't be able to run all the tests yourself since you don't have credentials to the official storage backends, but I can run those for you.

Lyndon-Li:
@jkowalski 
OK, let me check the existing UT code in `s3_storage_test.go` and see how to add UT for the new logics.

repo/blob/s3/s3_storage.go
```diff
@@ -312,6 +312,42 @@ func (s *s3Storage) FlushCaches(ctx context.Context) error {
 	return nil
 }
 
+func isAccessDenied(err error) bool {
+	var e minio.ErrorResponse
+	return errors.As(err, &e) && e.Code == "AccessDenied"
+}
+
+func (s *s3Storage) bucketExists(ctx context.Context) (bool, error) {
```
julio-lopez, line 320: Overall, it seems reasonable to do. The initial reaction or concern is about how this behaves with different S3 providers.

Let's make sure we provide information in the error that is useful to the user, and for debugging issues, etc.

Main comments are nits around some of the functionality and code structure.

- Make this a package (helper) function instead of a function with a receiver. Explicitly pass the `cli` client and the bucket name.
- Move (from the caller to this function) some of the error handling and wrapping.
- Return only error, instead of `bool, error`, simply check for error on the caller and propagate without wrapping at the caller

repo/blob/s3/s3_storage.go
```diff
@@ -312,6 +312,42 @@ func (s *s3Storage) FlushCaches(ctx context.Context) error {
...
+	ok, err := s.cli.BucketExists(ctx, s.Options.BucketName)
+	if err == nil {
+		return ok, nil
+	}
+
+	if !isAccessDenied(err) {
+		return false, errors.Wrap(err, "error checking bucket existence")
+	}
+
+	if s.Options.Prefix == "" {
```
julio-lopez, line 330: Can the check below (for a non-existing blob) be performed as well when the prefix is empty?

repo/blob/s3/s3_storage.go
```diff
@@ -408,12 +450,6 @@ func newStorageWithCredentials(ctx context.Context, creds *credentials.Credentia
 		return nil, errors.Errorf("bucket %q does not exist", opt.BucketName)
 	}
```
julio-lopez, line 451: Move these 2 error checks to the helper function and only return error from the helper, as mentioned above.

repo/blob/s3/s3_storage.go
```diff
@@ -312,6 +312,42 @@ func (s *s3Storage) FlushCaches(ctx context.Context) error {
...
+	var e minio.ErrorResponse
+	return errors.As(err, &e) && e.Code == "AccessDenied"
+}
+
+func (s *s3Storage) bucketExists(ctx context.Context) (bool, error) {
+	ok, err := s.cli.BucketExists(ctx, s.Options.BucketName)
+	if err == nil {
+		return ok, nil
+	}
+
```
julio-lopez, line 325: Check here whether or not the bucket exists, that is:
- when `err == nil && ok`, then `return nil`
- when `err == nil && !ok`, then `return errors.Errorf("bucket %q does not exist", bucketName)`.

Essentially, move here the check that was performed in the caller.

repo/blob/s3/s3_storage.go
```diff
@@ -312,6 +312,42 @@ func (s *s3Storage) FlushCaches(ctx context.Context) error {
 	return nil
 }
 
+func isAccessDenied(err error) bool {
+	var e minio.ErrorResponse
+	return errors.As(err, &e) && e.Code == "AccessDenied"
```
julio-lopez, line 317: Isn't there a way of using `errors.Is` here instead of explicitly comparing `e.Code`? Is the expectation that `err` will always be a `minio.ErrorResponse` and won't be wrapped?

repo/blob/s3/s3_storage.go
```diff
@@ -312,6 +312,42 @@ func (s *s3Storage) FlushCaches(ctx context.Context) error {
...
+	var blobOutput gather.WriteBuffer
+	defer blobOutput.Close()
+
+	err = translateError(s.GetBlob(ctx, blob.ID(nonExistentBlob), 0, -1, &blobOutput))
+
+	if !errors.Is(err, blob.ErrBlobNotFound) {
+		return false, errors.Wrap(err, "error getting blobs from bucket")
+	}
+
+	return true, nil
```
julio-lopez, line 348: Reverse the logic for readability, also wrap with `"unable to determine if bucket %q exists"` here.

```suggestion
	if errors.Is(err, blob.ErrBlobNotFound) {
		return nil
	}
	
	return errors.Wrapf(err, ""unable to determine if bucket %q exists", bucketName)
```

julio-lopez (commented):
@Lyndon-Li You should be able to test this locally at least against MinIO and S3 at least. I assume you have a MinIO instance for testing, or you could use the public one.

julio-lopez (changes requested):
Please address the requested changes

repo/blob/s3/s3_storage.go
julio-lopez: I do not see the changes (yet).

repo/blob/s3/s3_storage.go
Lyndon-Li: When the execution comes to this `if` block (`if s.Options.Prefix == ""`), it means that `s.cli.BucketExists` has returned an `AccessDenied` error. This already indicates that there is no permission to access the bucket itself. 
Then if the prefix is empty, it means users want to write data to the bucket directly (not to a sub dir), however, since we've proved that directly accessing to the bucket is impossible, it is safe to return an error here.

repo/blob/s3/s3_storage.go
Lyndon-Li: I have implemented the 2nd and 3rd suggestions above, but having difficulties for the 1st one -- making `bucketExists` as a helper function.
`bucketExists` needs to call `s3Storage.GetBlob` or `s3Storage.getBlobWithVersion`, which further depends on some functions from `s3Storage`, i.e., `getObjectNameString`.
I tried to modify the code, it would either cause code redundancies or a chain of changes to these functions.

Therefore, can we leave the function with `s3Storage`? If the name `bucketExists` is more like a helper function, how about change its name, e.g., reachBucket?

repo/blob/s3/s3_storage.go
Lyndon-Li: There is no wrap of `minio.ErrorResponse`, see the [minio code here](https://github.com/minio/minio-go/blob/8be815434d9eec0126e6559ea558ba60b18b8992/api-error-response.go#L120).
`minio.ErrorResponse` is not globally defined as errors, so we cannot use `errors.Is` on `minio.ErrorResponse`. We can translate it and define an global error in Kopia code like `translateError`, but eventually we need to check `e.Code` or `e.StatusCode` somewhere.

repo/blob/s3/s3_storage.go
Lyndon-Li: Done

repo/blob/s3/s3_storage.go
Lyndon-Li: Done, thanks

repo/blob/s3/s3_storage.go
Lyndon-Li: Done

repo/blob/s3/s3_storage.go
Lyndon-Li: For now, I've changed the function name to `bucketAccessible`. Let me know for any other suggestions.

Lyndon-Li:
@julio-lopez
> You should be able to test this locally at least against MinIO and S3. I assume you have a MinIO instance for testing, or you could use the public one.

Yes, I have local minio and I also have an AWS account (not sure if I can use it for Kopia test).
I am trying to figure out the logics of existing test cases, so as to decide whether to create a new case or reusing a existing one.

I may need some more time on creating the tests, meanwhile, we can continue discussing the code changes.

Lyndon-Li:
@julio-lopez 
Thanks for your review and suggestions!
I have addressed your comments, some have been reflected in the code and some need further discussion and I have added my comment inline.
Please have another look.

repo/blob/s3/s3_storage.go
julio-lopez: nit: what about `checkBucketAccess` ?

repo/blob/s3/s3_storage.go
```diff
@@ -312,6 +312,45 @@ func (s *s3Storage) FlushCaches(ctx context.Context) error {
 	return nil
 }
 
+func isAccessDenied(err error) bool {
+	var e minio.ErrorResponse
+	return errors.As(err, &e) && e.Code == "AccessDenied"
+}
+
+func (s *s3Storage) bucketAccessible(ctx context.Context) error {
```
julio-lopez, line 320: nit:
```suggestion
func (s *s3Storage) checkBucketAccess(ctx context.Context) error {
```

repo/blob/s3/s3_storage.go
```diff
@@ -399,21 +438,17 @@ func newStorageWithCredentials(ctx context.Context, creds *credentials.Credentia
...
-		return nil, errors.Errorf("bucket %q does not exist", opt.BucketName)
-	}
-
 	s := s3Storage{
 		Options:       *opt,
 		cli:           cli,
 		storageConfig: &StorageConfig{},
 	}
 
+	err = s.bucketAccessible(ctx)
```
julio-lopez, line 447: nit:
```suggestion
	err = s.checkBucketAccess(ctx)
```

julio-lopez (commented):
Provider tests for S3 stores passed.

See comment about eliminating the GetBlob check.

repo/blob/s3/s3_storage.go
Lyndon-Li: For this function (`checkBucketAccess`), the purpose still abides the original one --- check the existence of the bucket. Merely, because of the exception described in #3071, there is a false report if we only call `s.cli.BucketExists`. 
We read against the prefix, is not to assure the Get permission, but to assure the existence of the bucket dir -- `<bucket name>/<prefix>`. Basically, we can use any operation to do the check, i.e., `PutBlob`, `ListBlobs`, while get a non-existing blob is with the least cost.

Or in another word, this function is an EX version of the original `s.cli.BucketExists` function, which assures the existence of the bucket and works in all cases.

We don't aim to check all the permissions that will be used by the later repo operations. If there is indeed a permission error for further repo operations, the error will be returned along with the operation itself, e.g., following this function, we will call `GetBlob` to get the `.storageconfig` file, where the read object permission will be checked.

Lyndon-Li: Let me know if we want to do more checks based on the original purpose.

repo/blob/s3/s3_storage.go
```diff
@@ -312,6 +312,45 @@ func (s *s3Storage) FlushCaches(ctx context.Context) error {
...
+	// expecting an BlobNotFound error, in case that the account has no permission to the bucket itself
+	// but has permission to the prefix of the bucket
+	nonExistentBlob := fmt.Sprintf("kopia-s3-storage-initializing-%v", clock.Now().UnixNano())
+
+	var blobOutput gather.WriteBuffer
+	defer blobOutput.Close()
+
+	err = translateError(s.GetBlob(ctx, blob.ID(nonExistentBlob), 0, -1, &blobOutput))
+
+	if errors.Is(err, blob.ErrBlobNotFound) {
```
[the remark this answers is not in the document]
Lyndon-Li: I think the description from the document missed the factor of the prefix, in our scenarios it is like:

> If you have the s3:ListBucket permission on the bucket/prefix, Amazon S3 returns an HTTP status code 404 (Not Found) error.
> If you don’t have the s3:ListBucket permission on the bucket/prefix, Amazon S3 returns an HTTP status code 403 ("access denied") error.

At least, my local testing result is like this.
If so, this could meet our purpose, while 1 is our expectation the function returns `nil` for it:

1. If the account doesn't have permission to the bucket, but has permission to bucket/prefix, then 404 is returned
2. If the account doesn't have permission to the bucket and neither to bucket/prefix, then 403 is returned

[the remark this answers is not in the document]
Lyndon-Li: > Perhaps, if the error above (when checking for the bucket existence) is "access denied", simply log and return nil. Essentially, things "may or may not" work after that.

This is what the initial changes in this PR, while @jkowalski thought checking a non-existing object against the prefix  is better as it could also check whether the account has permission to the prefix dir. See here https://github.com/kopia/kopia/pull/3072#issuecomment-1596235663

repo/blob/s3/s3_storage.go
Lyndon-Li: Done, change it to `checkBucketAccess`

repo/blob/s3/s3_storage.go
Lyndon-Li: Done

repo/blob/s3/s3_storage.go
Lyndon-Li: Done

repo/blob/s3/s3_storage.go
julio-lopez: The request is to either:
- Preferred: not perform the blob existence check at all when `BucketExists(...)` returns access denied. Simply log, and `return nil`. See below.
- Or perform the blob existence check in all cases, not just when the prefix is empty. This is less preferred, because of the reasons explained in the comment below.


```Go
	if isAccessDenied(err) {
		// log.Infof("unable to determine if bucket %q exists, error: %s", s.Options.BucketName, err)
		return nil
	}
```

repo/blob/s3/s3_storage.go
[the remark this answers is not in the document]
julio-lopez: > This is what the initial changes in this PR, while @jkowalski thought checking a non-existing object against the prefix is better as it could also check the existence of the prefix dir. See here [#3072 (comment)](https://github.com/kopia/kopia/pull/3072#issuecomment-1596235663).

The suggestion is to check for a specific non-existing blob that should return "not found", because either: (1) the bucket does not exist; or (2) the blob does not exist. This does not check the existence of a prefix "directory" object.

At this point in this function, the return from `BucketExists(...)` is "access denied", which likely means the credential does not have permissions to list buckets or to list the name (not the contents) of this specific bucket.

According to the AWS S3 documentation, at this point doing a GET of a non-existent blob should always return "access denied" (here), and never "blob not found".

@Lyndon-Li Can you test this scenario?

- existing bucket
- credentials with:
  1. no access to listing this bucket, or all buckets
  2. access to listing, get, put etc of all blobs inside the bucket.
- perform a get of a non-existent blob with no prefix
- perform a get of a non-existent blob with an additional prefix

Also, what happens when the credential / principal does not have access to list the bucket and the bucket does not exist?
- Does `BucketExists(...)` return "access denied" or "not found"?
- What would the blob existence check return in that case?

Note:
There is no concept or distinction of directories in AWS S3, it is a flat key namespace. 
- An object named `dir/` is simply an object and it can have arbitrary contents.
- It is possible to have an object named `dir/object-1` without requiring the existence of another object with the same prefix, e.g. `dir`, or `dir/`.
This may not apply to other S3 "compatible" implementations.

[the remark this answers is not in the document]
Lyndon-Li: For sure, for an object store, there is no concept of dir. On the other hand, inside the AWS S3 policy, the permission of a bucket/prefix could be different from the bucket itself, like below configuration snippet:
        {
            "Effect": "Allow",
            "Action": [
                "s3:Get*",
                "s3:List*"
            ],
            "Resource": "arn:aws:s3:::bucket-123",
            "Condition": {
                "StringLike": {
                    "s3:prefix": "prefix-123/*"
                }
            }
        },
The above configuration means, the Get and List permission is only applied to `bucket-123/prefix-123`, which could be analogised as that the account has access to the sub-dir only.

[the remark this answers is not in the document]
Lyndon-Li: I have done the above tests through the current code in this PR, here are the results.
**no access to listing this bucket + perform a get of a non-existent blob with no prefix**
```
kopia repository create s3 --bucket=bucket-123
ERROR can't connect to storage: no permission to access bucket itself, bucket "bucket-123": Access Denied.
```

**no access to listing this bucket + perform a get of a non-existent blob with an additional prefix that the account has permission to access**
```
kopia repository create s3 --bucket=bucket-123 --prefix=prefix-123/test1/
Initializing repository with:
  block hash:          BLAKE2B-256-128
  encryption:          AES256-GCM-HMAC-SHA256
  splitter:            DYNAMIC-4M-BUZHASH
Connected to repository.
```

**no access to listing this bucket + perform a get of a non-existent blob with an additional prefix that the account has no permission to access**
```
kopia repository create s3 --bucket=bucket-123 --prefix=prefix-456/test1/
ERROR can't connect to storage: unable to determine if bucket dir is accessible, bucket "bucket-123", prefix "prefix-456/test1/": Access Denied
```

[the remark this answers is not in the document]
Lyndon-Li: Finally, I agree with this statement `This may not apply to other S3 "compatible" implementations`.
And though we have successfully tested the current code against AWS S3, we have no way to test it against all S3 compatible storages.

Therefore, looks like the preferred approach mentioned above is safer:

- Preferred: not perform the blob existence check at all when BucketExists(...) returns access denied. Simply log, and return nil. 

This will not able to capture **Test 3**, but it is totally OK, because later when the repository does further operations, the fail & abort will happen.

What do you think? @julio-lopez @jkowalski 
If you both agree, I will change the code accordingly.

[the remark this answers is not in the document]
julio-lopez: Hi @Lyndon-Li 
Thanks for running the tests, the explanation and your patience. This helped me understand the issues better.

After a conversation with Jarek, it seems that the best route is to simply remove the original bucket existence check and NOT perform any additional checks, that is, remove the helper altogether. The access check will be implicitly performed when getting the "storage config" blob below.

Thanks.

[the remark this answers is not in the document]
Lyndon-Li: Great! That is the simplest way, let me change the code accordingly.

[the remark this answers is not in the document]
Lyndon-Li: @julio-lopez @jkowalski 
I've changed the code accordingly. Please have another look.
Additionally, I didn't see any testing code is checking the return values of original bucket existence check, so I assume we don't need to change the testing code. And let me know any other thing I need to do in order to merge this PR.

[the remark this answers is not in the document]
Lyndon-Li: Here are the test results generated by the new code for the similar test cases above:
**Test 1: no access to this bucket + create repo to the bucket with no prefix**
```
kopia repository create s3 --bucket=bucket-123
ERROR can't connect to storage: error retrieving storage config from bucket "bucket-123": Access Denied
```

**Test 2: no access to this bucket + create repo with an additional prefix that the account has permission to access**
```
kopia repository create s3 --bucket=bucket-123 --prefix=prefix-123/test1/
Initializing repository with:
  block hash:          BLAKE2B-256-128
  encryption:          AES256-GCM-HMAC-SHA256
  splitter:            DYNAMIC-4M-BUZHASH
Connected to repository.
```

**Test 3: no access to listing this bucket + create a repo with an additional prefix that the account has no permission to access**
```
kopia repository create s3 --bucket=bucket-123 --prefix=prefix-456/test1/
ERROR can't connect to storage: error retrieving storage config from bucket "bucket-123": Access Denied
```

julio-lopez:
Thanks @Lyndon-Li

julio-lopez (approved):

PR #3078, opened 2023-06-18
chore(site): upgraded site to latest hugo and removed vendored files

jkowalski:
This switches to using github.com/google/docsy as a Hugo module, so we don't need to vendor the theme anymore or its dependencies.


basldfalksjdf:
I went through the updated site preview and did not see anything wrong or broken. Seems fine to me.

PR #3085, opened 2023-06-19
fix(cli): log cache full when on debug level

### LINKED ISSUE #3059: Maintenance job getting ~2000 times the same file, with huge performance and cost impact
Hello

While running a full maintenance job, we're having the following `GetBlob` call **1937** times : 

`2023-06-08T07:58:40.918587Z DEBUG kopia/repo [STORAGE] GetBlob	{"blobID":"q894b24492b1c49d925c43a851a35fc3e-s432977430d515f6311b","offset":0,"length":-1,"outputLength":21143278,"error":null,"duration":"2.0008945s"}`

The file is ~20Mb.
First log : `2023-06-08T07:12:34.026410Z`
Last log : `2023-06-08T07:58:40.918587Z`

So Kopia is spending **46 minutes** getting the same 20MB files 1937 times, consuming **38Gb** of bandwidth on our S3.
We are seeing this behavior for all the blob files.

The last successful maintenance job on this bucket have succed in more than 72 hours.

We are using Scaleway S3 Object Storage.

Is there a way to have a local cache for the maintenance job, and avoid having so many wasted time and bandwidth ?

arouene:
Simple debug log addition to help diagnose of duplicate downloading of blobs.
fixes #3059


arouene:
I was wondering, how about to set it to *warning* level instead of *debug* level ? As a Kopia user, I would like to know if the cache is full, leading to an increase cost of bandwidth, and an increase time to achieve the maintenance.

internal/cache/persistent_lru_cache.go
```diff
@@ -231,6 +231,7 @@ func (c *PersistentCache) Put(ctx context.Context, key string, data gather.Bytes
 		// MUST NOT go over the specified limit for the cache space to avoid
 		// snapshots/restores from getting affected by the cache's storage use.
 		if c.isCacheFullLocked() {
+			log(ctx).Warnf("Cache is full, unable to add %v into cache.", key)
```
jkowalski, line 234: we should add some kind of rate limiter here, for example by logging this message no more than once/minute, otherwise it will spam the logs.

jkowalski: alternatively make that configurable via a flag or environment variable.

arouene: I added a condition to limit to one warning max per minute, is the implementation ok ?

internal/cache/persistent_lru_cache.go
```diff
@@ -231,6 +235,11 @@ func (c *PersistentCache) Put(ctx context.Context, key string, data gather.Bytes
 		// MUST NOT go over the specified limit for the cache space to avoid
 		// snapshots/restores from getting affected by the cache's storage use.
 		if c.isCacheFullLocked() {
+			// Limit warnings to one per minute max.
+			if time.Now().Sub(lastCacheFullWarning) > time.Minute {
+				lastCacheFullWarning = time.Now()
```
julio-lopez, line 240: This is racy. Updating `lastCacheFullWarning needs a synchronization primitive.

julio-lopez: @arouene You are right, I missed the lock block there, so there was no race.

julio-lopez:
@arouene Please address the linter issues. Thanks.

arouene:
> @arouene Please address the linter issues. Thanks.

Done, thanks !

julio-lopez (approved):

internal/cache/persistent_lru_cache.go
```diff
@@ -231,6 +232,13 @@ func (c *PersistentCache) Put(ctx context.Context, key string, data gather.Bytes
 		// MUST NOT go over the specified limit for the cache space to avoid
 		// snapshots/restores from getting affected by the cache's storage use.
 		if c.isCacheFullLocked() {
+			// Limit warnings to one per minute max.
+			if clock.Now().Sub(c.lastCacheWarning) > time.Minute {
+				c.lastCacheWarning = clock.Now()
+
+				log(ctx).Warnf("Cache is full, unable to add %v into cache.", key)
```
julio-lopez, line 239: We should make this `Debug` down the road.

Shrekster: This can be pretty noisy, even once per minute. Some workflows take several hours.

If the cache is full, are not we making the situation worse by filling up the logs ??

julio-lopez: @Shrekster yes, we should
- reduce the frequency
- add counters

Shrekster: I discussed this partly offline. I think the write rate could be reduced, but even better I'd propose we just log once and then mute.

If we can settle on logging this once then this doesn't even need to be a debug log and that could be more useful upfront than having to turn on debug logs.

The alternate question to "how much logging is bad logging?" is maybe "what frequency is useful ?"

IMO once is ok for this becuase logging this the second time is only telling me that nothing got expired, but what debugging info do I get from it ? If cache was full once and the perf was crappy then that probably explains it.

Is the second log going to add more info ? Yes!

Does that more info help ? I don't know... doesn't seem like yet

julio-lopez: @Shrekster PTAL at #3193

Issue #3097, opened 2023-06-29
npm audit failures cause CI to fail for all PRs


ashmrtn:
CI is [failing](https://github.com/kopia/kopia/actions/runs/5362438323/jobs/9729302088?pr=3092#step:7:38) during setup. From some of the logs, it appears that some vulnerable packages are causing `npm audit` to fail which makes `make` fail while setting up the environment thus blocking all PRs. For the current issue, it seems like downgrading `electron-first-run` will solve things

Moving the `npm audit` check to some release workflow may lead to better dev experience as it would allows unrelated PRs to continue to merge while still ensuring kopia releases were blocked until nopde package versions without vulnerabilities were added

julio-lopez:
fixed

PR #3100, opened 2023-07-03
fix(ci): Downgrade electron to fix npm audit error

ashmrtn:
npm is reporting a vulnerability in electron-first run which is breaking CI. This downgrades the offending package to fix the audit report and unblock things

An alternative would be to update the npm commands in the makefile(s) to ignore moderate severity vulnerabilities. I'm open to other alternatives, my knowledge of npm and node in general is fairly limited

solves the immediate issue in #3097 but doesn't do any long-term fix for keeping CI unblocked if vulnerabilities pop up

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "auto-launch": "^5.0.5",
-        "electron-first-run": "^3.0.0",
+        "electron-first-run": "^2.0.0",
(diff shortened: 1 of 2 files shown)


jkowalski:
Downgrading seems counter-intuitive, let me see if there's a solution that moves us forward...

ashmrtn:
I agree it seems counter-intuitive. Unfortunately, this is what npm gives you if you run `npm audit fix --force --production`

ashmrtn:
closing as another PR has a better solution

PR #3103, opened 2023-07-03
chore(ci): group dependabot updates and make them weekly

jkowalski:
This generates larger PRs for multiple dependencies at a time, such as https://github.com/jkowalski/kopia/pull/476

It uses relatively new feature of Dependabot described here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,20 @@ updates:
     directory: "/"
     open-pull-requests-limit: 5
     schedule:
-      interval: daily
+      interval: weekly
+    groups:
+      # create large PR upgrading multiple infrastructure dependencies in one shot,
+      # only include upstream dependencies that are stable and have somewhat
+      # regular releases which would be otherwise hard to manually manage.
+      common-golang-dependencies:
+        patterns:
+        - "cloud.google.com/*"
+        - "github.com/aws/aws-sdk-go/*"
+        - "github.com/minio/minio-go/*"
+        - "golang.org/x/*"
+        - "google.golang.org/*"
+        - "github.com/prometheus/*"
+        - "go.opentelemetry.io/*"
   - package-ecosystem: github-actions
     directory: "/"
     open-pull-requests-limit: 3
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,7 +4,20 @@ updates:
     directory: "/"
     open-pull-requests-limit: 5
     schedule:
-      interval: daily
+      interval: weekly
+    groups:
+      # create large PR upgrading multiple infrastructure dependencies in one shot,
+      # only include upstream dependencies that are are stable, have somewhat
```
Shrekster, line 10:
```suggestion
      # only include upstream dependencies that are stable, have somewhat
```

jkowalski: done

.github/dependabot.yml
```diff
@@ -4,7 +4,20 @@ updates:
...
+      # regular releases which would be otherwise hard to manually manage.
+      common-golang-dependencies:
+        patterns:
+        - "cloud.google.com/*"
+        - "github.com/aws/aws-sdk-go/*"
+        - "github.com/minio/minio-go/*"
+        - "golang.org/x/*"
+        - "google.golang.org/*"
+        - "github.com/prometheus/*"
+        - "go.opentelemetry.io/*"
```
Shrekster, line 20: this is largely ok, just a minor issue ...

otel updates can particularly conflict for people those who import kopia SDK along with k8s imports in the same project. I think k8s also has an otel dependency, that if mismatches versions with kopia, could have weird compile time issues

@julio-lopez WDYT ?

Shrekster (approved):
LGTM, just a minor otel comment. can merge otherwise

.github/dependabot.yml
jkowalski: that is a problem regardless of this PR, right?

Shrekster: that is true 90% of the times. each update has been some manual effort so far

also, to be clear I don't think this is a kopia issue, just sdk inter dependency conflicts.

julio-lopez (commented):
+1: LG 👍

.github/dependabot.yml
julio-lopez: Yes. However, having the updates for opentelemetry as separate PRs may make it easier to "downgrade" if needed. Having said that, it is probably the same  effort as (or even more than) creating a separate commit with the required dep downgrades.

PR #3129, opened 2023-07-06
chore(general): cleanup "stale" error naming

julio-lopez:
Renames `IsESTALE(error)` to `IsStale(error)` to make it consistent with the conventions in the Go ecosystem.

Uses `errors.Is()` instead of `==` comparison. It is more robust.


Shrekster (approved):
LGTM! Thanks!

PR #3131, opened 2023-07-07
feat(providers): support for auth with Azure AD

### LINKED ISSUE #2800: Support authorize requests with Azure AD
[Azure AD](https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-requests-to-azure-storage) is the recommended way to authorize the requests to access the Azure storage. 

In some user cases, Azure AD is the only allowed way to do the auth, and getting the access key from Azure AD isn't allowed.

With Azure AD support, users can use Azure AD credentials directly and it is also helpful for integration with other products.

ywk253100:
Add support for auth with Azure AD

Fixes #2800


ywk253100:
We(Velero) consume the Kopia as a library, so I make the Azure provider exported in this PR so that we can wrap/extend it easily.

julio-lopez (changes requested):
This is not adding auth Azure AD functionality, it is only exporting internal symbols.

Please submit a PR with the enhancements you propose.

The internal implementation should remain unexported. You can copy the source and use it if needed.

repo/blob/azure/azure_storage.go
```diff
@@ -275,18 +276,23 @@ func New(ctx context.Context, opt *Options, isCreate bool) (blob.Storage, error)
 			fmt.Sprintf("https://%s/", storageHostname), cred, nil,
 		)
 
+	// Azure AD
 	default:
-		return nil, errors.Errorf("either storage key or SAS token must be provided")
+		cred, err := azidentity.NewDefaultAzureCredential(nil)
```
ywk253100, line 281: @julio-lopez Here is the code block for supporting Azure AD

Shrekster:
I agree with what @julio-lopez said above. There is no maintenance / work done to ensure that the external contract with the exposed objects will be honored.

ywk253100:
@julio-lopez @Shrekster Please take a look again

repo/blob/azure/azure_storage.go
```diff
@@ -274,9 +277,13 @@ func New(ctx context.Context, opt *Options, isCreate bool) (blob.Storage, error)
 		service, serviceErr = azblob.NewClientWithSharedKeyCredential(
 			fmt.Sprintf("https://%s/", storageHostname), cred, nil,
 		)
-
+	// Azure AD
```
julio-lopez, line 280: Should this should become an explicit option in the configuration, instead of being the default?

repo/blob/azure/azure_storage.go
julio-lopez: @ywk253100 Can you add a test for this?

take a look at the "storage provider" test in this package for an example of how to pass credentials to tests. Now, in this case this is falling back to the SDK default behavior, so that may require a different set of tests.

julio-lopez (commented):
@onkarbhat @leuyentran can you take a look please?

repo/blob/azure/azure_storage.go
leuyentran: hi @ywk253100 and @julio-lopez , the usage of `NewDefaultAzureCredential` is fine in this case. However, I'd like to point out that there might be an implementation bug in this function:
`NewDefaultAzureCredential` checks through 4 different possible Az credentials type and keeps track of the results via an array of type `TokenCredential` (an interface) called `creds`. The returned credentials in each check is an implementation of interface `TokenCredential`.  If a credential check is successful `NewDefaultAzureCredential` append  the returned credential type. If the credential check failed, `NewDefaultAzureCredential` appends an error reporter which is also an implementation of interface `TokenCredential".  
At the end of function, `NewDefaultAzureCredential` returns error if either of this is true:
1.  len of `creds` is 0 (in `defaultAzureCredentialConstructorErrorHandler` and also `NewChainedTokenCredential`) 
2.  any of the element in `creds` is `nil` (`NewChainedTokenCredential`) 

IIUC, len of `creds` will always be 4. Even if none of the check succeed, `creds` array would still contain 4 error reoporter which are 4 implementation of `TokenCredential`. 
What that means for us here is that we no longer fail early here. If we would like to continue early failure, my suggestion is to add 
```go
// get NewDefaultAzureCredential(nil) cred
token, err:= cred.GetToken(ctx, nil)
// then check for error
```

repo/blob/azure/azure_storage.go
ywk253100: The `DefaultAzureCredential` supports [several different credentials](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/environment_credential.go#L44-L68), how about that we only support some of them(e.g. only support `AZURE_TENANT_ID`, `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET`) and let users provide them via the CLI options just like `SASToken` and `StorageKey`?

repo/blob/azure/azure_storage.go
julio-lopez: @leuyentran Thanks for the deep and thorough analysis.

repo/blob/azure/azure_storage.go
julio-lopez: @ywk253100 Yes, explicitly adding the needed support as you suggest seems to be the desire approach, at least at first. Let's go down that path. And please add tests for the new functionality.

As @leuyentran pointed out, this is adding the ability to use various authentication methods (often referred to as SDK default creds), not only Azure AD. The problem I see with default credentials is that it is not explicit, which is often harder to track when there are issues.

> DefaultAzureCredential is a default credential chain for applications that will deploy to Azure. It combines credentials suitable for deployment with credentials suitable for local development. It attempts to authenticate with each of these credential types, in the following order, stopping when one provides a token:
>
> [EnvironmentCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#EnvironmentCredential)
> [WorkloadIdentityCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#WorkloadIdentityCredential), if environment variable configuration is set by the Azure workload identity webhook. Use [WorkloadIdentityCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#WorkloadIdentityCredential) directly when not using the webhook or needing more control over its configuration.
> [ManagedIdentityCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#ManagedIdentityCredential)
> [AzureCLICredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#AzureCLICredential)

If we decide to allow the usage of "default" creds, then the comment above (`// Azure AD`) should be updated. Currently it is misleading.

repo/blob/azure/azure_storage.go
julio-lopez: @ywk253100 you are right, the other two cases are not trying to get a token to validate the credentials. So, let's not add that here.

Having said that, @leuyentran point stands, the current implementation for `NewDefaultAzureCredential` does not return an error, even if there were errors attempting to obtain the different types of default credentials. It does not return an error and instead returns an error provider that will (lazily) fail later when `GetToken` is called.

leuyentran: I opened an issue for Azure SDK: https://github.com/Azure/azure-sdk-for-go/issues/21301

ywk253100:
The UT passed:
```
KOPIA_PROVIDER_TEST=true KOPIA_AZURE_TEST_CONTAINER=mgmt KOPIA_AZURE_TEST_STORAGE_ACCOUNT=xxx KOPIA_AZURE_TEST_TANENT_ID=xxx KOPIA_AZURE_TEST_CLIENT_ID=xxx KOPIA_AZURE_TEST_CLIENT_SECRET=xxx go test github.com/kopia/kopia/repo/blob/azure  -run TestAzureStorageClientSecret
ok  	github.com/kopia/kopia/repo/blob/azure	83.306s
```

repo/blob/azure/azure_storage_test.go
```diff
@@ -23,10 +23,13 @@ import (
...
 const (
-	testContainerEnv       = "KOPIA_AZURE_TEST_CONTAINER"
-	testStorageAccountEnv  = "KOPIA_AZURE_TEST_STORAGE_ACCOUNT"
-	testStorageKeyEnv      = "KOPIA_AZURE_TEST_STORAGE_KEY"
-	testStorageSASTokenEnv = "KOPIA_AZURE_TEST_SAS_TOKEN"
+	testContainerEnv           = "KOPIA_AZURE_TEST_CONTAINER"
+	testStorageAccountEnv      = "KOPIA_AZURE_TEST_STORAGE_ACCOUNT"
+	testStorageKeyEnv          = "KOPIA_AZURE_TEST_STORAGE_KEY"
+	testStorageSASTokenEnv     = "KOPIA_AZURE_TEST_SAS_TOKEN"
+	testStorageTanentIDEnv     = "KOPIA_AZURE_TEST_TANENT_ID"
```
julio-lopez, line 30:
```suggestion
	testStorageTenantIDEnv     = "KOPIA_AZURE_TEST_TENANT_ID"
```

repo/blob/azure/azure_storage_test.go
```diff
@@ -163,6 +166,44 @@ func TestAzureStorageSASToken(t *testing.T) {
 	require.NoError(t, providervalidation.ValidateProvider(ctx, st, blobtesting.TestValidationOptions))
 }
 
+func TestAzureStorageClientSecret(t *testing.T) {
+	t.Parallel()
+	testutil.ProviderTest(t)
+
+	container := getEnvOrSkip(t, testContainerEnv)
+	storageAccount := getEnvOrSkip(t, testStorageAccountEnv)
+	tanentID := getEnvOrSkip(t, testStorageTanentIDEnv)
```
julio-lopez, line 175:
```suggestion
	tenantID := getEnvOrSkip(t, testStorageTenantIDEnv)
```

repo/blob/azure/azure_storage_test.go
```diff
@@ -163,6 +166,44 @@ func TestAzureStorageSASToken(t *testing.T) {
...
+
+	ctx := testlogging.Context(t)
+
+	// use context that gets canceled after storage is initialize,
+	// to verify we do not depend on the original context past initialization.
+	newctx, cancel := context.WithCancel(ctx)
+	st, err := azure.New(newctx, &azure.Options{
+		Container:      container,
+		StorageAccount: storageAccount,
+		TenantID:       tanentID,
```
julio-lopez, line 190:
```suggestion
		TenantID:       tenantID,
```

repo/blob/azure/azure_storage.go
```diff
@@ -260,23 +261,32 @@ func New(ctx context.Context, opt *Options, isCreate bool) (blob.Storage, error)
...
 		service, serviceErr = azblob.NewClientWithSharedKeyCredential(
 			fmt.Sprintf("https://%s/", storageHostname), cred, nil,
 		)
+	// client secret
+	case opt.TenantID != "" && opt.ClientID != "" && opt.ClientSecret != "":
+		cred, err := azidentity.NewClientSecretCredential(opt.TenantID, opt.ClientID, opt.ClientSecret, nil)
+		if err != nil {
+			return nil, errors.Wrap(err, "unable to initialize client secret credential")
+		}
+		service, serviceErr = azblob.NewClient(fmt.Sprintf("https://%s/", storageHostname), cred, nil)
```
julio-lopez, line 286:
```suggestion

		service, serviceErr = azblob.NewClient(fmt.Sprintf("https://%s/", storageHostname), cred, nil)
```

julio-lopez (approved):

julio-lopez:
@ywk253100 Thank you

don4of4:
Hey all -- just an FYT that the docs don't appear to have this feature noted.   [A polish pass here might be helpful.](https://kopia.io/docs/repositories/#azure-blob-storage)

PR #3142, opened 2023-07-12
test(general): add crash consistency test to validate repository consistency when kopia stops abruptly

MinyiZou:
The main objective is to verify and ensure that the Kopia backup tool does not cause corruption in the repository when a crash occurs, particularly triggered by an unhandled SIGKILL signal. Most important thing is to support a dedicated pipeline for stress and longevity testing in Crash Consistency Testing.


tests/recovery/blobmanipulator/blobmanipulator.go
```diff
@@ -274,6 +274,21 @@ func (bm *BlobManipulator) SetUpSystemUnderTest() error {
 	return nil
 }
 
+// GenerateRandomFiles connects or creates a kopia repo, writes random data in source directory,
+// creates snapshots of the source directory.
```
Shrekster, line 278:
```suggestion
// GenerateRandomFiles connects or creates a Kopia repository that writes random data in source directory.
// Tests can later create snapshots from the source directory.
```

tests/robustness/crash_consistency_test/crash_consistency_test.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+func TestConsistencyWhenKill9AfterModify(t *testing.T) {
+	// create, connect repository
+	dataRepoPath := path.Join(*repoPathPrefix, dirPath+dataPath)
+	baseDir := makeBaseDir()
+	bm, err := blobmanipulator.NewBlobManipulator(baseDir, dataRepoPath)
+	if err != nil {
+		if errors.Is(err, kopiarunner.ErrExeVariableNotSet) {
+			log.Println("Skipping recovery tests because KOPIA_EXE is not set")
```
Shrekster, line 32: You want to mark this test as skipped instead of success, so better to use the testing skip interface for that.

```suggestion
			t.Skip("Skipping recovery tests because KOPIA_EXE is not set")
```

tests/robustness/crash_consistency_test/crash_consistency_test.go
```diff
@@ -0,0 +1,262 @@
...
+		} else {
+			log.Println("Error creating Blob Manipulator:", err)
+		}
+		os.Exit(0)
+	}
```
Shrekster, line 37: There is a nice error checking interface from `github.com/stretchr/testify/require` package which you can import and use to make this error handling a bit simpler:

```suggestion
	if errors.Is(err, kopiarunner.ErrExeVariableNotSet) {
		t.Skip("Skipping recovery tests because KOPIA_EXE is not set")
	}

	require.NoError(t, err)
```

tests/robustness/crash_consistency_test/crash_consistency_test.go
```diff
@@ -0,0 +1,262 @@
...
+	bm.DataRepoPath = dataRepoPath
+
+	// generate 10M * 10 dataset
+	fileSize := 1 * 1024 * 1024
+	numFiles := 100
+	err = bm.GenerateRandomFiles(fileSize, numFiles)
+	if err != nil {
+		log.Println("Error in creating snapshot data", err)
+		t.FailNow()
+	}
```
Shrekster, line 48: Same suggestion as above, I'll stop point these out now. You can use the `require` interface on most other changes as well.

```suggestion
	require.NoError(t, err, "failed creating source data for the snapshot operation")
```

tests/robustness/crash_consistency_test/crash_consistency_test.go
```diff
@@ -0,0 +1,262 @@
...
+	// create snapshot successfully at the first time
+	// populate the kopia repo under test with random snapshots
+	snapID, _, err := bm.TakeSnapshot(bm.PathToTakeSnapshot)
+	if err != nil {
+		log.Println("Error in creating snapshot", err)
+		t.FailNow()
+	}
+	dst := getRootDir(bm.PathToTakeSnapshot)
+
+	preResotrePath := filepath.Join(dirPath, "restore_pre")
```
Shrekster, line 59:
```suggestion
	preRestorePath := filepath.Join(dirPath, "restore_pre")
```

tests/robustness/crash_consistency_test/crash_consistency_test.go
```diff
@@ -0,0 +1,262 @@
...
+	fileSize := 1 * 1024 * 1024
+	numFiles := 100
+	err = bm.GenerateRandomFiles(fileSize, numFiles)
+	if err != nil {
+		log.Println("Error in creating snapshot data", err)
+		t.FailNow()
+	}
+
+	// create snapshot successfully at the first time
+	// populate the kopia repo under test with random snapshots
```
Shrekster, line 51:
```suggestion
	// Create the first snapshot and expect it to run successfully.
	// This will populate the repository under test with initial data snapshot.
```

tests/robustness/crash_consistency_test/crash_consistency_test.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	dst := getRootDir(bm.PathToTakeSnapshot)
+
+	preResotrePath := filepath.Join(dirPath, "restore_pre")
+	// try to restore a snapshot named restore_pre
+	stdout, err := bm.RestoreGivenOrRandomSnapshot("", preResotrePath)
+	if err != nil {
+		log.Println("Error restoring the kopia repository:", stdout, err)
+		t.FailNow()
+	}
```
Shrekster, line 65: Hmmm. I am trying to understand this step a little better. What do you think is the purpose of trying to perform a restore before the second snapshot ?

I think you maybe have some valid reason here, I just would like to understand it better.

chaitalisg:
A lot of the code here is similar to the `recovery` test package. Can you move the test to that package and pull out the common code into functions?

tools/crash-consistency-job.sh
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env bash
```
chaitalisg, line 1: Is this needed now that the test is moved under recovery_test package?

tests/robustness/filehandler/filehandler.go
```diff
@@ -0,0 +1,142 @@
+//go:build darwin || (linux && amd64)
```
chaitalisg, line 1: This should be moved under `recovery` as well.

tests/recovery/recovery_test/main_test.go
```diff
@@ -13,6 +13,8 @@ import (
 
 const (
 	dataSubPath = "recovery-data"
+	dirPath     = "kopia_dummy_repo"
```
chaitalisg, line 16: Can you use a more descriptive name? Something like consistencyTestRepo or similar? Same thing goes for the path `kopia_dummy_repo`.

tests/recovery/blobmanipulator/blobmanipulator.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/snapshot"
 	"github.com/kopia/kopia/tests/robustness"
+	"github.com/kopia/kopia/tests/robustness/filehandler"
```
chaitalisg, line 19: Filehandler is no longer under robustness package

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,167 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestConsistencyWhenKill9AfterModify(t *testing.T) {
```
chaitalisg, line 175: Add a comment describing the test function.

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,167 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+	err = bm.VerifySnapshot()
+	require.NoError(t, err)
+
+	// Create a temporary dir to restore a snapshot
+	restoreDir := t.TempDir()
+	if restoreDir == "" {
+		t.FailNow()
+	}
+
+	// try to restore a snapshot, this should error out
```
chaitalisg, line 253: Comment says this should error out but the next line checks for `NoError`. Old comment?

Makefile
```diff
@@ -307,6 +307,13 @@ recovery-tests: build-integration-test-binary $(gotestsum)
 	FIO_DOCKER_IMAGE=$(FIO_DOCKER_TAG) \
 	$(GO_TEST) -count=$(REPEAT_TEST) github.com/kopia/kopia/tests/recovery/recovery_test $(TEST_FLAGS)
 
+crash-consistency-test: export KOPIA_EXE ?= $(KOPIA_INTEGRATION_EXE)
```
chaitalisg, line 310: Remove this if the target is no longer used

tests/recovery/recovery_test/main_test.go
chaitalisg: Please address this. The variable name `dirPath` does not describe its purpose. How about crashConsistencyDirPath or whatever the purpose of this variable is.
Similarly, `kopia_dummy_repo` does not describe its purpose. How about something like - crash-consistency-test-repo

chaitalisg:
Please add a detailed PR description.

chaitalisg (approved):
LGTM!

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+		t.FailNow()
+	}
+
+	bm, err := blobmanipulator.NewBlobManipulator(baseDir, dataRepoPath)
+	if err != nil {
+		if errors.Is(err, kopiarunner.ErrExeVariableNotSet) {
+			t.Skip("Skipping crash consistency tests because KOPIA_EXE is not set")
+		} else {
+			t.Skip("Error creating SnapshotTester:", err)
+		}
```
Shrekster, line 191: You do not need the else clause because Skip() will exit the test.

```suggestion

		t.Skip("Error creating SnapshotTester:", err)
```

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+	}
+
+	bm.DataRepoPath = dataRepoPath
+
+	// create a snapshot for initialized data
+	snapID, err := bm.SetUpSystemWithOneSnapshot()
+	if err != nil {
+		t.FailNow()
+	}
```
Shrekster, line 200: How about using require package here too ? The advantage is that it'll actually also print the error which is good for debugging later.

```suggestion
	require.NoError(t, err)
```

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+
+	cmpDir := bm.PathToTakeSnapshot
+
+	copyDir := t.TempDir()
+	if copyDir == "" {
+		t.FailNow()
+	}
```
Shrekster, line 207:
```suggestion
	require.NotEmpty(copyDir, "TempDir() did not generate a valid dir")
```

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+	err = bm.FileHandler.CopyAllFiles(bm.PathToTakeSnapshot, copyDir)
+	require.NoError(t, err)
+
+	// modify original files
+	content := "\nthis is a test for TestConsistencyWhenKill9AfterModify\n"
+	err = bm.FileHandler.ModifyDataSetWithContent(copyDir, content)
+	require.NoError(t, err)
+	log.Println("Copy content and modify the files.")
+
+	// kill the kopia command before it exits
```
Shrekster, line 228: is this comment relevant here ?

You do not want to killl the `connect` command, but the snapshot command. Maybe move this comment to the snapshot command below ?

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+	log.Println("Verify snapshot corruption:")
+	// verify snapshot corruption
+	err = bm.VerifySnapshot()
+	require.NoError(t, err)
+
+	// Create a temporary dir to restore a snapshot
+	restoreDir := t.TempDir()
+	if restoreDir == "" {
+		t.FailNow()
+	}
```
Shrekster, line 253: same as before use the `require` package with a message explaining that TempDir() failed

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+	require.NoError(t, err)
+
+	err = bm.FileHandler.CopyAllFiles(bm.PathToTakeSnapshot, copyDir)
+	require.NoError(t, err)
+
+	// modify original files
+	content := "\nthis is a test for TestConsistencyWhenKill9AfterModify\n"
+	err = bm.FileHandler.ModifyDataSetWithContent(copyDir, content)
+	require.NoError(t, err)
+	log.Println("Copy content and modify the files.")
```
Shrekster, line 226: You should use `t.Log()` and `t.Logf()` where possible to avoid test noise.

SEE: https://pkg.go.dev/testing#B.Logf

These logs are generally not printed on the console, unless you specify the verbose `-v` flag to the test. Like `go test -v`.

The reason for using `t.Log` and `t.Logf` over `log.PrintX` is because developers usually run all the tests quite often. Every time we push anything to Kopia, we'd run all these tests locally. And if these tests would keep printing noise to the screen then it is hard to follow progress of all the tests. If anyone is interested in watching what a test is doing then they can specify the verbose flag or if the test fails then the test framework will dump the Log() statements anyway, so that they are there when you need them.

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+	env := []string{"KOPIA_PASSWORD=" + testenv.TestRepoPassword}
+	cmd.Env = append(os.Environ(), env...)
+
+	o, err := cmd.CombinedOutput()
+	require.NoError(t, err)
+	log.Println(string(o))
+
+	cmd = exec.Command(kopiaExe, "snap", "create", copyDir, "--json", "--parallel=1")
+
+	log.Println("Kill the kopia command before it exits:")
```
Shrekster, line 241: Same suggestion as above, you can replace all your `log.PrintX` statements with `t.Log`.

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+
+				break
+			}
+		}
+	}()
+
+	stdoutPipe, err := cmd.StdoutPipe()
+	if err != nil {
+		panic(err)
+	}
```
Shrekster, line 308: No need to panic() just fail the test. So, just use require.NoError().

tests/recovery/recovery_test/recovery_test.go
MinyiZou: I deleted the else clause. Thanks~

tests/recovery/recovery_test/recovery_test.go
MinyiZou: Thanks, I would like to use `require` interface instead.

tests/recovery/recovery_test/recovery_test.go
MinyiZou: Yes, you are right. I do the kill action before connect, so I should add connect with repository with the environment configuration, otherwise it will display "ERROR open repository: repository is not connected.kopia connect repo".

tests/recovery/recovery_test/recovery_test.go
MinyiZou: That's why I find it confusing when I see `t.Log` used in some places and `log.Println` used in others. I appreciate your response.

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+			output := scanner.Text()
+			log.Println(output)
+			errOut.Write(scanner.Bytes())
+			errOut.WriteByte('\n')
+
+			log.Println(output)
+			// Check if the output contains the "hashing" etc.
+			if strings.Contains(output, "hashing") && strings.Contains(output, "hashed") && strings.Contains(output, "uploaded") || strings.Contains(output, "Snapshotting") {
+				log.Println("Detaching and terminating target process")
+				cmd.Process.Kill()
```
Shrekster, line 292: This will work most of the times but will fail some of the times. The problem is that your strings.Contains() will not match every word every time this test runs.

`scanner := bufio.NewScanner(stderrPipe)` will allocate the default buffer size of 64k. What this means is that it MIGHT read upto those many characters before the first `scanner.Scan()` succeeds. In an unexpected scenario for your code, it might succeed once with half the word "uploa" first and the remaining "ded" in the next iteration, which will not make your test trigger the kill() action.

Which means that for your experiment, the kopia subcommand could actually print everything and then exit before the first iteration even triggering the kill action. Which is too late for you. Instead, I think what you are trying to do is to kill the subprocess as soon as you see lines being printed with the words you want.

So, you might be able to set a word split using `scanner.Split(bufio.ScanWords)` and check for words "uploaded" in that or you may also use `scanner.Split(bufio.ScanLines)` and then check for all the words in each line.

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+
+	wg.Add(1)
+
+	o := bytes.Buffer{}
+
+	go func() {
+		defer wg.Done()
+
+		// Create a scanner to read from stdoutPipe
+		scanner := bufio.NewScanner(stdoutPipe)
```
Shrekster, line 310: Do you really need to check BOTH stdout and stderr pipes ? I think you should be able to drop one of them.

tests/recovery/recovery_test/recovery_test.go
MinyiZou: Fix it with `require.NoError(t, err)`

tests/recovery/recovery_test/recovery_test.go
MinyiZou: fix it with `require.NotEmpty(copyDir, "TempDir() did not generate a valid dir")`

tests/recovery/recovery_test/recovery_test.go
MinyiZou: fix it with require.NotEmpty(restoreDir, "TempDir() did not generate a valid dir")

tests/recovery/recovery_test/recovery_test.go
MinyiZou: Fix it with `t.Log`

tests/recovery/recovery_test/recovery_test.go
MinyiZou: fix it with `require.NoError()`

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+			// Check if the output contains the "copying" text
+			if strings.Contains(output, "hashing") && strings.Contains(output, "hashed") && strings.Contains(output, "uploaded") {
+				cmd.Process.Kill()
+				break
+			}
+		}
+	}()
+
+	// Run the command
+	_ = cmd.Run()
```
Shrekster, line 328: cmd.Run() will wait for the process to exit and it will also close the StdErr pipe. So, you should use use `cmd,Start()` instead and then wait for your go routine `wg.Wait()` and finally, wait for the command to exit using `cmd.Wait()`.

tests/recovery/recovery_test/recovery_test.go
MinyiZou: I believe it's still necessary to have the stdoutpipe because there's a possibility that the snapshot will be created successfully. Although it's unlikely for the stdoutpipe to be executed in such cases, there are scenarios where the test environment is not cleaned properly in pipeline and the previous dataset is used for running the test. In such cases, it's important to know whether the snapshot was created successfully, even though the occurrence of this situation is unlikely.

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,172 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+			output := scanner.Text()
+			log.Println(output)
+			o.Write(scanner.Bytes())
+			o.WriteByte('\n')
+
+			log.Println("snapshot create successfully", output)
+			// Check if the output contains the "copying" text
+			if strings.Contains(output, "hashing") && strings.Contains(output, "hashed") && strings.Contains(output, "uploaded") {
+				cmd.Process.Kill()
+				break
```
Shrekster, line 322: You should drain the `stdoutPipe` before stopping the goroutine or your subcommand can get deadlocked. If a subprocess is blocked on writing to the std-err before of buffer limitations then you will not be able to kill it and hence your `cmd.Wait()` or `cmd.Run()` will get stuck forever.

tests/recovery/recovery_test/recovery_test.go
Shrekster: It should not matter whether if snapshot was created successfully. If Kopia prints to stderr then it will always print that line to stderr even if it succeeded.

tests/recovery/recovery_test/recovery_test.go
MinyiZou: Since the log messages related to uploading are only printed to the stderrPipe, I believe there is no need to include the stdoutPipe anymore. Therefore, I think there is no requirement to drain the stdoutPipe in this scenario.

tests/recovery/recovery_test/recovery_test.go
MinyiZou: I decide to delete stdoutPipe because the log about uploading only print at stderrPipe.

tests/recovery/recovery_test/recovery_test.go
MinyiZou: I used `cmd.Start()` rather than `cmd.Run()` in this case

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,143 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+func killOnCondition(t *testing.T, cmd *exec.Cmd) {
+	t.Helper()
+
+	stderrPipe, err := cmd.StderrPipe()
+	require.NoError(t, err)
+
+	// execute kill -9 while receive ` | 1 hashing, 0 hashed (65.5 KB), 0 cached (0 B), uploaded 0 B, estimating...` message
+	var wg sync.WaitGroup
+
+	var mu sync.Mutex
```
Shrekster, line 269: why do you need a mutex ?

tests/recovery/recovery_test/recovery_test.go
```diff
@@ -168,6 +172,143 @@ func TestSnapshotFixInvalidFiles(t *testing.T) {
...
+
+	// Add a WaitGroup counter for the first goroutine
+	wg.Add(1)
+
+	errOut := bytes.Buffer{}
```
Shrekster, line 274: is anyone using the errOut variable? If not then please drop it

tests/recovery/filehandler/filehandler.go
```diff
@@ -0,0 +1,155 @@
...
+		sourceFile.Close()
+
+		destinationFile.Close()
+	}
+
+	return nil
+}
+
+// CompareDirs examines and compares the quantities and contents of files in two different folders.
+func (handler *FileHandler) CompareDirs(source, destination string) error {
```
Shrekster, line 115: You can use the directory comparator fro the Kopia's internal `diff` module:

```go
	c, err := diff.NewComparer(&buf)
	require.NoError(t, err)

	t.Cleanup(func() {
		_ = c.Close()
	})

	err = c.Compare(ctx, dir1, dir2)
	require.NoError(t, err)
```

See this file: internal/diff/diff.go

tests/recovery/filehandler/filehandler.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// FileHandler implements a FileHandler struct.
+type FileHandler struct{}
+
+// ErrFileNumberInconsistency is utilized to raise an error when the number of items in the two folders differs.
+var ErrFileNumberInconsistency = errors.New("the number of items in the two folders is not the same")
+
+// ErrFilesInconsistency is utilized to raise an error when the name of item in the two folders differs.
+var ErrFilesInconsistency = errors.New("the name of the item in the source folder does not match the name in the duplicated folder")
```
Shrekster, line 22: you can drop this error now

tests/recovery/filehandler/filehandler.go
```diff
@@ -0,0 +1,111 @@
...
+
```
Shrekster, line 23: you can drop these errors now

```suggestion
```

tests/recovery/blobmanipulator/blobmanipulator.go
```diff
@@ -274,6 +276,54 @@ func (bm *BlobManipulator) SetUpSystemUnderTest() error {
...
+
+	// create snapshot of the data
+	log.Printf("Creating snapshot of directory %s", bm.PathToTakeSnapshot)
+
+	snapshotID, _, err := bm.TakeSnapshot(bm.PathToTakeSnapshot)
+	if err != nil {
+		return "", err
+	}
+
+	log.Printf("Creating snapshot of snapshot ID %s", snapshotID)
```
Shrekster, line 302:
```suggestion
	log.Printf("Created snapshot of snapshot ID %s", snapshotID)
```

tests/recovery/blobmanipulator/blobmanipulator.go
```diff
@@ -274,6 +276,54 @@ func (bm *BlobManipulator) SetUpSystemUnderTest() error {
...
+	numFiles := 50
+	ctx := context.Background()
+
+	err := bm.writeRandomFiles(ctx, fileSize, numFiles)
+	if err != nil {
+		return "", err
+	}
+
+	// create snapshot of the data
+	bm.PathToTakeSnapshot = bm.FileHandler.GetRootDir(bm.fileWriter.DataDirectory(ctx))
```
Shrekster, line 292: why not set the `bm.PathToTakeSnapshot` to the data dir ? We'd like to take the snapshot of the entire directory, right ?

```suggestion
	bm.PathToTakeSnapshot = bm.fileWriter.DataDirectory(ctx)
```

tests/recovery/blobmanipulator/blobmanipulator.go
```diff
@@ -274,6 +276,54 @@ func (bm *BlobManipulator) SetUpSystemUnderTest() error {
...
+// Tests can later create snapshots from the source directory.
+func (bm *BlobManipulator) GenerateRandomFiles(fileSize, numFiles int) error {
+	ctx := context.Background()
+
+	err := bm.writeRandomFiles(ctx, fileSize, numFiles)
+	if err != nil {
+		return err
+	}
+
+	bm.PathToTakeSnapshot = bm.FileHandler.GetRootDir(bm.fileWriter.DataDirectory(ctx))
```
Shrekster, line 317: same suggestions as above here:

```suggestion
	bm.PathToTakeSnapshot = bm.fileWriter.DataDirectory(ctx)
```

tests/recovery/blobmanipulator/blobmanipulator.go
Shrekster: Maybe I am missing something here. @MinyiZou I am assuming hat we'll ike to snapshot the entire directory and not just a single file.

MinyiZou: The structure of directories is as follows: /a/b/. In the "b" folder, there are 200 files. However, when using the bm.fileWriter.DataDirectory(ctx) method, I only receive a folder path (/a/). For this particular test scenario, it should be acceptable to capture either the entire directory as a snapshot or the 200 individual files. In case we choose to snapshot the 200 files, I will send a signal. `bm.FileHandler.GetRootDir` will return /a/b/ path if I input /a/.

Shrekster (approved):
LGTM! Nice work!

A couple things around the raciness attribute of the current design of the crash test, but this a good first iteration.

Issue #2822, opened 2023-03-14
.cache/kopia isn't fully tagged with CACHEDIR.TAG


Juerd:
Kopia will put a CACHEDIR.TAG file in a blob cache to facilitate skipping its own directory when creating a snapshot, but the file is not present in the base of the cache directory (e.g. `~/.cache/kopia`), which results in backups of `cli-logs`, `content-logs` and `<hash>/server-contents`. While I feel that none of these should be backed up by default, at least the `server-contents` cache should probably be treated the same way as the blob cache, because it is large and redundant with the contents of the backup itself.

jkowalski:
+1. That should be a quick and easy fix. PR welcome.

israel-lugo:
I just recently started using Kopia, and was surprised to see that `--ignore-cache-dirs` doesn't make Kopia ignore its own cache.

I'm much more worried about the <hash>/server-contents than the logs; on the first machine I looked at, server-contents is over 1 GiB. Logs themselves, I don't see why they shouldn't be backed up (although it's weird that they're stored in a cache directory). But server-contents really shouldn't be.

Could someone please take a second look at the PR or create a new one?

PR #3147, opened 2023-07-15
fix(cli): write CACHEDIR.TAG to log directories

NickIAm:
Write a CACHEDIR.TAG file to the ~/.cache/kopia directory to exclude cli-logs and address #2822. Get the path to write the cachetag file by going up 1 level from the content cache where it was previously written.


jkowalski:
can we write the marker to the logs directory directly instead of the parent?

NickIAm:
Yeah I can change it. Didn't realize that logs are in a different directory on Mac, I don't have access to one.

PR #3157, opened 2023-07-18
fix(ci): add test for theme selection

lupusA:
Dear all, 

this PR contains:

- Added test for changing themes
- Encapsulated the creation of a test repository in a function
- Added test for changing the byte representation
- Added test for changing the pagination

The PR is currently a draft. While the test for selecting different themes works, the tests for testing pagination and byte representation still needs to be finished. I was not able to extract the text within the cell, where the size of the snapshot is located. 

The idea was to get the size in both byte representations and compare them. 
Feedback is welcomed. 

Cheers,


lupusA:
The test for the pagination settings needs a proper ID for the dropdown menu in html-ui. Will make a pr for this. 

Cheers,

jkowalski:
this is awesome, thank you!

jkowalski (approved):

PR #3163, opened 2023-07-20
fix(cli): preserve error exit code when '--json' output is specified

julio-lopez:
Changes kopia's behavior to match the exit code that would have been returned when the `--json` flag was not specified.

`kopia snapshot create my/path --json` terminates with a 0 status code in cases where
`kopia snapshot create my/path` terminates with a non-zero exit code. One such case is when there are permissions errors reading files or directories to snapshot.

### DIFF
--- a/tests/end_to_end_test/snapshot_fail_test.go
+++ b/tests/end_to_end_test/snapshot_fail_test.go
@@ -76,6 +76,8 @@ func cond(c bool, a, b int) int {
 }
 
 func testSnapshotFailText(t *testing.T, isFailFast bool, snapshotCreateFlags []string, snapshotCreateEnv map[string]string) {
+	t.Helper()
+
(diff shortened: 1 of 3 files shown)


tests/end_to_end_test/snapshot_fail_test.go
```diff
@@ -36,22 +37,27 @@ func TestSnapshotNonexistent(t *testing.T) {
 
 func TestSnapshotFail_Default(t *testing.T) {
 	t.Parallel()
-	testSnapshotFail(t, false, nil, nil)
+	testSnapshotFailText(t, false, nil, nil)
+}
+
+func TestSnapshotFail_DefaultJSONOutput(t *testing.T) {
```
julio-lopez, line 43: This is the new test for the change.

Everything else is refactoring to parse the JSON output in tests.

denisvmedia (approved):
LGTM

jkowalski (approved):

PR #3164, opened 2023-07-20
refactor(general): Cleaner error checking in retention tests

ashmrtn:
Two minor cleanups for tests around retention:
* define an error and use `assert.ErrorIs` rather than comparing error messages when a blobtesting backend is used
* switch to `require.NoError` instead of an `if err != nil`/`t.Fatal` block

### DIFF
--- a/internal/blobtesting/object_locking_map.go
+++ b/internal/blobtesting/object_locking_map.go
@@ -14,6 +14,8 @@ import (
 	"github.com/kopia/kopia/repo/blob"
 )
 
+var ErrBlobLocked = errors.New("cannot alter object before retention period expires")
+
 type entry struct {
 	value          []byte
 	mtime          time.Time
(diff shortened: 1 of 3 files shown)


internal/blobtesting/object_locking_map.go
```diff
@@ -60,7 +62,7 @@ func (s *objectLockingMap) getLatestForMutationLocked(id blob.ID) (*entry, error
 	}
 
 	if !e.retentionTime.IsZero() && e.retentionTime.After(s.timeNow()) {
-		return nil, errors.New("cannot alter object before retention period expires")
+		return nil, errors.WithStack(ErrBlobLocked)
```
julio-lopez, line 65: In the long term, we are going to move away from `pkg/errors`. It's deprecated.

julio-lopez (approved):

internal/blobtesting/object_locking_map.go
ashmrtn: ah right, forgot about that thanks for the reminder. Well, at least until the switch this will give us stack trace info about where the error came from which is always sort of nice 😅

PR #3168, opened 2023-07-21
fix(ci): avoid hung CI jobs while notarizing the kopia app on MacOS 

julio-lopez:
This happens on PR from branches in the kopia repo.

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -36,6 +36,7 @@ else
 # notarize.
 unexport CSC_LINK
 unexport CSC_KEY_PASSWORD
+unexport KOPIA_UI_NOTARIZE
 
 endif
 endif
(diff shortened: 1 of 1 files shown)


app/Makefile
```diff
@@ -25,11 +25,20 @@ endif
...
 ifneq ($(FORCE_KOPIA_UI_SIGN),)
 electron_builder_flags+=--dir
-endif
 
+else
+
+# unset CSC_LINK completely, otherwise electron builder hangs attempting to
+# notarize.
+unexport CSC_LINK
+unexport CSC_KEY_PASSWORD
```
julio-lopez, line 38: @jkowalski Do these 2 need to be unset or is it enough to unset `KOPIA_UI_NOTARIZE` ?

julio-lopez:
@jkowalski CI passed, I'm going to merge this. If it causes issues we can always revert it.

PR #3170, opened 2023-07-22
fix(kopiaui): mac tray icon

### LINKED ISSUE #2356: Mac tray icon should have a transparent background
The mac tray icon should have a transparent section in the middle and border, instead of black, in order to match the theme of the built-in icons. See image:

Screen Shot 2022-09-01 at 10 12 52

rsammelson:
Fixes #2356


jkowalski:
Thank you, this looks awesome!

jkowalski (approved):

app/public/electron.js
```diff
@@ -337,7 +337,7 @@ app.on('ready', () => {
   tray = new Tray(
     path.join(
       resourcesPath(), 'icons',
-      selectByOS({ mac: 'kopia-tray.png', win: 'kopia-tray.ico', linux: 'kopia-tray.png' })));
+      selectByOS({ mac: 'kopiaTrayTemplate.png', win: 'kopia-tray.ico', linux: 'kopia-tray.png' })));
```
julio-lopez, line 340: @rsammelson mainly curious, was it necessary to rename this file? why was that?

rsammelson: The "Template" at the end of the filename is how it knows it should render as a template (separate light and dark versions) in electron.

I don't really think that is a great system but I guess it works.

julio-lopez: Ah! got it. Thanks for the explanation.

PR #3190, opened 2023-08-03
fix(ui): Task summary should show fatal errors

lupusA:
Hi, 

currently, the task summary shows the total number of ignored errors. Since the user is interested in non ignored errors, the summary should show fatal errors (see https://kopia.discourse.group/t/kopiaui-bug-on-log/2171/2)

Fatal errors are already shown when browsing the snapshot (indicated by a warning sign). However, the task summary showed ignored errors causing inconsistencies between both views. 

image

image

This PR contains the following fix:

- Store and send the total number of fatal errors

It would also be possible to extend the task summary to show ignored and fatal errors separately.

Feedback is welcomed. 

Cheers,

### DIFF
--- a/snapshot/snapshotfs/upload_progress.go
+++ b/snapshot/snapshotfs/upload_progress.go
@@ -257,15 +257,17 @@ func (p *CountingUploadProgress) Snapshot() UploadCounters {
 	defer p.mu.Unlock()
 
 	return UploadCounters{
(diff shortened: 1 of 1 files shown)


julio-lopez (approved):

julio-lopez:
@jkowalski FYI

PR #3192, opened 2023-08-03
fix(repository): Storage config blob caching

ashmrtn:
Fix minor issue with caching the storage blob config blob settings in-memory which could cause other blob PUTs after a reconfiguration to have incorrect settings

Most of PR is tests similar to the retention extension tests to ensure changing repo settings works as expected

### DIFF
--- a/repo/format/format_manager_test.go
+++ b/repo/format/format_manager_test.go
@@ -174,8 +174,7 @@ func TestInitialize(t *testing.T) {
 func TestInitializeWithRetention(t *testing.T) {
 	ctx := testlogging.Context(t)
 
-	startTime := time.Date(2020, 1, 1, 12, 0, 0, 0, time.UTC)
-	ta := faketime.NewTimeAdvance(startTime, 0)
+	ta := faketime.NewClockTimeWithOffset(0)
 	nowFunc := ta.NowFunc()
 
 	st := blobtesting.NewVersionedMapStorage(nowFunc).(cache.Storage)
@@ -212,13 +211,32 @@ func TestInitializeWithRetention(t *testing.T) {
 
 	_, err = st.TouchBlob(ctx, format.KopiaBlobCfgBlobID, time.Minute)
 	assert.ErrorIs(t, err, blobtesting.ErrBlobLocked, "Altering locked blob storage config should fail")
+
+	// Advance the clock to just before the retention period expires and check
+	// if we can modifiy the blob. Blobs should still be protected so TouchBlob
+	// should still fail.
+	ta.Advance((time.Hour * 48) - time.Minute)
+	_, err = st.TouchBlob(ctx, format.KopiaRepositoryBlobID, time.Minute)
+	assert.ErrorIs(t, err, blobtesting.ErrBlobLocked, "Altering locked repo blob should fail")
+
+	_, err = st.TouchBlob(ctx, format.KopiaBlobCfgBlobID, time.Minute)
+	assert.ErrorIs(t, err, blobtesting.ErrBlobLocked, "Altering locked blob storage config should fail")
+
+	// Advance the clock so the retention period has passed. Calling TouchBlob
+	// should now succeed.
+	ta.Advance(time.Minute)
+
+	_, err = st.TouchBlob(ctx, format.KopiaRepositoryBlobID, time.Minute)
+	assert.NoError(t, err, "Altering expired repo blob failed")
+
+	_, err = st.TouchBlob(ctx, format.KopiaBlobCfgBlobID, time.Minute)
+	assert.NoError(t, err, "Altering expired blob storage config failed")
 }
 
 func TestUpdateRetention(t *testing.T) {
 	ctx := testlogging.Context(t)
 
-	startTime := time.Date(2020, 1, 1, 12, 0, 0, 0, time.UTC)
-	ta := faketime.NewTimeAdvance(startTime, 0)
+	ta := faketime.NewClockTimeWithOffset(0)
 	nowFunc := ta.NowFunc()
 
 	st := blobtesting.NewVersionedMapStorage(nowFunc).(cache.Storage)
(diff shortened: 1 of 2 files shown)


repo/format/format_manager_test.go
```diff
@@ -53,15 +55,15 @@ func TestFormatManager(t *testing.T) {
 	startTime := time.Date(2020, 1, 1, 12, 0, 0, 0, time.UTC)
 	ta := faketime.NewTimeAdvance(startTime, 0)
 	nowFunc := ta.NowFunc()
-	cache := format.NewMemoryBlobCache(nowFunc)
+	blobCache := format.NewMemoryBlobCache(nowFunc)
```
julio-lopez, line 58: @ashmrtn Can you split this PR in 2 please?
Have the blobCache renaming refactoring in one, and the functional change with the tests in another one? Thanks.

ashmrtn: unfortunately I can't :/ the renaming is only to get around linting errors and should be in the final commit in the series. Without it the linter complains there's a variable named the same as an import (the cache package since that's what I use in the newer tests

repo/format/format_manager_test.go
```diff
@@ -169,13 +171,148 @@ func TestInitialize(t *testing.T) {
...
+	assert.Equal(t, mode, blobCfg.RetentionMode)
+	assert.Equal(t, period, blobCfg.RetentionPeriod)
+
+	// Attempting to touch the blobs the format manager writes should return
+	// errors as they should have retention enabled. Mod time adjustment (duration
+	// param) doesn't matter in this context.
+	_, err = st.TouchBlob(ctx, format.KopiaRepositoryBlobID, time.Minute)
+	assert.ErrorIs(t, err, blobtesting.ErrBlobLocked, "Altering locked repo blob should fail")
+
+	_, err = st.TouchBlob(ctx, format.KopiaBlobCfgBlobID, time.Minute)
```
Shrekster, line 213: Also, It'd be desired to verify that the initial format blob PUT was set with the desired `RetainUntilDate` and `Mode`.

Shrekster (approved):
Just one comment, LGTM otherwise.

repo/format/format_manager_test.go
ashmrtn: I can indirectly check the retain until date by advancing the clock, similar to the [tests](https://github.com/kopia/kopia/blob/22cba7003d789b71a9512a66e52781eeb4346033/repo/maintenance/blob_retain_test.go#L59) for retention extension. There doesn't seem to be a way to directly check those properties though. I'm going to save the larger change of checking mode and retain until for another PR but will advance the clock and try `TouchBlob` again in this PR if that's alright with you

Shrekster: Can we access the object-locking map underneath and add a method to it to expose full metadata ?

repo/format/format_manager_test.go
julio-lopez: Ashlie, I took a quick stab and split it two:
1. #3211
2. julio-lopez/kopia/pull/332

I'll merge #3211 soon. After that you can pull the commits from the second PR and either push them to this PR (preferred) or open a new PR. Your choice.

Thanks.

julio-lopez: Well, I actually missed a few renames in the first PR :(

ashmrtn: ah, sorry, I misunderstood the ordering request in your first comment (I guess that's what I get for reading comments before my morning caffeine). Thanks for the assist on that one. I'll pull the new commits in to this one

julio-lopez: No worries. I had not realized that the renaming was prompted by the linter and then necessary to get the PR passing. 😃

repo/format/format_manager_test.go
ashmrtn: yea, something like that should be doable, though it will require a slightly bigger change to store retention mode in the blobtesting implementation. Can I do that in another PR so this one doesn't get too big? If we're going to do that we should probably make similar checks in other retention-related tests

julio-lopez: These tests are (more than) sufficient to verify the fix introduced in the PR

julio-lopez (approved):
+1: LG 👍
🥇Thanks for doing this.

repo/format/format_manager_test.go
```diff
@@ -169,6 +171,179 @@ func TestInitialize(t *testing.T) {
...
+	assert.ErrorIs(t, err, blobtesting.ErrBlobLocked, "Altering locked blob storage config should fail")
+
+	// Advance the clock to just before the retention period expires and check
+	// if we can modifiy the blob. Blobs should still be protected so TouchBlob
+	// should still fail.
+	ta.Advance((time.Hour * 48) - time.Minute)
+	_, err = st.TouchBlob(ctx, format.KopiaRepositoryBlobID, time.Minute)
+	assert.ErrorIs(t, err, blobtesting.ErrBlobLocked, "Altering locked repo blob should fail")
+
+	_, err = st.TouchBlob(ctx, format.KopiaBlobCfgBlobID, time.Minute)
```
julio-lopez, line 222: @Shrekster separate from this PR.
The semantics of `TouchBlob` do not match my mental model and understanding of object locking. Having said that, it is sufficient for testing this particular set of changes.

repo/format/format_manager_test.go
Shrekster: > it will require a slightly bigger change to store retention mode in the blobtesting implementation

Makes sense. Ok to take this in without the metadata evaluation. The blobtesting can probably be improved going forward:

>  If we're going to do that we should probably make similar checks in other retention-related tests

Yes. Separate PR.

https://github.com/kopia/kopia/issues/3218

repo/format/format_manager_test.go
Shrekster: Agreed. We might be able to deprecate `TouchBlob` API with some caching changes, I believe that is the only storage piece that ever needs it.

My only concern with extended reliance on `TouchBlob` here was that it builds up the technical debt, but as @ashmrtn explained it is too much work for this PR otherwise.

PR #3193, opened 2023-08-03
fix(general): change cache full logging frequency & message

julio-lopez:
Followups to #3085

Change cache full logging frequency to once every 10 minutes. Addresses concerns about too many messages in the logs in #3085 

Avoid logging cache key. Instead log the cache description, which provides information about the types of contents being cached.


julio-lopez:
@arouene FYI

Shrekster (approved):

PR #3212, opened 2023-08-11
fix(general): cache storage config blob

julio-lopez:
Fix minor issue with caching the storage blob config blob settings in-memory which could cause other blob PUTs after a reconfiguration to have incorrect settings

Most of PR is tests similar to the retention extension tests to ensure changing repo settings works as expected


Shrekster:
@ashmrtn recently took this in another PR I believe. Might be ok to close this one now.

PR #3216, opened 2023-08-13
feat(server): add server support for unix domain sockets

PhracturedBlue:
Enable using unix domain sockets with kopia server.

unix sockets can be useful to allow connecting services without going throgh the network interface.  They tend to be faster than using tcp sockets, and can be restricted via file-access permissions.

I'm using unix sockets in a container solution using rootless containers with networking disabled via podman + systemd socket-activated-sockets.  This means that the reverse-proxy (nginx) doesn't need any network interface, as it speaks to the outside world through a socket-activated-socket, and speaks to other services via unix sockets.

The patch also implements the ability to use unixsockets for client connections, but  I'm not sure what the use of that is other than enabling testing

### DIFF
--- a/cli/command_server_control_test.go
+++ b/cli/command_server_control_test.go
@@ -1,6 +1,7 @@
 package cli_test
 
 import (
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


jkowalski:
This is a great idea, the PR looks very good too, please fix all the failures reported by automatic checkers (and likely also disable the test on Windows) and we can merge it.

PhracturedBlue:
Hopefully the tests all pass  now

PhracturedBlue:
This test runs fine on debian when I run 'make test'   It appears that there is no address given for --address when I look at the test results, but I have no idea how to reproduce this

PhracturedBlue:
Hopefully 3rd time is the charm.  The last issue appears to be that unix sockets must be < 108 (104 on Mac) bytes long.  and the tmpdir was too long.  I wasn't previously running with ENABLE_LONG_FILENAMES=1 so I missed it.

jkowalski (approved):

PR #3220, opened 2023-08-14
test(repository): More precise check for retention in tests

### LINKED ISSUE #3218: Preserve the retention settings in the blob testing framework
... to allow immutability tests to make assertions about on blobs getting the correct mode and retention settings.

https://github.com/kopia/kopia/pull/3192#discussion_r1291501273

ashmrtn:
Store and return information about retention in the blobtesting storage used for retention tests. This allows more precise checks on when a blob expires and the retention mode applied to the blob

closes #3218

### DIFF
--- a/internal/blobtesting/object_locking_map.go
+++ b/internal/blobtesting/object_locking_map.go
@@ -213,8 +213,11 @@ func (s *objectLockingMap) ExtendBlobRetention(ctx context.Context, id blob.ID, 
 		return blob.ErrBlobNotFound
 	}
 
-	// Use the clock instead of mtime since that aligns with how the S3 blob
-	// storage acts.
+	// Update the retention time from now to the given retention period in the
+	// future. Note that we do not bump the existing time on the element `e.mtime`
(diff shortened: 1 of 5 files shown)


internal/blobtesting/object_locking_map.go
```diff
@@ -199,8 +213,10 @@ func (s *objectLockingMap) ExtendBlobRetention(ctx context.Context, id blob.ID,
 		return blob.ErrBlobNotFound
 	}
 
+	// Use the clock instead of mtime since that aligns with how the S3 blob
+	// storage acts.
```
Shrekster, line 217: nit; Comment can be confusing for readers who are not looking at this change/PR and are only reading the comment in this file:

```suggestion
	// Update the retention time from now to the given retention period in the future.
	// Note that we do not bump the existing time on the element `e.mtime` by the
	// given delta because we'd like to align with the S3 storage's current time.
```

Shrekster (approved):
Thank you!

PR #3225, opened 2023-08-15
fix(repository): skip epoch compaction and advancement on read-only storage

### LINKED ISSUE #3224: loading epoch manager index with read-only storage may get stuck
Similar to #3219, there may be cases where attempting to load the index with read-only storage has error (which can technically safely be ignored) or get stuck (hasn't been verified, but would be an edge case)

The heart of the issue here is again that the epoch manager doesn't know the storage is read-only but does some opportunistic operations like [advancing the write epoch](https://github.com/kopia/kopia/blob/bb27e3147eaa155d7f46f181248d3a1d7e19b5b5/internal/epoch/epoch_manager.go#L697) and [compacting/cleaning up epochs in the background](https://github.com/kopia/kopia/blob/bb27e3147eaa155d7f46f181248d3a1d7e19b5b5/internal/epoch/epoch_manager.go#L711)

Background compaction will result in error messages in the log but no failures. It's unclear if being unable to create a new epoch marker will result in getting [stuck](https://github.com/kopia/kopia/blob/bb27e3147eaa155d7f46f181248d3a1d7e19b5b5/internal/epoch/epoch_manager.go#L459) or not. Triggering that specific case would require a previous instance of the epoch manager to have written enough index blobs that a new epoch would need to be created but the previous instance crashed or was unable to create the new marker

One solution to this would be to thread a read-only parameter through the stack to the epoch manager. The content manager (which contains the epoch manager) already has an options struct passed through the stack and easily accessible. The epoch manager could then check the read-only variable prior to attempting opportunistic actions like the above

ashmrtn:
Thread a `readOnly` variable down to the epoch manager and use it to determine if background compaction and epoch advancement should take place. The value of the variable is taken from the `ReadOnly` client option that can be passed to kopia during `repository.Open` calls

Included unit test highlights a (highly unlikely) case where having the epoch manager use read-only storage and require advancing to a new epoch causes it to get stuck during `Refresh`. This could occur if the previous epoch manager wrote enough new index blobs in the current epoch to require an epoch advancement but never wrote the marker to advance the epoch

closes #3224

### DIFF
--- a/internal/epoch/epoch_manager.go
+++ b/internal/epoch/epoch_manager.go
@@ -698,7 +698,7 @@ func (e *Manager) refreshAttemptLocked(ctx context.Context) error {
 		len(ues[cs.WriteEpoch+1]),
 		cs.ValidUntil.Format(time.RFC3339Nano))
 
-	if !e.readOnly && shouldAdvance(cs.UncompactedEpochSets[cs.WriteEpoch], p.MinEpochDuration, p.EpochAdvanceOnCountThreshold, p.EpochAdvanceOnTotalSizeBytesThreshold) {
+	if !e.st.IsReadOnly() && shouldAdvance(cs.UncompactedEpochSets[cs.WriteEpoch], p.MinEpochDuration, p.EpochAdvanceOnCountThreshold, p.EpochAdvanceOnTotalSizeBytesThreshold) {
 		if err := e.advanceEpoch(ctx, cs); err != nil {
 			return errors.Wrap(err, "error advancing epoch")
 		}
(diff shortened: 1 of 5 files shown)


internal/epoch/epoch_manager.go
```diff
@@ -186,6 +186,10 @@ type Manager struct {
 	log      logging.Logger
 	timeFunc func() time.Time
 
+	// whether the underlying storage should be considered read only. Influences
+	// whether index compaction occurs automatically.
+	readOnly bool
```
jkowalski, line 191: shouldn't this be exposed on the `blob.Storage` itself?

julio-lopez (approved):

internal/epoch/epoch_manager.go
julio-lopez: Ignore the comment above, I had not seen #3230

ashmrtn:
this has been rebased on top of the storage provider changes and is ready for another review

julio-lopez (approved):

julio-lopez:
Thanks @ashmrtn

PR #3226, opened 2023-08-16
fix(repository): Disable manifest manager compaction when in read-only mode

### LINKED ISSUE #3219: Listing snapshots in kopia/S3 with a Point-in-Time may fail due to attempted manifest compaction
Kopia with S3 allows opening the repo at a specific point in time assuming the bucket has versioning enabled. Opening the repo in this mode puts the blob storage layer in read-only mode because kopia doesn't support writes in a "view" of the repo

Unfortunately, higher layers of kopia, including the manifest manager are unaware of the fact that blob storage is read-only and may attempt write operations like manifest compaction. This leads to failures because kopia is unable to write the compacted set of manifest blobs and there's no way for the manifest manger to check if kopia is read-only

A similar issue may exist for index compaction as if I remember right that is also done on an as-needed basis rather than as part of a maintenance command

ashmrtn:
Add a new field to the manifest manager that denotes whether it should act in read-only mode or not. The main difference between read-only and normal modes for the manager is that manifest compaction is disabled in read-only mode

This allows kopia to successfully read manifest information in the repo when either the repo was opened in read-only mode or the repo uses a versioned S3 bucket and was opened at a specific point in time with the `--point-in-time` flag

The value of this new field is taken from the local client options `ReadOnly` parameter. Client options can be provided during `repository.Open` calls to kopia giving the user control over this if needed

closes #3219

### DIFF
--- a/repo/manifest/committed_manifest_manager.go
+++ b/repo/manifest/committed_manifest_manager.go
@@ -357,7 +357,7 @@ func loadManifestContent(ctx context.Context, b contentManager, contentID conten
 	return man, errors.Wrapf(err, "unable to parse manifest %q", contentID)
 }
 
-func newCommittedManager(b contentManager, readOnly bool) *committedManifestManager {
+func newCommittedManager(b contentManager) *committedManifestManager {
 	debugID := ""
 	if os.Getenv("KOPIA_DEBUG_MANIFEST_MANAGER") != "" {
 		debugID = fmt.Sprintf("%x", rand.Int63()) //nolint:gosec
@@ -368,6 +368,5 @@ func newCommittedManager(b contentManager, readOnly bool) *committedManifestMana
 		debugID:             debugID,
 		committedEntries:    map[ID]*manifestEntry{},
 		committedContentIDs: map[content.ID]bool{},
-		readOnly:            readOnly,
 	}
 }
--- a/repo/manifest/committed_manifest_manager.go
+++ b/repo/manifest/committed_manifest_manager.go
@@ -33,6 +33,13 @@ type committedManifestManager struct {
 	committedEntries map[ID]*manifestEntry
 	// +checklocks:cmmu
 	committedContentIDs map[content.ID]bool
+
+	// readOnly denotes whether this instance of the manifest manager only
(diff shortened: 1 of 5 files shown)


repo/manifest/committed_manifest_manager.go
```diff
@@ -216,7 +223,9 @@ func (m *committedManifestManager) compact(ctx context.Context) error {
 func (m *committedManifestManager) maybeCompactLocked(ctx context.Context) error {
 	m.verifyLocked()
 
-	if len(m.committedContentIDs) < autoCompactionContentCount {
+	// Don't attempt to compact manifests if the repo was opened in read only mode
+	// since we'll just end up failing.
+	if m.readOnly || len(m.committedContentIDs) < autoCompactionContentCount {
```
jkowalski, line 228: I think we should determine readonly-ness from the underlying content manager, which should in turn read it from the underlying blob.Storage.

ashmrtn: seems reasonable. I've opened #3230 to expose that information. If that looks fine I can switch this and #3225 to use that instead of passing a value from the local client config down the stack

ashmrtn:
this has been rebased on top of the storage provider changes and is ready for another review

julio-lopez (approved):
LGTM  ✅

repo/manifest/manifest_manager_test.go
```diff
@@ -381,3 +395,34 @@ func TestManifestAutoCompaction(t *testing.T) {
 		require.NoError(t, mgr.b.Flush(ctx))
 	}
 }
+
+func TestManifestAutoCompactionWithReadOnly(t *testing.T) {
+	ctx := testlogging.Context(t)
+	data := blobtesting.DataMap{}
+
+	bm := newContentManagerForTesting(ctx, t, data, false)
```
julio-lopez, line 403: nit: is there a way of making the `readOnly` parameter more explicit?

Either using a named `const`, or maybe changing the function signature to take an options struct or similar. Thoughts?

julio-lopez (approved):
+1: LG 👍

Minor nit

PR #3230, opened 2023-08-18
feat(providers): Create IsReadOnly API for blob storage

ashmrtn:
Create an API that allows higher-layers to determine if the storage is in read-only mode or not. This can then be leveraged by higher-layers to determine if they should execute operations like manifest compaction

related to:
* #3225
* #3226

### DIFF
--- a/repo/content/committed_read_manager.go
+++ b/repo/content/committed_read_manager.go
@@ -112,12 +112,6 @@ type SharedManager struct {
 	metricsStruct
 }
 
-// IsReadOnly returns whether this instance of the SharedManger supports
-// mutations to the index.
-func (sm *SharedManager) IsReadOnly() bool {
-	return sm.st.IsReadOnly()
-}
-
 // LoadIndexBlob return index information loaded from the specified blob.
 func (sm *SharedManager) LoadIndexBlob(ctx context.Context, ibid blob.ID, d *gather.WriteBuffer) ([]Info, error) {
 	err := sm.st.GetBlob(ctx, ibid, 0, -1, d)


jkowalski:
we keep adding more and more of these methods, instead of repeating all this implementation can we rename`UnsupportedBlobRetention` to something like`DefaultProviderImplementation` which will provide sensible default implementations for less-commonly implemented methods:

* `IsReadOnly` - returns false
* `ExtendBlobRetention` - returns `ErrUnsupportedObjectLock`
* `Close` - noop
* `FlushCaches` - noop

perhaps also `GetCapacity` - returns `ErrNotAVolume`

All providers will embed `DefaultProviderImplementation` and they are free to override just the methods they care about.

jkowalski:
also can this PR be only about `blob.Storage` changes and not include any `content`-level changes?

julio-lopez:
> All providers will embed `DefaultProviderImplementation` and they are free to override just the methods they care about.

Separate PR though.

repo/content/committed_read_manager.go
```diff
@@ -112,6 +112,12 @@ type SharedManager struct {
 	metricsStruct
 }
 
+// IsReadOnly returns whether this instance of the SharedManger supports
+// mutations to the index.
+func (sm *SharedManager) IsReadOnly() bool {
+	return sm.st.IsReadOnly()
+}
+
```
julio-lopez, line 120: Remove content-level change to keep changes contained to the blob layer

```suggestion
```

julio-lopez (approved):
LGTM modulo suggested change. Feel free to merge after applying the suggestion.

jkowalski:
> > All providers will embed `DefaultProviderImplementation` and they are free to override just the methods they care about.
> 
> Separate PR though.

Can we simply rename `UnsupportedBlobRetention` => `DefaultProviderImplementation` and add only `IsReadOnly` there, the rest of the refactoring can be in separate PR (although I'm fine do to it in one PR, since we'd be only removing stub methods)

PR #3231, opened 2023-08-19
feat(providers): add IsReadOnly

julio-lopez:
To help with #3225

### DIFF
--- a/repo/blob/readonly/readonly_storage.go
+++ b/repo/blob/readonly/readonly_storage.go
@@ -66,7 +66,22 @@ func (s readonlyStorage) FlushCaches(ctx context.Context) error {
 	return s.base.FlushCaches(ctx)
 }
 
+func (s readonlyStorage) IsReadOnly() bool {
+	return true
+}
+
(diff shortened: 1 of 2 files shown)


repo/blob/readonly/readonly_storage.go
```diff
@@ -66,7 +66,22 @@ func (s readonlyStorage) FlushCaches(ctx context.Context) error {
...
+	return true
+}
+
 // NewWrapper returns a readonly Storage wrapper that prevents any mutations to the underlying storage.
 func NewWrapper(wrapped blob.Storage) blob.Storage {
 	return &readonlyStorage{base: wrapped}
 }
+
+// IsReadOnly returns whether the storage is read only.
+func IsReadOnly(st blob.Storage) bool {
```
jkowalski, line 79: can we move this to `kopia/kopia/repo/blob` package? otherwise LGTM

jkowalski (approved):

julio-lopez:
Let's uses @ashmrtn approach

PR #3235, opened 2023-08-20
feat(providers): improve performance of SFTP uploads

jkowalski:
This improves the performance of gather writes (p and q blobs) in particular on high-latency links by reducing round-trips.

Tested by uploading Kopia working directory
(3 GB, files:32157 dirs:6051) to SFTP server on LAN (TrueNAS) over WiFi:

Before: 2m4s (~24 MB/s)
After: 1m13s (~41 MB/s)

### DIFF
--- a/repo/blob/sftp/sftp_storage.go
+++ b/repo/blob/sftp/sftp_storage.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/kopia/kopia/internal/connection"
 	"github.com/kopia/kopia/internal/dirutil"
+	"github.com/kopia/kopia/internal/gather"
 	"github.com/kopia/kopia/internal/iocopy"
 	"github.com/kopia/kopia/internal/ospath"
 	"github.com/kopia/kopia/repo/blob"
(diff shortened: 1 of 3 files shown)


repo/blob/sftp/sftp_storage.go
```diff
@@ -194,6 +195,16 @@ func (s *sftpImpl) PutBlobInPath(ctx context.Context, dirPath, fullPath string,
...
 	}
 
+	// SFTP client Write() does not do any buffering leading to sub-optimal
+	// performance of gather writes, so we copy the data to a contiguous
+	// temporary buffer first.
+	contig := gather.NewWriteBufferMaxContiguous()
+	defer contig.Close()
+
+	if _, err := data.WriteTo(contig); err != nil {
+		return errors.Wrap(err, "can't write to comtiguous buffer")
```
julio-lopez, line 205: s/comtiguous/contiguous/

julio-lopez:
@jkowalski I do not understand (yet) where the overhead was coming from or what was causing the slowdown. Could you elaborate?

PR #3238, opened 2023-08-20
feat(repository): added hard size limit to the on-disk cache

jkowalski:
Before this change the cache size limit was only "soft", since "minSweepAge" would take priority over size. This is useful for metadata caching where we want to keep all recently used metadata to speed up incremental snapshots, but sometimes we want to enforce hard limits on the cache size.

This PR adds support for optional hard limits and refactors cache code for better readability.

### DIFF
--- a/cli/command_cache_set.go
+++ b/cli/command_cache_set.go
@@ -12,13 +12,15 @@ import (
 )
 
 type commandCacheSetParams struct {
-	directory              string
-	contentCacheSizeMB     int64
-	maxMetadataCacheSizeMB int64
-	maxListCacheDuration   time.Duration
-	contentMinSweepAge     time.Duration
-	metadataMinSweepAge    time.Duration
-	indexMinSweepAge       time.Duration
+	directory                   string
+	contentCacheSizeMB          int64
+	maxMetadataCacheSizeMB      int64
+	contentCacheSizeLimitMB     int64
+	maxMetadataCacheSizeLimitMB int64
+	maxListCacheDuration        time.Duration
+	contentMinSweepAge          time.Duration
+	metadataMinSweepAge         time.Duration
+	indexMinSweepAge            time.Duration
 
 	svc appServices
 }
@@ -32,10 +34,15 @@ func (c *commandCacheSetParams) setup(svc appServices, parent commandParent) {
 	c.maxListCacheDuration = -1
 
 	cmd.Flag("cache-directory", "Directory where to store cache files").StringVar(&c.directory)
-	cmd.Flag("content-cache-size-mb", "Size of local content cache").PlaceHolder("MB").Default("-1").Int64Var(&c.contentCacheSizeMB)
+
+	cmd.Flag("content-cache-size-mb", "Desired size of local content cache").PlaceHolder("MB").Default("-1").Int64Var(&c.contentCacheSizeMB)
+	cmd.Flag("content-cache-size-limit-mb", "Maximum size of local content cache").PlaceHolder("MB").Default("-1").Int64Var(&c.contentCacheSizeLimitMB)
 	cmd.Flag("content-min-sweep-age", "Minimal age of content cache item to be subject to sweeping").DurationVar(&c.contentMinSweepAge)
-	cmd.Flag("metadata-cache-size-mb", "Size of local metadata cache").PlaceHolder("MB").Default("-1").Int64Var(&c.maxMetadataCacheSizeMB)
+
+	cmd.Flag("metadata-cache-size-mb", "Desired size of local metadata cache").PlaceHolder("MB").Default("-1").Int64Var(&c.maxMetadataCacheSizeMB)
+	cmd.Flag("metadata-cache-size-limit-mb", "Size of local metadata cache").PlaceHolder("MB").Default("-1").Int64Var(&c.maxMetadataCacheSizeLimitMB)
 	cmd.Flag("metadata-min-sweep-age", "Minimal age of metadata cache item to be subject to sweeping").DurationVar(&c.metadataMinSweepAge)
+
 	cmd.Flag("index-min-sweep-age", "Minimal age of index cache item to be subject to sweeping").DurationVar(&c.indexMinSweepAge)
 	cmd.Flag("max-list-cache-duration", "Duration of index cache").DurationVar(&c.maxListCacheDuration)
 	cmd.Action(svc.repositoryWriterAction(c.run))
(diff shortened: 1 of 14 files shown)


cli/command_cache_set.go
```diff
@@ -4,21 +4,42 @@ import (
...
 	"github.com/kopia/kopia/repo"
 	"github.com/kopia/kopia/repo/content"
 )
 
+type cacheSizeFlags struct {
+	contentCacheSizeMB          int64
+	maxMetadataCacheSizeMB      int64
+	contentCacheSizeLimitMB     int64
+	maxMetadataCacheSizeLimitMB int64
+	maxListCacheDuration        time.Duration
```
Shrekster, line 20: nit; Might be better to rename `maxMetadataCacheSizeLimitMB` -> `metadataCacheSizeLimitMB` and `maxMetadataCacheSizeMB ` -> `metadataCacheSizeMB ` to be consistent with the other names.

cli/command_repository_connect.go
```diff
@@ -66,9 +66,12 @@ func (c *connectOptions) setup(svc appServices, cmd *kingpin.CmdClause) {
...
 	// we must use *Var() methods, otherwise one of the commands would always get default flag values.
 	cmd.Flag("cache-directory", "Cache directory").PlaceHolder("PATH").Envar(svc.EnvName("KOPIA_CACHE_DIRECTORY")).StringVar(&c.connectCacheDirectory)
-	cmd.Flag("content-cache-size-mb", "Size of local content cache").PlaceHolder("MB").Default("5000").Int64Var(&c.connectMaxCacheSizeMB)
-	cmd.Flag("metadata-cache-size-mb", "Size of local metadata cache").PlaceHolder("MB").Default("5000").Int64Var(&c.connectMaxMetadataCacheSizeMB)
-	cmd.Flag("max-list-cache-duration", "Duration of index cache").Default("30s").Hidden().DurationVar(&c.connectMaxListCacheDuration)
+
+	c.maxListCacheDuration = 30 * time.Second //nolint:gomnd
+	c.contentCacheSizeMB = 5000
+	c.maxMetadataCacheSizeMB = 5000
+	c.cacheSizeFlags.setup(cmd)
```
Shrekster, line 73: Instead of setting the defaults inline in all places, I think if you can pass the defaults to the setup() function as parameters which will then set it up on the CLI. That way, even the CLI help will see the defaults for connect/create/set commands.

internal/blobtesting/map.go
```diff
@@ -25,7 +25,22 @@ type mapStorage struct {
 	keyTime map[blob.ID]time.Time
 	// +checklocks:mutex
 	timeNow func() time.Time
-	mutex   sync.RWMutex
+	// +checklocks:mutex
+	totalBytes int64
+	// +checklocks:mutex
+	limit int64
```
Shrekster, line 31: `limit` doesn't change after the construction of the map, so it doesn't need to be under the lock

internal/blobtesting/map.go
```diff
@@ -100,7 +115,13 @@ func (s *mapStorage) PutBlob(ctx context.Context, id blob.ID, data blob.Bytes, o
 
 	data.WriteTo(&b)
 
+	if s.limit >= 0 && s.totalBytes+int64(b.Len()) > s.limit {
```
Shrekster, line 118:
```suggestion
	if s.limit >= 0 && s.totalBytes+int64(b.Len()) >= s.limit {
```

internal/blobtesting/map.go
```diff
@@ -100,7 +115,13 @@ func (s *mapStorage) PutBlob(ctx context.Context, id blob.ID, data blob.Bytes, o
...
+		return errors.Errorf("exceeded limit, unable to add %v bytes, currently using %v/%v", b.Len(), s.totalBytes, s.limit)
```
Shrekster, line 119: We should not update the `s.keyTime` if the above check fails. This will mess up the key timestamps.

internal/cache/mutex_map.go
```diff
@@ -0,0 +1,89 @@
...
+	ent.refCount++
+
+	return ent.mut
+}
+
+func (m *mutexMap) getMutexAndReleaseRef(key string) *sync.RWMutex {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	ent := m.entries[key]
```
Shrekster, line 81: nit; probably should `panic()` if `m.entries` has not been initialized, which would mean that someone called lock/unlock out of order.

internal/blobtesting/map.go
jkowalski: switched to ignore

internal/blobtesting/map.go
```diff
@@ -25,7 +25,22 @@ type mapStorage struct {
...
+func (s *mapStorage) GetCapacity(ctx context.Context) (blob.Capacity, error) {
+	if s.limit >= 0 {
+		return blob.Capacity{
+			SizeB: uint64(s.limit),
+			FreeB: uint64(s.limit - s.totalBytes),
+		}, nil
+	}
+
+	return blob.Capacity{}, blob.ErrNotAVolume
 }
```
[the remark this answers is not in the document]
jkowalski: done

internal/blobtesting/map.go
jkowalski: i don't think this is correct, map of size 10 can fit 10 bytes, not 9.

internal/blobtesting/map.go
```diff
@@ -100,7 +115,13 @@ func (s *mapStorage) PutBlob(ctx context.Context, id blob.ID, data blob.Bytes, o
 
 	data.WriteTo(&b)
 
+	if s.limit >= 0 && s.totalBytes+int64(b.Len()) > s.limit {
+		return errors.Errorf("exceeded limit, unable to add %v bytes, currently using %v/%v", b.Len(), s.totalBytes, s.limit)
+	}
+
+	s.totalBytes -= int64(len(s.data[id]))
 	s.data[id] = b.Bytes()
+	s.totalBytes += int64(b.Len())
```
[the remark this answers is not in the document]
jkowalski: not sure i follow, but fixed that.

internal/cache/mutex_map.go
jkowalski: done

internal/blobtesting/map.go
jkowalski: done

cli/command_cache_set.go
jkowalski: done

internal/blobtesting/map.go
Shrekster: Ah, I see this check is with the incoming blob length. Makes sense.

internal/cache/persistent_lru_cache.go
```diff
@@ -295,86 +233,92 @@ func newContentMetadataHeap() contentMetadataHeap {
...
-		(c.listCache.DataSize()+unsuccessfulDeletesSize) > c.sweep.MaxSizeBytes {
-		oldest := heap.Pop(&c.listCache).(blob.Metadata) //nolint:forcetypeassert
+	for len(c.listCache.data) > 0 && (c.aboveSoftLimit(unsuccessfulDeleteBytes) || c.aboveHardLimit(unsuccessfulDeleteBytes)) {
+		// examine the oldest cache item without removing it from the heap.
+		oldest := c.listCache.data[0]
 
-		// stop here if the oldest item is below the specified minimal age
-		if age := now.Sub(oldest.Timestamp); age < c.sweep.MinSweepAge {
-			heap.Push(&c.listCache, oldest)
+		if age := now.Sub(oldest.Timestamp); age < c.sweep.MinSweepAge && !c.aboveHardLimit(unsuccessfulDeleteBytes) {
```
Shrekster, line 305: nothing to change here; just leaving a quick note about possible cache attrition

So, with this PR we'll override the minimum age rule if we are above hard-limit. This helps maintaining the cache hard limit invariant at all times and simplify things.

I am not 100% sure yet which is a better override ? I mean, overriding min-age vs overriding the hard-limit and allowing blobs with the min-age to decay in timely fashion. Could overriding min-age lead to cache attrition ? Do you know the original intent of having a min-age limit for the kopia cache ?

Shrekster (approved):
LGTM, just a concern about cache attrition, might be quite unfounded really

PR #3240, opened 2023-08-21
feat(repository): allow configuring auto compaction count

vkamra:
Adds a manager option to configure the threshold after which auto compaction kicks in.

This is useful when there are large number of concurrent writers in a repository and the default threshold
causes auto-compaction to thrash.

In a follow-up PR - will allow configuring this in repository open.


julio-lopez (commented):
+1: LG 👍

@jkowalski may have comments, so letting it to him to approve.

jkowalski:
LGTM as-is, but i'd like to rethink the whole auto-compaction business and perhaps move it to maintenance. I don't want to block what you're trying to do here, so let's work on maintenance separately.

jkowalski (approved):

vkamra:
> LGTM as-is, but i'd like to rethink the whole auto-compaction business and perhaps move it to maintenance. I don't want to block what you're trying to do here, so let's work on maintenance separately.

Agree with moving to maintenance. @ashmrtn and I were looking into this with our workload (lots of concurrent writers) and the approach we've been discussing is leveraging the flag here to disable compaction during backup and let maintenance run with default parameters.

Maintenance does end up triggering manifest compaction today (indirectly) as a result of [this](https://github.com/kopia/kopia/blob/master/repo/maintenance/maintenance_params.go#L117) call

jkowalski:
yeah, we should make it explicit, so that only maintenance ever calls it.

julio-lopez:
@vkamra can you rebase please? We are not able to do it.

vkamra:
> @vkamra can you rebase please? We are not able to do it.

Yep - I got this. Will push shortly.

julio-lopez (approved):

julio-lopez:
Will fix linter annotation issue in a sec

PR #3247, opened 2023-08-23
fix(general): add missing checklocks annotation

julio-lopez:


vkamra:
Thank you @julio-lopez

julio-lopez:
@vkamra I was going to tag you once CI passed ;)

Issue #3227, opened 2023-08-16
Snapshots with APPEND ACL fail when checkpoints exist


timoschwarzer:
I have the following setup:

- Computer 1 running Kopia in server mode and hosts the repository
  - This computer has ACLs set so users can only create snapshots but not delete them. This is that in case a remote computer is compromised, the attacker cannot delete snapshots.
    ```
    id:9e6a130b7e07167188ce384ce49c62c5 user:*@* access:APPEND target:type=content
    id:47f006af2a17b4de65b86eac0cd715c4 user:*@* access:READ target:type=policy,policyType=global
    id:2112facc91cc36387031894b2d5262b6 user:*@* access:READ target:type=policy,hostname=OWN_HOST,policyType=host
    id:afefbc72b6a0a0494f6cbd9359eea608 user:*@* access:FULL target:type=policy,hostname=OWN_HOST,username=OWN_USER
    id:bf65d63b9ac6f8456c3fbb664a66b3d0 user:*@* access:APPEND target:type=snapshot,hostname=OWN_HOST,username=OWN_USER
    ```
- Computer 2 is also running Kopia in server mode (for scheduled tasks to run automatically) but is connected to the Kopia server on Computer 1

Now Computer 2 starts making a new snapshot but the internet connection is slow. Due to the slow internet connection, the snapshot takes a while and Kopia creates `incomplete` snapshots as checkpoints in the process.

Now when Computer 2 is about to finish creating the snapshot it tries to delete the `incomplete` snapshots, which the repository server denies because of the ACLs. And because of that, the whole snapshot task fails and the final snapshot is never created.

I would expect Kopia to at least create the final snapshot regardless when it cannot delete the `incomplete` snapshots. It would be better if there was some way to only allow deleting `incomplete` snapshots, or even optimal if the `incomplete` snapshots would be deleted by maintenance tasks instead of the Kopia clients.

Gui13:
I think this issue is linked to the (automatically, probably wrongly) closed issue #2023 

@jkowalski do you think your change will allow the #2023 to be fixed as well?

mega12345mega:
I think this issue is now fixed. Here's the log for a snapshot in v0.22.3 from a Windows client to an Ubuntu server that took over an hour:
```
20:38:10.504 establishing new GRPC streaming session (purpose=Source Manager Uploader)
20:38:10.680 uploading me@computer:C:\Users\Me\Test
[...]

PR #3249, opened 2023-08-24
feat(repository): apply retention policies server-side

### LINKED ISSUE #2023: Snapshot migrate or create errors when using server with APPEND only ACL for snapshots
Hello,

When using kopia server with this policies:
```
id:d62f711dd6c06759bb461c6b236b1fb0 user:*@* access:APPEND target:type=content
id:a0517d9124156b1cc86f83d09a0492e0 user:*@* access:READ target:type=policy,policyType=global
id:f1458d3121e5deefbb8cd6157b9aa820 user:*@* access:READ target:type=policy,hostname=OWN_HOST,policyType=host
id:f97b833fef122f824fd8be61c02fe0de user:*@* access:READ target:type=user,username=OWN_USER@OWN_HOST
id:dc3379322d56c3d6a29aaa1e14637240 user:*@* access:READ target:type=policy,username=OWN_USER,hostname=OWN_HOST
id:57691fe221d611e5df75feb9090f8c6e user:*@* access:APPEND target:type=snapshot,hostname=OWN_HOST,username=OWN_USER
```

On `snapshot migrate` or `snapshot create` error occurs because kopia tries to apply retention policy, but ACL denies it.
In my case it occured during checkpoint process, and investigation showed that reason is possible [here](https://github.com/kopia/kopia/blob/master/snapshot/snapshotfs/upload.go#L433).

It will be nice to have a flag which will totally disable retention policy apply for commands mentioned above.

jkowalski:
This allows append-only snapshots where the client can never delete arbitrary manifests and policies are maintained on the server.

The client only needs permissions to create snapshots in a given source (APPEND), which automatically gives them permission to invoke the server-side method for their own snapshots only. The client still needs `FULL` access to delete snapshots.

Added `--overwrite` method to `kopia server acl add` that allows reducing access level for a combination of user/target.

Fixes #2023


jkowalski:
note this needs more work before it can get merged, we need to add a bunch of tests, in particular make sure this support #3227.

cli/command_acl_add.go
```diff
@@ -11,16 +11,18 @@ import (
...
+	level     string
+	overwrite bool
 }
 
 func (c *commandACLAdd) setup(svc appServices, parent commandParent) {
 	cmd := parent.Command("add", "Add ACL entry")
 	cmd.Flag("user", "User the ACL targets").Required().StringVar(&c.user)
 	cmd.Flag("target", "Manifests targeted by the rule (type:T,key1:value1,...,keyN:valueN)").Required().StringVar(&c.target)
 	cmd.Flag("access", "Access the user gets to subject").Required().EnumVar(&c.level, acl.SupportedAccessLevels()...)
+	cmd.Flag("overwrite", "Overwrite existing rule with the same user and targe").BoolVar(&c.overwrite)
```
Gui13, line 25:
```suggestion
	cmd.Flag("overwrite", "Overwrite existing rule with the same user and target").BoolVar(&c.overwrite)
```

internal/server/api_manifest.go
```diff
@@ -123,3 +126,47 @@ func handleManifestCreate(ctx context.Context, rc requestContext) (interface{},
...
+	}
+
+	usernameAtHostname, _, _ := rc.req.BasicAuth()
+
+	parts := strings.Split(usernameAtHostname, "@")
+	if len(parts) != 2 { //nolint:gomnd
+		return nil, requestError(serverapi.ErrorMalformedRequest, "malformed username")
+	}
+
+	// only allow users with apply retention policy if they have permission to add snapshots
```
Gui13, line 149:
```suggestion
	// only allow users to apply retention policy if they have permission to add snapshots
```

internal/server/api_manifest.go
```diff
@@ -123,3 +126,47 @@ func handleManifestCreate(ctx context.Context, rc requestContext) (interface{},
...
+	// for a particular parth.
```
Gui13, line 150:
```suggestion
	// for a particular path.
```

internal/server/grpc_session.go
```diff
@@ -440,6 +445,47 @@ func handlePrefetchContentsRequest(ctx context.Context, rep repo.Repository, aut
...
+
+	parts := strings.Split(usernameAtHostname, "@")
+	if len(parts) != 2 { //nolint:gomnd
+		return errorResponse(errors.Errorf("invalid username@hostname: %q", usernameAtHostname))
+	}
+
+	username := parts[0]
+	hostname := parts[1]
+
+	// only allow users with apply retention policy if they have permission to add snapshots
```
Gui13, line 460:
```suggestion
	// only allow users to apply retention policy if they have permission to add snapshots
```

snapshot/policy/expire.go
```diff
@@ -7,11 +7,24 @@ import (
...
+func ApplyRetentionPolicy(ctx context.Context, rep repo.RepositoryWriter, sourceInfo snapshot.SourceInfo, reallyDelete bool) ([]manifest.ID, error) {
+	// it is desired to not allow snapshots to be deleted by repository clients,
+	// while still maintain the ability to apply snapshot retention policies server-side.
+	if remote, ok := rep.(repo.RemoteRetentionPolicy); ok {
+		if sourceInfo.UserName == rep.ClientOptions().Username && sourceInfo.Host == rep.ClientOptions().Hostname {
+			// for repository clients, apply retention policy on the server.
+			log(ctx).Debug("applying retention policy on the server")
+
+			//nolint:wrapcheck
+			return remote.ApplyRetentionPolicy(ctx, sourceInfo.Path, reallyDelete)
```
Gui13, line 24: That's the crux, thanks for that @jkowalski ! :heart:

PR #3262, opened 2023-09-02
feat(server): fixed server logging to file & console

jkowalski:
Previously some logs from a running server were only kept in memory (including storage activity logs) which was confusing to many folks.

This changes the behavior so that logs are sent to their regular (console/file) file locations in addition to the UI tasks.

Old behavior can be restored by adding `--no-persistent-logs` to server.


julio-lopez (commented):
Questions related to the server in-memory logs.

- How were in-memory logs visible? via a server endpoint call that would retrieve them?
- Is there a rotation / expiration mechanism for the in-memory logs? or how is the memory consumption capped? How are these log entries cleared?

PR #3264, opened 2023-09-02
breaking(cli): added '--otlp-trace' flag that replaces Jaeger tracing with OTLP over gRPC

jkowalski:
Jaeger exporter support is now deprecated upstream, so we need to move to OTLP. Fortunately Jaeger supports OTLP out of the box.
See https://www.jaegertracing.io/docs/1.48/deployment/

To configure the exporter, use environment variables as described in
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options


julio-lopez (commented):
+1: LG 👍

PR #3265, opened 2023-09-03
feat(server): reduce server polling

jkowalski:


akash07k:
@jkowalski  Thanks for this.
Can you please provide any test build? I'm unable to find it in the PR.

jkowalski:
> @jkowalski Thanks for this.
> Can you please provide any test build? I'm unable to find it in the PR.

Can you ping me on slack.kopia.io so I can make the build for you.

cli/command_server_start.go
```diff
@@ -97,6 +99,8 @@ func (c *commandServerStart) setup(svc advancedAppServices, parent commandParent
 	cmd.Flag("server-control-password", "Server control password").PlaceHolder("PASSWORD").Envar(svc.EnvName("KOPIA_SERVER_CONTROL_PASSWORD")).StringVar(&c.serverControlPassword)
 
 	cmd.Flag("auth-cookie-signing-key", "Force particular auth cookie signing key").Envar(svc.EnvName("KOPIA_AUTH_COOKIE_SIGNING_KEY")).Hidden().StringVar(&c.serverAuthCookieSingingKey)
+	cmd.Flag("log-scheduler", "Enable logging of scheduler actions").Hidden().Default("true").BoolVar(&c.debugScheduler)
+	cmd.Flag("min-maintenance-interval", "Minimal maintenance interval").Hidden().Default("60s").DurationVar(&c.minMaintenanceInterval)
```
Shrekster, line 103:
```suggestion
	cmd.Flag("min-maintenance-interval", "Minimum maintenance interval").Hidden().Default("60s").DurationVar(&c.minMaintenanceInterval)
```

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+var log = logging.Module("scheduler")
+
+// GetItemsFunc is a callback that returns items for the scheduler to consider.
+type GetItemsFunc func(ctx context.Context, now time.Time) []Item
+
+// Item is a describes item that can be scheduled with a function invoked the next time the item is due.
+type Item struct {
+	Description string
+	NextTime    time.Time
+	Trigger     func()
```
julio-lopez, line 26: nit: naming, the executed function is not really the trigger for the `Item`, it is the function to be executed when this item is due. I cannot think of a good name though ATM.

Are items going to be persisted at any point? Should JSON tags be preemptively added?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// Scheduler is a scheduler triggers the next upcoming events based on the list of items returned by listItems().
```
julio-lopez, line 29: this comment is confusing, please reword and clarify.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+	TimeNow func() time.Time
+	Debug   bool
+
+	refreshRequested chan string
+	getItems         GetItemsFunc
+	closed           chan struct{}
+	wg               sync.WaitGroup
+}
+
+// Options provides options for the scheduler.
```
julio-lopez, line 40: does this work?
```suggestion
// Options for the scheduler.
```

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+type Options struct {
+	TimeNow        func() time.Time
+	Debug          bool
+	RefreshChannel chan string
```
julio-lopez, line 44: Wondering, why is the channel provided to the scheduler? Can't it be owned by the scheduler and instead returned by a function? Or does it even need to be exposed at all? Can it be a function that is called on the scheduler such as:
`s.DoRefresh(string) error` or similar?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+	for {
+		now := s.TimeNow()
+		nextTriggerTime, toTrigger := s.upcomingItems(ctx, now)
+
+		sleepTimeUntilNextTrigger := sleepTimeOrDefault(now, nextTriggerTime, sleepTimeWhenNoUpcomingSnapshots)
+		if sleepTimeUntilNextTrigger < 0 {
+			sleepTimeUntilNextTrigger = 0
+		}
+
+		if s.Debug {
```
julio-lopez, line 125: Does it make sense to print this message when `sleepTimeUntilNextTrigger`, maybe only print it when it is not

```suggestion
		if s.Debug && sleepTimeUntilNextTrigger > 0 {
```

julio-lopez (commented):
Few comments

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/kopia/kopia/internal/ctxutil"
+	"github.com/kopia/kopia/repo/logging"
+)
+
+var log = logging.Module("scheduler")
+
+// GetItemsFunc is a callback that returns items for the scheduler to consider.
+type GetItemsFunc func(ctx context.Context, now time.Time) []Item
+
+// Item is a describes item that can be scheduled with a function invoked the next time the item is due.
```
Shrekster, line 22:
```suggestion
// Item describes an item that can be scheduled with a function that is invoked the next time the item is due.
```

internal/scheduler/scheduler.go
Shrekster, line 29:
```suggestion
// Scheduler triggers the next upcoming events based on the list of items returned by the getItems function.
```

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+
+const sleepTimeWhenNoUpcomingSnapshots = 8 * time.Hour
+
+func (s *Scheduler) upcomingItems(ctx context.Context, now time.Time) (nextTriggerTime time.Time, toTrigger []Item) {
+	allsm := s.getItems(ctx, now)
+
+	for _, t := range allsm {
+		if nextTriggerTime.IsZero() || t.NextTime.Before(nextTriggerTime) {
+			nextTriggerTime = t.NextTime
+			toTrigger = nil
```
Shrekster, line 82: Can getItems() return items which were supposed to be triggered in past ?

So, even when we have found a new smaller trigger time here, don't we at-least want to keep those items which have a trigger time already in the past from now ? (i.e. t.NextTime < now)

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+
+	return nextTriggerTime, toTrigger
+}
+
+func sleepTimeOrDefault(now, t time.Time, def time.Duration) time.Duration {
+	if t.IsZero() {
+		return def
+	}
+
+	return t.Sub(now)
```
Shrekster, line 106: just for clarification, would be good to add a comment here

This could be a negative value, right ? if `getItems()` returns items to be triggered a few seconds in the past then the sleep time would become negative.

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+		case <-time.After(sleepTimeUntilNextTrigger):
+			for _, sm := range toTrigger {
+				log(ctx).Debugf("triggering %v", sm.Description)
+
+				sm.Trigger()
+			}
+
+		case reason := <-s.refreshRequested:
+			if s.Debug {
+				log(ctx).Debugw("schedule reevaluation requested", "reason", reason)
```
Shrekster, line 147:
```suggestion
				log(ctx).Debugw("schedule re-evaluation requested", "reason", reason)
```

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,181 @@
...
+				TriggerNames(toTrigger),
+			)
+		}
+
+		select {
+		case <-s.closed:
+			// stopping, just exit
+			return
+
+		case <-time.After(sleepTimeUntilNextTrigger):
```
Shrekster, line 138: This is a potential memory-leak if we end up triggering too many refreshes. Given the potential wait times could be hours, this could accumulate easily:

```
The underlying Timer is not recovered by the garbage collector
 until the timer fires. If efficiency is a concern, use `NewTimer`
 instead and call `Timer.Stop` if the timer is no longer needed.
```
REF: https://pkg.go.dev/time#After

You can use Timer.Reset() to set a new wait.

internal/scheduler/scheduler.go
Shrekster: I agree with @julio-lopez  comment here. A function would prevent the channel from being exposed to the caller. Similar to how Stop() works.

internal/server/server.go
```diff
@@ -450,6 +480,10 @@ func (s *Server) Refresh(ctx context.Context) error {
 		return errors.Wrap(err, "unable to sync sources")
 	}
 
+	if dr, ok := s.rep.(repo.DirectRepository); ok && s.maint != nil {
+		s.maint.refresh(ctx, dr, false)
```
Shrekster, line 484: probably should check if `s.maint` is nil ?

internal/server/server.go
```diff
@@ -450,6 +480,10 @@ func (s *Server) Refresh(ctx context.Context) error {
 		return errors.Wrap(err, "unable to sync sources")
 	}
 
+	if dr, ok := s.rep.(repo.DirectRepository); ok && s.maint != nil {
```
Shrekster, line 483: Ideally, when `ok` is true, `s.maint != nil` is always true as well.

internal/server/server.go
```diff
@@ -450,6 +480,10 @@ func (s *Server) Refresh(ctx context.Context) error {
...
+		s.maint.refresh(ctx, dr, false)
+	}
```
Shrekster, line 485: When `ok` is true, `s.maint != nil` is always true as well, so isn't the interface check for `s.rep.(repo.DirectRepository)` is redundant ?

internal/scheduler/scheduler.go
jkowalski: that's not something that needs to be persistent.

internal/scheduler/scheduler.go
jkowalski: the channel needs to be external, since it simplifies handling refreshes while we recreated - i was struggling with resolving a deadlock here and this was the only reliable choice I found, we can change it however we want since it's internal.

internal/scheduler/scheduler.go
jkowalski: yes, there's a test for it.

internal/scheduler/scheduler.go
jkowalski: done in base PR.

internal/scheduler/scheduler.go
jkowalski: done in base PR.

internal/scheduler/scheduler.go
jkowalski: done in base PR.

internal/scheduler/scheduler.go
jkowalski: done in base PR.

internal/scheduler/scheduler.go
jkowalski: done in base PR.

internal/scheduler/scheduler.go
jkowalski: done in base PR

repo/maintenance/maintenance_schedule.go
```diff
@@ -72,34 +73,38 @@ func getAES256GCM(rep repo.DirectRepository) (cipher.AEAD, error) {
...
+// if the maintenance is not owned by this user, returns time.Time{}.
+func TimeToAttemptNextMaintenance(ctx context.Context, rep repo.DirectRepository) (time.Time, error) {
 	mp, err := GetParams(ctx, rep)
 	if err != nil {
 		return time.Time{}, errors.Wrap(err, "unable to get maintenance parameters")
 	}
 
+	// if maintenance is not owned by this user, do not run maintenance here.
+	if !mp.isOwnedByByThisUser(rep) {
+		return time.Time{}, nil
```
Shrekster, line 85: I prefer if we return a permission error here instead of time.Time{} as it fails silently

Shrekster: I thought about this twice after reading the full change, I think you can still return a `time.Time{}` object that `IsZero()` for when maintenance is not scheduled. So, I think it is better to stay consistent with that behavior here. No need to return the error, just zero time when not an owner is ok.

internal/server/server.go
```diff
@@ -1025,6 +1015,57 @@ func (s *Server) allSourceManagers() map[snapshot.SourceInfo]*sourceManager {
...
+
+	if s.maint != nil {
+		// if we have a direct repository, add an item to run maintenance (if we're the owner)
+		if nextMaintenanceTime := s.maint.nextMaintenanceTime(); !nextMaintenanceTime.IsZero() {
+			result = append(result, scheduler.Item{
+				Description: "maintenance",
+				Trigger:     s.maint.trigger,
+				NextTime:    nextMaintenanceTime,
+			})
+		}
```
Shrekster, line 1039: probably don't add to the schedule if we are not the owner ?

internal/server/server.go
```diff
@@ -1025,6 +1015,57 @@ func (s *Server) allSourceManagers() map[snapshot.SourceInfo]*sourceManager {
...
+
+	// add a scheduled item to refresh all sources and policies
+	result = append(result, scheduler.Item{
+		Description: "refresh",
+		Trigger:     s.refreshAsync,
+		NextTime:    s.nextRefreshTime,
+	})
+
+	if s.maint != nil {
+		// if we have a direct repository, add an item to run maintenance (if we're the owner)
```
Shrekster, line 1032:
```suggestion
		// If we have a direct repository, add an item to run maintenance.
		// If we're the owner then nextMaintenanceTime will be zero.
```

internal/scheduler/scheduler.go
jkowalski: fixed in previous PR

internal/scheduler/scheduler.go
jkowalski: added comment in previous PR

internal/server/server_maintenance.go
```diff
@@ -0,0 +1,154 @@
...
+	s.wg.Wait()
+
+	log(ctx).Debug("maintenance manager stopped")
+}
+
+func (s *srvMaintenance) beforeRun() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// make sure we're not schduling next maintenance until we refresh
```
Shrekster, line 60:
```suggestion
	// make sure we're not scheduling next maintenance until we refresh
```

repo/maintenance/maintenance_schedule.go
jkowalski: ack

internal/server/server.go
jkowalski: done, simplified the code even more

internal/scheduler/scheduler.go
julio-lopez: @jkowalski This was already merged with the scheduler PR.

What about the comment regarding the naming of the `Trigger func()`? Any ideas of a more appropriate name for it?

Shrekster: I am also in two minds about this. The trigger() is actually being used like a trigger in maintenance scheduler instead of it being an interleaving operation.

I think that is the correct usage because otherwise you could end up blocking the scheduler. Let keep the name for now as it signals the desired usage.

Shrekster (approved):
Looks good. I won't be able to comment about the source-manager changes too much because I am unfamiliar with that code, but the changes looks solid to me.

akash07k:
Wow, thanks for merging. @jkowalski  I'm testing PR since around 2 weeks now, it is impressive.

PR #3283, opened 2023-09-09
feat(server): Add socket-activation support

PhracturedBlue:
I'm not sure how much interest there is in this, but the attached patch adds systemd socket-activation support for Kopia.  Normally socket-activation is used for on-demand application startup, but that probably isn't an ideal setup for Kopia as it could prevent the running of maintenance tasks.  However socket activation has a few other uses:

* Allows configuring permissions of a unix-domain-socket
* Allows exposing ports when running in a rootless podman container without the need for slirp4netns (socket-activation does not work with Docker as far as I know)
* Can be used to allow running of privileged ports as non-root user (additional setup required)

My personal use is the 1st one, as it provides a race-free way to ensure my reverse-proxy can access the unix-domain-socket I use for Kopia.

The test is a little ugly, as the `go-systemd/activation` module provides no means of changing the starting FD when scanning for socket-activated sockets, which requires that we enable activation on fd 3 which may already be in use when the test runs.

### DIFF
--- a/cli/command_server_control_linux_test.go
+++ b/cli/command_server_control_linux_test.go
@@ -1,0 +1,104 @@
+//go:build linux
+// +build linux
+
+package cli_test
+
+import (
+	"net"
+	"os"
+	"strconv"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/kopia/kopia/internal/testutil"
+	"github.com/kopia/kopia/tests/testenv"
+)
+
+func TestServerControlSocketActivated(t *testing.T) {
+	var port int
+
+	env := testenv.NewCLITest(t, testenv.RepoFormatNotImportant, testenv.NewInProcRunner(t))
+
+	dir0 := testutil.TempDirectory(t)
+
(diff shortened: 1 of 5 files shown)


jkowalski:
Looks like there are some build issues, in principle I like the change, happy to merge when it passes all checks.

PhracturedBlue:
Thanks for the feedback.  I didn't want to spend too much time on it if there wasn't interest.  I don't have access to the failing platforms, so I need to rely on the ci tests to validate.  I tried to make the socket-activation tests linux only.  The code itself doesn't need to be gated, as it should disable itself on unsupported platforms, but the tests rely on linux-only behavior due to the ugly syscall thing.

PhracturedBlue:
It looks like all tests passed except the code-coverage percent.
It is not possible to hit line 46 with the current activation module, but I'd like to keep the check as that may change in the future

line 58 is not new (just a reformatting of the existing code).  I can write a test for that if you want, but it is mostly unrelated to this change

line 63 is also difficult to check as it requires also duping fd=4 and I'm already scared of the code that handled fd=3.

If fixing the code-coverage is critical to merging, I probably need to roll my own implementation of socket-activation, as the go-systemd module is not very amenable to this type of testing.

jkowalski (approved):

jkowalski:
we're not very rigid about code coverage for cases like this, thank you!

Issue #2656, opened 2022-12-25
GDrive Delegate domain-wide authority/Service Account Quota


gallexme:
Hey when using native gdrive i get this error:

Despite 10 retries: Create in PutBlob(......): unexpected Google Drive error: googleapi: Error 403: The user's Drive storage quota has been exceeded., storageQuotaExceeded

thats because  the service account has their own limits, and to be able to use the whole google drives account storage you would need to upload as the actual domain/user instead

ex: https://stackoverflow.com/a/68313988

Skylinar:
I'm facing the same issue.

oinegue:
In our organization, we worked around that limitation by using a Shared Drive, that is shared with manager role to the service account and to my account.
As the documentation of Google Cloud says that "All files in a shared drive are owned collectively by your organization", the shared account limit isn't considered.
Unfortunately this is possible only if you have a Google Drive subscription that allows you to create Shared Drives.

Skylinar:
This sounds promising. I will give it a try. Thanks for the hint!

Skylinar:
Thanks @oinegue, your hint works as a charm.

PR #3288, opened 2023-09-11
feat(providers): Support interactive oauth setup for gdrive.

### LINKED ISSUE #3047: [design] Migrate GDrive from service account to OAuth
Hi @jkowalski,

As people reported in https://github.com/kopia/kopia/issues/2656, the current GDrive auth doesn't really work as the service account's quota is used instead of the user's. To fix this, we need to migrate to OAuth-based auth. I'm interested in taking on this work, but I'd like your review on this design.

We have two options in terms of [grant scope](https://developers.google.com/drive/api/guides/api-specific-auth):
- Use https://www.googleapis.com/auth/drive - Kopia will have all access to user's Drive files, which I'm not very comfortable with.
- Use https://www.googleapis.com/auth/drive.file - Kopia will have access to specific folders user has chosen, which is closer to the current model and more secure. I lean towards this option.

With `drive.file` scope, we need to ask the user to grant us access to one Drive folder. As far as my research goes, the only way is through [the Drive File Picker API](https://developers.google.com/drive/api/guides/picker), which must be embedded in a web page. One way to do it is to run a temporary HTTP server for this.

The upside of this approach:
- The user selects a folder and grants us access in one go. It's slightly more streamlined than the current GDrive onboarding, where they need to create a service account on GCloud, share the folder on Drive, and then come onto the CLI.
- We can integrate this whole flow into the Kopia React app for UI users.

The downsides:
- `repository create` needs to show a link that the user needs to visit and interact with the Drive widget. If the user runs this command from a headless server behind a NAT, they may need to port-forward the server to their desktop. Alternatively, they can run `repository create` on a desktop and copy the files over. `repository connect` is unaffected.

Questions:
- Is it weird to run a HTTP server as part of `repository create`? Can you think of a better way through this?

Best,

xkxx

xkxx:
Fixes https://github.com/kopia/kopia/issues/3047, https://github.com/kopia/kopia/issues/2656

Hi @jkowalski,

Would you take an early look at this PR:
- The changes in repo/ and cli/ are small and ready more or less.
- The changes in internal/gdriveauth/ is more POC-quality (see below).

I would like your opinion on how the ephemeral http server is structured. You mentioned that you may want to add more use cases in the future, so maybe a consistent architecture would help? I also don't work with go serving on my day job, so feedback is much appreciated here.

I also marked the existing service account-based auth as deprecated in the CLI, although it should work. It's my opinion that we should mark it obsolete and remove it soon, since the gotchas with the quota is great and confusing.

The overall flow of the code:
- In `repository create`, we launch a http server that serves 2 ajax endpoints and a html file.
- The user is asked to authenticate and then authorize kopia. If it goes well, we receive an auth code.
- The auth code is sent back to the server, which it uses to exchange for a long-term refresh token and a short-term access token. The access token is returned to the client.
- The client initializes a gdrive picker widget with the access token.
- The user is asked to pick a folder to give kopia access to. The folder id is then sent back to the server.
- The token and folder id are persisted, the server is shutdown.

Edge cases:
- If the user lingers on the picker for >1hr without doing anything, the access token expires. We don't currently handle that, the user needs to re-run the command.
- If the refresh token expires for any reason, we need to prompt the user to re-run the initial setup.

Thanks,

xkxx


petefinesse:
@jkowalski Can we prioritize it, please? service-account quotas are confusing for everyone. There are multiple topics/slack questions about it

jkowalski:
Sorry I've been sitting on this for too long, I don't really use google drive myself so am not very familiar with it and this wasn't most straightforward to test.

Overall UI and the approach looks reasonable. I have some questions:

1. we need step-by-step instructions for testing it, in particular which keys do I need to generate beforehand? Can you add a doc page?
2. We need some kind of automated tests
3. How do we deal with access tokens persistence? My understanding is they should be somewhat persisted/cached so we don't exchange refresh tokens for access tokens on each session.
4. How would refresh token be extended, are those eternal or will refresh tokens expire after a while?
5. Would it be practical to register app/client credentials and embed those in the app so that folks don't need to do all those manual steps?
6. How would this extend to KopiaUI? How can we integrate the authorization flow into the web app we already have and the desktop app?

PR #3289, opened 2023-09-11
feat(ui): Allow specifying whether local or all snapshots are shown by default

PhracturedBlue:
When I connect to the server http page, I prefer to see all snapshots shown by default.  every time I switch away and come back I need to reselect 'All' instead of 'Local'.  This just makes the default a preference option.

Requires https://github.com/kopia/htmlui/pull/186

### DIFF
--- a/internal/serverapi/serverapi.go
+++ b/internal/serverapi/serverapi.go
@@ -289,7 +289,8 @@ type CLIInfo struct {
 
 // UIPreferences represents JSON object storing UI preferences.
 type UIPreferences struct {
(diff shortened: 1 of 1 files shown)


PhracturedBlue:
I noticed that this will result in a configuration setting that has no effect when using the kopiaUI application (since multiUser is false there).

Solutions:
* keep the code as is
* leave the setting always available and place it in a 'multi user' section
* query multi-user mode on the preferences page and hide the config option
* remove the config option altogether and default to 'All Snapshots' if there is no local snapshot

The last option cold be done entirely in the UI without a need for any server-side changes, but it is the least flexible.

jkowalski (approved):

PR #3294, opened 2023-09-12
feat(cli): add more info to "snapshot create" json

schra:
I'd like to expose how many bytes were uploaded to the json of `kopia snapshot create --json`. I tested my diff and it works as expected.

I think this diff is rather hacky. I am motivated to improve this diff, if you let me know what you would like me to change. I wasn't sure how to cleanly implement this and I am also new to Go.

### DIFF
--- a/cli/command_snapshot_create.go
+++ b/cli/command_snapshot_create.go
@@ -286,6 +286,19 @@ func (c *commandSnapshotCreate) snapshotSingleSource(ctx context.Context, fsEntr
 	manifest.Tags = tags
 	manifest.UpdatePins(c.pins, nil)
 
+	progress := c.svc.getProgress()
+
+	manifest.Progress = map[string]int64{
+		"uploadedBytes":     progress.uploadedBytes.Load(),
+		"cachedBytes":       progress.cachedBytes.Load(),
+		"hashedBytes":       progress.hashedBytes.Load(),
+		"cachedFiles":       int64(progress.cachedFiles.Load()),
+		"hashedFiles":       int64(progress.hashedFiles.Load()),
+		"uploadedFiles":     int64(progress.uploadedFiles.Load()),
+		"ignoredErrorCount": int64(progress.ignoredErrorCount.Load()),
+		"fatalErrorCount":   int64(progress.fatalErrorCount.Load()),
+	}
+
 	startTimeOverride, _ := parseTimestamp(c.snapshotCreateStartTime)
 	endTimeOverride, _ := parseTimestamp(c.snapshotCreateEndTime)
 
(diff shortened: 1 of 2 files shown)


cli/command_snapshot_create.go
```diff
@@ -286,6 +286,19 @@ func (c *commandSnapshotCreate) snapshotSingleSource(ctx context.Context, fsEntr
 	manifest.Tags = tags
 	manifest.UpdatePins(c.pins, nil)
 
+	progress := c.svc.getProgress()
+
+	manifest.Progress = map[string]int64{
```
schra, line 291: I've listed all integer fields of `cliProgress` here, except `inProgressHashing` because it is "in progress" and thus not interesting to expose to the json

cli/command_snapshot_create.go
```diff
@@ -336,7 +349,7 @@ func (c *commandSnapshotCreate) snapshotSingleSource(ctx context.Context, fsEntr
 		}
 	}
 
-	c.svc.getProgress().Finish()
+	progress.Finish()
```
schra, line 352: not totally sure if I should first `Finish()` before I access the values in `progress`. to me it seems like that the `Finish` function only changes what is displayed in the cli but doesn't affect the integer values, thus I think it doesn't matter when `Finish` is called

snapshot/manifest.go
```diff
@@ -18,9 +18,10 @@ type Manifest struct {
 	ID     manifest.ID `json:"id"`
 	Source SourceInfo  `json:"source"`
 
-	Description string          `json:"description"`
-	StartTime   fs.UTCTimestamp `json:"startTime"`
-	EndTime     fs.UTCTimestamp `json:"endTime"`
+	Description string           `json:"description"`
+	StartTime   fs.UTCTimestamp  `json:"startTime"`
+	EndTime     fs.UTCTimestamp  `json:"endTime"`
+	Progress    map[string]int64 `json:"progress"`
```
schra, line 24: I wasn't sure how exactly to expose `progress` here. Would it be better to introduce a new struct? Can I somehow directly use the `cliProgress` struct? Is using a map acceptable here?

jkowalski: doesn't `Stats` below have what you need?

schra: `Stats` doesn't contain an uploadedBytes field. The only byte fields it has are `TotalFileSize` and `ExcludedTotalFileSize`. What would be the best option to continue? For example, I could try to add the new fields `UploadedTotalFileSize`  and ` CachedTotalFileSize`   to `Stats`. Or I could keep the map that I introduced. Or I could create a new Stats struct that is only used to report the data in the json file. Thoughts :)?

jkowalski: Let's do Stats. I would prefer to be intentional and only add 1 or 2 field as needed, I know some folks manage millions of manifests and bloat can already be a problem.

Issue #3197, opened 2023-08-07
[Feature Request]: Automatically run overdue snapshots


DarkArc:
Currently, Kopia doesn't seem to do any "cathcup" snapshots if they're overdue. 

# Observed Behavior

In other words, if your backups run at night at 1am, and you have you computer off when the snapshot should've kicked off, Kopia will not run until 1am the next day.

# Expected Behavior

Kopia should (at least optionally) immediately start any overdue snapshots that missed their run window when the computer is booted.

# Rational

Nightly backups are often optimal in terms of avoiding impacting users. However, nightly backups in desktop and laptop form factors can easily result in a user missing a backup if their computer happens to not be on at the "right" time for an unexpectedly long period.

This can in theory even affect more frequent uses if you get "unlucky" (as your device is regularly being put to sleep/shutdown) and thus isn't online "on the hour", when the various backup(s) are scheduled to run.

Kopia having a nice GUI client positions itself as a great easy to use solution for these sort of "desktop"/"laptop" form factors. However, the lack of any kind of "catch up" makes this quite a dangerous option; particularly when deploying to users that aren't as tech-savy and might not think to check in on the status of their backups.

lupusA:
Hi @DarkArc ,

Can you post some information about your setup ? I am running a repository server with a schedule of 12 hours. When I start my computer, Kopia immediately creates a new snapshot when it is overdue. 

Cheers,

DarkArc:
I have a time of day schedule as noted (1am):
![image](https://github.com/kopia/kopia/assets/778012/3c68a75d-650a-4be8-a048-fc09b7237346)

lupusA:
Hi @DarkArc ,

Would you mind to set the schedule to a fix 6 or 12 hours period instead of a fix time of day and test again? Maybe a schedule with a fix time of day is not working correctly.

Cheers,

DarkArc:
Yes, it looks like Kopia correctly handles this for 12 hour snapshots, but not time based snapshots.

DarkArc:
Closing in favor of https://github.com/kopia/kopia/issues/3319

PR #3311, opened 2023-09-16
feat(cli): Added option to allow running missed snapshots on Kopia start

PhracturedBlue:
This PR adds a configuration option to create an immediate snapshot if a previous time-of-day snapshot was missed.  Today, interval snapshots will automatically be taken immediately if they were missed, but a time-of-day snapshot will not be taken until the next schedule.

 made this a configuration item instead of just enabling it because it may be that some users would prefer to miss a snasphot and take the next one off-hours rather than doing it immediately on startup.

This implements https://github.com/kopia/kopia/issues/3197

This is associated wit the htmlui patch: https://github.com/kopia/htmlui/pull/190

### DIFF
--- a/cli/command_policy_set_scheduling.go
+++ b/cli/command_policy_set_scheduling.go
@@ -16,12 +16,14 @@ type policySchedulingFlags struct {
 	policySetTimesOfDay []string
 	policySetCron       string
 	policySetManual     bool
+	policySetRunMissed  bool
 }
 
 func (c *policySchedulingFlags) setup(cmd *kingpin.CmdClause) {
 	cmd.Flag("snapshot-interval", "Interval between snapshots").DurationListVar(&c.policySetInterval)
 	cmd.Flag("snapshot-time", "Comma-separated times of day when to take snapshot (HH:mm,HH:mm,...) or 'inherit' to remove override").StringsVar(&c.policySetTimesOfDay)
 	cmd.Flag("snapshot-time-crontab", "Semicolon-separated crontab-compatible expressions (or 'inherit')").StringVar(&c.policySetCron)
+	cmd.Flag("run-missed", "Run missed time-of-day snapshots (has no effect on interval snapshots)").BoolVar(&c.policySetRunMissed)
 	cmd.Flag("manual", "Only create snapshots manually").BoolVar(&c.policySetManual)
 }
 
@@ -91,6 +93,8 @@ func (c *policySchedulingFlags) setScheduleFromFlags(ctx context.Context, sp *po
 		}
 	}
 
+	c.setRunMissedFromFlags(ctx, sp, changeCount)
+
 	if sp.Manual {
 		*changeCount++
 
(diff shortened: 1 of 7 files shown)


PhracturedBlue:
Sorry...I somehow screwed up the final rebase and made a mess of things.  It should all be ready to go now I think

lupusA:
Hi, do we need a configuration item for this? I was taken a look at the code two weeks ago, and thought that it could be handled by rewriting that method. 

What are your plans?

Cheers

PhracturedBlue:
As per the kopia PR, I think it should have a configuration.  A server may be configured such that the only appropriate time for a snapshot is off-hours, and it is better to miss the snapshot than to take one at whatever time Kopia starts.  So I think both models are valid (just as in linux we have cron and anacron which provide both behaviors, and at least on debian, are used depending on where a cron statement is placed)

PhracturedBlue:
Ugh...it looks like the test for socket-activation isn't robust enough.  I added a bug to discuss options: https://github.com/kopia/kopia/issues/3313
This issue shouldn't be related to the PR though.

jkowalski:
LGTM, but can you please do:

``` 
go get github.com/kopia/htmluibuild@9880605
go mod tidy
```

and add commit to this PR, this way we can merge both CLI and UI in one shot.

PhracturedBlue:
Thanks, done.
For future reference, do you prefer multiple commits (which may or may-not be squashed at the end) or a single rebased commit?  I've been mostly doing the latter, but wasn't sure if there was a preference.

jkowalski:
re: commits it does not matter, we squish them anyway

multiple commits make it easier to see the changes over time

jkowalski (approved):

DarkArc:
This doesn't seem like it's working in 1.14.1. I set up a time of day snapshot for 1am, shut down the computer at 12:50, turned the computer on today, and it did not run.

DarkArc:
Also, it seems weird to me as a user that this is a setting, or at least, that it's a setting that _only_ applies to time of day snapshots. What about the cron expressions, do they count?

Also why is it the default to do catch up snapshots for x-hour snapshot types, but (presumably) not time of day or cron?

DarkArc:
> Did you enable the option before starting?

Yes.

> There is some heuristic logic that prevents running a snapshot on startup if the next snapshot would run within the next 30 minutes to prevent running 2 snapshots back-to-back.

The next snapshot was ~14 hours away.

> It is an option because in same server settings running a backup in the middle of the day si a bad idea. It is better to skip a backup than to run on launch

I would expect those environments to drive this externally via the CLI and something like cron (where they have more control anyways, and will get email reports/centralized logging/etc).

> Maybe the default should be to enabled vs disabled as that is probably more common. However, the interval snapshots just say 'run this every n minutes' which is quite different than 'run this at xx:yy o'clock. I don't see why the default of one has anything to do with the default of the other.

The relation is consistency across the feature set (time based backups). Why does one set of time based backups have a different "catch up" behavior than another? It adds complexity, and with complexity comes user error and mistakes.

I still don't think this needs to be an optional thing (and I apologize for not commenting sooner, for whatever reason this wasn't flagged for me -- the person that filed https://github.com/kopia/kopia/issues/3197), but if it does, I'd strongly advocate for it being the default behavior.

PR #3314, opened 2023-09-17
fix(ui): fixed dark mode via https://github.com/kopia/htmlui/pull/191

jkowalski:


jkowalski:
Verified that the dark mode showing every other row as whit in a table is now fixed. Thanks @lupusA

PR #3323, opened 2023-09-18
feat(server): improve scheduler algorithm to run missed snapshots 

### LINKED ISSUE #3319: Automatically run overdue snapshots
https://github.com/kopia/kopia/pull/3311 (implementing https://github.com/kopia/kopia/issues/3197) doesn't seem to work.

image

I set this all up last night, shortly after midnight, shutdown the computer, and then booted it this morning. The snapshot did not run.

PhracturedBlue:
The current algorithm assumes a singe snapshot per-day.  It doesn't work well with either >1 time-of-day snapshots or with cron snapshots.

This PR calculates the next snapshot time that should have happened based on the previous snapshot.  If that time is before now, then we missed a snapshot and should do a catchup snapshot.

I kept the heuristic of not running a snapshot if the next snapshot would occur within the next 30 minutes.  I'm no longer sure if that is a good policy or not.

Fixes #3319

### DIFF
--- a/snapshot/policy/scheduling_policy.go
+++ b/snapshot/policy/scheduling_policy.go
@@ -94,8 +94,6 @@ func (p *SchedulingPolicy) NextSnapshotTime(previousSnapshotTime, now time.Time)
 		return time.Time{}, false
 	}
 
-	const oneDay = 24 * time.Hour
-
 	var (
 		nextSnapshotTime time.Time
 		ok               bool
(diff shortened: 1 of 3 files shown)


snapshot/policy/scheduling_policy_test.go
```diff
@@ -225,14 +225,15 @@ func TestNextSnapshotTime(t *testing.T) {
 			wantOK:               true,
 		},
 		{
-			// Don't run immediately even though RunMissed is set because last run was not missed
+			// Run immediately because Cron was missed
 			pol: policy.SchedulingPolicy{
-				TimesOfDay: []policy.TimeOfDay{{10, 0}},
+				TimesOfDay: []policy.TimeOfDay{{11, 55}},
```
jkowalski, line 230: can you add a test with multiple times of day?

snapshot/policy/scheduling_policy.go
```diff
@@ -147,22 +181,34 @@ func (p *SchedulingPolicy) NextSnapshotTime(previousSnapshotTime, now time.Time)
...
-
 	const halfhour = 30 * time.Minute
 
-	return (len(p.TimesOfDay) > 0 || len(p.Cron) > 0) && p.RunMissed && previousSnapshotTime.Add(oneDay-halfhour).Before(now) && nextSnapshotTime.After(now.Add(halfhour))
+	if !p.RunMissed {
+		return false
+	}
+
+	nextSnapshot := nextSnapshotTime
+	todSnapshot, todOk := p.getNextTimeOfDaySnapshot(previousSnapshotTime.Add(time.Second))
```
jkowalski, line 196: why do we need to add `time.Second`? Is that really needed?

snapshot/policy/scheduling_policy_test.go
PhracturedBlue: Done

snapshot/policy/scheduling_policy.go
DarkArc: It's a small thing, but I'd suggest extracting `previousSnapshotTime.Add(time.Second)` into its own variable (`momentAfterSnapshot` or something :) )

PhracturedBlue:
Should I be trying to debug the mac or ubuntu test failures?  neither seem to have anything to do with scheduling policy, so I was assuming they are flaky tests unrelated to my change.  But if that isn't true, please let me know, and I'll try to dig into them.

jkowalski (approved):

PR #3330, opened 2023-09-19
fix(cli): RunMissed policy now defaults to true

PhracturedBlue:
I've seen a few comments now indicating that defaulting the RunMissed policy option to 'true' may be a better option.  This PR also converts it into a Tri-State which makes setting/clearing the value more intuitive from the cmdline (in current production, --run-missed must be specified whenever changing the schedule, whereas when using the tristate, it is only modified when the user specifically requests it.

Note that this PR will conflict with #3323, so if both are accepted, one of them will need to be reconciled with the other.

### DIFF
--- a/snapshot/policy/scheduling_policy.go
+++ b/snapshot/policy/scheduling_policy.go
@@ -75,7 +75,7 @@ type SchedulingPolicyDefinition struct {
 }
 
 // defaultRunMissed is the value for RunMissed.
-const defaultRunMissed = false
+const defaultRunMissed = true
 
 // Interval returns the snapshot interval or zero if not specified.
 func (p *SchedulingPolicy) Interval() time.Duration {


snapshot/policy/policy_tree.go
```diff
@@ -46,7 +46,7 @@ var (
 		KeepWeekly:               newOptionalInt(defaultKeepWeekly),
 		KeepMonthly:              newOptionalInt(defaultKeepMonthly),
 		KeepAnnual:               newOptionalInt(defaultKeepAnnual),
-		IgnoreIdenticalSnapshots: newOptionalBool(defaultIgnoreIdenticalSnapshots),
+		IgnoreIdenticalSnapshots: NewOptionalBool(defaultIgnoreIdenticalSnapshots),
 	}
 
 	defaultSchedulingPolicy = SchedulingPolicy{}
```
jkowalski, line 52: should we add this to defaultSchedulingPolicy ?

PhracturedBlue: Yes, thanks, I was looking for something like this and missed it here.

PhracturedBlue:
Note.  This may be a breaking change.  Converting the boolean to a tristate will reset runMissed on any existing policies.

jkowalski:
> Note. This may be a breaking change. Converting the boolean to a tristate will reset runMissed on any existing policies.

@julio-lopez any concerns here?

@PhracturedBlue I think we can elimiate the breaking change by using `OrDefault(false)`. The default global policy for new repos will have global default of true and pre-existing repos will have a global defaulf of "not set" and thus "false".

PhracturedBlue:
after further testing, it isn't a breaking change in that any overrides that have been created so far will continue to work.  the only 'breaking' part is that everyone using the global policy will now have 'run-missed' default to 'true'.  I don't know if that is breaking as much as just an intended change in policy.  I can indeed make the change you mentioned and then only new repos would get the default set to true which is more conservative.  Either way is fine, but I'd like guidance on which way you prefer.

jkowalski:
I think if we change to OrDefault(false) it's certainly safer as it won't introduce any surprises for existing users.

jkowalski (approved):
LGTM

PhracturedBlue:
I tested using OrDefault(false) and it doesnt' actually change the behavior at all.  The reason is because the current code uses a bool which ends up resulting in false==inherit.  so if you haven't set it you get the default.  if you set it to false, you also get the default.  This means the current implementation is pretty broken, in that if you do:
```
policy set --run-missed /dir
policy set --no-run-missed /dir/subdir
```
that the 2nd command doesn't work and instead /dir/subdir will use 'run-missed' because false==inherit (which is true)

Once this patch is applied the behavior becomes what you'd expect.  It doesn't seem like there is really any way to have the default be true and to not make it the default for everyone's existing repos.  I'm open to other suggestions.

jkowalski:
That's the point, when you `inherit` all the way to the top you get what's stored in a default policy. Only if that is not set `false` will be used. So old repositories will get `false` (because of OrDefault(false)) and new repositories will get `true` (because of `DefaultPolicy.Scheduling.RunMissed`)

PhracturedBlue:
I'm missing something then.  because even after changing the OrDefault code to use `false`, my 'old' repos are inheriting from the new default, not from `false`.  I think this is because the current `runmissed=false` seems to map to `runmissed=nil` when the bool is upgraded to a Enum, so the --global policy picks up the new default (which is `true`).  I don't see a way to change that behavior.

julio-lopez (changes requested):
Just catching up. Let's hold off merging this.

IINM, this will break some use cases because it changes the assumption for new repos / policies, etc.

julio-lopez:
Adding @redgoat650 here.

The main concern is that now for new repos is that `RunMissed` will be `true` for new repos. This will break various use cases and will be surprising for some users. In particular, for a use case where the the execution of kopia commands is explicitly triggered externally (by the user or by an orchestration system), will this result in an additional missed snapshot being taken?

Note: Also, I do not yet fully understand the behavior for old repos when `RunMissed` is not set.

PhracturedBlue:
What if I split this into 2 PRs?  I think the change from a bool to enum is actually a bug-fix.  Today you cannot unset the run-missed policy on a child dir due to false==inherit.  I can Make a pr which converts the bool to an enum (and keeps the default policy of disabled), and then a separate 1-line PR which converts the default from false to true.

julio-lopez:
> What if I split this into 2 PRs? I think the change from a bool to enum is actually a bug-fix. Today you cannot unset the run-missed policy on a child dir due to false==inherit. I can Make a pr which converts the bool to an enum (and keeps the default policy of disabled), and then a separate 1-line PR which converts the default from false to true.

Perfect! please do. Thanks a lot.

redgoat650:
The change from bool to enum looks great. Adding my thoughts on the default change discussion:

1. The advantage of changing the default to `true` is clear: one might argue that since kopia is a backup tool, it should by default do everything in its power to secure live data, erring on the side of "when in doubt, create more snapshots rather than fewer".

2. The advantage of keeping the default as `false` is also clear: user expectations and workflows will require no changes, and kopia already provides and documents the tools needed to give "run-missed-time-of-day" functionality.

3. The disadvantage of changing the default to `true` is less clear: will it break any existing user workflows? I can certainly think of contrived cases where the behavior will diverge depending on the default value, though none of what I can think of seems particularly compelling or useful (note: some compelling/useful cases may exist, even if I don't happen to think of them!). Do any users depend on such cases? Seems unlikely, but we don't have enough information to answer that with certainty.

4. The disadvantage of keeping the default `false` seems pretty minimal: current behavior is clearly described; users may initially expect a different default behavior but aligning their expectations is a matter of providing one bool flag. There may be some situations where a user assumes the wrong default behavior, and would have benefitted from sneaking in that extra missed-snapshot run. The users affected are those that didn't read or understand kopia's documented options. 
An additional disadvantage is if we continually need to address posts asking why run-missed isn't default (but if it were the other way around, would these be replaced by questions like "I told kopia to only snapshot every day at midnight, why did it trigger one when I started up the server at 5pm?"?).

Overall, to me, the absolute safest option is to keep the default unchanged. The very-likely-also-safe option is to change the default to true, but there are unknown risks behind that door.

julio-lopez:
Thanks Nick.

- It makes sense to turn the default to "true" for new policies, so missed snapshots are taken.
- What about existing policies that do not have the value set? should the default be true as well? That changes behavior for existing users.

Thoughts?

redgoat650:
> What about existing policies that do not have the value set? should the default be true as well? That changes behavior for existing users.

Well (from my testing at least) this shouldn't happen for existing policies; the behavior won't change in an existing repo, only for new repos created.

The existing repo has "false" affirmatively defined in its global policy by default, and that's written into the policy itself when the repo is created, not reliant on whatever default value kopia has defined. So when I created a repository with a false-default-kopia-build, then queried the global policy with a true-default-kopia-build, the value still returns as `false` - meaning that any policies currently inheriting from global are also still false (behavior unchanged).

Following #3337's merge, I can now also modify the global policy setting to `inherit`, which will compute the value based on whatever default is compiled into the kopia I use to query it. This requires the user to actively override the value to `inherit` though.

All new repos created with the true-default-kopia-build will also directly define the global policy to its default, which will be true for new repos. Again, by affirmatively setting to `inherit`, it'll evaluate to whatever's compiled into the querying kopia binary.

In short, whatever the default value is in kopia at the time the repo is created is "saved" in the global policy - either true or false. Now users can actively set it to "inherit" if they want, which will instead use the current compiled default value at runtime. Otherwise, the policies in existing repos should behave the same as they always did. Policies in new repos would use the new default true.

So the user that gets affected is one that creates repos frequently and creates a time-of-day policy for each repo and does not explicitly set the `--run-missed` flag when doing so. Existing repos and policies should be unaffected.

PhracturedBlue:
@redgoat650 My testing of this PR indicates that the policy of a pre-existing repo IS changed when this PR is applied (which is not how I expected it to behave).  Things will be a bit more complicated with #3337 applied but it should still be easy to test a pre-3337 repo vs one with this PR applied.  I don't know WHY I was seeing that behavior, as it was not what I thought the code would do.  Unless the desire is to make all old repos default runmissed to 'true' I think we need to understand why I don't see the expected behavior before moving forward.

redgoat650:
I see, I was testing post-#3337, so the behavior I described isn't quite accurate for existing repos/policies. When I create a repo pre-#3337, the `runMissed` option is empty for the global policy, which only gets set to the binary default when connecting after #3337! Alright yeah this may be an issue for existing policies.

PhracturedBlue:
I rebased the change to make this into a 1-liner.  I am not sure if there is a way to set the default policy to 'false' for existing repos, but to 'true' for new repos.  It could probably be done with some code in LoadConfigFromFile() but that feels really ugly to me.  I guess there could be some kind of 'upgrade' routine for the config file, but that leads to versioning the config file, and seems like a huge amount of work to enable this feature.

julio-lopez:
My questions above about the default behavior are really about what's best for (the majority of) the users.

Personally, either way works for me.
Let's merge this, and we can tweak it if we see that it would make sense to change the  behavior.

julio-lopez (approved):

julio-lopez:
@jkowalski PTAL and comment if there are any concerns

PR #3332, opened 2023-09-19
test(snapshots): Name policy tests to make debugging easier

PhracturedBlue:
After debugging several snapshot-policies, I found it would be nicer to have names for each test rather than needing to count them.

### DIFF
--- a/snapshot/policy/scheduling_policy_test.go
+++ b/snapshot/policy/scheduling_policy_test.go
@@ -1,7 +1,6 @@
 package policy_test
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
@@ -10,17 +9,19 @@ import (
 	"github.com/kopia/kopia/snapshot/policy"
 )
 
+//nolint:maintidx
 func TestNextSnapshotTime(t *testing.T) {
 	cases := []struct {
+		name                 string
 		pol                  policy.SchedulingPolicy
 		now                  time.Time
 		previousSnapshotTime time.Time
 		wantTime             time.Time
 		wantOK               bool
 	}{
-		{}, // empty policy, no snapshot
+		{name: "empty policy, no snapshot"},
 		{
-			// next snapshot is 1 minute after last, which is in the past
+			name:                 "next snapshot is 1 minute after last, which is in the past",
 			pol:                  policy.SchedulingPolicy{IntervalSeconds: 60},
 			now:                  time.Date(2020, time.January, 1, 12, 3, 0, 0, time.Local),
 			previousSnapshotTime: time.Date(2020, time.January, 1, 11, 50, 0, 0, time.Local),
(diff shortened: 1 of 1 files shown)


snapshot/policy/scheduling_policy_test.go
```diff
@@ -237,8 +254,8 @@ func TestNextSnapshotTime(t *testing.T) {
 		},
 	}
 
-	for i, tc := range cases {
-		t.Run(fmt.Sprintf("case-%v", i), func(t *testing.T) {
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
```
julio-lopez, line 258: This becomes part of the test name, which makes it hard to select.

What about changing `name` to description, and logging the description at the beginning of the test (via `t.Log[f]`).

It may be a good idea to still have a `name` for the test, but with something shorter in CamelCase.

WDYT?

PhracturedBlue: I didn't think about the ease of selecting a single test.  I think a name is just redundant if t.Log shows up.  Let me try it with the original numeric names and with a log message per test.  I saw it was done like that in the cli policy-set test, but didn't realize why

PhracturedBlue: I tried it with t.Log, but it prints all the test logs (not just the failing one). But adding the name to the 'message' field of the require.xxx seems to work really well, so that is what I did.

PhracturedBlue:
Example failure with the latest change:
```
--- FAIL: TestNextSnapshotTime (0.00s)
    --- FAIL: TestNextSnapshotTime/case-20 (0.00s)
        scheduling_policy_test.go:261:
                Error Trace:    /go/src/kopia/snapshot/policy/scheduling_policy_test.go:261
                Error:          Not equal:
                                expected: time.Date(2020, time.January, 2, 11, 55, 30, 0, time.Local)
                                actual  : time.Date(2020, time.January, 3, 11, 55, 0, 0, time.Local)

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                -(time.Time) 2020-01-02 11:55:30 +0000 UTC
                                +(time.Time) 2020-01-03 11:55:00 +0000 UTC

                Test:           TestNextSnapshotTime/case-20
                Messages:       Run immediately since last run was missed and RunMissed is set
FAIL
FAIL    github.com/kopia/kopia/snapshot/policy  0.014s
FAIL
```

julio-lopez (approved):

snapshot/policy/scheduling_policy_test.go
```diff
@@ -10,54 +10,59 @@ import (
 	"github.com/kopia/kopia/snapshot/policy"
 )
 
+//nolint:maintidx
```
julio-lopez, line 13: is this `//nolint` directive still needed?
```suggestion
```

PhracturedBlue: yes.  apparently adding 'name' to each section is enough to put us over.  I could split the test up into multiple functions if that is preferred.

PR #3336, opened 2023-09-20
fix(kopiaui): Fix missing window when external displays are detached

lupusA:
This PR contains the following fix:

- Fixed missing window in cases where external displays are detached
- Maximize the window now works as expected

This PR is an **emergency fix** for #3315. If an external display is detached after the kopia-ui window has been closed on this display, the ui won't open anymore. 

The logic now checks if the scaling factors of attached displays are the same (covers https://github.com/electron/electron/issues/10862). Additionally it checks, whether the window bounds lie within a display's working area. Both conditions has to be true, so that the stored window bounds are restored. 

If not, the window will default to the center of the primary screen with it's default width and height. 

Please review. 

Cheers,

### DIFF
--- a/app/public/electron.js
+++ b/app/public/electron.js
@@ -51,12 +51,14 @@ function isWithinBounds(winBounds) {
       isFactorEqual = false
       break;
     }
+    /*
     //Check if window bounds lie within at least one display's workarea
     if ((winBounds.x >= workArea.x && winBounds.y >= workArea.y)
       && (winBounds.x < (workArea.x + winBounds.width)
-      && winBounds.y < (workArea.y + winBounds.height))) {
+        && winBounds.y < (workArea.y + winBounds.height))) {
       isWithinBounds = true
     }
+    */
     prevFactor = factor
   }
   return isFactorEqual && isWithinBounds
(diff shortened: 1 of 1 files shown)


jkowalski:
it seems to be that restoring window position is not as easy as we though, we should probably add a way to shut it globally off in the app menu?

lupusA:
> it seems to be that restoring window position is not as easy as we though, we should probably add a way to shut it globally off in the app menu?

Hi @jkowalski, 

I would not add an option. Reason is that restoring the window should be a default behavior for applications. Other applications do it and Kopia should do it too. 

The implementation in this PR should be robust enough for common situations. If it is not covered it defaults to the primary display, which is fine. 

However, in Version 0.14.1, the window can get out of bounds forever. You need to attach the external monitor again to get it back.

Cheers

lupusA:
Hi @jkowalski, 

i've refactored the method. It should now work as expected. The code does the following:

- Saving the window bounds by using a hash of the current display configuration as a key
- If the window is shown, it retrieves the stored window bounds for this configuration
- If the scaling factors are equal, it restores the window bounds. 

Cheers,

jkowalski (approved):

PR #3337, opened 2023-09-21
fix(cli): Convert --run-missed from bool to Enum

PhracturedBlue:
As per discussion on #3330 this converts --run-missed from a bool to a Enum to allow 'inherit' to properly work.  It keeps the the default as 'false', and so will not change the behavior of existing or new repositories.

Changing the default to 'true' can be decided on independently.

### DIFF
--- a/snapshot/policy/scheduling_policy.go
+++ b/snapshot/policy/scheduling_policy.go
@@ -74,6 +74,9 @@ type SchedulingPolicyDefinition struct {
 	RunMissed       snapshot.SourceInfo `json:"runMissed,omitempty"`
 }
 
+// DefaultRunMissed is the value for RunMissed.
+const DefaultRunMissed = true
+
 // Interval returns the snapshot interval or zero if not specified.
 func (p *SchedulingPolicy) Interval() time.Duration {
 	return time.Duration(p.IntervalSeconds) * time.Second
@@ -162,7 +165,7 @@ func (p *SchedulingPolicy) checkMissedSnapshot(now, previousSnapshotTime, nextSn
 
 	const halfhour = 30 * time.Minute
 
-	if !p.RunMissed.OrDefault(true) {
+	if !p.RunMissed.OrDefault(DefaultRunMissed) {
 		return false
 	}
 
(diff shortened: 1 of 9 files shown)


snapshot/policy/scheduling_policy.go
```diff
@@ -162,7 +165,11 @@ func (p *SchedulingPolicy) checkMissedSnapshot(now, previousSnapshotTime, nextSn
 
 	const halfhour = 30 * time.Minute
 
-	return (len(p.TimesOfDay) > 0 || len(p.Cron) > 0) && p.RunMissed && previousSnapshotTime.Add(oneDay-halfhour).Before(now) && nextSnapshotTime.After(now.Add(halfhour))
+	if !p.RunMissed.OrDefault(false) {
```
julio-lopez, line 168: nit: should this be the following for consistency?
That is, `p.RunMissed` will get the _default_ value when not set?

```suggestion
	if !p.RunMissed.OrDefault(defaultRunMissed) {
```

julio-lopez (commented):
+1: LG 👍

The ARM64 server scheduling test failed. @PhracturedBlue can you take a look please?

@jkowalski any concerns merging this one?

julio-lopez:
The ARM64 test job passed after re-triggering, ❄️ ?

julio-lopez:
Let's merge this and address the current review comment separately. Also, if Jarek has additional comments we can address those as well.

julio-lopez (approved):
See previous comment

snapshot/policy/scheduling_policy.go
PhracturedBlue: This was done on purpose for the #3330 PR feedback.  Basically by using `OrDefault(false)` we ensure that if a user is upgrading a repo that doesn't have --run-missed that it will never be enabled, regardless of whether the default is true or false.

If you want me to make the change, please confirm and I will, but I think there is a difference between the default for a newly created repo, and the default for an existing repo.

julio-lopez: I guess it depends on what we decide based on the discussion on #3330.

PR #3345, opened 2023-09-23
feat(cli): Initial secrets implementation

PhracturedBlue:
This is an implementation of a secrets framework to allow various ways to specify and store secrets.  It is eventually meant to implement #3312 when complete.

The implementation creates a Secret object which stores any secret.  it can be prefixed with various options to allow reading the secret from an env variable, cmdline, vault, etc.  Or the secret can be passed in directly.  If a secret is provided directly on the cmdline, it will be encrypted using a signing key and will be stored in the config file.  If the secret is provided indirectly (cmdline, vault, env-variable) it is not stored.

The signing key is a random 32-byte string, and is stored in the config file after encrypting it with a key derived from the repo password.  This makes it easy to change the repo password without needing to update any stored secrets.  ~~Currently the code is using fernet for the symmetric-key encryption because it was easy to get working, but likely this should reuse the same algorithms used for file encryption.~~ The code has been updated to share the Aes encryption code between the repo and the secrets.

~~The signing key is currently stored in a context, as I needed it to be globally available with the initial implementation.  However, I think I have mostly eliminated that need, so I am going to try to eliminate the Context stuff.~~
I have eliminated the need for the context, though there is still some code reformatting needed after the fact.

I am using reflection to walk the local-config struct to find Secrets.  This provides a clean interface to the rest of the code in that variables do not need to be individually decoded after the password is available, but I'm not sure it is the ideal solution.

I am posting to see if I can get any feedback on whether this is a reasonable solution, or if something else is desired.

At the moment, only sftp is supported.

Usage:
* --sftp-password <password>  --> password is stored encrypted in config file
* --sftp-password envvar:ENVIRONMENT_VARIABLE --> password will be read from $ENVIRONMENT_VARIABLE
* --sftp-password command:"command to execute" --> "command to execute" will be executed and password will be read from stdout
* --sftp-password file:<filename> --> password will be read from the contents of <filename>

This implementation does mean that there are some illegal password strings (anything starting with one of the keywords followed by a colon).  There may be alternatives if this is undesirable.

To Do:
- [x] Remove context manager
- [x] Fix existing tests 
- [x] Implement other repo types
  - [x] Implement Azure
  - [x] Implement B2
  - [x] Implement  GCS
  - [x] Implement GDrive
  - ~~Implement Rclone~~
  - [x] Implement S3
  - [x] Implement Webdav
- [x] Implement upgrading from current password-store
- [x] Implement keyring
- [x] Write tests
- [x] Improve code organization now that the manager is gone
- [x] Use a secret for the `--password` and `--server-password` cmdline argument
- [x] Figure out if we should remove the gnome-keyring options in favor of the Secret equivalents
- [x] Handle password change
- [x] Share the Key management between the repo and the secrets?


Scrumplex:
> This implementation does mean that there are some illegal password strings (anything starting with one of the keywords followed by a colon). There may be alternatives if this is undesirable.

What if you also add a `text:` or `plaintext:` prefix too?

PhracturedBlue:
> What if you also add a `text:` or `plaintext:` prefix too?
That is an option.  It changes the behavior of kopia, but it is not likely that users are crating/connecting new repos every day, so it probably wouldn't break user's workflows.  The one case I am concerned with is whether this can/should be extended for `--password`  While clearly we wouldn't want to store password in the config file, using a secret argument would allow using all the same methods of retrieval which could be nice.  If we want to do that, than I don't think I want to change the semantics since that would be a breaking change.

Scrumplex:
Oh I should clarify that `text:` / `plaintext:` should be in addition to just specifying a plaintext string

Examples:
- `password` -> `password`
- `envvar:SECRET_TOKEN` -> value of `$SECRET_TOKEN`
- `file:/run/secrets/token` -> contents of `/run/secrets/token`
- `plaintext:file:123` -> `file:123`

PhracturedBlue:
> Oh I should clarify that `text:` / `plaintext:` should be in addition to just specifying a plaintext string

Yeah, I realized that is what you meant after I thought about it some more.  I implemented exactly that, and converted `--password` and `--server-password` to use secrets to ensure it worked as intended

jkowalski:
this is looking great already, may I suggest separating non-functional changes (code movement and renames) into separate PR that can be trivially merged so that this one becomes smaller and easier to review?

internal/secrets/secrets.go
```diff
@@ -0,0 +1,219 @@
...
+	err := s.Set(value)
+	if err != nil {
+		return nil
+	}
+
+	return &s
+}
+
+// Set will set the Secret type.
+func (s *Secret) Set(value string) error {
```
jkowalski, line 50: can we follow the pattern we have elsewhere (encryption, hash, compression, blob storage, etc.) where there's package-wide registry of supported algorithms and a way to register one on startup

The use case here is to allow folks extending Kopia to register completely new secret storage mechanisms that should just work.

internal/secrets/secrets.go
```diff
@@ -0,0 +1,219 @@
...
+	switch {
+	case strings.HasPrefix(value, "envvar:"):
```
jkowalski, line 52: repeating the same string twice is usually a recipe for copy/paste mistakes, better to user a map/slice and iterate it

internal/secrets/secrets.go
```diff
@@ -0,0 +1,219 @@
...
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/alecthomas/kingpin/v2"
+)
+
+type keyType int
```
jkowalski, line 15: probably better for this to be a string, makes it easier to extend

internal/secrets/secrets.go
```diff
@@ -0,0 +1,219 @@
...
+import (
+	"encoding/hex"
+	"encoding/json"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/alecthomas/kingpin/v2"
```
jkowalski, line 12: dependency on kingpin should be avoided as much as possible from low-level packages like this, kingpin is really only meant to be used from cli and testing packages

internal/secrets/secrets.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	s.StoreValue = d
+	s.Type = Config
+
+	return nil
+}
+
+// SecretVar is called by kingpin to handle Secret arguments.
+func SecretVar(s kingpin.Settings, target **Secret) {
```
jkowalski, line 171: I think this method can be just moved to `cli` (and made unexported) as can the next one.

internal/secrets/secrets.go
PhracturedBlue: I prefer the switch, but I think I can kill 2 birds with 1 stone by defining the consts as the prefix, and using them here.

PhracturedBlue:
@jkowalski I implemented the changes based on your code review.  I will start cleaning up the broken tests and implementing tests for the new functionality next.

PhracturedBlue:
One thing I noticed is that because we call `DeriveKeyFromPassword` twice (once for secrets, and once for the repo) and this function is designed to be slow, it adds several seconds to kopia startup time.  Currently the salt is different for both cases, so the result is different, and they can't be shared.
I don't have access to the repo when decrypting secrets, so if we wanted to share the salt, we'd need to store it it the local config, which I don't think is ideal..  Not sure how to proceed.

internal/secrets/secrets.go
jkowalski: what I meant was exposing something like `secrets.Register(prefix, impl)` so you can plug in various implementations:

```
secrets.Register("command", &commandProvider{})
secrets.Register("envvar", &envvarProvider{})
secrets.Register("vault", &vaultProvider{})
secrets.Register("keyring", &keyringProvider{})
```

cli/storage_s3.go
```diff
@@ -24,8 +24,8 @@ func (c *storageS3Flags) Setup(svc StorageProviderServices, cmd *kingpin.CmdClau
 	cmd.Flag("endpoint", "Endpoint to use").Default("s3.amazonaws.com").StringVar(&c.s3options.Endpoint)
 	cmd.Flag("region", "S3 Region").Default("").StringVar(&c.s3options.Region)
 	cmd.Flag("access-key", "Access key ID (overrides AWS_ACCESS_KEY_ID environment variable)").Required().Envar(svc.EnvName("AWS_ACCESS_KEY_ID")).StringVar(&c.s3options.AccessKeyID)
-	cmd.Flag("secret-access-key", "Secret access key (overrides AWS_SECRET_ACCESS_KEY environment variable)").Required().Envar(svc.EnvName("AWS_SECRET_ACCESS_KEY")).StringVar(&c.s3options.SecretAccessKey)
-	cmd.Flag("session-token", "Session token (overrides AWS_SESSION_TOKEN environment variable)").Envar(svc.EnvName("AWS_SESSION_TOKEN")).StringVar(&c.s3options.SessionToken)
+	secretVarWithEnv(cmd.Flag("secret-access-key", "Secret access key (overrides AWS_SECRET_ACCESS_KEY environment variable)"), svc.EnvName("AWS_SECRET_ACCESS_KEY"), &c.s3options.SecretAccessKey)
```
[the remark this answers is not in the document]
ghaushe-ampere: yes, i added the check below instead.  I don't see a good way to support the required field along with using a custom flag.  I can try to rethink how the env variable case is handled, as this is the primary reason required doesn't work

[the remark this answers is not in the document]
PhracturedBlue: Yes, I added the check below instead. I don't see a good way to support the required field along with using a custom flag. I can try to rethink how the env variable case is handled, as this is the primary reason required doesn't work

jkowalski:
I'm not sure I understand the need to have yet another encryption scheme for embedding encryopted literals in a config file when we have external secret storage systems for fetching those dynamically. 

I was hoping we can simplify the solution by separating 2 concerns - one being access to all these runtime providers of secrets and the other - the problem of storing localconfig unencrypted.

1. I was hoping `secrets.Secret` could simply be a subtype of string (not encrypted at all), sometimes prefixed with a `source:` 

Wen can then have something like:

```
type Provider interface {
  Fetch(ctx context.Context, param string) (string, error)
}

var Providers = map[string]Provider{}

func (s Secret) Evaluate(ctx context.Context) (string, error) {
   for prefix, impl := range Providers {
     if strings.HasPrefix(string(s), prefix + ":") {
       return impl.Fetch(ctx, s[len(prefix)+1:])
    }
  }

   return string(s), nil
}
```

This would check if the string starts with any of supported providers and invioke the appropriate provider to get the value, or return the literal value if no prefix matches.

2. The above would be dramatically simpler at the expense of sometimes having raw secrets in the config file, but this could be handled in a completely orthogonal way, namely by encrypting the entire `LocalConfig` prior to storing it. For encrypting local config we can derive the whole-file encryption key from password using some random salt and encryption parameters that will be embedded in JSON in addition to ciphertext. This would be similar to `KopiaRepositoryJSON.EncryptedFormatBytes`.

PhracturedBlue:
That's the kind of feedback that I'd have preferred before doing a full implementation.  I don't have the time to rework this completly right now.  I'll just leave it until I have time to come back to it.

jkowalski:
> That's the kind of feedback that I'd have preferred before doing a full implementation. I don't have the time to rework this completly right now. I'll just leave it until I have time to come back to it.

Yeah, sorry about that. I've been traveling the whole last week and did not have a chance to look into that properly. I think 85% of this PR is still exactly what we need, we just need to simplify the secrets package and perhaps remove a few things.

linust:
having proper support for secrets would make for a more secure and clean docker configuration. How much is missing to get this completed? (Besides a rebase from hell? ;-)

Issue #3344, opened 2023-09-22
Bug? - Manual Snapshot Only Setting Not Overriding Global Policy


apb2019:
My global policy for a repository has "Manual Snapshots Only" set to "yes". However for a particular folder I want to run regular 10min snapshots so during the snapshot setup I set the "Snapshot Frequency" to "every 10 Minutes" and set "Manual Snapshots Only" to "yes" instead of "inherit from parent" then save the policy but when I reopen the policy, the "Manual snapshots Only" setting has been reset back to "inherit from parent" and the blue tick mark next to defined by Global policy is still set.
Is the a bug?

apb2019:
Should have mentioned that when I set the local policy "Snapshot Frequency" to "every 10 Minutes" there was no error message  ("invalid scheduling policy: manual cannot be combined with other scheduling policies") that I would have expected due to the scheduling being a conflict with the Manual Snapshots Only=yes in the global policy.
When the local policy is reopened the "Snapshot Frequency" setting is still showing "every 10 Minutes" though no "upcoming snapshots" are listed.

PhracturedBlue:
How did you set 'manual policy only' for the global policy?
When I tried to reproduce your case, I get:
```
kopia policy set --global --manual
Setting policy for (global)
 - setting manual snapshot field to true

ERROR can't save policy for (global): failed to validate policy: invalid scheduling policy: invalid scheduling policy: manual cannot be combined with other scheduling policies
```

Also, what version of Kopia are you running?

apb2019:
Policy was set via UI. I get a similar message if the snapshot frequency for the global policy is set to something other than the default (none) and then try to set manual snapshots to yes. Originally saw this in v13 but  just recreated the problem in V14.1 as follows
Using UI (14.1)
Create new local repository without changing any of the defaults.
edit/save global policy to change "Manual snapshots only" to yes - leaving all other options unchanged ie snapshot frequency = (none) etc
Create new snapshot - changing Snapshot frequency from (none) to "every 10 mins" and Manual snapshots only from "inherit from parent" to "no"
take snapshot
If I the open policy for snapshot just taken. I see Snapshot frequency still says "every 10 Min" but Manual snapshot only has reverted back to "no" and no upcoming snapshots are listed
Hope this helps

PhracturedBlue:
I was able to reproduce this on v0.14.1.  But apparently I introduced an issue in the changes to --run-missed that makes it impossible to set 'run manually' to true on --global.

[...]

PR #3349, opened 2023-09-24
fix(cli): Convert policy set --manual to use a tri-state

PhracturedBlue:
As per issue #3344, If Manual is set to true in a more-general policy, it cannot be reset to false in a lower policy.  This is due to manual being a bool, and there being no way to distinguish between 'inherit' and 'false'

This PR converts Manual into an OptionalBool so that it can hold values of true, false, and inherit, and can thus override both from false->true and from true->false

There is an included test to validate as well.

This PR includes #3348 as inheritance cannot be properly tested until that issue is addressed.  I recommend committing #3348 1st, and then this PR

### DIFF
--- a/cli/command_policy_set_scheduling.go
+++ b/cli/command_policy_set_scheduling.go
@@ -15,7 +15,7 @@ type policySchedulingFlags struct {
 	policySetInterval   []time.Duration // not a list, just optional duration
 	policySetTimesOfDay []string
 	policySetCron       string
-	policySetManual     bool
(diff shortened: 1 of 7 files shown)


lupusA:
This PR will close #3344

ImBearChild:
This is a friendly ping. I believe this fix is very useful for users who only want to take scheduled snapshots for certain paths. It seems that this PR has passed all tests. Could it be merged now?

PR #3358, opened 2023-09-26
(docs) Add 'How do I automate backups?' to FAQ

jmbrunskill:
Adds an FAQ section on how to automate backups based on discourse answer here:
https://kopia.discourse.group/t/basics-questions-thanks-for-your-time/258/5

A colleague spent a lot of time trying to figure out how to automate kopia backups, which makes me think adding some basic docs & explaination is useful.
The design is a little confusing because policies can setup scheduled, but they don't run without the server running or a cron job.

I've put in very minimal examples cron jobs and systemd service, neither of which have had much testing.

Perhaps it's better to explain the socket based systemd approach mentioned here?
https://kopia.io/docs/repository-server/#kopia-with-systemd
I don't really understand it though, so would need some input on that.

This docs current focus on linux, but I think it would be helpful to have similar docs for windows, and mac.

### DIFF
--- a/site/content/docs/FAQs/_index.md
+++ b/site/content/docs/FAQs/_index.md
@@ -129,3 +129,75 @@ For more information on the `checkpoint interval`, please refer to the [command-
 #### What is a Kopia Repository Server?
 
 See the [Kopia Repository Server help docs](../repository-server) for more information.
+
+#### How do I automate backups?
+
+There are two main pathways to automated backups with kopia:
+
+1. Run kopia cli with a scheduling tool such as cron.
+1. Start kopia server, and use a policy to schedule backups.
+
+##### Running kopia cli with cron
+
+Most linux based systems have [cron](https://en.wikipedia.org/wiki/Cron) installed.
+This can be used to trigger kopia cli commands on a schedule.
+A simple backup script could look something like this:
+
+```bash
+#!/bin/bash
+# This script will backup the home directory for myusername
+kopia snapshot create /home/myusername
+```
+
+You'll need to make this script executable:
+
+```bash
+chmod +x /path/to/backup_script.sh
+```
+
+To schedule this script to run every day at 3am, you could add the following line to your crontab:
+
+```bash
+0 3 * * * /path/to/backup_script.sh
+```
+
+You may want to add logging to your script to capture any errors that occur or set it up notify you if a backup fails, etc.
+
+##### Running kopia server as a linux service
+
+If you are running kopia server, it will attempt to backup your files based on configured policies.
+
+Kopia server can be started with the following command:
+
+```bash
+kopia server start
+```
+
+> Note: (if you haven't setup SSL, you may need to use `kopia server start --insecure`)
+
+On linux, you can use [systemd](https://en.wikipedia.org/wiki/Systemd) to create and manage the kopia server service.
+
+Create a service file at `/etc/systemd/system/kopia.service` something like this:
+
+```bash
+[Unit]
+Description=Kopia Backup Service.
+
+[Service]
+ExecStart=/usr/bin/kopia server start --insecure
+Environment=HOME=/root
+
+[Install]
+WantedBy=multi-user.target
+```
+
+> Warning: This will run kopia as the root user and without ssl certificates, which may be a security risk.
+> You may want to create a dedicated user for kopia to run as, and setup ssl certificates.
+
+After creating the service file, you can start the service and set it to run on startup using the following commands:
+
+```bash
+systemctl daemon-reload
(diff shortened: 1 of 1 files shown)


basldfalksjdf:
A few thoughts on this:

(1) IMO, if a guide or tip or whatever is fairly long, such as the proposed changed, it should be on it sown page rather than in the FAQs. We can create a FAQ for it but then use it to link to the new page rather than within the FAQ. Otherwise, the FAQ will get way too long

(2) The FAQ is incomplete as it does not discuss KopiaUI, which you don't need to do anything to automate snapshots

(3) I don't use Kopia CLI, but I thought that too allowed you to automate backups by setting up policies? Without the need to use cron?

jmbrunskill:
Many thanks for your feedback!

> (1) IMO, if a guide or tip or whatever is fairly long, such as the proposed changed, it should be on its own page rather than in the FAQs. We can create a FAQ for it but then use the FAQ to link to the new page rather than within the FAQ. Otherwise, the FAQ will get way too long

Fair enough, I agree, do you have a suggestion on where it would go?
Maybe we leave the FAQ (How do I automate backups) and add a cron example under the CLI section here?
https://kopia.io/docs/getting-started/

> (2) The proposed change is incomplete as it does not discuss KopiaUI, which you don't need to do anything to automate snapshots

If I understand correctly, KopiaUI runs `kopia server` which (if running) can automate snapshots.
I haven't done enough with it to really understand the ins and outs of it, but pretty confident it can be run without actually serving the UI component.

> (3) I don't use Kopia CLI, but I thought that too allowed you to automate backups by setting up policies? Without the need to use cron?

This is exactly the kind of confusion I would like to see cleared up and clarified in the docs, and I'm very happy to be pointed at the information about how to do it if so.

Kopia docs are unclear as it seems to imply automation that doesn't come out of the box if you install on linux at least.

From what I can tell you have to run kopia server, which is what I tried to suggest using `systemd`.
I'm guessing a lot of linuxy people are probably going to use cron though, so maybe it's just a matter of pointing to the idea rather than giving two options?

basldfalksjdf:
@jkowalski can clarify about whether Kopia CLI explicitly needs to be set to run via cron or setting a policy is enough to automate backups. Im pretty sure setting a policy is enough, but as I said I don't use Kopia CLI.

If setting the policy is enough, then we don't really need an extra page for this. We could create a FAQ and do a short one paragraph writeup in that FAQ that says something like "Kopia automatically runs backups based on the policies that you create." The getting started guide already has information on setting up policies.

If you do need to use cron to automate, then I think a dedicated page is the way to go and then just link to that page in the FAQ.

RayZ0rr:
> @jkowalski can clarify about whether Kopia CLI explicitly needs to be set to run via cron or setting a policy is enough to automate backups. Im pretty sure setting a policy is enough, but as I said I don't use Kopia CLI.

Kopia CLI doesn't imply any sort of automatic backup. I can clearly see the confusion that @jmbrunskill raises as I had the same confusion when setting up automatic kopia backups in my cronjobs.

Maybe there is an easier way by running `kopia server start` or something but a FAQ or docs page like this PR suggests is very helpful.

I now use
```bash
export KOPIA_CONFIG_PATH="${HOME}/path/to/config"
# need to securely provide password:
export KOPIA_PASSWORD="some-password"

# snip

${KOPIA_CMD} snapshot create --tags="${BACKUP_TAGS}" "${BACKUP_TARGET}" | tee -a "${logFile}"
```

jkowalski:
note that passing kopia password is generally only needed on `repository connect`. The password is then stored in keystore/keychain or in a local file if those are unavailable.

The exception is when you pass `--no-persist-credentials` on `connect`, in such case you need to provide the password each time.

Kumen (approved):
A good starting point.

fgimian:
This is a very useful addition to the docs as it's an area I too have found confusing. I think having any sort of scheduling configuration via the CLI implies that the CLI itself can perform this task, even though this isn't possible since it's not a daemon.

Is it also worth mentioning `kopia snapshot create --all` which will automatically snapshot all directories which have previously had a snapshot taken in the repository?

To me it seems the simplest way would be that you first manually perform these steps:

1. Create your repository
2. Configure your policies as required
3. Perform an initial snapshot of all directories you want to include in the repository

Then you can simply configure your cron tool of choice and just run `kopia snapshot create --all`.

One aspect which I haven't looked into yet is how we can ensure that two snapshots are not running at once. In restic, I would attempt to first run `unlock` and then check if any locks remain. If so, I worked on the assumption that another backup job was already running, and thus would skip the run with an error.

Cheers
Fotis

PR #3364, opened 2023-09-28
refactor(snapshots): Minor reorder of defer funcs for streaming files

ashmrtn:
Reorder streaming file upload defer functions slightly so the reader is closed prior to calling FinishedFile. This allows folks who hook into those callbacks to make stronger assumptions about the ordering/state of things if needed.

### DIFF
--- a/snapshot/snapshotfs/upload.go
+++ b/snapshot/snapshotfs/upload.go
@@ -342,13 +342,12 @@ func (u *Uploader) uploadStreamingFileInternal(ctx context.Context, relativePath
 		return nil, errors.Wrap(err, "unable to get streaming file reader")
 	}
 
-	defer reader.Close() //nolint:errcheck
-
 	var streamSize int64
 
 	u.Progress.HashingFile(relativePath)
 
 	defer func() {
+		reader.Close() //nolint:errcheck
 		u.Progress.FinishedHashingFile(relativePath, streamSize)
 		u.Progress.FinishedFile(relativePath, ret)
 	}()


jkowalski (approved):

snapshot/snapshotfs/upload.go
```diff
@@ -342,13 +342,12 @@ func (u *Uploader) uploadStreamingFileInternal(ctx context.Context, relativePath
...
 	}
 
-	defer reader.Close() //nolint:errcheck
-
 	var streamSize int64
 
 	u.Progress.HashingFile(relativePath)
 
 	defer func() {
+		reader.Close() //nolint:errcheck
```
julio-lopez, line 350: @ashmrtn @jkowalski I am not sure I understand the effect of this change nor the motivation.

- This changes the guarantees about closing the reader. While uncommon / unexepected, if `HashingFile` were to `panic`, then `reader.Close()` would not be called. Granted, kopia does not catch the panics, so if kopia panics here it would terminate the process. Nevertheless, this changes the guarantees.
- In either case, (before or after this change) the ordering in which these functions are called is stable.
- With this change, the order for `reader.Close()` has been changed. Is that reordering needed? what is the use case?

Before the progress handler deferred call would be called first and then `reader.Close()`. With this change `reader.Close()` will be called before the progress handlers deferred calls are called. Is this guarantee needed? It seems that deferring `reader.Close()` is needed for "data" correctness and avoiding leaking resources. Whereas the deferred calls for the progress handlers are needed to ensure the progress metrics and indicators are updated. Why is a particular ordering between these calls needed?

ashmrtn: The main goal of the change is to enforce ordering between the `Close()` call and the `FinishedFile()` call. Specifically, we want to ensure that no other data from the file will be read when `FinishedFile()` is called and that any cleanup tasks done in `Close()` have already completed when `FinishedFile()` is called. This makes it easier to reason about state if someone overrides/wraps `FinishedFile()` or other progress handlers. It's also more inline with the call ordering of other file types (i.e. `Close()` then progress handlers)

PR #3365, opened 2023-10-01
refactor(repository): added fs.DirectoryIterator

jkowalski:
This significantly reduces number of small allocations while taking snapshots of lots of files, which leads to faster snapshots.

```
$ runbench --kopia-exe ~/go/bin/kopia \
   --compare-to-exe ~/go/bin/kopia-baseline --min-duration 30s \
   ./snapshot-linux-parallel-4.sh
DIFF duration: current:5.1 baseline:5.8 change:-13.0 %
DIFF repo_size: current:1081614127.6 baseline:1081615302.8 change:-0.0 %
DIFF num_files: current:60.0 baseline:60.0 change:0%
DIFF avg_heap_objects: current:4802666.0 baseline:4905741.8 change:-2.1 %
DIFF avg_heap_bytes: current:737397275.2 baseline:715263289.6 change:+3.1 %
DIFF avg_ram: current:215.0 baseline:211.5 change:+1.6 %
DIFF max_ram: current:294.8 baseline:311.4 change:-5.3 %
DIFF avg_cpu: current:167.3 baseline:145.3 change:+15.1 %
DIFF max_cpu: current:227.2 baseline:251.0 change:-9.5 %
```


ashmrtn (commented):
Generally looks good. A few things that stand out that could use some more discussion prior to merging I think in addition to some of the more open ended comments/questions in the review:
* are there going to be scalability concerns when it comes to large directories or source snapshots? In some places we seem to have moved from a pattern that materializes only a single directory entry/source snapshot and runs a callback to a pattern that materializes _all_ directory entries or source snapshots and returns a static iterator
* do we want "sticky" errors at the end of iteration? Some implementations may provide this already (i.e. `staticIterator`) but I'm not sure if that's intentional or just how the code worked out

jkowalski:
> * are there going to be scalability concerns when it comes to large directories or source snapshots? In some places we seem to have moved from a pattern that materializes only a single directory entry/source snapshot and runs a callback to a pattern that materializes _all_ directory entries or source snapshots and returns a static iterator

This is mostly for simplificity, the scenarios are really about filesystem mount where I think it expected that the user is mostly interactive and will no be simultaneously scanning many large directories. We may optimize for this case later if needed.

> * do we want "sticky" errors at the end of iteration? Some implementations may provide this already (i.e. `staticIterator`) but I'm not sure if that's intentional or just how the code worked out

I clarified end-of-iteration behavior in the comments. Basically iteration should not be continued after an error. By this definition StaticIterator returns its error once (if not nil) or never.

fs/entry.go
```diff
@@ -59,13 +59,43 @@ type StreamingFile interface {
...
+
+		cur, err = iter.Next(ctx)
+	}
+
+	return err //nolint:wrapcheck
+}
+
+// DirectoryIterator iterates entries in a directory.
+type DirectoryIterator interface {
+	Next(ctx context.Context) (Entry, error) // returns nil to signal end of iteration
```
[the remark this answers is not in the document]
jkowalski: Iteration errors are expected to terminate the iteration loop as clarified in the comment. 

I think this leaves some room for creativity for the implementations of the iterator:

* for example they can return `fs.ErrorEntry` with err=nil to signal that unreadable entry was found, which will cause uploader to report an error without stopping the upload, 

* alternatively the implementation can skip over invalid entries (what local_fs.go is doing to resolve NOT_FOUND race) or 

* buffer any errors and return them together as the last iteration result from Next().

internal/webdavmount/webdavmount.go
```diff
@@ -111,29 +111,38 @@ var symlinksAreUnsupportedLogged = new(int32)
...
+	iter, err := d.entry.Iterate(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "error reading directory")
+	}
+
+	defer iter.Close()
+
+	e, err := iter.Next(ctx)
+	for e != nil {
+		if n > 0 && foundEntries >= n {
```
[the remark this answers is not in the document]
jkowalski: thanks for spotting this, I realized this function had a pre-existing bug which can now be fixed with the iterator pattern.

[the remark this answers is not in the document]
jkowalski: also, n<=0 is not going to be a problem, it will return all entries, n=1 will return one

cli/command_ls.go
```diff
@@ -54,9 +54,22 @@ func (c *commandList) run(ctx context.Context, rep repo.Repository) error {
...
-		return c.printDirectoryEntry(innerCtx, e, prefix, indent)
-	}); err != nil {
+	iter, err := d.Iterate(ctx)
+	if err != nil {
+		return err //nolint:wrapcheck
+	}
+	defer iter.Close()
+
+	e, err := iter.Next(ctx)
+	for e != nil {
```
julio-lopez, line 64: should err be explicitly checked here?

```suggestion
	for err == nil && e != nil {
```

fs/entry.go
```diff
@@ -59,13 +59,50 @@ type StreamingFile interface {
...
+	iter, err := dir.Iterate(ctx)
+	if err != nil {
+		return err //nolint:wrapcheck
+	}
+
+	defer iter.Close()
+
+	cur, err := iter.Next(ctx)
+
+	for cur != nil {
```
julio-lopez, line 82: explicitly check for `err == nil` ?

fs/entry.go
```diff
@@ -59,13 +59,50 @@ type StreamingFile interface {
...
+
+	return err //nolint:wrapcheck
+}
+
+// DirectoryIterator iterates entries in a directory.
+//
+// The client is expected to call Next() in a loop until it returns a nil entry to signal
+// end of iteration or until an error has occurred.
+//
+// The behavior of calling Nex() after iteration has signaled its end is undefined.
```
julio-lopez, line 98: Should it return an error?

fs/entry_dir_iterator.go
```diff
@@ -0,0 +1,29 @@
...
+		it.cur++
+
+		return v, it.err
+	}
+
+	return nil, nil
+}
+
+// StaticIterator returns a DirectoryIterator which returns the provided
+// entries in order followed by a given final error.
```
julio-lopez, line 26:
```suggestion
// entries in order followed by a given final error.
// It is not safe to concurrently access directory iterator.
```

julio-lopez (commented):
Only a few minor comments / questions.

Relying on @ashmrtn for the review here. Thanks Ashlie.

fs/entry.go
```diff
@@ -91,8 +91,15 @@ func IterateEntries(ctx context.Context, dir Directory, cb func(context.Context,
 }
 
 // DirectoryIterator iterates entries in a directory.
+//
+// The client is expected to call Next() in a loop until it returns a nil entry to signal
+// end of iteration or until an error has occurred.
+//
+// The behavior of calling Nex() after iteration has signaled its end is undefined.
```
ashmrtn, line 98: nit: spelling fix
```suggestion
// The behavior of calling Next() after iteration has signaled its end is undefined.
```

fs/localfs/local_fs_os.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		return nil, errors.Wrap(err, "unable to get child")
+	}
+
+	return entryFromDirEntry(st, fullPath+string(filepath.Separator)), nil
+}
+
+func toDirEntryOrNil(dirEntry os.DirEntry, prefix string) (fs.Entry, error) {
+	fi, err := os.Lstat(prefix + dirEntry.Name())
```
ashmrtn, line 93: nit: seems like this should be using `filepath.Join`?

fs/entry.go
```diff
@@ -91,8 +91,15 @@ func IterateEntries(ctx context.Context, dir Directory, cb func(context.Context,
 }
 
 // DirectoryIterator iterates entries in a directory.
+//
+// The client is expected to call Next() in a loop until it returns a nil entry to signal
+// end of iteration or until an error has occurred.
```
ashmrtn, line 96: Do we also want to clarify that code expects either (non-nil entry, error, end of iterator) since much of the code loops on `e := Next(); e != nil` without checking the error value. I'm not really sure under what conditions you could end up with a non-nil entry and error, but it could cause problems if that happens

ashmrtn (approved):
Beyond a few nits and open questions from others I think things look good

fs/localfs/local_fs_os.go
jkowalski: prefix already includes trailing "/" or "\" so this is not needed here. This whole business of managing prefix needs to be reworked in separate PR (originally it was to save on memory allocation, but is not really working that well. I have some idea to dramatically improve it).

fs/entry.go
jkowalski: not requiring that makes implementations of iterator simpler.

fs/entry.go
jkowalski: Added comment to clarify valid results:

```
(nil,nil) - end of iteration, success
(entry,nil) - iteration in progress, success
(nil,err) - iteration stopped, failure
```

cli/command_ls.go
jkowalski: `e != nil` implies `err == nil` per comment on `Next()`, additional check would be redundant.

fs/entry.go
jkowalski: `e != nil` implies `err == nil` per comment on `Next()`, additional check would be redundant.

fs/entry.go
julio-lopez: Can we explicitly add this to the function documentation?

PR #3394, opened 2023-10-20
feat(repository): Add support for Azure `DeleteBlob` operations where immutability protection is on

KastenMike:
Currently kopia does not support Azure where blobs are protected by something like a policy. A session file might be prevented by a 5 day protection window but it wants to be deleted immediately. This gets around that issue e.g.

image

image

Other related Azure changes in a PR train:
- [feat(repository): Add ExtendBlobRetention support for Azure](https://github.com/KastenMike/kopia/pull/1)
- [feat(repository): Add support for Azure immutability protection](https://github.com/KastenMike/kopia/pull/2)
- [feat(repository): Add Azure PIT support](https://github.com/KastenMike/kopia/pull/3)
- [docs(site): Add Azure ransomware protection section](https://github.com/kopia/kopia/pull/3398)


repo/blob/storage.go
```diff
@@ -21,6 +21,9 @@ var ErrSetTimeUnsupported = errors.Errorf("SetTime is not supported")
 // ErrInvalidRange is returned when the requested blob offset or length is invalid.
 var ErrInvalidRange = errors.Errorf("invalid blob offset or length")
 
+// ErrBlobImmutableDueToPolicy is returned when a policy prevented an action on an Azure blob.
+var ErrBlobImmutableDueToPolicy = errors.Errorf("blob immutable due to policy")
```
jkowalski, line 25: i don't think we need to add another error here unless it's expected that callers will need to deal with it somehow.

KastenMike: It's just to make `DeleteBlob` cleaner which uses it. Otherwise we'd need to repeat the content of `translateError` inside `DeleteBlob` again 

e.g. inside `DeleteBlob`, instead of just `err = translateError(err)` and the switch with a new case it'd need to be something like:

```
err = translateError(err)

if errors.Is(err, blob.ErrBlobNotFound) {
    // don't return error if blob is already deleted
    return nil
}

var re *azcore.ResponseError

if errors.As(err, &re) && re.ErrorCode == bloberror.BlobImmutableDueToPolicy {
    return az.retryDeleteBlob(ctx, b)
}

```

KastenMike: updated

repo/blob/azure/azure_storage.go
```diff
@@ -226,6 +216,85 @@ func (az *azStorage) DisplayName() string {
...
+		NewContainerClient(az.container).
+		NewBlobClient(az.getObjectNameString(b)).
+		WithVersionID(*resp.VersionID)
+	if err != nil {
+		return errors.Wrap(err, "failed to get versioned blob client")
+	}
+
+	_, err = bc.Delete(ctx, nil)
+	if err != nil {
+		return errors.Wrap(err, "failed to delete the delete marker blob version")
```
julio-lopez, line 292: Same here, if the 0-byte blob version could not be deleted, it seems OK to not fail the blob delete operation. Removing the specific blob version could happen in the background via a 3rd party process such as lifecycle policy in the blob store.

```suggestion
		log(ctx).Info("failed to delete the delete marker blob version", err)
```

julio-lopez (commented):
@KastenMike See inline comments. Thanks.

repo/blob/storage.go
julio-lopez: At this point this error seems to be specific to Azure.

julio-lopez:
Running provider tests as well. https://github.com/kopia/kopia/actions/runs/6636153104/job/18028167562?pr=3394

repo/blob/azure/azure_storage.go
```diff
@@ -226,6 +216,85 @@ func (az *azStorage) DisplayName() string {
...
+
+	_, err = az.service.DeleteBlob(ctx, az.container, az.getObjectNameString(b), nil)
+	if err != nil {
+		return errors.Wrap(err, "failed to soft delete blob")
+	}
+
+	bc, err := az.service.ServiceClient().
+		NewContainerClient(az.container).
+		NewBlobClient(az.getObjectNameString(b)).
+		WithVersionID(*resp.VersionID)
```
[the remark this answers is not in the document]
KastenMike: Since this path is only reached for those with versioning + immutability support enabled the `VersionID` can assumed to be populated like how things like `*it.Properties.ContentLength` are also not checked for...but will add some checks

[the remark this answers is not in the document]
julio-lopez: Understood.
Nevertheless,
- a client should always validate the response and never trust that the server is sending a valid response;
- similarly, a server should never trust that a client is sending a correct request.

repo/blob/azure/azure_storage.go
```diff
@@ -226,6 +216,85 @@ func (az *azStorage) DisplayName() string {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to get versioned blob client")
```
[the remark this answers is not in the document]
julio-lopez: The odds of the error occurring seem low, however at a large enough scale, the error will certainly happen. Silently ignoring the error makes it really hard to debug when the error does occur. We can add the log message. This is unlikely to become noisy, but if it does, we can change the level to Debug, or we can consider other alternatives such as adding a counter for "handled" errors, or some other mechanism.

repo/blob/azure/azure_storage.go
```diff
@@ -258,38 +256,43 @@ func (az *azStorage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes, op
...
 	}
 
-	_, err = az.service.DeleteBlob(ctx, az.container, az.getObjectNameString(b), nil)
+	_, err = az.service.DeleteBlob(ctx, az.container, blobName, nil)
 	if err != nil {
 		return errors.Wrap(err, "failed to soft delete blob")
 	}
 
+	if resp.VersionID == nil || *resp.VersionID == "" {
+		// shouldn't happen
```
julio-lopez, line 282: log it? it shouldn't happen, that does not mean it won't happen.

julio-lopez (approved):

repo/blob/azure/azure_storage.go
```diff
@@ -226,6 +214,90 @@ func (az *azStorage) DisplayName() string {
...
+
+	if resp.VersionID == nil || *resp.VersionID == "" {
+		// shouldn't happen
+		return nil
+	}
+	bc, err := az.service.ServiceClient().
+		NewContainerClient(az.container).
+		NewBlobClient(blobName).
+		WithVersionID(*resp.VersionID)
+	if err != nil {
```
julio-lopez, line 289: We should log here and below. instead of silently swallowing the errors.
We can merge this as is, and add the logging statements separately.

repo/blob/azure/azure_storage.go
KastenMike: Added a log although I'm not sure if they're readable in some way (e.g. if the call stack higher up already has "maintenance" log

repo/blob/azure/azure_storage.go
```diff
@@ -278,20 +280,26 @@ func (az *azStorage) retryDeleteBlob(ctx context.Context, b blob.ID) error {
 		return errors.Wrap(err, "failed to soft delete blob")
 	}
 
+	log := logging.Module("azure-immutability")
```
julio-lopez, line 283: This works, however the pattern elsewhere is to have a package-level variable.

repo/blob/azure/azure_storage.go
```diff
@@ -278,20 +280,26 @@ func (az *azStorage) retryDeleteBlob(ctx context.Context, b blob.ID) error {
...
+		log(ctx).Info("VersionID not returned, exiting without deleting the delete marker version")
 		return nil
 	}
+
 	bc, err := az.service.ServiceClient().
 		NewContainerClient(az.container).
 		NewBlobClient(blobName).
 		WithVersionID(*resp.VersionID)
 	if err != nil {
+		log(ctx).Infof("Issue preparing versioned blob client: %v", err)
```
julio-lopez, line 296: isn't this going to require a blank line before the `return` statement?
Same below.

julio-lopez: Apparently not.

julio-lopez:
Core provider tests => https://github.com/kopia/kopia/actions/runs/6646559684/job/18060262479?pr=3394

julio-lopez:
Thanks @KastenMike

PR #3405, opened 2023-10-25
feat(providers): Add `ExtendBlobRetention` support for Azure

KastenMike:
Adds `ExtendBlobRetention` code for Azure and fixes a couple typos I saw.


repo/blob/s3/s3_storage_test.go
```diff
@@ -119,7 +119,7 @@ func getProviderOptions(tb testing.TB, envName string) *Options {
 	}
 
 	if o.Prefix != "" {
-		tb.Fatalf("options providd in '%v' must not specify a prefix", envName)
+		tb.Fatalf("options provided in '%v' must not specify a prefix", envName)
```
KastenMike, line 122: Note: tb.Fatalf doesn't actually make the test fail. For that it should be followed by a tb.FailNow()

repo/blob/storage.go
```diff
@@ -107,6 +108,9 @@ const (
 
 	// Compliance - compliance mode.
 	Compliance RetentionMode = "COMPLIANCE"
+
+	// Locked - Locked policy mode for Azure.
```
julio-lopez, line 112: We should include _"unlocked"_ as well.

julio-lopez (approved):
This PR should go in with the one that depends on it as well, since that one has the tests :)

Merging this one now to move forward ;)

@KastenMike For future reference, it often makes sense to submit small trivial changes, such as typo corrections in their own PR. The reasons being - those are trivial to review and thus can be easily merged right away;
- and then they reduce noise in the PRs that have functional changes, so those become easier to review.

julio-lopez:
@KastenMike also, thanks again for breaking things into smaller pieces to facilitate reviewing.

julio-lopez:
@redgoat650 FYI

PR #3407, opened 2023-10-26
feat(repository): Add Azure PIT support

KastenMike:
Adds support for Azure PIT


julio-lopez:
@KastenMike see initial comments

repo/blob/azure/azure_storage.go
```diff
@@ -56,7 +62,15 @@ func (az *azStorage) GetBlob(ctx context.Context, b blob.ID, offset, length int6
...
 
-	resp, err := az.service.DownloadStream(ctx, az.container, az.getObjectNameString(b), opt)
+	bc, err := az.service.ServiceClient().
+		NewContainerClient(az.container).
+		NewBlobClient(az.getObjectNameString(b)).
+		WithVersionID(versionID)
+	if err != nil {
+		return errors.Wrap(err, "failed to get versioned blob client")
+	}
+
```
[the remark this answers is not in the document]
KastenMike: It shouldn't be needed. `WithVersionID` has a comment `Pass "" to remove the versionID returning a URL to the base blob`

repo/blob/azure/azure_storage.go
```diff
@@ -215,6 +228,27 @@ func (az *azStorage) DisplayName() string {
 	return fmt.Sprintf("Azure: %v", az.Options.Container)
 }
 
+func (az *azStorage) getBlobName(it *azblobmodels.BlobItem) string {
+	n := *it.Name
+	return n[len(az.Prefix):]
+}
+
+func (az *azStorage) getBlobMeta(it *azblobmodels.BlobItem) blob.Metadata {
```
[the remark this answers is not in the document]
KastenMike: The same as above~ the code has been in since February without issue so I guess there is always a blob item if the loop is entered

repo/blob/azure/azure_pit.go
```diff
@@ -0,0 +1,199 @@
...
+	// IsImmutableStorageWithVersioning is needed for PutBlob with RetentionPeriod being set.
+	props, err := s.service.ServiceClient().NewContainerClient(s.container).GetProperties(ctx, nil)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not get determine if container '%s' supports versioning", s.container)
+	}
+
+	if props.IsImmutableStorageWithVersioningEnabled == nil || !*props.IsImmutableStorageWithVersioningEnabled {
+		return nil, errors.Errorf("cannot create point-in-time view for non-versioned bucket '%s'", s.container)
+	}
+
```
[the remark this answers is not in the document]
KastenMike: Why? Properties is a fast check and a common way to check if a blob or container exists also. I was tempted to replace the `ListBlobs` in `New` with `GetProperties` for the same reason

repo/blob/azure/azure_pit.go
```diff
@@ -0,0 +1,199 @@
...
+		return s, nil
+	}
+
+	// IsImmutableStorageWithVersioning is needed for PutBlob with RetentionPeriod being set.
+	props, err := s.service.ServiceClient().NewContainerClient(s.container).GetProperties(ctx, nil)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not get determine if container '%s' supports versioning", s.container)
+	}
+
+	if props.IsImmutableStorageWithVersioningEnabled == nil || !*props.IsImmutableStorageWithVersioningEnabled {
```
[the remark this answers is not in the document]
KastenMike: Can be an `IsImmutableStorageWithVersioningEnabled` check to be used in conjunction with immutability protection for now~ to check for Versioning only it seems would need either access to the related StorageAccount or to query a random blob and look for a VersionID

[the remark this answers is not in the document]
KastenMike: Switched to versioning. If someone deleted `KopiaRepositoryBlobID` then it stops working though without user intervention to undelete it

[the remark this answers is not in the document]
KastenMike: Potentially we could keep `IsImmutableStorageWithVersioningEnabled` as a fallback in case `KopiaRepositoryBlobID` is deleted but they have `IsImmutableStorageWithVersioningEnabled`, wdyt?

repo/blob/azure/azure_pit.go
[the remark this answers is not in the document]
julio-lopez: Because (it seems that) it'd be good to reduce the set of permissions needed by kopia, in particular avoid requiring permissions to check bucket properties.

What is needed for kopia to work, is for the credentials to have the permissions to perform the actual operations, (e.g., list with versions), and not necessarily the permissions for checking bucket properties.
Anyway, this is something that we can look into later, separate from this PR.

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/kopia/kopia/internal/testutil"
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/blob/azure"
+)
+
+func TestGetBlobVersions(t *testing.T) {
+	t.Parallel()
+	testutil.ProviderTest(t)
+
+	// must be with Immutable Storage with Versioning enabled
```
julio-lopez, line 25: This should be asserted in the test (after getting the environment and creating the bucket) and fail the test if the container does not have the required properties?

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,218 @@
...
+	container := getEnvOrSkip(t, testContainerEnv)
+	storageAccount := getEnvOrSkip(t, testStorageAccountEnv)
+	storageKey := getEnvOrSkip(t, testStorageKeyEnv)
+
+	// create container if does not exist
+	createContainer(t, container, storageAccount, storageKey)
```
julio-lopez, line 31: A different `createContainer` function may be needed for this test to ensure that the bucket is created with the required settings / properties.

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	ctx := testlogging.Context(t)
+	data := make([]byte, 8)
+	rand.Read(data)
+	// use context that gets canceled after opening storage to ensure it's not used beyond New().
+	newctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
julio-lopez, line 38: `t.Cleanup()` is preferred in tests

```suggestion
	t.Cleanup(cancel)
```

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	prefix := fmt.Sprintf("test-%v-%x-", clock.Now().Unix(), data)
```
julio-lopez, line 40: nit: does it make sense to group all test data in a logical directory?
```suggestion
	prefix := fmt.Sprintf("test-%v-%x/", clock.Now().Unix(), data)
```

repo/blob/azure/azure_pit.go
```diff
@@ -0,0 +1,203 @@
...
+			return vs[:i]
+		}
+
+		// The DeleteMarker blob takes the Timestamp of the previous version but has its own Version.
+		// If there was a Kopia Delete Marker (the blob was protected) it will be caught above but if
+		// the container has versioning enabled but no blob retention protection (or the blob was deleted outside
+		// of the protection window) then we need to check the time of the VersionID because there could be a situation
+		// where Azure's DeleteMarker version has Timestamp 2023-10-20 but Version 2023-10-27.
+		if vs[i].IsDeleteMarker {
+			// Assumes the Put and deletion didn't happen within the same second.
```
julio-lopez, line 99: Does this always hold? What are the consequences when the put and delete happen in the same second?

repo/blob/azure/azure_pit.go
```diff
@@ -0,0 +1,203 @@
...
+			versionLimit := t.Add(1 * time.Second)
+
+			versionTime, err := time.Parse(time.RFC3339Nano, vs[i].Version)
+			if err != nil {
+				return nil
+			}
+
+			if versionTime.After(versionLimit) {
```
julio-lopez, line 107: Why not simply include the deletion marker based on the time in the Version? That is, without adding a second to the time limit.

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	const blobName = "TestGetBlobVersions"
+	blobID := blob.ID(blobName)
+	dataTimestamps, err := putBlobs(ctx, st, blobID, dataBlobs)
+	require.NoError(t, err)
+
+	defer st.DeleteBlob(ctx, blobID)
+
+	pastPIT := clock.Now().Add(-1 * time.Hour).UTC()
+	futurePIT := clock.Now().Add(1 * time.Hour).UTC()
```
julio-lopez, line 68: Should the bounds for the edge cases be narrower here?
```suggestion
	pastPIT := clock.Now().Add(-1 * time.Second).UTC()
	futurePIT := clock.Now().Add(1 * time.Second).UTC()
```

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,218 @@
...
+	t.Parallel()
+	testutil.ProviderTest(t)
+
+	// must be with Immutable Storage with Versioning enabled
+	container := getEnvOrSkip(t, testContainerEnv)
+	storageAccount := getEnvOrSkip(t, testStorageAccountEnv)
+	storageKey := getEnvOrSkip(t, testStorageKeyEnv)
+
+	// create container if does not exist
+	createContainer(t, container, storageAccount, storageKey)
```
julio-lopez, line 135: Similar comments to the ones above.

Also, these tests run in parallel, so container creation is going to race.

julio-lopez (commented):
LG overall.

I am not completely sure about the handling of the deletion marker.

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	ctx := testlogging.Context(t)
+	data := make([]byte, 8)
+	rand.Read(data)
+	// use context that gets canceled after opening storage to ensure it's not used beyond New().
+	newctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	prefix := fmt.Sprintf("test-%v-%x-", clock.Now().Unix(), data)
```
julio-lopez, line 144: nit:

```suggestion
	prefix := fmt.Sprintf("test-%v-%x/", clock.Now().Unix(), data)
```

julio-lopez:
Provider tests are failing for Azure Blob
https://github.com/kopia/kopia/actions/runs/6751751502/job/18356268523?pr=3407

repo/blob/azure/azure_versioned_test.go
KastenMike: I guess I'll leave it for now since azure_storage_test.go and s3_storage_test.go are using defer also

repo/blob/azure/azure_versioned_test.go
KastenMike: Versioning is enabled on the storage account so the `createContainer` can be the same

repo/blob/azure/azure_versioned_test.go
KastenMike: Updated. Also updated it on existing Azure tests

repo/blob/azure/azure_versioned_test.go
KastenMike: I think it doesn't matter how wide they are, as long as they're guaranteed to be in the past/future but I've updated based on the dataTimestamps

repo/blob/azure/azure_versioned_test.go
KastenMike: Removed the `createContainer` commands since they're only in 1 other test

repo/blob/azure/azure_versioned_test.go
KastenMike: If versioning is not enabled it'll fail on `st, err = azure.New` with `cannot create point-in-time view for non-versioned container`

repo/blob/azure/azure_pit.go
KastenMike: Removed this code

repo/blob/azure/azure_pit.go
KastenMike: done

repo/blob/azure/azure_versioned_test.go
julio-lopez: Mike: Please use `t.Cleanup()` here, we can clean up the other instances separately.

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/kopia/kopia/internal/clock"
+	"github.com/kopia/kopia/internal/gather"
+	"github.com/kopia/kopia/internal/testlogging"
+	"github.com/kopia/kopia/internal/testutil"
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/blob/azure"
+	"github.com/kopia/kopia/repo/format"
+)
+
+func TestGetBlobVersions(t *testing.T) {
```
julio-lopez, line 22: @KastenMike This test and the one below are failing in the "test/providers" job. The likely cause is that the buckets used in the test do not have the required setup, that is version-level immutability enabled.

repo/blob/azure/azure_versioned_test.go
julio-lopez: That's error prone, brittle and not future proof.
If the container does not exist and the test that does not create the container runs first, then it will fail.

julio-lopez:
@redgoat650 Do you have any comments?

repo/blob/azure/azure_versioned_test.go
KastenMike: The container will be created in `TestAzureStorage` if not already created by human intervention. The 2nd run it would pass and I guess the container creation is not frequent. Similarly the tests will fail if the StorageAccount is updated but nobody creates it.

I've re-added the `createContainer` command anyway. The race does not matter since already exists error is ignored.

repo/blob/azure/azure_versioned_test.go
KastenMike: Did you re-run the tests? I cannot see run history & you didn't link a new run. Now it doesn't check for version-level immutability

KastenMike: OK I see it in Actions. Someone should change enable versioning on the related storage account for it to work

repo/blob/azure/azure_versioned_test.go
KastenMike: imo it looks worse when needing to pass params but ok

repo/blob/azure/azure_versioned_test.go
KastenMike: In the process whoever sets it could enable version-level immutability on the container since it seems the SDK doesn't support it

KastenMike: @jkowalski is it something you'd be able to enable?

julio-lopez: Mike: this requires a little more than that.

These 2 tests need to run in a different setup. That means:
- setting up a **separate** storage account with version-level immutability enabled (as you mentioned).
- create a new bucket with version-level immutability, or have the test create it with the desired settings.
- these tests need to use a separate 'credentials setting' from the environment. The actual credentials may be the same, but the storage account and bucket would be different.
- the GH environment needs to be updated with these new cred information.

KastenMike: Why does it require a separate storage account? you could simply have the same one running new + existing tests but with 2 new settings changed

julio-lopez: @KastenMike For what I understand, the storage account currently used in the tests does not have immutability enabled. I thought that version immutability needed to be enabled at the storage account level. This is what I recall but I could be mistaken. I'll double check.

If no separate storage account is needed, the it is a matter of simply creating a separate container with the appropriate settings.

PTAL at https://github.com/kopia/kopia/compare/azure-immutability-setup

This contains the setup code for the storage account and the container. PLMK if that makes sense, if there's anything missing or that needs to be changed.

Thanks.

KastenMike: Setting `ImmutableStorageWithVersioning` to true via CLI is automatically making versioning be enabled too? In the UI they make you do 2 separate actions. A policy isn’t needed on the container~ only versioning + ImmutableStorageWithVersioning
If the intention would be to run it repeatedly like the `createContainer` currently then it needs checks for "already exists" errors also on the SA & container

julio-lopez: Behavior observed using the console: attempting to enable "version immutability" in the container requires enabling "versioning" in the storage account. It wasn't clear at that moment what the implication was for existing buckets, so we created a separate storage account and bucket.

See inline comment as well.

repo/blob/azure/azure_versioned_test.go
```diff
@@ -0,0 +1,227 @@
...
+	t.Parallel()
+	testutil.ProviderTest(t)
+
+	// must be with Immutable Storage with Versioning enabled
+	container := getEnvOrSkip(t, testContainerEnv)
```
julio-lopez, line 27: Use the `KOPIA_AZURE_TEST_IMMUTABLE_*` variant of the environment variables for these tests. We'll run the provider tests before merging to make sure the tests won't break.

KastenMike: updated

julio-lopez: Mike: apologies, my previous comment was ambiguous.
Use the `KOPIA_AZURE_TEST_IMMUTABLE_*` variant of the environment variables **only** in the new tests that are added here.

KastenMike: updated

julio-lopez (approved):
@KastenMike 🥇Thanks for doing this.

LGTM  ✅

PR #3411, opened 2023-10-27
fix(cli): Don't return error when parameters unchanged

KastenMike:
Currently there is an [inconsistency](https://github.com/search?q=repo%3Akopia%2Fkopia%20%22no%20change%22&type=code) between what to do when `!anyChanges`...some return an error while others log it and return nil. I'm not sure if in any of the commands it makes sense to return an error in such cases (I would guess always nil) but in the case of SetParameters I think it's not needed. People want a desired state and they have the desired state so all is good; similar to how DeleteBlob returns nil when the blob doesn't exist

At least one path where it is reached is setting `retentionMode` to `none` but retention is already disabled; whereas it's not the case if you set it to `COMPLIANCE` although the value is also not changed. ~~Maybe the logging should also be removed due to this (it only logs in 1 or a couple edge cases)~~

https://github.com/kopia/kopia/blob/b7f3e17c9febe120c5c122ba33753eb86ca63251/cli/command_repository_set_parameters.go#L210-L214

### DIFF
--- a/cli/command_repository_set_parameters.go
+++ b/cli/command_repository_set_parameters.go
@@ -228,8 +228,6 @@ func (c *commandRepositorySetParameters) run(ctx context.Context, rep repo.Direc
 	requiredFeatures = c.addRemoveUpdateRequiredFeatures(requiredFeatures, &anyChange)
 
 	if !anyChange {
-		log(ctx).Infof("no changes from set-parameters")
-
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


cli/command_repository_set_parameters.go
```diff
@@ -228,7 +228,7 @@ func (c *commandRepositorySetParameters) run(ctx context.Context, rep repo.Direc
 	requiredFeatures = c.addRemoveUpdateRequiredFeatures(requiredFeatures, &anyChange)
 
 	if !anyChange {
-		return errors.Errorf("no changes")
+		return nil
```
julio-lopez, line 231: Mike: can you add a test for this?

PR #3412, opened 2023-10-27
feat(repository): Add support for Azure immutability protection

KastenMike:
Adds the option to set retention for Azure. They'd need to set the mode to Locked and run the command `kopia maintenance set --extend-object-locks` true like on S3.

Only includes `Locked` since `Unlocked` is not recommended. `PutBlob` with an invalid type (e.g. Locked for S3) will fail. The same validation check could also be done on [ExtendBlobRetention ](https://github.com/kopia/kopia/pull/3405) or it can be left to fail naturally...technically `PutBlob` can also be left to fail naturally rather than having the guard clause like `opts.HasRetentionOptions() && !opts.RetentionMode.IsValidAzure()`...instead put on Azure CLI would just fail like "COMPLIANCE is unknown retention type"

### DIFF
--- a/cli/command_repository_create.go
+++ b/cli/command_repository_create.go
@@ -49,7 +49,7 @@ func (c *commandRepositoryCreate) setup(svc advancedAppServices, parent commandP
 	cmd.Flag("object-splitter", "The splitter to use for new objects in the repository").Default(splitter.DefaultAlgorithm).EnumVar(&c.createSplitter, splitter.SupportedAlgorithms()...)
 	cmd.Flag("create-only", "Create repository, but don't connect to it.").Short('c').BoolVar(&c.createOnly)
 	cmd.Flag("format-version", "Force a particular repository format version (1, 2 or 3, 0==default)").IntVar(&c.createFormatVersion)
-	cmd.Flag("retention-mode", "Set the blob retention-mode for supported storage backends.").EnumVar(&c.retentionMode, blob.Governance.String(), blob.Compliance.String())
+	cmd.Flag("retention-mode", "Set the blob retention-mode for supported storage backends.").EnumVar(&c.retentionMode, blob.Governance.String(), blob.Compliance.String(), blob.Locked.String())
 	cmd.Flag("retention-period", "Set the blob retention-period for supported storage backends.").DurationVar(&c.retentionPeriod)
 
 	c.co.setup(svc, cmd)
--- a/repo/blob/azure/azure_storage.go
+++ b/repo/blob/azure/azure_storage.go
@@ -13,6 +13,7 @@ import (
 	azblobblob "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/bloberror"
 	azblockblob "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
+	azblobmodels "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
 	"github.com/pkg/errors"
 
 	"github.com/kopia/kopia/internal/clock"
@@ -120,8 +121,8 @@ func translateError(err error) error {
 
 func (az *azStorage) PutBlob(ctx context.Context, b blob.ID, data blob.Bytes, opts blob.PutOptions) error {
 	switch {
-	case opts.HasRetentionOptions():
-		return errors.Wrap(blob.ErrUnsupportedPutBlobOption, "blob-retention")
+	case opts.HasRetentionOptions() && !opts.RetentionMode.IsValidAzure():
+		return errors.Wrap(blob.ErrUnsupportedPutBlobOption, "blob retention mode is not valid for Azure")
 	case opts.DoNotRecreate:
 		return errors.Wrap(blob.ErrUnsupportedPutBlobOption, "do-not-recreate")
 	}
@@ -157,7 +158,7 @@ func (az *azStorage) getObjectNameString(b blob.ID) string {
 
 // ListBlobs list azure blobs with given prefix.
 func (az *azStorage) ListBlobs(ctx context.Context, prefix blob.ID, callback func(blob.Metadata) error) error {
-	prefixStr := az.Prefix + string(prefix)
+	prefixStr := az.getObjectNameString(prefix)
 
 	pager := az.service.NewListBlobsFlatPager(az.container, &azblob.ListBlobsFlatOptions{
 		Prefix: &prefixStr,
@@ -173,19 +174,7 @@ func (az *azStorage) ListBlobs(ctx context.Context, prefix blob.ID, callback fun
 		}
 
 		for _, it := range page.Segment.BlobItems {
-			n := *it.Name
-
-			bm := blob.Metadata{
-				BlobID: blob.ID(n[len(az.Prefix):]),
-				Length: *it.Properties.ContentLength,
-			}
-
-			// see if we have 'Kopiamtime' metadata, if so - trust it.
-			if t, ok := timestampmeta.FromValue(stringDefault(it.Metadata["kopiamtime"], "")); ok {
-				bm.Timestamp = t
-			} else {
-				bm.Timestamp = *it.Properties.LastModified
-			}
+			bm := az.getBlobMeta(it)
 
 			if err := callback(bm); err != nil {
 				return err
@@ -213,6 +202,27 @@ func (az *azStorage) ConnectionInfo() blob.ConnectionInfo {
 
 func (az *azStorage) DisplayName() string {
 	return fmt.Sprintf("Azure: %v", az.Options.Container)
+}
+
+func (az *azStorage) getBlobName(it *azblobmodels.BlobItem) string {
+	n := *it.Name
+	return n[len(az.Prefix):]
+}
+
+func (az *azStorage) getBlobMeta(it *azblobmodels.BlobItem) blob.Metadata {
+	bm := blob.Metadata{
+		BlobID: blob.ID(az.getBlobName(it)),
+		Length: *it.Properties.ContentLength,
+	}
+
+	// see if we have 'Kopiamtime' metadata, if so - trust it.
+	if t, ok := timestampmeta.FromValue(stringDefault(it.Metadata["kopiamtime"], "")); ok {
+		bm.Timestamp = t
+	} else {
+		bm.Timestamp = *it.Properties.LastModified
+	}
+
+	return bm
 }
 
 func (az *azStorage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes, opts blob.PutOptions) (azblockblob.UploadResponse, error) {
--- a/repo/blob/azure/azure_storage_test.go
+++ b/repo/blob/azure/azure_storage_test.go
@@ -7,8 +7,10 @@ import (
 	"net/url"
 	"os"
 	"testing"
+	"time"
 
-	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	legacyazblob "github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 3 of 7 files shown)


repo/blob/azure/azure_storage_test.go
```diff
@@ -270,3 +273,125 @@ func TestAzureStorageInvalidCreds(t *testing.T) {
...
+	if !blobRetention.After(currentTime) {
+		t.Fatalf("blob retention period not in the future: %v", blobRetention)
+		t.FailNow()
+	}
+
+	extendOpts := blob.ExtendOptions{
+		RetentionMode:   blob.Locked,
+		RetentionPeriod: 4 * time.Second,
+	}
+	err = st.ExtendBlobRetention(ctx, dummyBlob, extendOpts)
```
KastenMike, line 341: This will not pass until https://github.com/kopia/kopia/pull/3412 is merged

KastenMike: It will also not pass until https://learn.microsoft.com/en-us/azure/storage/blobs/immutable-policy-configure-version-scope?tabs=azure-portal#enable-support-for-version-level-immutability is enabled - would you be able to do this for the relevant provider @jkowalski?

repo/blob/azure/azure_storage_test.go
```diff
@@ -270,3 +274,134 @@ func TestAzureStorageInvalidCreds(t *testing.T) {
...
+	t.Parallel()
+	testutil.ProviderTest(t)
+
+	// must be with ImmutableStorage with Versioning enabled
+	container := getEnvOrSkip(t, testContainerEnv)
+	storageAccount := getEnvOrSkip(t, testStorageAccountEnv)
+	storageKey := getEnvOrSkip(t, testStorageKeyEnv)
+
+	// create container if does not exist
+	createContainer(t, container, storageAccount, storageKey)
```
KastenMike, line 289: note: this addition is basically pointless since we are not deleting/creating containers each time I assume; otherwise there is no point in having the cleanup step for data older than 1 hour since there wouldn't be any

cli/command_repository_create.go
```diff
@@ -49,7 +49,7 @@ func (c *commandRepositoryCreate) setup(svc advancedAppServices, parent commandP
 	cmd.Flag("object-splitter", "The splitter to use for new objects in the repository").Default(splitter.DefaultAlgorithm).EnumVar(&c.createSplitter, splitter.SupportedAlgorithms()...)
 	cmd.Flag("create-only", "Create repository, but don't connect to it.").Short('c').BoolVar(&c.createOnly)
 	cmd.Flag("format-version", "Force a particular repository format version (1, 2 or 3, 0==default)").IntVar(&c.createFormatVersion)
-	cmd.Flag("retention-mode", "Set the blob retention-mode for supported storage backends.").EnumVar(&c.retentionMode, blob.Governance.String(), blob.Compliance.String())
+	cmd.Flag("retention-mode", "Set the blob retention-mode for supported storage backends.").EnumVar(&c.retentionMode, blob.Governance.String(), blob.Compliance.String(), blob.Locked.String())
```
Shrekster, line 52: While Kopia works with multiple tiers underneath, but the storage APIs and CLIs are designed to work seamlessly across all with minimal storage specific couplings. There are a few places where it is just unavoidable to expose storage specific parameters to the CLI/API.

Maybe we should rename the retention-modes in Kopia from "governance" / "compliance" to something more generic like "WORM Mode" / "RBAC Mode" or similar to keep things generic between S3/GCS/Azure/etc. However, this can be done in a separate PR; no need to overload this PR.

I am wondering whether we need to make an exception for Azure immutability as well. For all intents and purposes of Azure usage within Kopia, ain't Azure "locked" mode is just the same as the governance mode ? So, for Azure can we use the existing mode of Kopia called "governance" as its locked mode ? That way, we'll not need to introduce a new mode and going forward we can make things more generic and rename governance to "WORM Mode" or something more generic so it doesn't sound like AWSish.

repo/blob/azure/azure_storage_test.go
```diff
@@ -270,3 +274,134 @@ func TestAzureStorageInvalidCreds(t *testing.T) {
...
 	}
 }
+
+// TestAzureStorageImmutabilityProtection runs through the behavior of Azure immutability protection.
+func TestAzureStorageImmutabilityProtection(t *testing.T) {
+	t.Parallel()
+	testutil.ProviderTest(t)
+
+	// must be with ImmutableStorage with Versioning enabled
+	container := getEnvOrSkip(t, testContainerEnv)
```
Shrekster, line 284: @KastenMike don't we need to setup the container here with immutability ?

repo/blob/azure/azure_storage_test.go
Shrekster: Yes, we should create the container once manually and then keep reusing it. Creating and deleting containers for this test may not be a good idea because we'll need to create unique containers and then we might end up in a situation where we start leaking these containers. Plus, deleting buckets can take longer if there are locked objects underneath.

Shrekster: I think @julio-lopez said that he is preparing a script that could be run one-time during provisioning to setup the required buckets in Kopia test env which we can keep repurposing.

repo/blob/azure/azure_storage_test.go
Shrekster: Correct. @julio-lopez said he is working on a script from this.

repo/blob/azure/azure_storage_test.go
KastenMike: it should already be prepared beforehand

cli/command_repository_create.go
KastenMike: Locked is more like Compliance mode (Governance is not so secure as certain users can override the protections).
The main reasons against using Compliance:
1. Azure is kind of like a 2nd class citizen then
2. In the `azure_storage` we'd need to replace Compliance with Locked in put & extend methods
Can be done but looks a bit strange imo - if GCP eventually adds support then they may also have different names at that point too

"WORM Mode" / "RBAC Mode" I guess would be a breaking change for existing setups

KastenMike: I've updated it to use COMPLIANCE although I didn't update https://github.com/kopia/kopia/pull/3405 - it could be updated when this is merged to make use of the `blob.Locked` from this PR

Shrekster: Thanks Mike. You have valid points. I am not hard on it, but thanks for making the change. Next time when we make the change of "WORM Mode" / "RBAC Mode" then we'll have to do less number of these.

KastenMike: I'm leaning more to going back to Locked...it also need adding to `putBlob` too which I forgot about and that's part of the problem of using S3 names where they don't belong...you need to remember to update them everywhere

Shrekster: No worries, thanks for trying.

Shrekster: What is the ugly IF check BTW ?

repo/blob/azure/azure_storage.go
```diff
@@ -232,6 +244,7 @@ func (az *azStorage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes, op
 	}
 
 	if opts.HasRetentionOptions() {
+		// kopia delete marker blob can be Unlocked rather than Compliance
```
julio-lopez, line 247: Not sure this comment is relevant. Whether a blob is locked or not depends on how the user configured it, that is, the option specified in the configuration should be respected.

repo/blob/azure/azure_storage_test.go
```diff
@@ -270,3 +274,134 @@ func TestAzureStorageInvalidCreds(t *testing.T) {
...
+	cli := getAzureCLI(t, storageAccount, storageKey)
+
+	count := getBlobCount(ctx, t, st, content.BlobIDPrefixSession)
+	require.Equal(t, 1, count)
+
+	currentTime := clock.Now().UTC()
+
+	blobRetention := getBlobRetention(ctx, t, cli, container, blobNameFullPath)
+	if !blobRetention.After(currentTime) {
+		t.Fatalf("blob retention period not in the future: %v", blobRetention)
```
julio-lopez, line 336: We may want to add a note here that this test may flake in cases where the initial retention time is not long enough. 3 seconds may work ~100% on a local dev setup. In CI, shared VM with competing tests, etc. exceeding this threshold may be common.
We may want to increase the initial retention period as well.

repo/blob/azure/azure_storage_test.go
julio-lopez: Container has been set up, and has been used for the Azure PITR functionality.

julio-lopez (commented):
Let's merge this after addressing the comments about:
- preserving the configured retention mode (locked vs. unlocked)
- the changes for the test

and making sure tests pass in the test/providers job.

Let's keep the (kopia) CLI changes out and do that in a separate PR.

cli/command_repository_create.go
julio-lopez: We should try to provide an uniform interface in the CLI. We only had a single immutability implementation (for S3) and unfortunately we did not anticipate creating that interface.

Here's a straw-man proposal, please beat it and find holes.

What about accepting all these (case-insensitive) values in the CLI for both S3 and Azure.
1. "Locked"
2. "Compliance"
3. "Unlocked"
4. "Governance"

And having an internal translation to a canonical `{locked, unlocked}` representation.
And then in each provider translate from the canonical representation to the provider specific setting.

For S3 object locking:
- `locked` => `"Compliance"`
- `unlocked` => `"Governance"`

For Azure immutability:
- `locked` => `"Locked"`
- `unlocked` => `"Unlocked"`

Thoughts?

Again, let's do this in a separate PR, and merge this one after making the requested changes.

repo/blob/azure/azure_immu_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/kopia/kopia/internal/gather"
+	"github.com/kopia/kopia/internal/testlogging"
+	"github.com/kopia/kopia/internal/testutil"
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/blob/azure"
+	"github.com/kopia/kopia/repo/content"
+)
+
+// TestAzureStorageImmutabilityProtection runs through the behavior of Azure immutability protection.
+func TestAzureStorageImmutabilityProtection(t *testing.T) {
```
julio-lopez, line 24: Test looks good.
Filed #3476 to followup on the test cleanup.

repo/blob/azure/azure_storage_test.go
julio-lopez: Filed #3477 to track the container creation for these tests.

repo/blob/azure/azure_storage.go
```diff
@@ -119,14 +120,16 @@ func translateError(err error) error {
...
-	case opts.HasRetentionOptions():
-		return errors.Wrap(blob.ErrUnsupportedPutBlobOption, "blob-retention")
-	case opts.DoNotRecreate:
+	if opts.DoNotRecreate {
 		return errors.Wrap(blob.ErrUnsupportedPutBlobOption, "do-not-recreate")
 	}
 
-	_, err := az.putBlob(ctx, b, data, opts)
+	_, err := az.putBlob(ctx, b, data, blob.PutOptions{
+		RetentionMode:   blob.Locked,
```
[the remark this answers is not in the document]
KastenMike: good point :) seems I went to extreme trying to remove the if statement before 😄 fixed

repo/blob/azure/azure_storage.go
```diff
@@ -133,14 +133,21 @@ func translateError(err error) error {
...
-	switch {
-	case opts.HasRetentionOptions():
-		return errors.Wrap(blob.ErrUnsupportedPutBlobOption, "blob-retention")
-	case opts.DoNotRecreate:
+	if opts.DoNotRecreate {
 		return errors.Wrap(blob.ErrUnsupportedPutBlobOption, "do-not-recreate")
 	}
 
-	_, err := az.putBlob(ctx, b, data, opts)
+	hardcodedOpts := blob.PutOptions{
```
julio-lopez, line 140: nit: not sure about the variable name.
Anyway, merging now and we'll modify it separately.

repo/blob/azure/azure_storage.go
julio-lopez:
```suggestion
		// kopia delete marker blob must be "Unlocked", so it cannot be overridden to "Locked" here.
```

julio-lopez (approved):
Thanks Mike

Issue #1962, opened 2022-05-21
cross compile not working


durdin85:
Hi, kopia looks like nice tool so I wanted to give it a try, but to cover all my needs it need to support also 32-bit systems. It might sound weird in 2022, but it's still true. Anyway as there are no 32-bit binaries available, I was going to build kopia on my own.
What I've tested and works : cross compiling kopia on 64-bit linux resulting 32-bit windows binary.
What does not work for me : cross compiling kopia-UI on 64-bit linux resulting 32-bit windows/electron app.

Maybe I am doing it wrong, this is just my first day with kopia, but seems it assumes I am on windows, when I am trying to build binaries for windows, and so it tries to pick the windows-style paths, and assumes "powershell" exists. On my system I indeed have powershell, but it's called pswh instead. I've tried setting up the alias powershell=pwsh, even tried putting fake powershell binary into PATH, but with no luck. I've even tried tampering with Makefiles, modifying my environment, but it gets only worse. Last attempt:

```
$ env LANG=C GOOS=windows GOARCH=386 make kopia-ui
/bin/sh: 1: powershell: not found
/bin/sh: 1: powershell: not found
make -C app build-electron
make[1]: Entering directory '/home/durdin/dev/kopia/app'
/bin/bash: go: command not found
/bin/bash: go: command not found
go run github.com/kopia/kopia/tools/gettool -tool node:16.13.0 --output-dir /home/durdin/dev/kopia/tools/.tools/node-16.13.0
/bin/bash: go: command not found
make[1]: *** [../tools/tools.mk:138: /usr/local/bin/npm] Error 127
make[1]: Leaving directory '/home/durdin/dev/kopia/app'
make: *** [Makefile:114: kopia-ui] Error 2

$ # doing some steps manually works
$ go run github.com/kopia/kopia/tools/gettool -tool node:16.13.0 --output-dir /home/durdin/dev/kopia/tools/.tools/node-16.13.0
Downloading node version 16.13.0 from https://nodejs.org/dist/v16.13.0/node-v16.13.0-linux-x64.tar.gz...
2022/05/21 21:46:51 https://nodejs.org/dist/v16.13.0/node-v16.13.0-linux-x64.tar.gz checksum ok
$ which npm
/usr/bin/npm
$ which go
/opt/go/bin/go

```

jkowalski:
Windows is not the main development OS for most folks here, so it's lagging behind.

[...]

PR #3431, opened 2023-11-05
feat(ci): Add Windows 386 architecture support

davidberardozzi:
Refers to #1962.

The goal is to add support for Windows 386 architectures.


julio-lopez:
Windows on 32-bit architectures is not supported by kopia. This platform is considered deprecated by Microsoft and it is no longer receiving support.

Issue #3426, opened 2023-11-02
Switching a large blob store to Storj is problematic


rjkroege:
I attempted to switch to Storj ([Globally Distributed Cloud Object Storage](https://www.storj.io/)) as a blobstore. This doesn't currently work very well. Two observations about this:

* `kopia repository sync-to  rclone` seems to work but is very slow
* the Storj S3 compatibility interface works just fine for *small* blobstores but is not adequate for production use because the Storj S3 gateway limits `s3_storage.go:/ListBlobs()` to a thousand blobs (per documentation) total if using prefix matching.

I see two solutions:

* run a local Storj S3 gateway with much larger limits
* write a "native" Storj blob implementation in Kopia.

I would expect that the "native" implementation (i.e. Storj blob implementation) is preferable for its performance and robustness. 

@jkowalski et al: are you open to PRs implementing such a feature? If so, I will attempt to write this.

mathrick:
Any progress? This makes backups on Storj work fine at first, but then stop working as time goes on, which is very, very bad. I just discovered that my backups have been silently failing for a while now, and I can't connect to the repository anymore, with errors during connection initialisation:

```
20:06:03.690 finished initial cache scan {"cache":"contents","duration":"3.34742068s","totalRetainedSize":5242839371,"tooRecentBytes":0,"tooRecentCount":0,"maxSizeBytes":5242880000,"limitBytes":0,"inUsePercent":99}
20:06:03.965 finished initial cache scan {"cache":"metadata","duration":"268.039301ms","totalRetainedSize":5241761666,"tooRecentBytes":0,"tooRecentCount":0,"maxSizeBytes":5242880000,"limitBytes":0,"inUsePercent":99}
20:06:06.857 finished initial cache scan {"cache":"index-blobs","duration":"2.891745439s","totalRetainedSize":4094244378,"tooRecentBytes":0,"tooRecentCount":0,"maxSizeBytes":5242880000,"limitBytes":0,"inUsePercent":78}
20:06:06.869 [STORAGE] concurrency level reached {"maxConcurrency":1}
20:06:06.887 [STORAGE] concurrency level reached {"maxConcurrency":2}
20:06:06.891 [STORAGE] concurrency level reached {"maxConcurrency":3}
20:06:06.897 [STORAGE] concurrency level reached {"maxConcurrency":4}
[...]

PR #3445, opened 2023-11-11
Add a blob implementation for Storj

rjkroege:
This CL adds a blob storage implementation for the Storj cloud service.
It helps implement #3426.


rjkroege:
Some notes to the reviewer:

* I have some questions about implementation noted as `TODO`
* I am wondering how to manage the secret keys for running the tests. I haven't looked into that yet.
* The command line integration is in another PR.
* I'd prefer to write the shard support and multipart uploads as separate PRs (mostly) for review convenience. I could add to this PR too though as separate commits.

rjkroege:
I note in passing: I did some tests and shard support is required for performant operation.

Issue #3372, opened 2023-10-06
S3 ransomware protection leads to ugly errors when creating repository and snapshots


NiklausHofer:
I am particularly interested in Kopia's ransomware protection as described here: https://kopia.io/docs/advanced/ransomware-protection/

To this end, I have set-up an S3 storage using MinIO.

I have set-up both a user and a bucket, both called "example", with the user having the following policies applied:
```json
# readwrite_personal_bucket:
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::${aws:username}/*"
            ]
        }
    ]
}

# no_delete:
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Deny",
            "Action": [
                "s3:DeleteBucket",
                "s3:DeleteBucketPolicy",
                "s3:DeleteObject",
                "s3:DeleteObjectVersion"
            ],
            "Resource": [
                "arn:aws:s3:::*"
            ]
        }
    ]
}
```

The `no_delete` policy is adapted from the above linked Ransomawre Protection page. As far as I can tell, it works as intended. The user can upload data, but not delete it.

Next, I went to create a repository, using Kopia on my PC:
```bash
kopia repo create s3 \
    --endpoint=server.example.com \
    --bucket=example \
    --access-key=example \
    --secret-access-key=******************************** \
    --retention-mode COMPLIANCE \
    --retention-period 92d
```
The bucket gets created as expected, however, at the end of the run, an ugly error message is printed:
```
ERROR error populating repository: error flushing writer: error flushing contents: error flushing indexes: unable to commit session: failed to delete session marker s5d4f094ae11129f410b25139e08c864a-s0714add1b92f1bb0121: unable to complete DeleteBlob(s5d4f094ae11129f410b25139e08c864a-s0714add1b92f1bb0121) despite 10 retries: Access Denied.
```

Creating snapshots works, but it too, prints out this error message:
```bash
kopia snapshot create /etc/
```
```
Snapshotting root@octarine:/etc ...
 * 0 hashing, 0 hashed (0 B), 2521 cached (20.7 MB), uploaded 0 B, estimating...
Created snapshot with root k8f3138572a7a2c0d9cd8fe0c906ac8b1 and ID b76cc12f5de90b54684021bbf0c9acca in 0s
[...]

PR #3451, opened 2023-11-17
docs(providers): fix documentation for S3 permissions

julio-lopez:
DeleteObject needs to be allowed. This is needed for kopia to work correctly.

The assumption is that S3 versioning is enabled.


julio-lopez:
#3372

PR #3454, opened 2023-11-20
feat(cli): A: Add ability to dump pprof data to logs ...

aaron-kasten:
This PR adds the ability to dump pprof data to logs for debugging. 

Usage
=====

example:

```
[1] 77308
```

```
$ kill %1
```

```
saving PEM buffers for output
dumping PEM for "PPROF MEM"
-----BEGIN PPROF MEM-----
.
.
.
//8psEjOrZ4AAA==
-----END PPROF MEM-----
```

```
$ ./go/bin/kats dump.b64
```

```
 ./go/bin/kats --help
 Usage of ./go/bin/kats:
  -verbose
    	verbose outout
```

```
File: pprof_mem.bin
Type: inuse_space
(pprof) tree
```

Configuration Options
=====

TBD


julio-lopez:
@aaron-kasten The steps below allow you to directly get pprof files in (binary) protobuf format that can be then directly loaded in `go tool pprof`. This allows getting the required pprof files, already separated without the need to encode them and then decode them.

Would that approach work for your use case? if not, why not?


```
# 1. start the kopia process with a metrics listener and enable pprof. For example:

kopia --metrics-listen-addr localhost:10005 --enable-pprof snapshot create &

# 2. periodically get various pprof dumps from the metrics endpoint using go tool pprof

go tool pprof -proto  http://localhost:10005/debug/pprof/profile

# output:
#
# The obtained pprof file can be examined directly with go pprof
# Fetching profile over HTTP from http://localhost:10005/debug/pprof/profile
# Saved profile in ${HOME}/pprof/pprof.samples.cpu.002.pb.gz
# Generating report in profile001.pb.gz

# 3. load pprof file, for example, using the built-in web UI

go tool pprof -http localhost:5000 $HOME/pprof/pprof.samples.cpu.001.pb.gz

```

Other kopia pprof endpoints:
`/debug/pprof/cmdline`
`/debug/pprof/profile`
`/debug/pprof/symbol`
`/debug/pprof/trace`

aaron-kasten:
> @aaron-kasten The steps below allow you to directly get pprof files in (binary) protobuf format that can be then directly loaded in `go tool pprof`. This allows getting the required pprof files, already separated without the need to encode them and then decode them.
> 
> Would that approach work for your use case? if not, why not?
> 
> ```
> # 1. start the kopia process with a metrics listener and enable pprof. For example:
> 
> kopia --metrics-listen-addr localhost:10005 --enable-pprof snapshot create &
> 
> # 2. periodically get various pprof dumps from the metrics endpoint using go tool pprof
> 
> go tool pprof -proto  http://localhost:10005/debug/pprof/profile
> 
> # output:
> #
> # The obtained pprof file can be examined directly with go pprof
> # Fetching profile over HTTP from http://localhost:10005/debug/pprof/profile
> # Saved profile in ${HOME}/pprof/pprof.samples.cpu.002.pb.gz
> # Generating report in profile001.pb.gz
> 
> # 3. load pprof file, for example, using the built-in web UI
> 
> go tool pprof -http localhost:5000 $HOME/pprof/pprof.samples.cpu.001.pb.gz
> ```
> 
> Other kopia pprof endpoints: `/debug/pprof/cmdline` `/debug/pprof/profile` `/debug/pprof/symbol` `/debug/pprof/trace`

There are disadvantages to using the publish URL and poll vs. the method proposed in this PR train:
- Polling can miss state that is important to diagnosing some pod failures.  The proposed method dumps pprof information on process exit, capturing state at process termination.  Repeated pod failures will create multiple dumps, removing guesswork of process lifecycle events (when debugging).
- Setting up network/tunnel to allow polling is "complicated".  There are several hurdles to accessing published network ports depending on environment: tunnels and/or port forwarding may need to be setup, a destination for the captured files needs to be configured.  In a Kubernetes environment this may mean configuration of additional containers, pods, volumes, etc... these changes present their own risks.  The proposed method uses an existing, reliable, facility for capturing data - log capture.  Many systems already have log capture facilities in the cluster or on the node, if not `kubectl logs` can be used to capture the output.  This removes any new subsystems for file capture by using tools known to already have a high level of reliability and fidelity.
- The output from the proposed system may be automated with little support infrastructure.  The process is limited to capturing log information.  Conversion from PEM to binary form can be done off-line.  As already mentioned, developing an automated system for capturing pprof data using published network endpoints will be complicated by network access and delineation of process lifecycle.
- This same facility is already a supported subsystem in other software.  Having this implemented in Kopia gives the ability to capture pprof data across the entire application stack (not just export of snapshots).
-

julio-lopez:
@aaron-kasten I see your point. It makes sense.

debug/debug.go
```diff
@@ -0,0 +1,380 @@
...
+// ProfileConfig configuration flags for a profile.
+type ProfileConfig struct {
+	flags []string
+	buf   *bytes.Buffer
+}
+
+// ProfileConfigs configuration flags for all requested profiles.
+type ProfileConfigs struct {
+	mu  sync.Mutex
+	pcm map[ProfileName]*ProfileConfig
```
Shrekster, line 72:
```suggestion

	// +checklocks:mu
	pcm map[ProfileName]*ProfileConfig
```

debug/debug.go
```diff
@@ -0,0 +1,380 @@
...
+		def: DefaultDebugProfileRate,
+	},
+	ProfileNameMutex: {
+		fn:  func(x int) { runtime.SetMutexProfileFraction(x) },
+		def: DefaultDebugProfileRate,
+	},
+}
+
+// GetValue get the value of the named flag, `s`.  false will be returned
+// if the flag value does not exist.
```
Shrekster, line 96:
```suggestion
// GetValue get the value of the named flag, `s`.  False will be returned
// if the flag does not exist. True will be returned if flag exists without
// a value.
```

debug/debug.go
```diff
@@ -0,0 +1,380 @@
...
+
+	return "", false
+}
+
+func parseProfileConfigs(bufSizeB int, ppconfigs string) map[ProfileName]*ProfileConfig {
+	pbs := map[ProfileName]*ProfileConfig{}
+	allProfileOptions := strings.Split(ppconfigs, ":")
+
+	for _, profileOptionWithFlags := range allProfileOptions {
+		// of those, see if any have profile-specific settings
```
Shrekster, line 119:
```suggestion
		// of those, see if any have profile specific settings
```

debug/debug.go
```diff
@@ -0,0 +1,380 @@
...
+	mu  sync.Mutex
+	pcm map[ProfileName]*ProfileConfig
+}
+
+//nolint:gochecknoglobals
+var pprofConfigs = &ProfileConfigs{}
+
+type pprofSetRate struct {
+	fn  func(int)
+	def int
```
Shrekster, line 80: nit; maybe rename `fn` -> `setter` and `def` -> `defaultValue` for readability. It took me multiple hoops to realize this. Or you could add a comment here.

debug/debug.go
```diff
@@ -0,0 +1,380 @@
...
+			continue
+		}
+
+		log(ctx).Debugf("setting PPROF rate, %d, for %s", n1, k)
+		pprofset.fn(n1)
+	}
+}
+
+// ClearProfileFractions set the profile fractions to their zero values.
+func ClearProfileFractions(profileBuffers map[ProfileName]*ProfileConfig) {
```
Shrekster, line 181: What do you think about making `ClearProfileFractions()` as method on a profileBuffers instance and make `map[ProfileName]*ProfileConfig` a type instead ?

debug/debug.go
```diff
@@ -0,0 +1,380 @@
...
+
+	flgs := strings.Split(ppconfig, ",")
+	if len(flgs) > 0 && flgs[0] != "" { // len(flgs) > 1 && flgs[0] == "" should never happen
+		q.flags = flgs
+	}
+
+	return q
+}
+
+func setupProfileFractions(ctx context.Context, profileBuffers map[ProfileName]*ProfileConfig) {
```
Shrekster, line 148: if you could make this a type `profileBuffers map[ProfileName]*ProfileConfig`, then `setupProfileFractions()` could be the called right after `parseProfileConfigs()`

Shrekster (approved):
LGTM!

debug/debug.go
aaron-kasten: There's a rule in K10 that helper methods don't have receivers... so this is an extension of that:  having helpers without receivers.  I see it as a stylistic issue .. I ask that we not change it for this PR - we can change it if there's a compelling reason.

debug/debug.go
aaron-kasten: Similarly this seems stylistic ... I'd as that we postpone those changes until the PR is in and/or there's a concrete justification.

debug/debug.go
```diff
@@ -0,0 +1,386 @@
...
+
+	pprofConfigs.pcm = parseProfileConfigs(bufSizeB, ppconfigs)
+
+	// profiling rates need to be set before starting profiling
+	setupProfileFractions(ctx, pprofConfigs.pcm)
+
+	// cpu has special initialization
+	v, ok := pprofConfigs.pcm[ProfileNameCPU]
+	if ok {
+		err := pprof.StartCPUProfile(v.buf)
```
julio-lopez, line 231: Are there any concerns about large memory consumption here?
How large can the profile get and thus `v.buf` ?

debug/debug.go
```diff
@@ -0,0 +1,386 @@
...
+		Bytes: bs,
+	}
+	// wrt is likely a line oriented writer, so writing individual lines
+	// will make best use of output buffer and help prevent overflows or
+	// stalls in the output path.
+	pr, pw := io.Pipe()
+	// encode PEM in the background and output in a line oriented
+	// fashion - this prevents the need for a large buffer to hold
+	// the encoded PEM.
+	go func() {
```
julio-lopez, line 255: what is the rationale for doing the encoding in a separate go routine?

Would it be possible to write to `wrt` either directly or through a `bufio.Writer`?

debug/debug.go
```diff
@@ -0,0 +1,386 @@
...
+		err := pprof.StartCPUProfile(v.buf)
+		if err != nil {
+			log(ctx).With("cause", err).Warn("cannot start cpu PPROF")
+			delete(pprofConfigs.pcm, ProfileNameCPU)
+		}
+	}
+}
+
+// DumpPem dump a PEM version of the byte slice, bs, into writer, wrt.
+func DumpPem(bs []byte, types string, wrt *os.File) error {
```
julio-lopez, line 240: Does `wrt` need to be an `*os.File` ? or could it be an `io.Writer` ?

debug/debug.go
```diff
@@ -0,0 +1,386 @@
...
+		if k == ProfileNameCPU {
+			pprof.StopCPUProfile()
+			continue
+		}
+
+		_, ok := v.GetValue(KopiaDebugFlagForceGc)
+		if ok {
+			log(ctx).Debug("performing GC before PPROF dump ...")
+			runtime.GC()
+		}
```
julio-lopez, line 345: Should `KopiaDebugFlagForceGc` be a "global" flag that applies to all profiles?
or

Should `runtime.GC()` be only called once even if multiple profiles specify the flag?

Thoughts?

julio-lopez (approved):
No request for immediate changes.

@aaron-kasten There are a number of inline questions, PTAL.

Merging as is, we can come back and iterate.

debug/debug.go
aaron-kasten: No, it doesn't :)

aaron-kasten: Fixed.

debug/debug.go
aaron-kasten: The background encoding is a design to mitigate pass-through delays and conserve memory.

The buffered reader will block when its full.. this prevents using additional memory to keep the entire output.  The method of writing (lines at a time) is to prevent output from being clipped (outputting an entire PEM at once encourages clipping of the output - the process is much more reliable a line at a time).

debug/debug.go
julio-lopez: Asking because it is not clear to me whether a GC is needed before dumping each profile, or what difference it makes. To be clear, I don't have a good answer.

When stopping and dumping the profiles, it seems that a possible sequence would be

`GC` -> `dump profile 1` -> GC -> `dump profile 2` -> `GC` -> `dump profile 3`

In this case profile 2 would contain some information (allocations or side effects) coming from performing `dump profile 1`. Similarly, profile 3 would contain information coming from dumping profiles 1 and 2.

An alternative sequence would be.

`GC` -> `dump profile 1`, `dump profile 2` -> `dump profile 3`

Not sure if it would make a difference, or if it matters

debug/debug.go
julio-lopez: Wondering if this could be simplified, for example, with something along the lines of the following:

```go
	bw := bufio.NewWriterSize(wrt, someBufSize)

	if err := pem.Encode(bw, blk); err != nil {
		return err // TODO: wrap error
	}

	if err := bw.Flush(); err != nil {
		return err // TODO: wrap error
	}

	return nil
```

debug/debug.go
julio-lopez: Anyway 🤷🏼‍♂️

PR #3462, opened 2023-11-22
test(general): Fixup error output when server startup fails in robustness test 

aaron-kasten:
Fixup error output when server startup fails in robustness test


Shrekster (approved):

Shrekster:
The `TestIndexBlobManagerStress` test failure is a known flake. Rerunning.

PR #3465, opened 2023-11-23
fix(snapshots): optimise memory consumption for restores

### LINKED ISSUE #3460: High memory consumption at restore
Hello,

When using `kopia restore` we had up to about **20 GB of used memory**, which prevented us to restore.

Here is a capture of the heap allocation while restoring (from a subsequent restoration which is taking a bit less memory).

Screenshot from 2023-11-21 11-15-48

About the process of restoration: We are migrating the last snapshot  on the disk, and then we restore that snapshot. 

Why so much memory is used while restoring ? What can we do about it ?

arouene:
Code was stacking closures of metadata of the whole snapshot before starting to process the files.
This PR invert the logic by processing the directories breadth-first, freeing memory as the files are processed.

Resolves #3460.


renardguill:
👍

jkowalski:
In principle I like this, the only concern is progress reporting - it will likely mess up ETA estimation for restores.

Can you run this change before/after on some large directory and perhaps capture couple screenshots when this is running?

arouene:
I understand for the ETA, I have no stats on this for now. But on memory the win is clear. 

As I said, my playground is about 5TB, 
Screenshot from 2023-12-01 17-19-34

Our bigger folder is: 1 478 455 files
image

The memory consumption was about 10GB at least

image

And is now about 1GB (the screenshot is taken from the last restore today, which says 701MiB of heap)



Screenshot from 2023-12-01 11-45-40

julio-lopez:
@arouene what do the stats shown above belong to? (~5 TB, ~26 million files, ~18K directories)

Is that the total content in the repository or something else?

arouene:
We are running without the patch, and the memory consumption is acceptable. Besides in the current architecture, this change would be challenging.

RobKenis:
Hello @arouene 
I am currently running in this issue https://github.com/vmware-tanzu/velero/issues/8582 where memory consumption is not acceptable. 
Is the problem I am seeing related to this PR?

PR #3471, opened 2023-11-28
test(providers): add variables for Azure immutability testing

julio-lopez:


julio-lopez:
@KastenMike @Shrekster @jkowalski FYI

PR #3473, opened 2023-11-29
feat(cli): C: Add ability to dump pprof data to logs ...

aaron-kasten:
This PR adds the ability to dump pprof data to logs for debugging. 

This is a continuation of https://github.com/kopia/kopia/pull/3454

Usage
=====

example:

```
[1] 77308
```

```
$ kill %1
```

```
saving PEM buffers for output
dumping PEM for "PPROF MEM"
-----BEGIN PPROF MEM-----
.
.
.
//8psEjOrZ4AAA==
-----END PPROF MEM-----
```

```
$ ./go/bin/kats dump.b64
```

```
 ./go/bin/kats --help
 Usage of ./go/bin/kats:
  -verbose
    	verbose outout
```

```
File: pprof_mem.bin
Type: inuse_space
(pprof) tree
```

Configuration Options
=====

TBD


cli/command_snapshot_create.go
```diff
@@ -118,6 +119,7 @@ func (c *commandSnapshotCreate) run(ctx context.Context, rep repo.RepositoryWrit
 		return errors.New("description too long")
 	}
 
+	//nolint:contextcheck
```
denisvmedia, line 122: Why do we need this nolint directive here? it seems to be disabling the check for a non-inherited context, but where do we use it?

debug/debug.go
```diff
@@ -27,7 +27,7 @@ type ProfileName string
 
 const (
 	pair = 2
-	// PPROFDumpTimeout when dumping PPROF data, set an upper bound on the time it can take to log.
+	// PPROFDumpTimeout when dumping PPROF data, set an upper bound on the time it can take to log. (default 15s).
```
denisvmedia, line 30: nit: the newly added text seems to be excessive, considering that we can immediately the value of the const.

debug/debug.go
```diff
@@ -293,7 +302,7 @@ func DumpPem(bs []byte, types string, wrt *os.File) error {
 			return fmt.Errorf("could not write PEM: %w", err2)
 		}
 
-		return io.EOF
+		return nil
```
denisvmedia, line 305: Can you please explain this change? Why io.EOF is not ok anymore?

cli/command_server_start.go
```diff
@@ -217,25 +218,28 @@ func (c *commandServerStart) run(ctx context.Context) error {
...
-		}
-	})
+	c.svc.onSigTerm(func() { shutdownServer(ctx, httpServer, srv) })
 
-	c.svc.onRepositoryFatalError(func(_ error) {
-		if serr := httpServer.Shutdown(ctx); serr != nil {
-			log(ctx).Debugf("unable to shut down: %v", serr)
-		}
+	c.svc.onSigDump(func() {
+		debug.StopProfileBuffers(ctx)
```
Shrekster, line 237: Maybe log right before dumping to give a feedback that we got the signal ?

cli/command_snapshot_create.go
```diff
@@ -233,7 +235,39 @@ func (c *commandSnapshotCreate) setupUploader(rep repo.RepositoryWriter) *snapsh
...
+		// use new context as old one may have already errored out
+		// test changing this to run() context in future
+		ctx0, canfn := context.WithTimeout(context.Background(), debug.PPROFDumpTimeout)
+		defer canfn()
+
+		debug.StopProfileBuffers(ctx0)
+	})
+
+	c.svc.onSigDump(func() {
+		// no context passed into function.  Consider adding context to this function call
```
Shrekster, line 250: this is more of a GitHub issue than code comment ?

cli/command_server_start.go
```diff
@@ -217,25 +218,28 @@ func (c *commandServerStart) run(ctx context.Context) error {
 			return errors.Wrap(httpServer.Close(), "close")
 		}
 
+		rep := srv.GetRepository()
+		if rep != nil {
+			//nolint:errcheck
+			rep.Close(ctx)
```
Shrekster, line 224: I discussed with @aaron-kasten offline, just leaving a note here. This is probably going to race with the async init setup for openRepository() before this.

cli/command_server_start.go
```diff
@@ -273,6 +277,24 @@ func (c *commandServerStart) run(ctx context.Context) error {
...
+	if serr := httpServer.Shutdown(ctx); serr != nil {
+		log(ctx).Debugf("unable to shut down http server: %v", serr)
+	}
+
+	rep := srv.GetRepository()
+	if rep == nil {
+		return
+	}
+
+	if rerr := rep.Close(ctx); rerr != nil {
```
Shrekster, line 293: I am not sure whether Close() would be balanced with shutdownServer() teardown path. How important is it to close the repo during teardown ? What happens if we do not call close() ?

IMO a Close() should only be performed by someone who either does an addRef() o the repo object or who creates the repo object. Since, shutdownServer() did not open() it, so it should not be calling close() on it either. What do you think about this style of maintaining the resource construction and destruction layering, Aaron ?

cli/command_snapshot_create.go
```diff
@@ -233,7 +235,39 @@ func (c *commandSnapshotCreate) setupUploader(rep repo.RepositoryWriter) *snapsh
 		u.CheckpointInterval = interval
 	}
 
-	c.svc.onCtrlC(u.Cancel)
+	c.svc.onCtrlC(func() {
+		u.Cancel()
+
+		// use new context as old one may have already errored out
+		// test changing this to run() context in future
```
Shrekster, line 242: > test changing this to run() context in future

Lets file a GitHub issue for this :-)

cli/command_snapshot_create.go
```diff
@@ -233,7 +235,39 @@ func (c *commandSnapshotCreate) setupUploader(rep repo.RepositoryWriter) *snapsh
...
+		debug.StopProfileBuffers(ctx0)
+		// restart profile buffers as this does not kill the process
+		debug.StartProfileBuffers(ctx0)
+	})
+
+	c.svc.onSigTerm(func() {
+		u.Cancel()
+
+		// use new context as old one may have already errored out
+		// test changing this to run() context in future
```
Shrekster, line 265: > test changing this to run() context in future

same suggestion as above, belongs in a GitHub issue

cli/command_snapshot_migrate.go
```diff
@@ -44,13 +45,19 @@ func (c *commandSnapshotMigrate) setup(svc advancedAppServices, parent commandPa
...
 func (c *commandSnapshotMigrate) run(ctx context.Context, destRepo repo.RepositoryWriter) error {
 	sourceRepo, err := c.openSourceRepo(ctx)
 	if err != nil {
 		return errors.Wrap(err, "can't open source repository")
 	}
 
-	defer sourceRepo.Close(ctx) //nolint:errcheck
+	defer func() {
+		//nolint:errcheck
+		destRepo.Close(ctx)
```
Shrekster, line 57: We should let `destRepo` be closed by `app.go:App::maybeRepositoryAction`. WDYT ?

cli/command_snapshot_migrate.go
```diff
@@ -69,16 +76,41 @@ func (c *commandSnapshotMigrate) run(ctx context.Context, destRepo repo.Reposito
...
+		if canceled {
+			return
+		}
+
+		canceled = true
+
+		for s, u := range activeUploaders {
+			log(ctx).Infof("canceling active uploader for %v", s)
+			u.Cancel()
 		}
```
Shrekster, line 98: Nice refactor!

cli/config.go
```diff
@@ -8,6 +8,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
+	"syscall"
```
Shrekster, line 11: Use of syscall package in a multi-platform app might cause problems. You might want to create separate implementations for each platform or use generic signals I guess ?

cli/config.go
```diff
@@ -29,14 +30,26 @@ func (c *App) onRepositoryFatalError(f func(err error)) {
...
+}
+
+func onSig(chn chan bool, sig os.Signal, f func()) {
 	s := make(chan os.Signal, 1)
-	signal.Notify(s, os.Interrupt)
+	signal.Notify(s, sig)
 
 	go func() {
-		// invoke the function when either real or simulated Ctrl-C signal is delivered
+		// invoke the function when either real or simulated SIGTERM signal is delivered
```
Shrekster, line 50:
```suggestion
		// invoke the function when either real or simulated signal interrupt is delivered
```

debug/debug.go
```diff
@@ -361,22 +370,28 @@ func StopProfileBuffers(ctx context.Context) {
...
 		}
 	}
 	// dump the profiles out into their respective PEMs
 	for k, v := range pprofConfigs.pcm {
+		// process context
+		if ctx.Err() != nil {
+			log(ctx).With("cause", ctx.Err()).Error("cannot write PEM")
+			return
+		}
+
```
Shrekster, line 383: nit; What happens if we do not check the process ctx for an error in this loop ?

internal/server/server.go
```diff
@@ -563,6 +563,12 @@ func (s *Server) endUpload(ctx context.Context, src snapshot.SourceInfo) {
 	s.parallelSnapshotsChanged.Signal()
 }
 
+// GetRepository get the repository for this server.  This is used to ensure
+// repositories are closed on shutdown.  Also symmetric to SetRepository.
+func (s *Server) GetRepository() repo.Repository {
+	return s.rep
```
Shrekster, line 569: My suggestion would be to try to solve the `ensure repositories are closed on shutdown` problem in a separate PR, so that it doesn't get mixed u with your signal handling changes. It seems like the imbalanced closing of repos is an existing problem. So, unless we are introducing a new problem, we might eb able to do it separately without causing any breakage, what do you think ?

repo/api_server_repository.go
```diff
@@ -346,6 +355,13 @@ func openRestAPIRepository(ctx context.Context, si *APIServerInfo, password stri
 
 	rr.omgr = omgr
 
+	par.registerEarlyCloseFunc(func(ctx context.Context) error {
+		// release debugging resources as part of early debug.  This leaves the debugger
+		// a chance to use runtime resources before they too are released
+		rr.CloseDebug(ctx)
+		return nil
+	})
+
```
Shrekster, line 364: since `openRestAPIRepository()` calls `debug.StartProfileBuffers(ctx)`, I think it'd be okay to stop the profile buffers using `debug.StopProfileBuffers(ctx)`, that'd keep the layering consistent.

repo/grpc_repository_client.go
```diff
@@ -855,16 +862,17 @@ func openGRPCAPIRepository(ctx context.Context, si *APIServerInfo, password stri
...
 	if err != nil {
 		return nil, err
 	}
 
+	par.refCountedCloser.registerEarlyCloseFunc(
+		func(ctx context.Context) error {
+			rep.CloseDebug(ctx)
+			return errors.Wrap(conn.Close(), "error closing GRPC connection")
+		})
+
```
Shrekster, line 875: `newGRPCAPIRepositoryForConnection()` is the who initializes the profile buffers internally, correct ?

If so, would it be better if `newGRPCAPIRepositoryForConnection` does the callback registration for stopping the buffers internally ?

repo/open.go
```diff
@@ -367,6 +370,11 @@ func openWithConfig(ctx context.Context, st blob.Storage, cliOpts ClientOptions,
 		},
 	}
 
+	dr.registerEarlyCloseFunc(func(ctx context.Context) error {
+		dr.CloseDebug(ctx)
+		return nil
+	})
```
Shrekster, line 376: same suggestion as in the other place where CloseDebug() is used. Since, we are calling `debug.StartProfileBuffers` as this layer, it should be ok to call `debug.StopProfileBuffers` fro this layer as well. Thoughts ?

repo/repository.go
```diff
@@ -137,6 +139,10 @@ type directRepository struct {
 	afterFlush []RepositoryWriterCallback
 }
 
+func (r *directRepository) CloseDebug(ctx context.Context) {
+	debug.StopProfileBuffers(ctx)
+}
```
Shrekster, line 144: given how we have been calling `StartProfileBuffers`, I am not sure if you'll actually need to wrap `debug.StopProfileBuffers` in the repository object.

Shrekster (commented):
Changes look great, maybe a minor simplification could be done to get the layering consistent. Pretty solid PR otherwise! Will look again after your next update!

cli/command_server_start.go
```diff
@@ -212,6 +212,8 @@ func (c *commandServerStart) run(ctx context.Context) error {
 		// wait for all connections to finish for up to 5 seconds
 		log(ctx2).Debugf("attempting graceful shutdown for %v", c.shutdownGracePeriod)
 
+		debug.StopProfileBuffers(ctx)
```
denisvmedia, line 215: Is it not a mistake that ctx and not ctx2 is used?

debug/debug.go
aaron-kasten: I wanted to move away from `io.EOF`.  It had the advantage that it simulated the error response from a reader of the PEM data, but looked counter-intuitive to me.  So I finally moved to `nil` having the opinion was clearer.

cli/command_server_start.go
aaron-kasten: @Shrekster feedback is give within `StopProfileBuffers`

Shrekster (approved):
LGTM, please fix the tests and coverage issues.

cli/command_server_start.go
```diff
@@ -273,6 +273,19 @@ func (c *commandServerStart) run(ctx context.Context) error {
 	return errors.Wrap(srv.SetRepository(ctx, nil), "error setting active repository")
 }
 
+// shutdownServer shutdown http server and close the repository.
+//
+//nolint:revive,unparam
```
denisvmedia, line 278: Note, disabling revive completely disables _all_ the rules of revive. We should only disable particular rules, if we need. See the docs: https://github.com/mgechev/revive#comment-directives

aaron-kasten:
> @aaron-kasten I am going to review this PR before we merge it.
> 
> I would like to understand well the startup and shutdown approach here and see if it can be simplified.
> 
> Blocking merging for now.

@julio-lopez yes, please make your proposal!  The approach here is the "best" combination of tractability (understandable) and minimal code I could find.  Please feel free to propose another approach.

julio-lopez:
@aaron-kasten First a couple of questions to make sure I understand the issue(s) and then some suggestions and feedback.

*Question*: IIUC, the main objectives with the changes in this PR are:
- hook/wire the initialization and shutdown of pprof logging (facilities);
- set up a signal handler to trigger pprof dumps on demand.

is that correct? anything else I am missing?

*Feedback and suggestions*

- A requirement is that the pprof logging functionality should be _disabled by default_ and only be explicitly enabled when the environment variable is specified.
- In particular, the signal handler to trigger pprof dumps should only be installed if profiling is enabled, and not be installed otherwise.

- The initialization, shutdown and signal handling should be simplified if possible. See below.
- Denis has a couple of comments, please reply to those.
- Please make sure that linters are not broadly and unnecessarily disabled. The linters are in place to aid with the maintainability of the code base.
- Is the following configuration string valid? `cpu,heap=debug=1,mutex=debug=1:rate=1000`  Please add a unit test for it.

Regarding the initialization, PTAL at the POC kastenhq/kopia#536
Feel free to copy and use that as a basis in this PR.
I ran a "smoke" test, and for the most part seems to work. Here's what I did:

```
KOPIA_DEBUG_PPROF="cpu=,heap=debug=1,mutex=debug=1:rate=1000" ./kopia server start --insecure
```

and then `kill -USR1 <kopia-server-pid>` and was able to get a dump of one profile.

Also, terminating the process dumps to the logs one of the pprofs before the process exits.

Let's find a time to jump on a call and discuss further.

denisvmedia (changes requested):
Please use a proper linter rule for revive.

cli/app.go
```diff
@@ -486,6 +490,9 @@ func (c *App) runAppWithContext(command *kingpin.CmdClause, cb func(ctx context.
 		releasable.EnableTracking(releasable.ItemKind(r))
 	}
 
+	pproflogging.MaybeStartProfileBuffers(ctx)
+	defer pproflogging.MaybeStopProfileBuffers(ctx)
```
julio-lopez, line 494: on stop detach from the parent content
1. in case the parent context has errored out
2. to set a timeout for the dump

cli/config.go
```diff
@@ -30,8 +30,14 @@ func (c *App) onRepositoryFatalError(f func(err error)) {
 	c.onFatalErrorCallbacks = append(c.onFatalErrorCallbacks, f)
 }
 
+func (c *App) onDebugDump(f func()) {
+	onSig(c.simulatedSigDump, SignalDump, f)
+}
+
 func (c *App) onTerminate(f func()) {
-	s := make(chan os.Signal, 1)
+	// channel length of 2 to accommodate both signals
```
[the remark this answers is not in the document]
aaron-kasten: there's a note in `signal.c`

internal/pproflogging/debug.go
```diff
@@ -117,7 +320,8 @@ func (p ProfileConfig) GetValue(s string) (string, bool) {
 	return "", false
 }
 
-func parseProfileConfigs(bufSizeB int, ppconfigs string) map[ProfileName]*ProfileConfig {
+// parseProfileConfigs.
```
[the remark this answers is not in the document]
aaron-kasten: Good catch.  Done.

cli/config.go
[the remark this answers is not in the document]
julio-lopez: As you point out, this is a single-shot handler, so the rest of the comments below are not relevant to the change, mainly for discussion purposes.

This callback (channel) is only installed for `SIGTERM` and `SIGINT`.
`SIGKILL` is not interceptable as you know.

> ... _"has sufficient buffer space to keep up with the expected signal rate"_ ...

They way I understand the documentation you referenced above for `signal.Notify`, is that:
- The channel must be buffered (that is, buffer size > 0). If the channel is not buffered, then the message will not be added to the channel because the signal handler will not block on sending to the channel. That is, if the channel is unbuffered or full, then it will simply not send the message on the channel.
- If the routine reading from the channel cannot read fast enough to keep up with the signal rate, then messages will not be delivered to the channel, so the channel buffer size should be proportional to the maximum signal rate that is to be handled.

internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +103,228 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
...
+		log(ctx).Debug("no profile buffer configuration to start")
+		return
+	}
+
+	pprofConfigs.mu.Lock()
+	defer pprofConfigs.mu.Unlock()
+
+	pprofConfigs.pcm = pcm
+
+	log(ctx).Debug("no profile buffer configuration to start")
```
julio-lopez, line 126: Is this logging message accurate?

internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +103,228 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
...
+
+	pprofConfigs.StartProfileBuffersLocked(ctx)
+}
+
+// MaybeRestartProfileBuffers used by SIGQUIT signal handlers to output PPROF data without exiting.
+func MaybeRestartProfileBuffers(ctx context.Context) {
+	pprofConfigs.mu.Lock()
```
julio-lopez, line 133: This would `panic` when `pprofConfigs == nil`.

Having said that, `pprofConfigs` is set to a non-nil value at package initialization time, so it should never be `nil` unless it is changed later at runtime.

internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +103,228 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
...
+	defer pprofConfigs.mu.Unlock()
+
+	if pprofConfigs == nil || len(pprofConfigs.pcm) == 0 {
```
julio-lopez, line 136: This line would not be reached if `pprofConfigs == nil`, since it would have panicked above.

internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +103,228 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
...
+	pprofConfigs.StartProfileBuffersLocked(ctx)
+}
+
+// MaybeStopProfileBuffers stop and dump the contents of the buffers to the log as PEMs.  Buffers
+// supplied here are from MaybeStartProfileBuffers.
+func MaybeStopProfileBuffers(ctx context.Context) {
+	pprofConfigs.mu.Lock()
+	defer pprofConfigs.mu.Unlock()
+
+	if pprofConfigs == nil || len(pprofConfigs.pcm) == 0 {
```
julio-lopez, line 151: ditto regarding `pprofConfigs == nil`

tests/pprofdump/repositorymanipulator/repositorymanipulator.go
julio-lopez, line 1: This code is not being referenced anywhere else in this PR.
Let's leave it out of this PR and review it separately in its own PR.

Also, this looks like most of the functionality here is the same as the one in `tests/recovery/blobmanipulator/blobmanipulator.go`. Any reason for not using `blobmanipulator/BlobManipulator` ?

tests/pprofdump/pprofdump_test/main_test.go
julio-lopez, line 1: It looks like there are no tests yet in this package.
Please leave this file out of this PR.

julio-lopez (commented):
This is moving in the right direction and getting closer to "landing" it.

Please see inline comments.

Makefile
```diff
@@ -353,11 +353,11 @@ ifneq ($(GOOS),windows)
 endif
 
 htmlui-e2e-test:
-	HTMLUI_E2E_TEST=1 go test -timeout 600s github.com/kopia/kopia/tests/htmlui_e2e_test -v $(TEST_FLAGS)
+	HTMLUI_E2E_TEST=1 go test -timeout 900s github.com/kopia/kopia/tests/htmlui_e2e_test -v $(TEST_FLAGS)
 
 htmlui-e2e-test-local-htmlui-changes:
 	(cd ../htmlui && npm run build)
-	HTMLUI_E2E_TEST=1 HTMLUI_BUILD_DIR=$(CURDIR)/../htmlui/build go test -timeout 600s github.com/kopia/kopia/tests/htmlui_e2e_test -v $(TEST_FLAGS)
+	HTMLUI_E2E_TEST=1 HTMLUI_BUILD_DIR=$(CURDIR)/../htmlui/build go test -timeout 900s github.com/kopia/kopia/tests/htmlui_e2e_test -v $(TEST_FLAGS)
```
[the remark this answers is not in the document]
Shrekster: From the screenshots it seemed like a timeout, so he was trying out the bump. However, @aaron-kasten said there is a synchronization problem in the linux UI tests caused by his change. So, he will push another and probably drop this timeout bump now.

cli/app.go
```diff
@@ -486,6 +490,25 @@ func (c *App) runAppWithContext(command *kingpin.CmdClause, cb func(ctx context.
...
+	pproflogging.MaybeStartProfileBuffers(ctx)
+
+	defer func() {
+		ctx0, canfn := context.WithTimeout(ctx, pproflogging.PPROFDumpTimeout)
+		defer canfn()
+
+		pproflogging.MaybeStopProfileBuffers(ctx0)
+	}()
+
+	c.onDebugDump(func() {
```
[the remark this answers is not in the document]
Shrekster: Actually this seems like a simple change, why not take it in now ? If it is more than the env check then please defer.

cli/command_server_start.go
```diff
@@ -273,6 +274,15 @@ func (c *commandServerStart) run(ctx context.Context) error {
...
 
+// shutdownServer shutdown http server and close the repository.
+func shutdownServer(ctx context.Context, httpServer *http.Server) {
+	log(ctx).Infof("Shutting down...")
+
+	if serr := httpServer.Shutdown(ctx); serr != nil {
+		log(ctx).Debugf("unable to shut down http server: %v", serr)
+	}
+}
+
```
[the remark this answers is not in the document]
Shrekster: @aaron-kasten do you need this function for `onRepositoryFatalError` ? Can we split apart the `onRepositoryFatalError` trigger to a separate PR or is that required for this PR ?

internal/logfile/logfile_test.go
```diff
@@ -207,7 +207,7 @@ func verifyFileLogFormat(t *testing.T, fname string, re *regexp.Regexp) {
 	s := bufio.NewScanner(f)
 
 	for s.Scan() {
-		require.True(t, re.MatchString(s.Text()), "log line does not match the format: %v (re %v)", s.Text(), re.String())
+		require.True(t, re.MatchString(s.Text()), "log line does not match the format: %q (re %q)", s.Text(), re.String())
```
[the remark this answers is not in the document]
aaron-kasten: `s.Text()` and `re.String()` both return `string`s.  `%q` will quote the string, visually separating it from the rest of the message.

The old output looked something like:

```
log line does not match the format: example log line to check (re ^ check$)
```

the new log line looks like:

```
log line does not match the format: "example log line to check" (re "^ check$")
```

internal/pproflogging/pproflogging.go
aaron-kasten: Correct.  If someone made the change than the function would fail, and the tests would fail, signaling to the developer that they made an error - which seems correct.  The code seems safe to me.

tests/pprofdump/repositorymanipulator/repositorymanipulator.go
aaron-kasten: @julio-lopez correct.  blobmanipulaor is the source of pprofdumps blobmanipulator.  Its separate because I expected its implementation to deviate from the "recovery" blobmanipulator as the testing for pprofdumps expanded...

and Yes, this file is for future tests.... there are no current references.  I'll remove.

internal/logfile/logfile_test.go
```diff
@@ -110,8 +110,8 @@ func TestLogFileRotation(t *testing.T) {
 	tmpLogDir := testutil.TempDirectory(t)
 
 	env.RunAndExpectSuccess(t, "snap", "create", dir1,
-		"--file-log-local-tz", "--log-level=error", "--file-log-level=debug",
-		"--max-log-file-segment-size=1000", "--log-dir", tmpLogDir, "--log-dir-max-files=3", "--content-log-dir-max-files=4")
+		"--file-log-local-tz", "--log-level=error", "--file-log-level=info", "--wait-for-log-sweep",
```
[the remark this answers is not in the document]
aaron-kasten: This is a change to account for pprof logf output on termination.   This will be reverted.

cli/command_server_start.go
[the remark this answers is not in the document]
aaron-kasten: @julio-lopez @Shrekster yes... this can easily be put into a separate PR.  This is really here for better logging/notification in the logs. (I have a strong preference to capture all errors and log where appropriate).

internal/pproflogging/pproflogging.go
aaron-kasten: fixed

cli/command_repository_sync.go
```diff
@@ -242,6 +242,7 @@ func (c *commandRepositorySyncTo) runSyncBlobs(ctx context.Context, src blob.Rea
 
 				numBlobs, bytesCopied := totalCopied.Add(m.Length)
 				progressMutex.Lock()
+
```
Shrekster, line 245: nit; drop this addition of line

tests/pprofdump/pprofdump_test/main_test.go
```diff
@@ -0,0 +1,46 @@
+//go:build darwin || linux
```
Shrekster, line 1: This can become a separate PR.

tests/pprofdump/pprofdump_test/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+func TestMain(m *testing.M) {
+	dataRepoPath := path.Join(*repoPathPrefix, dataSubPath)
+
+	th := &kopiaPprofDumpRepositoryConnectTest{}
+	th.init(dataRepoPath)
+
+	// run the tests
+	result := m.Run()
+
+	os.Exit(result)
```
Shrekster, line 31: We should check that the dump was created after SIGTERM.

tests/testenv/cli_exe_runner.go
```diff
@@ -54,7 +54,6 @@ func (e *CLIExeRunner) Start(t *testing.T, args []string, env map[string]string)
 	return stdoutPipe, stderrPipe, c.Wait, func(sig os.Signal) {
 		if sig == os.Kill {
 			c.Process.Kill()
-
```
Shrekster, line 57: nit; drop this change

PR #3517, opened 2023-12-18
test(general): fix race in TestIndexEpochManager_NoCompactionInReadOnly

julio-lopez:


internal/epoch/epoch_manager_test.go
```diff
@@ -393,26 +393,26 @@ func TestIndexEpochManager_NoCompactionInReadOnly(t *testing.T) {
...
+		if err := te2.mgr.Refresh(ctx); err != nil {
+			loadedErr.Store(err)
+		}
 
-		loadedErr = te2.mgr.Refresh(ctx)
 		te2.mgr.backgroundWork.Wait()
+		loadedDone.Store(true)
 	}()
 
-	if !assert.Eventually(t, func() bool { return loadedDone }, time.Second*5, time.Second) {
```
julio-lopez, line 409: `loadDone` is being red here without any synchronization that ensures the value is propagated across goroutines.

internal/epoch/epoch_manager_test.go
```diff
@@ -393,26 +393,26 @@ func TestIndexEpochManager_NoCompactionInReadOnly(t *testing.T) {
...
+	if !assert.Eventually(t, loadedDone.Load, time.Second*5, time.Second) {
 		// Return early so we don't report some odd failure on the error check below
 		// when we just never managed to initialize the epoch manager.
 		return
 	}
 
-	assert.NoError(t, loadedErr, "refreshing read-only index")
+	assert.Nil(t, loadedErr.Load(), "refreshing read-only index")
```
julio-lopez, line 415: technically, `loadedErr` does not need to be an `atomic.Value` since this read _"happens after" (and only) after `loadedDone` has been set to true and `loadedErr` is set before `loadedDone` is set. Nevertheless, making this an atomic value since there are no explicit locks and the ordering of the reads could change in the future and a race could be inadvertently introduced.

julio-lopez:
@ashmrtn FYI

PR #3520, opened 2023-12-18
fix(providers): change versioning check for Azure storage

KastenMike:
It is unusable if `kopia.respository` is deleted. This changes it to work with deleted or not

### DIFF
--- a/repo/blob/azure/azure_pit.go
+++ b/repo/blob/azure/azure_pit.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/blob/readonly"
-	"github.com/kopia/kopia/repo/format"
 )
 
 type azPointInTimeStorage struct {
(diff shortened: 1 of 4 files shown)


repo/blob/azure/azure_pit.go
```diff
@@ -176,26 +175,13 @@ func (az *azPointInTimeStorage) isAzureDeleteMarker(it *azblobmodels.BlobItem) b
...
-	props, err := s.service.ServiceClient().
-		NewContainerClient(s.container).
-		NewBlobClient(s.getObjectNameString(format.KopiaRepositoryBlobID)).
-		GetProperties(ctx, nil)
-	if err != nil {
-		return nil, errors.Wrapf(err, "could not get determine if container '%s' supports versioning", s.container)
-	}
-
-	if props.VersionID == nil {
-		return nil, errors.Errorf("cannot create point-in-time view for non-versioned container '%s'", s.container)
```
julio-lopez, line 194: Can the check be replaced with a _"list versions"_ for the repro prefix?

julio-lopez (changes requested):

repo/blob/azure/azure_pit.go
```diff
@@ -181,21 +184,17 @@ func maybePointInTimeStore(ctx context.Context, s *azStorage, pointInTime *time.
...
-	if err != nil {
-		return nil, errors.Wrapf(err, "could not get determine if container '%s' supports versioning", s.container)
+	pit := &azPointInTimeStorage{
+		azStorage:   *s,
+		pointInTime: *pointInTime, // not used for the check
 	}
 
-	if props.VersionID == nil {
-		return nil, errors.Errorf("cannot create point-in-time view for non-versioned container '%s'", s.container)
+	err := pit.getBlobVersions(ctx, format.KopiaRepositoryBlobID, func(vm versionMetadata) error {
```
julio-lopez, line 192: Mike: how does these behave in the following scenarios:

- the container does not have versioning enabled?
- the container has versioning enabled?

Please add unit tests for these scenarios, for `maybePointInTimeStore` and `az.getVersionedBlobMeta`, to run them as part of the "core provider tests".

Thanks.

KastenMike: the container does not have versioning enabled? => PIT fails. Added a test case

the container has versioning enabled? => it works. the existing test cases cover this although I modified it to delete the kopia.repository file to ensure it still works when the file is deleted

repo/blob/azure/azure_versioned_test.go
```diff
@@ -19,6 +19,46 @@ import (
 	"github.com/kopia/kopia/repo/format"
 )
 
+func TestGetBlobVersionsFailsWhenVersioningDisabled(t *testing.T) {
```
julio-lopez, line 22: Test seems to have passed
https://github.com/kopia/kopia/actions/runs/7294187986

julio-lopez (approved):

PR #3527, opened 2023-12-19
build(site): upgrade Go version to 1.21 in Netlify

julio-lopez:


site/.go-version
```diff
@@ -1 +1 @@
-1.19
+1.21.5
```
julio-lopez, line 1: Note: it was necessary to specify the patch version (.5) for Netlify to actually use the version. Specifying `1.21` resulted in the version not being found and falling back to the default 1.19 version used to build the website.

julio-lopez:
@lupusA FYI

julio-lopez:
We'll add something to CI to ensure this is consistent with the `go.mod` file as we update it.

Issue #3449, opened 2023-11-16
Can't access restricted directories on win32 (easy fix)


carlfarrington:
Where available, Kopia should invoke its right to the SeBackupPrivilege token on win32, in the same way that Robocopy does for its Backup Mode.

For a powershell script that shows the Win32 calls requred, see https://github.com/gtworek/PSBits/blob/master/Misc/EnableSeBackupPrivilege.ps1

I am currently working around this by calling a similar PowerShell script to add this token to a PowerShell or cmd.exe session which then invokes Kopia.

Without this, Kopia can't be relied upon to backup a shared folder in a business environment. All it takes is for a person to set explicit ACLs on a subfolder of their choice, for Kopia to be denied acces. The users never add 'Administrator', or 'Backup User' when they are deciding to limit access to a folder within the main company shared folder.

carlfarrington:
Apologies, I mean 'easy fix for someone who has a suitable build environment and knows what they're doing' (not me !)

PR #3543, opened 2023-12-28
feat(snapshots): Implement volume shadow copy support on Windows

### LINKED ISSUE #2662: Seemingly Kopia locks files when it is backing them up
Thus, I saw the following - and discovered that a scheduled Kopia job was running:

image

This is bad, because (1) the user might be confused, at least initially, and (2) if the user wishes to proceed with the deletion, she will either have to wait or else abort the backup job.

Windows 10 Pro - 22H2 (2009: 19045)
Kopia 0.12.1

### LINKED ISSUE #3371: [Feature] Improve the creation of VSS Volume Shadow Copies
Enabling VSS can provide write-on-copy and improved backup consistency (from inconsistent-backup to crash-consistent-backup), it would be highly beneficial if improvements could be made or enabled by default

1. allow parallel creation of snapshots after enabling
2. compatible with unicode characters, it now always reports `ERROR upload error: The system cannot find the path specified` if the snapshot path contains some unicode characters
3. unneeded shadow copies are prone to residual
4. configure it in policy `--enable-vss=true`

mxk:
This is another attempt at adding Volume Shadow Copy Service (VSS) support on Windows. See #2334, #2662, and #3371 for background info and reasons why this is a desirable feature. This version is the most minimal implementation so that I can get an idea of whether this approach is acceptable. If so, I'll make a few improvements and add the policy option to the UI.

Unlike the [example action scripts](https://kopia.io/docs/advanced/actions/#windows-shadow-copy), this uses the `DeviceObject` path of the shadow copy directly, which removes the need to link the shadow copy anywhere in the file system, and makes the backup even more robust. To be fair, the linking step in the script probably could've been removed as well, assuming that Kopia doesn't have any issues handling `\\?\` paths (everything seems to work fine in my testing).

### Usage

```
$ kopia policy set <path> --enable-volume-shadow-copy=always
```

OR

```
$ kopia policy set <path> --enable-volume-shadow-copy=when-available
```

**Note:** One issue that affects all uses of VSS is the need to run kopia.exe as an elevated user who is a member of the Administrators group. Without that, the snapshot will fail (unless `when-available` is specified, in which case VSS won't be used at for the snapshot)

Fixes #2662, fixes #3371


snapshot/policy/actions_policy.go
```diff
@@ -11,12 +11,16 @@ type ActionsPolicy struct {
 	// commands run once before and after each snapshot root (can be inherited).
 	BeforeSnapshotRoot *ActionCommand `json:"beforeSnapshotRoot,omitempty"`
 	AfterSnapshotRoot  *ActionCommand `json:"afterSnapshotRoot,omitempty"`
+
+	// use Windows Volume Shadow Copy Service (can be inherited).
+	VSS *OptionalBool `json:"vss,omitempty"`
```
jkowalski, line 16: it would be good to consider future direction where we may want to add first-class support for ZFS, LVM and other types of snapshots (some of which may need more than a simple boolean to enabled/disable them).

I would recommend making this more generic and possible to add parameters, and perhaps create a whole new section for OS-specific point-in-time snapshotting of filesystems and block devices, like so:

```
"osSnapshot": {
  "volumeShadowCopy": {
    "enabled": true,
    // any VSS parameters go here.
  },
  "zfs": {
   "enabled": true,
   "mountPath": "..."
  },
  "lvm": {
   "enabled": true,
  },
  "btrsfs": {
   "enabled": true,
  }
},
```

This may be nested in `actions` or parallel to it, not sure which one is better.

cli/command_policy_set_actions.go
```diff
@@ -34,6 +35,7 @@ func (c *policyActionFlags) setup(cmd *kingpin.CmdClause) {
 	cmd.Flag("action-command-timeout", "Max time allowed for an action to run in seconds").Default("5m").DurationVar(&c.policySetActionCommandTimeout)
 	cmd.Flag("action-command-mode", "Action command mode").Default("essential").EnumVar(&c.policySetActionCommandMode, "essential", "optional", "async")
 	cmd.Flag("persist-action-script", "Persist action script").BoolVar(&c.policySetPersistActionScript)
+	cmd.Flag("vss", "Enable Volume Shadow Copy Service integration ('true', 'false', 'inherit')").EnumVar(&c.policySetVSS, booleanEnumValues...)
```
jkowalski, line 38: nit: it may be good to avoid acronyms, how about `enable-volume-shadow-copy`.

Also, should this be an enum:

- `always` - always uses VSS, fails if it can't create VSS snapshot
- `when-available` - uses VSS when available (running as admin), falls back to creating snapshot without VSS when not running as administrator
- `never` - never uses VSS
- `inherit`

cli/command_policy_show.go
```diff
@@ -420,6 +420,17 @@ func appendActionsPolicyRows(rows []policyTableRow, p *policy.Policy, def *polic
 		rows = append(rows, policyTableRow{"No actions defined.", "", ""})
 	}
 
+	if isWindows() {
```
jkowalski, line 423: we generally don't generate OS-specific output, one can be looking at a policy belonging to a different user/host, and output should not change whether we run on Windows or Linux.

jkowalski:
I took a pass at the PR, looks very promising, it's important to get CLI and policy configuration experience right so we can in the future expand it to support other snapshot mechanisms.

BTW thanks for creating go-vss. Many projects need something like this.

mxk:
@jkowalski Thanks for your feedback! I made all the suggested changes. Please take another look.

I set the default setting to `when-available` because I can't think of any reason why Kopia shouldn't use VSS by default, if possible. To ensure compatibility for all the existing users who have before/after snapshot actions configured, a shadow copy is created only if the root directory does not already refer to an existing shadow copy. A warning is logged in this case to get people to remove their action scripts.

jkowalski:
This is looking great.

You still have some linter failures. I recommend `make lint-all` locally which will run linter in all OS configurations (this takes a little while).

Also, we need more tests - related to CLI flag setting and corresponding `policy show` output - there is a number of those you can tweak.

I'm curious whether we can actually use VSS on Github Runner when running tests - it would be great do write a Windows-specific test that ensures we can snapshot files that are normally locked. A quick search reveals that elevated UAC is available - we just need to find a way to trigger this when running tests.

We should also add tests for the behavior when VSS is not available (not running elevated).

mxk:
Linter failures should be fixed and I added policy set/show tests.

So far, I have not been able to run kopia.exe with elevated privileges from an integration test. By using a manifest file, I can set `dist\testing_windows_amd64\kopia.exe` to require admin privileges, but when I try to run it from a test, I get "The requested operation requires elevation" error rather than a UAC prompt. I've also tried getting a linked token via [wintoken](https://github.com/FourCoreLabs/wintoken), but that didn't work either. If anyone has other ideas, please let me know.

mxk:
I added an integration test that changes behavior depending on whether it is running with elevated privileges or not, and then used [gsudo](https://github.com/gerardog/gsudo) in the action workflow to test both conditions.

mxk:
@jkowalski Could you please take another look?

LakeishaKowalczyk:
@jkowalski 
Hello, may I ask how the progress here is going? This feature is really excellent and I’ve been longingly anticipating this feature

jkowalski:
sorry, missed that, taking a look now...

snapshot/snapshotfs/upload_os_snapshot_windows.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/kopia/kopia/fs/localfs"
+	"github.com/kopia/kopia/snapshot/policy"
+)
+
+func osSnapshotMode(p *policy.OSSnapshotPolicy) policy.OSSnapshotMode {
+	return p.VolumeShadowCopy.Enable.OrDefault(policy.OSSnapshotNever)
+}
+
+//nolint:wrapcheck
+func createOSSnapshot(ctx context.Context, root fs.Directory, _ *policy.OSSnapshotPolicy) (newRoot fs.Directory, cleanup func(), err error) {
```
jkowalski, line 20: nit: can we rename 'err' to something unique here to avoid confusion something like `finalErr`, this would be more obvious in `defer`

jkowalski:
this is looking awesome, i have 2 nits - but there are still some errors in the windows run which look like real issues with long filenames

jkowalski:
cc @julio-lopez for visibility, once we merge this, we could then similarly add ZFS, BTRFS and LVM snapshots

mxk:
@jkowalski Thanks for taking a look! I made both changes. As far as I can tell, there are two Windows failures. The first has to do with the logfile_test using `DEBUG|INFO` regex to match log output:

https://github.com/kopia/kopia/actions/runs/7479769611/job/21139454308#step:8:62

Is it ok to update the [regex](https://github.com/mxk/kopia/blob/vss/internal/logfile/logfile_test.go#L23-L25) to also match `WARN`?

I believe the other tests fail because [`MaybePrefixLongFilenameOnWindows`](https://github.com/mxk/kopia/blob/vss/internal/atomicfile/atomicfile.go#L33-L38) is stripping away the `\\?\` prefix and then just returning that path to [`os.ReadFile`](https://github.com/mxk/kopia/blob/vss/fs/localfs/shallow_fs.go#L67), which fails.

https://github.com/kopia/kopia/actions/runs/7479769611/job/21139454308#step:8:1259
https://github.com/kopia/kopia/actions/runs/7479769611/job/21139454308#step:8:2439

I don't think I understand why this was being done in the first place, so I I'm not sure what the best fix is. Should I add a special case to leave the path alone if it starts with `\\?\GLOBALROOT\`?

jkowalski:
> @jkowalski Thanks for taking a look! I made both changes. As far as I can tell, there are two Windows failures. The first has to do with the logfile_test using `DEBUG|INFO` regex to match log output:
> 
> https://github.com/kopia/kopia/actions/runs/7479769611/job/21139454308#step:8:62
> 
> Is it ok to update the [regex](https://github.com/mxk/kopia/blob/vss/internal/logfile/logfile_test.go#L23-L25) to also match `WARN`?
> 
> I believe the other tests fail because [`MaybePrefixLongFilenameOnWindows`](https://github.com/mxk/kopia/blob/vss/internal/atomicfile/atomicfile.go#L33-L38) is stripping away the `\\?\` prefix and then just returning that path to [`os.ReadFile`](https://github.com/mxk/kopia/blob/vss/fs/localfs/shallow_fs.go#L67), which fails.
> 
> https://github.com/kopia/kopia/actions/runs/7479769611/job/21139454308#step:8:1259 https://github.com/kopia/kopia/actions/runs/7479769611/job/21139454308#step:8:2439
> 
> I don't think I understand why this was being done in the first place, so I I'm not sure what the best fix is. Should I add a special case to leave the path alone if it starts with `\\?\GLOBALROOT\`?

Let's try that, or perhaps simply return any paths that start with "\\?\" as-is?

mxk:
@jkowalski Done. Can you re-run the tests, please?

jkowalski:
looks like there's some windows-specific regression in tests now, it likely is related to long filename rewriting

mxk:
Yep, I somehow missed the call to `wrapIgnorefs` on `overrideDir`, so the path comparison didn't work correctly. I also wasn't running the test as an admin, so it was passing for me since it just didn't use VSS.

I'm not 100% sure how to handle `TestLoggingFlags` failing due to `Another shadow copy operation is already in progress`. I'm assuming it's an intermittent error that happens due to some other tests running in parallel. For now, I added some retry logic for this error, which is probably a good idea anyway, but the test may still fail. Another option is to just disable OS snapshots altogether for some tests.

jkowalski (approved):
LGTM, this is a really awesome contribution.

jkowalski:
@mxk looks like there is currently some test flakiness in the master branch after we merged this. Would you be able to investigate the failures and fix them? Also, would be great if you could join http://slack.kopia.io 

See #3628 and #3624

mxk:
@jkowalski I'm on Slack now and just looked through the issues and PRs.

As far as I can tell, the original issue was `TestServerScheduling` failing because VSS snapshots sometimes take 10+ seconds to create, which is normal, and this test was measuring the number of Kopia snapshots that can be taken in 10 seconds (#3624). I would prefer to disable VSS for this test and any others like it, or make `never` the default setting for just the integration tests without changing the actual default global policy.

Globally disabling VSS (#3625) defeats the purpose of `when-available` setting, which is to allow the use of VSS on a best-effort basis, especially for everyone who is currently using PowerShell actions for it. This change then caused `TestShadowCopy` to start failing since it was assuming a default of `when-available` (#3628 and #3629).

The original issue also reports a race failure in go-ole, which seems to be caused by how it handles `*string` output parameters in WMI calls. This is only used in one place when I call `Win32_ShadowCopy.Create` and can't be avoided. I don't think it is causing any real issues, but I definitely don't like it doing something that's not strictly correct as far as Go pointer rules are concerned. I'll see if I can contribute a fix for that.

Is there anything else that I missed?

@julio-lopez @aaron-kasten @Shrekster

jkowalski:
Can we track all these action items in a separate issue with ultimate goal of re-enabling 'when-available' once all the underlying failures are addressed and test stability is restored?

akash07k:
Awesome feature addition. I was eyeing on it since a long. where can I find the documentation for this new feature? I can't yet find it in official docs.

LinuxOnTheDesktop:
So, hold on (if I may chime in): is 'shadow copy' now enabled by default on Windows, or not?

mxk:
VSS is currently disabled by default because it was causing issues for some tests. Issue #3636 was created to reenable it. You must currently use ‘policy set --enable-volume-shadow-copy when-available’ (or ‘always’) to enable it. Documentation also still needs to be updated.

basldfalksjdf:
IMO vss should be left off by default. It is too new of a feature to force everyone to use it.

dbolton:
Is there a way to enable the volume shadow copy on KopiaUI? My current backups are incomplete without this. I tried the [before/after scripts](https://kopia.io/docs/advanced/actions/#windows-shadow-copy) but didn't quite understand how to set them up to work correctly.

mxk:
@dbolton Not yet. It can only be enabled via the command line for now.

carlfarrington:
I suppose (hope) this fixes #3449 too ?

carlfarrington:
> I suppose (hope) this fixes #3449 too ?

Now that I think about it, it probably doesn't.
I was previously using before.ps1 and after.ps1 to create VSS snapshots, and in that scenario, the snapshots still did not allow access to restricted directories.

I'm quite surprised nobody else runs into this issue, backing up Windows fileservers and Kopia not being able to access directories that the backup-user does not explicitly have access to.

My workaround to run Kopia with adjusted token privileges to include SeBackupPrivilege is .. quite complicated.

PR #3398, opened 2023-10-22
docs(site): Add Azure ransomware protection section

KastenMike:
Doc update for https://github.com/kopia/kopia/pull/3394

image

image

PR #3564, opened 2024-01-11
feat(repo)gcs: add support for versioning and retention

guestisp:
still a WIP but this PR aim to add support for versioning and retention for malware protection to the native Google Cloud Storage backend

backup and retention extension are working as expected
point in time still doesn't work, probably i need some help in this regard.


guestisp:
Can someone help a bit ?

repo/blob/gcs/gcs_storage.go
```diff
@@ -39,13 +39,28 @@ type gcsStorage struct {
 }
 
 func (gcs *gcsStorage) GetBlob(ctx context.Context, b blob.ID, offset, length int64, output blob.OutputBuffer) error {
+	return gcs.getBlobWithVersion(ctx, b, latestVersionID, offset, length, output)
+}
+
+// getBlobWithVersion returns full or partial contents of a blob with given ID and version.
+func (gcs *gcsStorage) getBlobWithVersion(ctx context.Context, b blob.ID, version string, offset, length int64, output blob.OutputBuffer) error {
+	fmt.Printf("getBlobWithVersion: blob %s versione %s\n", b, version)
```
KastenMike, line 47:
```suggestion
	fmt.Printf("getBlobWithVersion: blob %s version %s\n", b, version)
```

repo/blob/gcs/gcs_storage.go
```diff
@@ -122,28 +162,36 @@ func (gcs *gcsStorage) PutBlob(ctx context.Context, b blob.ID, data blob.Bytes,
 	writer.ContentType = "application/x-kopia"
 	writer.ObjectAttrs.Metadata = timestampmeta.ToMap(opts.SetModTime, timeMapKey)
 
+	if opts.RetentionPeriod != 0 {
+		retainUntilDate := clock.Now().Add(opts.RetentionPeriod).UTC()
+		writer.ObjectAttrs.Retention = &storage.ObjectRetention{
+			Mode:        "Unlocked",
```
KastenMike, line 168:
```suggestion
			Mode:        blob.Locked,
```

repo/blob/gcs/gcs_storage.go
```diff
@@ -155,6 +203,26 @@ func (gcs *gcsStorage) DeleteBlob(ctx context.Context, b blob.ID) error {
 	return err
 }
 
+func (gcs *gcsStorage) ExtendBlobRetention(ctx context.Context, b blob.ID, opts blob.ExtendOptions) error {
+	retentionMode := "Unlocked" // TODO: properly handle lock mode
```
KastenMike, line 207:
```suggestion
	retentionMode := blob.Locked
```

repo/blob/gcs/gcs_storage.go
```diff
@@ -264,12 +298,18 @@ func New(ctx context.Context, opt *Options, isCreate bool) (blob.Storage, error)
...
 	}
 
-	gcs := &gcsStorage{
+	st := &gcsStorage{
 		Options:       *opt,
 		storageClient: cli,
 		bucket:        cli.Bucket(opt.BucketName),
 	}
 
+	fmt.Println("Provo il PIT")
```
KastenMike, line 307: in general the Println's can go

repo/blob/gcs/gcs_pit.go
```diff
@@ -0,0 +1,174 @@
...
+		return versionMetadata{}, false
+	}
+
+	//v = vs[0] // versione s3
+	v = vs[len(vs)-1] // versione azure
+
+	fmt.Printf("newestAtUnlessDeleted del pit %s deleted=%t\n", v.Timestamp, v.IsDeleteMarker)
+
+	//return v, !v.IsDeleteMarker
+	return v, true
```
KastenMike, line 129: here you assume a blob is never deleted

repo/blob/gcs/gcs_pit.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// Removes versions that are newer than t. The filtering is done in place
+// and uses the same slice storage as vs. Assumes entries in vs are in descending
+// timestamp order.
+func getOlderThan(vs []versionMetadata, t time.Time) []versionMetadata {
+	fmt.Printf("getOlderThan %s \n", t)
+
+	sort.Slice(vs, func(i, j int) bool { return vs[i].Timestamp.Before(vs[j].Timestamp) })
```
KastenMike, line 138: You're sorting both here and in `newestAtUnlessDeleted`

repo/blob/gcs/gcs_storage.go
KastenMike: I'll look into it Today, see if I notice anything

kreare: main issue i had is that GCS doesn't allow sorting object versions by creation time so i had to list all versions, sort manually by creation time, cycle the result and keep the newestthan (or older then). but im not a Go programmer, i had huge issues trying to sort and debug this thing. i have an apparently working version on my laptop but i'm not confident on why It works 🤣

(probably on Discord i can explain better)

KastenMike: So a couple notes from looking at it.
- blob versions are automatically sorted from oldest to most recently created so no need to add sorting anywhere
- `IsDeleteMarker` needs to consider not only if `Deleted` is set but also if it was deleted after the PIT e.g. `!oi.Deleted.IsZero() && (gcs.PointInTime == nil || oi.Deleted.Before(*gcs.PointInTime))`. For convenience you can change `infoToVersionMetadata` to be a method instead of func
- `newestAtUnlessDeleted` needs to return `!v.IsDeleteMarker` as mentioned 
- `getOlderThan` needs `return vs[:i]` like Azure since they're ascending 

If you make these changes then PIT will work as expected

Not really to PIT in particular:
I guess better if Timestamp is .Created rather than .Updated

kreare: > blob versions are automatically sorted from oldest to most recently created so no need to add sorting anywhere

Not with Google Cloud Storage, that's why i wasn't unable to get this working. Google doesn't have a stable sort currently.

kreare: Something link this seem to fix the sorting and also fix getting ALL versions before PIT and the LATEST version BEFOR PIT (i did some test on simple arrays, not on object from google cloud). An unsorted array like `[]int{2, 7, 14, 4, 9, 22, 23, 6}` returns `[2,4,6]` from `getOlderThan` and `6` from `newestAtUnlessDeleted` supposing a PIT set to `7`
```
// newestAtUnlessDeleted returns the last version in the list older than the PIT.
func newestAtUnlessDeleted(vs []versionMetadata, t time.Time) (v versionMetadata, found bool) {
	fmt.Printf("newestAtUnlessDeleted %s\n", t)

	vs = getOlderThan(vs, t)

	if len(vs) == 0 {
		return versionMetadata{}, false
	}

	v = vs[len(vs)-1]
	fmt.Printf("newestAtUnlessDeleted del pit %s deleted=%t\n", v.Timestamp, v.IsDeleteMarker)

	//return v, !v.IsDeleteMarker
	return v, true
}

// Removes versions that are newer than t. The filtering is done in place
// and uses the same slice storage as vs. Assumes entries in vs are in descending
// timestamp order.
func getOlderThan(vs []versionMetadata, t time.Time) []versionMetadata {
	fmt.Printf("getOlderThan %s \n", t)

	sort.Slice(vs, func(i, j int) bool { return vs[i].Timestamp.Before(vs[j].Timestamp) })

	for i := range vs {
		fmt.Printf("getOlderThan %s: vs[i].Timestamp: %s \n", t, vs[i].Timestamp)
		if vs[i].Timestamp.After(t) {
			fmt.Printf("  Skippo versione %s del %s. Troppo nuova.\n", vs[i].Version, vs[i].Timestamp)
			return vs[:i]
		}
	}

	return vs
}
```

KastenMike: > > blob versions are automatically sorted from oldest to most recently created so no need to add sorting anywhere
> 
> Not with Google Cloud Storage, that's why i wasn't unable to get this working. Google doesn't have a stable sort currently.

Every time I call list blobs like this: `$HOME/go/bin/kopia blob list --prefix=untitled`
I get all 4 versions of the blob in ascending order at the point of "newestAtUnlessDeleted: PRE SORT"
If it wasn't stable then at least once it would vary unless you are sorting before `PRE SORT` also

KastenMike: You can see the changes I mentioned [here](https://github.com/KastenMike/kopia/pull/10/files)

`return v, true` is not correct since it would return deleted blobs e.g. PIT=15:00, blob deleted 14:00

KastenMike: I @'d you at the lines since I realised I didn't do them in a separate commit

KastenMike:
```
****found blob: untitled file -- om.BlobID: untitled file -- query.Prefix: untitled -- prefix: untitled
newestAtUnlessDeleted 2024-08-29 15:04:05 +0000 UTC with 6 versions
newestAtUnlessDeleted: PRE SORT: found version 1724921148084300 del 2024-08-29 08:45:48.088 +0000 UTC. deleted=true
newestAtUnlessDeleted: PRE SORT: found version 1724921200485667 del 2024-08-29 08:46:40.497 +0000 UTC. deleted=true
newestAtUnlessDeleted: PRE SORT: found version 1724921590145980 del 2024-08-29 08:53:10.156 +0000 UTC. deleted=true
newestAtUnlessDeleted: PRE SORT: found version 1724926790177147 del 2024-08-29 10:19:50.18 +0000 UTC. deleted=true
newestAtUnlessDeleted: PRE SORT: found version 1724935441257780 del 2024-08-29 12:44:01.268 +0000 UTC. deleted=true
newestAtUnlessDeleted: PRE SORT: found version 1724935889814920 del 2024-08-29 12:51:29.826 +0000 UTC. deleted=false
```

repo/blob/gcs/gcs_pit.go
```diff
@@ -0,0 +1,174 @@
...
+		return gcs, nil
+	}
+
+	// Does the bucket supports versioning?
+	attrs, err := gcs.bucket.Attrs(ctx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not get determine if bucket '%s' supports versioning", gcs.BucketName)
+	}
+
+	if !attrs.VersioningEnabled {
```
KastenMike, line 164: We also should error if ObjectRetentionMode != "Enabled"
Otherwise connecting/creating to a repo fails: `ERROR cannot initialize repository: unable to write blobcfg blob: PutBlob() failed for "kopia.blobcfg": unable to complete PutBlob(kopia.blobcfg) despite 10 retries: unexpected GCS error: googleapi: Error 400: Bucket does not have object retention enabled., invalid`

repo/blob/gcs/gcs_storage.go
kreare: > Every time I call list blobs like this: `$HOME/go/bin/kopia blob list --prefix=untitled`
> I get all 4 versions of the blob in ascending order at the point of "newestAtUnlessDeleted: PRE SORT"
> If it wasn't stable then at least once it would vary unless you are sorting before `PRE SORT` also

I'm 100% sure about this, i've opened, times ago a request on Google and they said that the sort is not stable. But it's easy to fix, just sort the versioning on Kopia side. There is a feature request opened for google to implement a custom sorting feature in the API.

KastenMike: They might not have custom sorting, but we don't need it. Or manual sorting. You can delete your sort calls and create 10 versions of a blob and every list command will have them in ascending order at the `PRE SORT` log you added

kreare: > You can see the changes I mentioned [here](https://github.com/KastenMike/kopia/pull/10/files)
> 
> `return v, true` is not correct since it would return deleted blobs e.g. PIT=15:00, blob deleted 14:00 even when it is the latest/only blob version

I've seen your branch. If it works, i can safely drop this PR and fork yours as mine was just a wip and not a working one.

kreare: > They might not have custom sorting, but we don't need it. Or manual sorting. You can delete your sort calls and create 10 versions of a blob and every list command will have them in ascending order at the `PRE SORT` log you added

When i did this PR it wasn't stable, that's why i've added a sort function, trying to make that stable. Probably they have changed something .....

repo/blob/gcs/gcs_pit.go
```diff
@@ -161,7 +145,7 @@ func maybePointInTimeStore(ctx context.Context, gcs *gcsStorage, pointInTime *ti
 		return nil, errors.Wrapf(err, "could not get determine if bucket '%s' supports versioning", gcs.BucketName)
 	}
 
-	if !attrs.VersioningEnabled {
+	if !attrs.VersioningEnabled || attrs.ObjectRetentionMode != "Enabled" {
 		return nil, errors.Errorf("cannot create point-in-time view for non-versioned bucket '%s'", gcs.BucketName)
```
KastenMike, line 149: A separate check would be better. That way you can highlight that ObjectRetention is the problem in the err

repo/blob/gcs/gcs_versioned.go
```diff
@@ -111,12 +111,12 @@ func infoToVersionMetadata(prefix string, oi *storage.ObjectAttrs) versionMetada
...
 		BlobID:    toBlobID(oi.Name, prefix),
 		Length:    oi.Size,
-		Timestamp: oi.Updated,
+		Timestamp: oi.Created,
 	}
 
 	return versionMetadata{
 		Metadata:       bm,
-		IsDeleteMarker: !oi.Deleted.IsZero(),
+		IsDeleteMarker: !oi.Deleted.IsZero() && (gcs.PointInTime == nil || oi.Deleted.Before(*gcs.PointInTime)),
```
KastenMike, line 119: it should be `func (gcs *gcsStorage) infoToVersionMetadata` for this to work

repo/blob/gcs/gcs_storage.go
```diff
@@ -122,28 +162,36 @@ func (gcs *gcsStorage) PutBlob(ctx context.Context, b blob.ID, data blob.Bytes,
 	writer.ContentType = "application/x-kopia"
 	writer.ObjectAttrs.Metadata = timestampmeta.ToMap(opts.SetModTime, timeMapKey)
 
+	if opts.RetentionPeriod != 0 {
+		retainUntilDate := clock.Now().Add(opts.RetentionPeriod).UTC()
+		writer.ObjectAttrs.Retention = &storage.ObjectRetention{
+			Mode:        blob.Locked,
```
KastenMike, line 168:
```suggestion
			Mode:        string(blob.Locked),
```

repo/blob/gcs/gcs_storage.go
```diff
@@ -155,6 +203,26 @@ func (gcs *gcsStorage) DeleteBlob(ctx context.Context, b blob.ID) error {
 	return err
 }
 
+func (gcs *gcsStorage) ExtendBlobRetention(ctx context.Context, b blob.ID, opts blob.ExtendOptions) error {
+	retentionMode := blob.Locked
+
+	retainUntilDate := clock.Now().Add(opts.RetentionPeriod).UTC().Truncate(time.Second)
+
+	ObjectRetention := &storage.ObjectRetention{
+		Mode:        retentionMode,
```
KastenMike, line 212:
```suggestion
		Mode:        string(blob.Locked),
```

repo/blob/gcs/gcs_versioned.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// versionMetadataCallback is called when processing the metadata for each blob version.
+type versionMetadataCallback func(versionMetadata) error
+
+// IsVersioned returns whether versioning is enabled in the store.
+// It returns true even if versioning is enabled but currently suspended for the
+// bucket. Notice that when object locking is enabled in a bucket, object
+// versioning is enabled and cannot be suspended.
+func (gcs *gcsStorage) IsVersioned(ctx context.Context) (bool, error) {
```
KastenMike, line 32: It seems this isn't used and can be removed

repo/blob/gcs/gcs_versioned.go
```diff
@@ -0,0 +1,122 @@
...
+
+	if !foundBlobs {
+		return blob.ErrBlobNotFound
+	}
+
+	return nil
+}
+
+// listBlobVersions lists all versions for all the blobs with the given blob ID prefix.
+func (gcs *gcsStorage) listBlobVersions(ctx context.Context, prefix blob.ID, callback versionMetadataCallback) error {
```
KastenMike, line 62:
```suggestion
func (gcs *gcsPointInTimeStorage) listBlobVersions(ctx context.Context, prefix blob.ID, callback versionMetadataCallback) error {
```
I guess everything in here should be `gcsPointInTimeStorage` rather than `gcsStorage`

repo/blob/gcs/gcs_versioned.go
```diff
@@ -0,0 +1,122 @@
...
+		if errCallback := callback(om); errCallback != nil {
+			return errors.Wrapf(errCallback, "callback failed for %q", attrs.Name)
+		}
+	}
+
+	return nil
+}
+
+func toBlobID(blobName, prefix string) blob.ID {
+	return blob.ID(prefix + strings.TrimPrefix(blobName, prefix))
```
KastenMike, line 107:
```suggestion
	return blob.ID(strings.TrimPrefix(blobName, prefix))
```

KastenMike:
Some follow up comments based on the current state of the PR, I also left a few more on this PR:
- Can you create an [isCreate](https://github.com/kopia/kopia/blob/master/cli/storage_azure.go#L52) check on Connect like Azure/AWS has
- separate out `!attrs.VersioningEnabled || attrs.ObjectRetentionMode != "Enabled"` checks
- `string(blob.Locked)` as mentioned in a comment to get around compilation issues- 
- remove the sorts
- remove the print statements

Otherwise it looks ok :+1:

Finally could you add some test coverage and a documentation section for this PR as 2 separate PRs? The test coverage can be a new branch based off this one (PR train~) to avoid the PR getting to big 
Docs can just be a small section added like [this](https://github.com/kopia/kopia/pull/3398) for Azure
For tests you could mimic `azure_versioned_test.go` and `azure_immu_test.go`

kreare:
> Some follow up comments based on the current state of the PR, I also left a few more on this PR:
> 
> * Can you create an [isCreate](https://github.com/kopia/kopia/blob/master/cli/storage_azure.go#L52) check on Connect like Azure/AWS has
> * separate out `!attrs.VersioningEnabled || attrs.ObjectRetentionMode != "Enabled"` checks
> * `string(blob.Locked)` as mentioned in a comment to get around compilation issues-
> * remove the sorts
> * remove the print statements
> 
> Otherwise it looks ok 👍
> 
> Finally could you add some test coverage and a documentation section for this PR as 2 separate PRs? The test coverage can be a new branch based off this one (PR train~) to avoid the PR getting to big Docs can just be a small section added like [this](https://github.com/kopia/kopia/pull/3398) for Azure For tests you could mimic `azure_versioned_test.go` and `azure_immu_test.go`

currently i'm a little bit busy at work and my current version rebuilt locally (a couple of days ago) with the new laptop and newer Go version doesn't work anymore, it crash and i don't have time to troubleshoot due to lack of time

repo/blob/gcs/gcs_storage.go
```diff
@@ -155,6 +203,26 @@ func (gcs *gcsStorage) DeleteBlob(ctx context.Context, b blob.ID) error {
 	return err
 }
 
+func (gcs *gcsStorage) ExtendBlobRetention(ctx context.Context, b blob.ID, opts blob.ExtendOptions) error {
+	retentionMode := blob.Locked
+
+	retainUntilDate := clock.Now().Add(opts.RetentionPeriod).UTC().Truncate(time.Second)
+
+	ObjectRetention := &storage.ObjectRetention{
```
KastenMike, line 211:
```suggestion
	objectRetention := &storage.ObjectRetention{
```

repo/blob/gcs/gcs_versioned.go
```diff
@@ -0,0 +1,122 @@
...
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return errors.Wrapf(err, "could not list objects with prefix %q", query.Prefix)
+		}
+
+		if onlyMatching && attrs.Name != query.Prefix {
+			return nil
```
KastenMike, line 90: I guess this should be `break` or `continue` instead of `return nil` depending on the sort order

julio-lopez:
Implemented.

PR #3616, opened 2024-02-05
Update _index.md: old link redirects to sales page

triatic:


lupusA (commented):
We should probably add the benchmark on our site and not refer to the web archive imo. 
What do you think?

triatic:
> We should probably add the benchmark on our site and not refer to the web archive imo.

Linking to an external source gives some assurance of trust/independence. Plus there may be copyright issues around copying the benchmark to another site (permission to re-publish would be needed).

lupusA:
> > We should probably add the benchmark on our site and not refer to the web archive imo.
> 
> Linking to an external source gives some assurance of trust/independence. Plus there may be copyright issues around copying the benchmark to another site (permission to re-publish would be needed).

We should ask about getting the permission then. Otherwise, we will have dead links here and there all the time :-)

triatic:
> We should ask about getting the permission then. Otherwise, we will have dead links here and there all the time :-)

Ok, but that might take some time. Accepting the PR would solve things before that is achieved.

PR #3629, opened 2024-02-07
fix(ci): fix shadow-copy test

### LINKED ISSUE #3628: Failure in tests/os_snapshot_test TestShadowCopy
Since #3625

`The process cannot access the file because it is being used by another process.`

Details: https://github.com/kopia/kopia/actions/runs/7808046134

Ref: #3624

aaron-kasten:
Fixes #3628

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -125,6 +125,7 @@ $ make -C site build
 
 This will auto-generate [Markdown](https://en.wikipedia.org/wiki/Markdown) files with documentation for currently supported Kopia CLI subcommands and store them under `site/content/docs/Reference/Command-Line` and then generate the website which is stored in `site/public`.
 
+
 To see the  website in a browser it's more convenient to use:
 
 ```shell
--- a/BUILD.md
+++ b/BUILD.md
@@ -125,7 +125,6 @@ $ make -C site build
 
 This will auto-generate [Markdown](https://en.wikipedia.org/wiki/Markdown) files with documentation for currently supported Kopia CLI subcommands and store them under `site/content/docs/Reference/Command-Line` and then generate the website which is stored in `site/public`.
 
-
 To see the  website in a browser it's more convenient to use:
 
 ```shell
--- a/tests/os_snapshot_test/os_snapshot_windows_test.go
+++ b/tests/os_snapshot_test/os_snapshot_windows_test.go
@@ -37,6 +37,7 @@ func TestShadowCopy(t *testing.T) {
 	e.RunAndExpectSuccess(t, "policy", "set", "--global", "--enable-volume-shadow-copy=when-available")
 
 	_, err = vss.Get("{00000000-0000-0000-0000-000000000000}")
+
 	isAdmin := !errors.Is(err, os.ErrPermission)
 	if isAdmin {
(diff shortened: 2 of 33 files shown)


BUILD.md
```diff
@@ -125,6 +125,7 @@ $ make -C site build
 
 This will auto-generate [Markdown](https://en.wikipedia.org/wiki/Markdown) files with documentation for currently supported Kopia CLI subcommands and store them under `site/content/docs/Reference/Command-Line` and then generate the website which is stored in `site/public`.
 
+
```
Shrekster, line 128: nit; drop this

maybe your editor does this automatically ? Some of my vim plugins used to do this until I configured them to not do auto formatting on existing files.

tests/os_snapshot_test/os_snapshot_windows_test.go
```diff
@@ -46,6 +48,10 @@ func TestShadowCopy(t *testing.T) {
 	}
 
 	sources := clitestutil.ListSnapshotsAndExpectSuccess(t, e)
+
+	require.NotEmpty(t, sources)
+	require.NotEmpty(t, sources[0].Snapshots)
```
Shrekster, line 53: good check

The existing code is a bit confusing though... would we get at-least one snapshot if we do not have the permission ?

Shrekster: Nothing to do here but I find a couple more issues with this test, for example getting any other error besides os.ErrPermission would be suppressed at first and fail later ... not a deal breaker, but I'd prefer to fail fast.

BUILD.md
aaron-kasten: @Shrekster I added to trigger a test.  I'll remove it.

julio-lopez (approved):
@aaron-kasten
🥇Thanks for doing this!!!

tests/os_snapshot_test/os_snapshot_windows_test.go
mxk: I probably should've added a comment to clarify what this is doing. An error is guaranteed here since the all-zero GUID is not a valid shadow copy ID. This is purely an "are you an admin?" check designed to match [what go-vss is doing][isAdmin] to know which version of the test to execute.

[isAdmin]: https://github.com/mxk/go-vss/blob/9b80ac5672d6a81d6a03347c18f2bc01a506c6e9/vss.go#L237-L260

Also, yes, a Kopia snapshot is created in both scenarios. The test would've simply panicked previously if that didn't happen.

julio-lopez: > I probably should've added a comment to clarify what this is doing. An error is guaranteed here since the all-zero GUID is not a valid shadow copy ID. This is purely an "are you an admin?" check designed to match [what go-vss is doing](https://github.com/mxk/go-vss/blob/9b80ac5672d6a81d6a03347c18f2bc01a506c6e9/vss.go#L237-L260) to know which version of the test to execute.

It makes sense to add the comment for clarity, it makes it easier to understand the test.

> a Kopia snapshot is created in both scenarios. The test would've simply panicked previously if that didn't happen.

Nevertheless, it makes sense to check explicitly than letting the test panic. It makes it easier to track down test failures.

PR #3643, opened 2024-02-12
feat(snapshots): add support for xattr

miekg:
This adds support for extended attributes on Unix - although I only tested it manually on Linux.

This should include all the needed bits, but lacks proper tests - I only did a manual tests, referenced in one of the commits, included here again:

~~~
% ./kopia repository create filesystem --path=/tmp/kopia

% pwd
/home/miek/src/github.com/miekg/tmp
% getfacl testfile1
# file: testfile1
# owner: miek
# group: miek
user::rw-
user:root:rw-
group::rw-
group:root:rw-
group:miek:rw-
mask::rw-
other::r--

% ./kopia --log-level=debug snapshot create /home/miek/src/github.com/miekg/tmp/

% ./kopia snapshot list
....
  2024-02-12 19:23:57 CET k0e335d9c23c5a20f78e358d1bd286865 2.1 MB drwxrwxr-x files:5 dirs:1 (latest-1..2,hourly-1,daily-1,weekly-1,monthly-1,annual-1)

% ./kopia restore k0e335d9c23c5a20f78e358d1bd286865 /tmp/XXXX
Restoring to local filesystem (/tmp/XXXX) with parallelism=8...
Processed 6 (2.1 MB) of 5 (2.1 MB).
Restored 5 files, 1 directories and 0 symbolic links (2.1 MB).

% cd /tmp/XXXX
% l
go.mod  go.sum  testfile1  tmp*  xattr.go
% getfacl testfile1
# file: testfile1
# owner: miek
# group: miek
user::rw-
user:root:rw-
group::rw-
group:root:rw-
group:miek:rw-
mask::rw-
other::r--
~~~


jkowalski:
this is promising but needs a lot more work:

* first and foremost we need a design/strategy for supporting other non-core attributes which are currently unsupported, such as Windows ACLs, additional stat_t fields, etc., perhaps 'attributes` needs to be a map of maps for extensibility, with first-level entry being the type of an attribute ("xattr", "winacl", something like that). Perhaps have a strongly-typed struct?

* we obviously need to get it to compile on all environments and address all linter issues  (`make lint-all` is a good check)

* we need comprehensive tests that pass on all supported 

* we will need a way to enable/disable support for xattr for both backup and restore based on policies because fetching them is expensive and/or folks may not want to restore them. I think the default should be to not include xattrs

* we will likely need some kind of filtering/selection so that only certain types of attributes are backed up and some can be ignored

miekg:
thanks for your comments

>  we obviously need to get it to compile on all environments and address all linter issues (make lint-all is a good check)

yep

> we need comprehensive tests that pass on all supported

yep, and adding support for openbsd and freebsd

> we will need a way to enable/disable support for xattr for both backup and restore based on policies because fetching them is expensive and/or folks may not want to restore them. I think the default should be to not include xattrs

It this with your proposed first point, or is this already expensive in it's current form, where it uploads a nil map if there are no attributes?

> first and foremost we need a design/strategy for supporting other non-core attributes which are currently unsupported, such as Windows ACLs, additional stat_t fields, etc., perhaps 'attributes` needs to be a map of maps for extensibility, with first-level entry being the type of an attribute ("xattr", "winacl", something like that). Perhaps have a strongly-typed struct?

can do

> we will likely need some kind of filtering/selection so that only certain types of attributes are backed up and some can be ignored

based on the first map key? I.e. like --no-axttr woudl exclude those, but leave, say, winacl? Or more detailed on specific axttr fields?

miekg:
I don't have the cycles to push this further.

Issue #3639, opened 2024-02-08
Make "read" commands/operations really read-only. 


julio-lopez:
IOW, ensure no writes are performed during read operations such as `snapshot list`, `snapshot restore`, `repo connect`, etc.

The execution of read commands, such as `snapshot list` among many others, is expected to only read data from the (kopia) repository and not perform any writes or change the repository in any way. 
- Performing read-only operations is safer overall.
- It allows for use cases where the repository is only accessible in read-only mode, for example, due to credentials that only have read permissions.
- Performing writes during read commands is often unexpected and may lead to bad user experience, such as taking a long time to complete simple commands such as `snapshot list` or `repo connect`.

At the moment, kopia performs a write operations, primarily related to opportunistic cleanup tasks:
- repository maintenance (#3174)
- index compaction (#3638)
- others ????

The objective is ensure only reads are issued to the storage during these commands.

The short term plan is to address solve the specific issues related to epoch index compaction (#3638) and repository maintenance (#3174).

Long term, we should consider adding the concept of an internal "read-only" session for operations to ensure (potentially at compile time) that no writes are performed in these read-only cases.

Issue #3638, opened 2024-02-08
Avoid index (epoch) cleanup and compaction during index reads


julio-lopez:
## Objectives
- Increase safety in terms of repository consistency.
- Improve user experience by providing a consistent expected behavior in terms of duration of operations and resource utilization.

## Current Issues:

Currently, kopia is performing writes on the index load path.
- This is not desirable for read-only operations. It is also surprising for the user (See #3639).
- It delays the completion of other commands such as `repository connect` or `snapshot list` among others.
- Results in unexpected high resource utilization due to aggressive parallelism.
- Has the potential for duplicated (and wasted) work. Two clients, or in some cases the same client may perform the same expensive work multiple times. For example, the same epoch could be compacted more than once.

julio-lopez:
Proposed approach

- [ ] Add a flag to disable index write operations during index loads and preserve current behavior.
- [ ] Explicitly perform index cleanup during maintenance:
  - single epoch compactions
  - epoch range compaction
  - epoch advancement
  - epoch marker cleanup
  - deletion watermark cleanup
- [ ] Disable by default write operations on index load using the mechanism implemented in 1.
- [ ] Long term: completely remove from index loading all the codepaths that perform index writes, and remove the flag that controls this behavior.

julio-lopez:
Index loading simplified call graph

![eim-call](https://github.com/kopia/kopia/assets/1953782/2b8ebd20-4789-42f4-a2c2-c893bc750f58)

ashmrtn:
silly question, but I could use a little extra context: what type(s) of maintenance will allow index compaction? Will it be full and quick maintenance or just one of them?

julio-lopez:
> silly question, but I could use a little extra context: what type(s) of maintenance will allow index compaction? Will it be full and quick maintenance or just one of them?

Good question.

The approach implemented in the current PR is as follows.

For quick maintenance:
- Single epoch compactions
- Epoch advancement

For full maintenance:
- Single epoch compactions
- Epoch range compaction
- Epoch advancement
- Epoch marker cleanup
- Deletion watermark cleanup

Does that make sense?

ashmrtn:
Thanks for the info Julio! The details you laid out make sense from an understanding perspective

[...]

PR #3645, opened 2024-02-13
refactor(general): allow disabling writes on index index loads

julio-lopez:
This is only a temporary mechanism for changing the default behavior.

Future PRs will:
- change default value
- remove index writes altogether from the write path

Followup to: #3603

Ref:
- #3638
- #3639

### DIFF
--- a/internal/epoch/epoch_manager.go
+++ b/internal/epoch/epoch_manager.go
@@ -186,6 +186,8 @@ type Manager struct {
 	log      logging.Logger
 	timeFunc func() time.Time
 
+	allowCleanupWritesOnIndexLoad bool
+
 	// wait group that waits for all compaction and cleanup goroutines.
 	backgroundWork sync.WaitGroup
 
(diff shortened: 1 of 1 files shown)


julio-lopez:
@ashmrtn FYI

internal/epoch/epoch_manager.go
```diff
@@ -1015,16 +1019,28 @@ func rangeCheckpointBlobPrefix(epoch1, epoch2 int) blob.ID {
 	return blob.ID(fmt.Sprintf("%v%v_%v_", RangeCheckpointIndexBlobPrefix, epoch1, epoch2))
 }
 
+func allowWritesOnIndexLoad() bool {
```
julio-lopez, line 1022: Note: this env variable mechanism to control the behavior is temporary.
We will remove it once the corresponding code paths are _explicitly_ hooked in maintenance.

redgoat650 (approved):
LGTM

PR #3646, opened 2024-02-13
fix(general): disable writes on index loads

julio-lopez:
Change default `allowWritesOnIndexLoad` to false when the corresponding environment variable is unset.

Add unit test to ensure epoch is not advanced on index loads by default

Followup to #3645

Ref:
- #3638
- #3639

### DIFF
--- a/internal/epoch/epoch_manager.go
+++ b/internal/epoch/epoch_manager.go
@@ -1133,11 +1133,6 @@ func rangeCheckpointBlobPrefix(epoch1, epoch2 int) blob.ID {
 func allowWritesOnIndexLoad() bool {
 	v := strings.ToLower(os.Getenv("KOPIA_ALLOW_WRITE_ON_INDEX_LOAD"))
 
-	if v == "" {
-		// temporary default to be changed once index cleanup is performed on maintenance
-		return true
-	}
-
 	return v == "true" || v == "1"
 }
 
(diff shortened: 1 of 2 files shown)


internal/epoch/epoch_manager.go
```diff
@@ -999,16 +1019,23 @@ func rangeCheckpointBlobPrefix(epoch1, epoch2 int) blob.ID {
 	return blob.ID(fmt.Sprintf("%v%v_%v_", RangeCheckpointIndexBlobPrefix, epoch1, epoch2))
 }
 
+func allowWritesOnIndexLoad() bool {
+	v := strings.ToLower(os.Getenv("KOPIA_ALLOW_WRITE_ON_INDEX_LOAD"))
```
jkowalski, line 1023: instead of that, can we pass this via Parameters ?

julio-lopez: @jkowalski I had a separate PR that wired this value, but decided against it.
As noted both in a previous PR and in the GH issue, the intent is for this to be a temporary "flag", so it should not make it into any (potentially) persistent data structure. We will remove this env var and epoch compaction code path from the read path, soon after enabling epoch compaction in maintenance. Does that make sense?

jkowalski: makes sense. LGTM

I reviewed other PRs and I must commend you on all the research and refactoring work you did there.

jkowalski (approved):

julio-lopez (commented):
OK, time to make this the default behavior 🤞🏼

julio-lopez:
@redgoat650 @Shrekster FYI

PR #3651, opened 2024-02-16
refactor(general): perform index compaction during repo maintenance

julio-lopez:
Perform index epoch compaction and cleanup during repository maintenance.

- [x] Advance epoch on quick maintenance #3709
- [x] Advance epoch on full maintenance
- [x] Compact a single epoch on quick maintenance #3728
- [x] Compact a single epoch on full maintenance
- [x] Range compaction on full maintenance #3727
- [x] Cleanup watermark and epoch markers on full maintenance #3726

Ref:
- #3638
- #3639

Followup to:
- #3603
- #3645

Related helpers and changes:
- #3721
- #3735

### DIFF
--- a/repo/maintenance/maintenance_run.go
+++ b/repo/maintenance/maintenance_run.go
@@ -46,6 +46,7 @@ const (
 	TaskExtendBlobRetentionTimeFull  = "extend-blob-retention-time"
 	TaskCleanupLogs                  = "cleanup-logs"
 	TaskEpochDeleteSupersededIndexes = "delete-superseded-epoch-indexes"
+	TaskEpochCleanupMarkers          = "cleanup-epoch-markers"
 )
 
 // shouldRun returns Mode if repository is due for periodic maintenance.
@@ -335,6 +336,16 @@ func runTaskEpochMaintenanceFull(ctx context.Context, runParams RunParameters, s
 
 	if !hasEpochManager {
 		return nil
+	}
+
+	// clean up epoch markers
+	err := ReportRun(ctx, runParams.rep, TaskEpochCleanupMarkers, s, func() error {
+		log(ctx).Infof("Cleaning up unneeded epoch markers...")
+
+		return errors.Wrap(em.CleanupMarkers(ctx), "error removing epoch markers")
+	})
+	if err != nil {
+		return err
 	}
 
 	return ReportRun(ctx, runParams.rep, TaskEpochDeleteSupersededIndexes, s, func() error {
--- a/repo/maintenance/maintenance_run.go
+++ b/repo/maintenance/maintenance_run.go
@@ -50,6 +50,7 @@ const (
 	TaskEpochDeleteSupersededIndexes = "delete-superseded-epoch-indexes"
 	TaskEpochCleanupMarkers          = "cleanup-epoch-markers"
 	TaskEpochGenerateRange           = "generate-epoch-range-index"
+	TaskEpochCompactSingle           = "compact-single-epoch"
 )
 
 // shouldRun returns Mode if repository is due for periodic maintenance.
@@ -350,6 +351,14 @@ func runTaskEpochMaintenanceQuick(ctx context.Context, runParams RunParameters, 
 
 	if !hasEpochManager {
 		return nil
+	}
+
+	err := ReportRun(ctx, runParams.rep, TaskEpochCompactSingle, s, func() error {
+		log(ctx).Infof("Compacting an eligible uncompacted epoch...")
+		return errors.Wrap(em.MaybeCompactSingleEpoch(ctx), "error compacting single epoch")
+	})
+	if err != nil {
+		return err
 	}
 
 	return runTaskEpochAdvance(ctx, em, runParams, s)
--- a/repo/maintenance/maintenance_run.go
+++ b/repo/maintenance/maintenance_run.go
@@ -36,16 +36,16 @@ type TaskType string
 
 // Task IDs.
 const (
-	TaskSnapshotGarbageCollection   = "snapshot-gc"
-	TaskDeleteOrphanedBlobsQuick    = "quick-delete-blobs"
-	TaskDeleteOrphanedBlobsFull     = "full-delete-blobs"
-	TaskRewriteContentsQuick        = "quick-rewrite-contents"
-	TaskRewriteContentsFull         = "full-rewrite-contents"
-	TaskDropDeletedContentsFull     = "full-drop-deleted-content"
-	TaskIndexCompaction             = "index-compaction"
-	TaskExtendBlobRetentionTimeFull = "extend-blob-retention-time"
-	TaskCleanupLogs                 = "cleanup-logs"
-	TaskCleanupEpochManager         = "cleanup-epoch-manager"
+	TaskSnapshotGarbageCollection    = "snapshot-gc"
+	TaskDeleteOrphanedBlobsQuick     = "quick-delete-blobs"
+	TaskDeleteOrphanedBlobsFull      = "full-delete-blobs"
+	TaskRewriteContentsQuick         = "quick-rewrite-contents"
+	TaskRewriteContentsFull          = "full-rewrite-contents"
+	TaskDropDeletedContentsFull      = "full-drop-deleted-content"
+	TaskIndexCompaction              = "index-compaction"
+	TaskExtendBlobRetentionTimeFull  = "extend-blob-retention-time"
+	TaskCleanupLogs                  = "cleanup-logs"
+	TaskEpochDeleteSupersededIndexes = "delete-superseded-epoch-indexes"
 )
 
 // shouldRun returns Mode if repository is due for periodic maintenance.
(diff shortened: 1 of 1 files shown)


internal/epoch/epoch_manager.go
```diff
@@ -950,6 +1023,101 @@ func (e *Manager) getCompleteIndexSetForCommittedState(ctx context.Context, cs C
...
+
+// Returns a continuous close-open epoch range for the keys, that is [lo, hi).
+// A range of the form [v,v) means the range is empty.
+// When the range is not continuous an error is returned.
+func getKeyRange[E any](m map[int]E) (intRange, error) {
+	keys := maps.Keys(m)
+
+	slices.Sort(keys)
+
+	keys = slices.Compact(keys)
```
plar, line 1048: I think `slices.Compact` is redundant here, as keys are inherited from `map[int]E`, which by definition should not have duplicates.

internal/epoch/epoch_manager.go
```diff
@@ -950,6 +1023,101 @@ func (e *Manager) getCompleteIndexSetForCommittedState(ctx context.Context, cs C
...
+		return intRange{}, nil
+	}
+
+	if lo, hi := keys[0], keys[l-1]; hi-lo > l-1 {
+		return intRange{lo: -1, hi: -1}, errors.Wrapf(errNonContiguousRange, "lo: %d, hi: %d, length: %d", lo, hi, l)
+	}
+
+	return intRange{lo: keys[0], hi: keys[l-1] + 1}, nil
+}
+
```
plar, line 1061: I'm not sure how big `m` could be but we can do slightly better w/o `slices.Sort`. The implementation below is more efficient for large `m`: O(n log n) -> O(n). It also removes dependencies on experimental `golang.org/x/exp/maps` package.

```
func getKeyRange[E any](m map[int]E) (intRange, error) {
    lo, hi, count := 0, 0, 0
    for k := range m {
        if count == 0 { // init lo and hi with the 1st key.
            lo, hi = k, k 
        } else {
            if k < lo {
                lo = k
            }
            if k > hi {
                hi = k
            }
        }
        count++
    }
    
    if count == 0 { // no keys were found
        return intRange{}, nil
    }

    if hi-lo > count-1 { // is the range contiguous?
        return intRange{lo: -1, hi: -1}, errors.Wrapf(errNonContiguousRange, "lo: %d, hi: %d, length: %d", r.lo, r.hi, count)
    }

    return intRange{lo: lo, hi: hi+1}, nil
}
```
I didn't test it :)

plar (commented):
Alternative implementation of `getKeyRange`.

internal/epoch/epoch_manager.go
```diff
@@ -1067,6 +1067,57 @@ func getCompactedEpochRange(cs CurrentSnapshot) (intRange, error) {
...
+		return nil
+	}
+
+	compacted, err := getCompactedEpochRange(cs)
+	if err != nil {
+		return err
+	}
+
+	if uncompacted.lo < compacted.hi {
+		uncompacted.lo = compacted.hi // avoid compacting already compacted epoch
```
redgoat650, line 1094: What scenario would lead to this (if any)? Committed state in an inconsistent/transitory state? Racing with other compactions?

repo/maintenance/maintenance_run.go
```diff
@@ -337,9 +374,40 @@ func runTaskCleanupEpochManager(ctx context.Context, runParams RunParameters, s
...
 
-	return ReportRun(ctx, runParams.rep, TaskCleanupEpochManager, s, func() error {
+	if err := runTaskEpochAdvance(ctx, em, runParams, s); err != nil {
+		return err
+	}
+
+	// compact a single epoch
+	if err := ReportRun(ctx, runParams.rep, TaskEpochCompactSingle, s, func() error {
+		log(ctx).Infof("Compacting an eligible uncompacted epoch...")
+		return errors.Wrap(em.MaybeCompactSingleEpoch(ctx), "error compacting single epoch")
```
redgoat650, line 384: Can you explain a bit about the ordering in full vs quick maintenance? Here we `runTaskEpochAdvance` then `MaybeCompactSingleEpoch`. In quick we do it in reverse order. Is there any significance, does the ordering matter at all? Will we be writing compacted blobs into old/new epochs depending on quick vs full?

internal/epoch/epoch_manager.go
```diff
@@ -950,6 +1023,101 @@ func (e *Manager) getCompleteIndexSetForCommittedState(ctx context.Context, cs C
...
+func (r intRange) isEmpty() bool {
+	return r.length() <= 0
+}
+
+var errNonContiguousRange = errors.New("non-contiguous range")
+
+// Returns a continuous close-open epoch range for the keys, that is [lo, hi).
+// A range of the form [v,v) means the range is empty.
+// When the range is not continuous an error is returned.
+func getKeyRange[E any](m map[int]E) (intRange, error) {
```
redgoat650, line 1043: (organization) Totally optional, but this implementation seems generic enough to be decomposed into its own helper package or file, along with its unit test, perhaps?

redgoat650 (approved):
Left a few questions, but otherwise it makes sense to me. Reviewing commit-by-commit was very helpful, thanks

internal/epoch/epoch_manager.go
julio-lopez: Yeah, we can do that in a separate PR.

Also, it'd be good to discuss a broader philosophical question here, since it is not clear to me what's the _threshold or breaking point_ between:
- Reuse (_"don't repeat yourself"_, etc) and
- _"A little copying is better than a little dependency."_

Anyway, let's discuss that separate from this PR.

internal/epoch/epoch_manager.go
julio-lopez: yeah, 🤦🏼

internal/epoch/epoch_manager.go
julio-lopez: Thanks, nice suggestion!
I'll put this in a separate PR with tests, and then use it here.

internal/epoch/epoch_manager.go
julio-lopez: Good question. Let me dig again.

IIRC:
- It is what you mention. this function is not assuming it will be called sequentially, in isolation, etc, and that instead there may be cross-client concurrency. This is even if the intent is to only call this from maintenance to guarantee serial execution.
- It is also due to the way the "committed state" is loaded. Again, need to double check.

repo/maintenance/maintenance_run.go
```diff
@@ -327,7 +336,35 @@ func runTaskCleanupLogs(ctx context.Context, runParams RunParameters, s *Schedul
...
+	})
+	if err != nil {
+		return err
+	}
+
+	return runTaskEpochAdvance(ctx, em, runParams, s)
+}
+
+func runTaskEpochMaintenanceFull(ctx context.Context, runParams RunParameters, s *Schedule) error {
 	em, ok, emerr := runParams.rep.ContentManager().EpochManager(ctx)
```
julio-lopez, line 368:
```suggestion
	em, hasEpochManager, emerr := runParams.rep.ContentManager().EpochManager(ctx)
```

repo/maintenance/maintenance_run.go
julio-lopez: There is no constraint in the ordering. We should be consistent though.
It may make sense to advance first.

internal/epoch/epoch_manager.go
julio-lopez: In general, the length of these ranges, and thus the size of the input maps, is expected to be in the single digits. If the number of uncompacted epochs grows too large, then there will be other performance issues, and the algorithmic complexity of this function will be irrelevant even in those cases.

The suggested implementation with tests is in #3721 @plar @redgoat650 PTAL. This PR (#3651) now depends on that one. Thanks.

internal/epoch/epoch_manager.go
julio-lopez: For now, it's been moved it to the `_utils` file in the same package.

internal/epoch/epoch_manager.go
julio-lopez: The implementation changed to account for the case where the list of blobs for some uncompacted epochs is not included in the _"current snapshot"_ (`cs`).

See https://github.com/kopia/kopia/pull/3728/files#diff-0fc9e0e50903701c369ab11147ea2f21b5c9756abac3299c28aeab8bc628ce41

repo/maintenance/maintenance_run.go
```diff
@@ -337,9 +353,61 @@ func runTaskCleanupEpochManager(ctx context.Context, runParams RunParameters, s
...
+		return err
+	}
+
+	if err := runTaskEpochAdvance(ctx, em, runParams, s); err != nil {
+		return err
+	}
+
+	// compact range
+	if err := ReportRun(ctx, runParams.rep, TaskEpochGenerateRange, s, func() error {
+		log(ctx).Infof("Attempting to compact a range of epoch indexes ...")
```
redgoat650, line 391: super nit: the other similar info logs have no space before the ellipsis. Remove for consistency?
```suggestion
		log(ctx).Infof("Attempting to compact a range of epoch indexes...")
```

redgoat650 (approved):
Looks good

PR #3654, opened 2024-02-18
refactor(repository): added context to potentially blocking repository methods

jkowalski:


repo/format/format_manager.go
```diff
@@ -80,7 +80,7 @@ func (m *Manager) maybeRefreshNotLocked() error {
 	}
 
 	// current format not valid anymore, kick off a refresh
-	return m.refresh(m.ctx)
```
julio-lopez, line 83: `Manager.ctx` is no longer used and should be removed for clarity.

repo/format/format_manager.go
```diff
@@ -80,7 +80,7 @@ func (m *Manager) maybeRefreshNotLocked() error {
...
+	return m.refresh(ctx)
```
julio-lopez, line 83: It seems that one of the primary changes here is to stop using a `ctx` field in the Manager and instead use a context that is passed in. Most (if not all) of the changes can be traced to wiring the context to `content.Manager.maybeRefreshNotLocked`
Is that right?

repo/content/content_manager.go
```diff
@@ -924,9 +929,9 @@ func (bm *WriteManager) lock() {
 }
 
 // +checklocksrelease:bm.mu
-func (bm *WriteManager) unlock() {
+func (bm *WriteManager) unlock(ctx context.Context) {
```
julio-lopez, line 932: thinking out loud: this seems like an _artificial artifact_, not clear anything simple can be done about it.

julio-lopez (approved):

repo/format/format_manager.go
julio-lopez: Addressed in #3664 3664

PR #3655, opened 2024-02-18
refactor(general): Increase restore progress granularity

e-sumin:
When restoring huge file(s), the progress reporting is done in a bit weird way:

```
kopia_test % kopia snapshot restore ka2084d263182164b6cf3456668e6b6da /Users/eugen.sumin/kopia_test/2
Restoring to local filesystem (/Users/eugen.sumin/kopia_test/2) with parallelism=8...
Processed 6 (5.4 GB) of 5 (5.4 GB) 1.6 MB/s (100.0%) remaining 0s.
Processed 6 (5.4 GB) of 5 (5.4 GB) 1.6 MB/s (100.0%) remaining 0s.
Processed 6 (5.4 GB) of 5 (5.4 GB) 1.6 MB/s (100.0%) remaining 0s.
Processed 6 (5.4 GB) of 5 (5.4 GB) 1.5 MB/s (100.0%) remaining 0s.
Processed 6 (5.4 GB) of 5 (5.4 GB) 1.5 MB/s (100.0%) remaining 0s.
Processed 6 (5.4 GB) of 5 (5.4 GB) 1.5 MB/s (100.0%) remaining 0s.
Restored 5 files, 1 directories and 0 symbolic links (5.4 GB).
```
In fact, the amount of restored data is dumped when particular file completely restored.

This PR contains the least invasive change, which allows us to see progress update while file is downloaded from object storage.
```
Restoring to local filesystem (/Users/eugen.sumin/kopia_test/55) with parallelism=8...
Processed 2 (3.1 MB) of 5 (1.8 GB).
Processed 4 (459.6 MB) of 5 (1.8 GB) 270.3 MB/s (25.2%) remaining 4s.
Processed 4 (468.7 MB) of 5 (1.8 GB) 269 MB/s (25.7%) remaining 4s.
Processed 4 (741.6 MB) of 5 (1.8 GB) 269 MB/s (40.6%) remaining 3s.
Processed 4 (1.1 GB) of 5 (1.8 GB) 280 MB/s (57.6%) remaining 2s.
Processed 5 (1.4 GB) of 5 (1.8 GB) 291.1 MB/s (75.2%) remaining 1s.
Processed 5 (1.4 GB) of 5 (1.8 GB) 289.8 MB/s (75.6%) remaining 1s.
Processed 5 (1.6 GB) of 5 (1.8 GB) 270.2 MB/s (85.3%) remaining 0s.
Processed 5 (1.7 GB) of 5 (1.8 GB) 256.3 MB/s (95.0%) remaining 0s.
Processed 6 (1.8 GB) of 5 (1.8 GB) 251 MB/s (100.0%) remaining 0s.
Processed 6 (1.8 GB) of 5 (1.8 GB) 251 MB/s (100.0%) remaining 0s.
Restored 5 files, 1 directories and 0 symbolic links (1.8 GB).
```


e-sumin:
@KastenMike @julio-lopez could you please review ?

repo/object/object_reader.go
```diff
@@ -35,6 +35,10 @@ func VerifyObject(ctx context.Context, cr contentReader, oid ID) ([]content.ID,
 	return tracker.contentIDs(), nil
 }
 
+// FileReadingProgressCallback is a callback intended to be used during file copying
+// to report amount of data sent to destination.
+type FileReadingProgressCallback func(chunkSize int64)
```
jkowalski, line 40: why do we need to change object reader at all, can't we just emit the progress as we're reading the bytes during restore?

snapshot/restore/local_fs_output.go
```diff
@@ -403,6 +404,14 @@ func (o *FilesystemOutput) copyFileContent(ctx context.Context, targetPath strin
 	}
 	defer r.Close() //nolint:errcheck
 
+	type withProgressTracking interface {
```
jkowalski, line 407: I think it's easier to wrap "r" so that it reports bytes written in the callback.

e-sumin: Thank you for the idea, I've changed the code, is it what you meant ?

repo/object/object_reader.go
jkowalski: this does not seem used anywhere, the rest looks good

e-sumin:
@jkowalski do you think the changes could be merged ?

e-sumin:
> @jkowalski do you think the changes could be merged ? Or something else has to be done before ?

Don't want to be annoying, just a friendly reminder :)

denisvmedia:
@jkowalski @redgoat650 @julio-lopez could you please check this PR?

snapshot/restore/restore.go
```diff
@@ -203,19 +229,27 @@ func (c *copier) copyEntryInternal(ctx context.Context, e fs.Entry, targetPath s
 	case fs.File:
 		log(ctx).Debugf("file: '%v'", targetPath)
 
-		c.stats.RestoredFileCount.Add(1)
-		c.stats.RestoredTotalFileSize.Add(e.Size())
+		bytesExpected := e.Size()
+		bytesWritten := int64(0)
+		progressCallback := func(chunkSize int64) {
+			bytesWritten += chunkSize
+			c.stats.RestoredTotalFileSize.Add(chunkSize)
```
Shrekster, line 236: nit; this callback is hit each time we have read `chunkSize` from the repository, not necessarily `RestoredTotalFileSize`. So, `RestoredTotalFileSize ` might become equal to `EnqueuedTotalFileSize` and we could still be writing. Maybe we can introduce a different stat here like `ProcessingTotalFileSize` or `PrefetchedTotalFileSize` ?

snapshot/restore/local_fs_output.go
```diff
@@ -402,16 +430,20 @@ func (o *FilesystemOutput) copyFileContent(ctx context.Context, targetPath strin
...
 	targetPath = atomicfile.MaybePrefixLongFilenameOnWindows(targetPath)
 
 	if o.WriteFilesAtomically {
 		//nolint:wrapcheck
-		return atomicfile.Write(targetPath, r)
+		return atomicfile.Write(targetPath, wr)
 	}
 
-	return write(targetPath, r, f.Size(), o.copier)
+	return write(targetPath, wr, f.Size(), o.copier)
```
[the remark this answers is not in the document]
e-sumin: That's true, but this way of progress update was chosen because it is less intrusive. Also, when reading, usually, the block is very small in comparison with total amount of data. So, we, anyway, can get 100% earlier than last block will be read. So, I'd not consider this as critical flaw.

snapshot/restore/restore.go
e-sumin: Yes, that's correct. This progress calculation way treats bytes read as bytes written.

Shrekster:
@e-sumin and I discussed this over a call and he explained why the reader based approach is correct / simpler for now. I also agree with his comments above.

@e-sumin is looking into writing a unit-test coverage with a custom test-only callback plumbing to make it non-flaky.

e-sumin:
I did some refactoring of snapshot restore:

- progress reporting interval is taken from command line, as it was done for upload progress reporting
  - currently upload progress instance is keeping progress flag and interval and restore progress just utilizes its value (this was originally discussed with @Shrekster as less invasive thing)
  - cli progress refactoring should be done (but probably it would be better to have it as a separate PR). We need to extract flag and interval and throttling mechanism to reused part and create progress instance depending on operation (upload or restore)
    - this also will allow to write separate tests for progress reporting format and progress reporting intervals
- probably it would be better to pass restore progress interface to `copier` and increase counters separately when needed instead of keeping another copy of `stats` in `copier` and passing its values to progress reporter.

Some of changes I've done was done to be able to have an unit test which verifies that progress is reported properly.

e-sumin:
I've fixed the test and merged master into my branch. The tests should pass now.

tests/end_to_end_test/restore_test.go
```diff
@@ -82,7 +121,26 @@ func TestRestoreCommand(t *testing.T) {
...
+	runner.CustomizeApp = func(a *cli.App, kp *kingpin.Application) {
+		origCustomizeApp(a, kp)
+		a.SetRestoreProgress(frp)
+	}
+
+	e.RunAndExpectSuccess(t, "restore", snapID, restoreFailDir, "--progress-update-interval", "1ms")
+
+	runner.CustomizeApp = origCustomizeApp
+
+	// Expecting progress to be reported multiple times and flush to be invoked ad the end
```
Shrekster, line 140:
```suggestion
	// Expecting progress to be reported multiple times and flush to be invoked at the end
```

snapshot/restore/local_fs_output.go
```diff
@@ -402,16 +430,20 @@ func (o *FilesystemOutput) copyFileContent(ctx context.Context, targetPath strin
 		return errors.Wrap(err, "unable to open snapshot file for "+targetPath)
 	}
 	defer r.Close() //nolint:errcheck
+	wr := &progressReportingReader{
```
Shrekster, line 433:
```suggestion

	wr := &progressReportingReader{
```

Shrekster (approved):

snapshot/restore/restore_progress.go
```diff
@@ -0,0 +1,10 @@
+package restore
+
+// Progress is invoked by copier to report status of snapshot restoration.
+type Progress interface {
+	SetCounters(
```
denisvmedia, line 5: Too late to comment here, but I believe having an args struct would be better.

julio-lopez (commented):
@e-sumin see couple of minor comments

snapshot/restore/local_fs_output.go
```diff
@@ -54,6 +54,34 @@ func getStreamCopier(ctx context.Context, targetpath string, sparse bool) (strea
...
+
+func (r *progressReportingReader) Seek(offset int64, whence int) (int64, error) {
+	return r.r.Seek(offset, whence) //nolint:wrapcheck
+}
+
+func (r *progressReportingReader) Close() error {
+	return r.r.Close() //nolint:wrapcheck
+}
+
+func (r *progressReportingReader) Read(p []byte) (int, error) {
```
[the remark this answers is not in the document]
julio-lopez: Addressed in #4003

PR #3668, opened 2024-02-21
feat(cli): 1B: PPROF extensions: load profile config

aaron-kasten:
This is a continuation of https://github.com/kopia/kopia/pull/3454

This is one of 8 PRs in a PR train:
`aaron-kasten/kopia:pprof-extensions-c-1a`
`aaron-kasten/kopia:pprof-extensions-c-1b`
`aaron-kasten/kopia:pprof-extensions-c-1c`
`aaron-kasten/kopia:pprof-extensions-c-1d`
`aaron-kasten/kopia:pprof-extensions-c-1e`
`aaron-kasten/kopia:pprof-extensions-c-1f`
`aaron-kasten/kopia:pprof-extensions-c-1g`

Usage
=====

example:

```
[1] 77308
```

```
$ kill %1
```

```
saving PEM buffers for output
dumping PEM for "PPROF MEM"
-----BEGIN PPROF MEM-----
.
.
.
//8psEjOrZ4AAA==
-----END PPROF MEM-----
```

```
$ ./go/bin/kats dump.b64
```

```
 ./go/bin/kats --help
 Usage of ./go/bin/kats:
  -verbose
    	verbose outout
```

```
File: pprof_mem.bin
Type: inuse_space
(pprof) tree
```

Configuration Options
=====

TBDThis PR adds the ability to dump pprof data to logs for debugging. 

This is a continuation of https://github.com/kopia/kopia/pull/3454

Usage
=====

example:

```
[1] 77308
```

```
$ kill %1
```

```
saving PEM buffers for output
dumping PEM for "PPROF MEM"
-----BEGIN PPROF MEM-----
.
.
.
//8psEjOrZ4AAA==
-----END PPROF MEM-----
```

```
$ ./go/bin/kats dump.b64
```

```
 ./go/bin/kats --help
 Usage of ./go/bin/kats:
  -verbose
    	verbose outout
```

```
File: pprof_mem.bin
Type: inuse_space
(pprof) tree
```

Configuration Options
=====

TBD


internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +105,44 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
...
+		wrt: wrt,
+	}
+
+	return q
+}
+
+// LoadProfileConfig configure PPROF profiling from the config in ppconfigss.
+func LoadProfileConfig(ctx context.Context, ppconfigss string) (map[ProfileName]*ProfileConfig, error) {
+	// if empty, then don't bother configuring but emit a log message - use might be expecting them to be configured
+	if ppconfigss == "" {
```
julio-lopez, line 119: maybe add the debug log message as noted in the comment above?

internal/pproflogging/pproflogging.go
```diff
@@ -220,7 +267,7 @@ func StartProfileBuffers(ctx context.Context) {
 	pprofConfigs.mu.Lock()
 	defer pprofConfigs.mu.Unlock()
 
-	pprofConfigs.pcm = parseProfileConfigs(bufSizeB, ppconfigs)
+	pprofConfigs.pcm, _ = parseProfileConfigs(bufSizeB, ppconfigs)
```
julio-lopez, line 270: Should this bail out when `parseProfileConfigs` returns an error?

internal/pproflogging/pproflogging_test.go
```diff
@@ -124,3 +197,137 @@ func TestDebug_newProfileConfigs(t *testing.T) {
...
+	oldEnv = os.Getenv(EnvVarKopiaDebugPprof)
+}
+
+// +checklocksignore
+//
+//nolint:gocritic
+func restoreUnlockEnv(t *testing.T) {
+	t.Helper()
+
+	t.Setenv(EnvVarKopiaDebugPprof, oldEnv)
```
julio-lopez, line 295: I may be missing something here. I am not sure what this is achieving.

IIUC `t.Setenv()` sets the env var during the execution of the test and automatically restores the original value when the test finishes.
https://pkg.go.dev/testing#T.Setenv

`restoreUnlockEnv()` is being called at the end of various tests. It seems that would cause :
- to save whatever value (`ev`) the variable has at that point (after it has been modified in the test);
- then, the briefly change the env var value to the value the var had when the test started, that is `oldEnv`; and
– then reset the value to `ev`, the value the env var had before calling `t.Setenv`.

It seems that `t.Setenv()` should be called at the beginning of the test to set the desired env var value for the test and let the testing framework perform its cleanup and reset the value by itself at the end.

internal/pproflogging/pproflogging_test.go
```diff
@@ -124,3 +197,137 @@ func TestDebug_newProfileConfigs(t *testing.T) {
...
+	}
+}
+
+// +checklocksignore
+//
+//nolint:gocritic
+func saveLockEnv(t *testing.T) {
+	t.Helper()
+
+	mu.Lock()
```
julio-lopez, line 285: Why is this lock needed?
Test execution is serialized unless explicitly made parallel via `t.Parallel()`

internal/pproflogging/pproflogging_test.go
```diff
@@ -124,3 +197,137 @@ func TestDebug_newProfileConfigs(t *testing.T) {
...
+				return
+			}
+			require.Equal(t, tc.expectProfileFlagValue, flagValue)
+		})
+	}
+}
+
+// +checklocksignore
+//
+//nolint:gocritic
```
julio-lopez, line 281: why are these linter suppression annotations and the ones below needed?

internal/pproflogging/pproflogging_test.go
```diff
@@ -124,3 +197,137 @@ func TestDebug_newProfileConfigs(t *testing.T) {
...
+
+func TestWriter(t *testing.T) {
+	eww := &ErrorWriter{mx: 5, err: io.EOF}
+	n, err := eww.WriteString("Hello World")
+	require.ErrorIs(t, io.EOF, err)
+	require.Equal(t, 5, n)
+	require.Equal(t, "Hello", string(eww.bs))
+}
+
+type ErrorWriter struct {
```
julio-lopez, line 307: Will this be used to simulate errors in future tests? It'd be a good idea to add a comment explaining the intent. WDYT?

internal/pproflogging/pproflogging_test.go
aaron-kasten: See changes for my interpretation of what you mean?

internal/pproflogging/pproflogging_test.go
aaron-kasten: Good point!  The tests are probably never be parallelized.

internal/pproflogging/pproflogging_test.go
aaron-kasten: @julio-lopez I'm unsure.  Likely legacy.  Linter doesn't seem to complain without it.

internal/pproflogging/pproflogging_test.go
aaron-kasten: Yes... I was thinking that when I was working on a future PR.  I had forgotten hot to use it and needed to reference the Unit test.

internal/pproflogging/pproflogging.go
```diff
@@ -127,14 +164,19 @@ func parseProfileConfigs(bufSizeB int, ppconfigs string) map[ProfileName]*Profil
 		flagValue := ""
 
 		if len(profileFlagNameValuePairs) > 1 {
+			// only <key>=<value? allowed
 			flagValue = profileFlagNameValuePairs[1]
 		}
 
-		flagKey := ProfileName(strings.ToLower(profileFlagNameValuePairs[0]))
+		flagKey := ProfileName(profileFlagNameValuePairs[0])
+		if flagKey == "" {
```
Shrekster, line 172: is it required to strip the value before checking for empty ?

internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +103,41 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
 	},
 }
 
+func newProfileConfigs(wrt Writer) *ProfileConfigs {
+	q := &ProfileConfigs{
+		wrt: wrt,
+	}
```
Shrekster, line 109: Shouldn't this also allocate the map q.pcm ?

aaron-kasten: No. The nil value is used to detect if there are any profiles configured.

Shrekster: Ok, never mind. I noticed that your `parseProfileConfigs()` does allocate a map and assign internally. So, this is not required.

internal/pproflogging/pproflogging.go
aaron-kasten: The values should be stripped before they reach this point.  Regardless, I can add a  unit test later to ensure they are stripped.

internal/pproflogging/pproflogging.go
```diff
@@ -63,23 +63,29 @@ const (
...
+	io.Writer
+	io.StringWriter
 }
 
 // ProfileConfigs configuration flags for all requested profiles.
 type ProfileConfigs struct {
 	mu sync.Mutex
-
+	// +checklocks:mu
+	wrt Writer
```
Shrekster, line 84: Could you please add a comment right above `wrt` to help the readers understand what is it about ?

internal/pproflogging/pproflogging.go
Shrekster: Sounds good. Thanks for the clarification.

internal/pproflogging/pproflogging.go
aaron-kasten: I can add that in a followon PR with some additional unit testing and comments.

internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +103,41 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
...
+	q := &ProfileConfigs{
+		wrt: wrt,
+	}
+
+	return q
+}
+
+// LoadProfileConfig configure PPROF profiling from the config in ppconfigss.
+func LoadProfileConfig(ctx context.Context, ppconfigss string) (map[ProfileName]*ProfileConfig, error) {
+	// if empty, then don't bother configuring but emit a log message - use might be expecting them to be configured
```
Shrekster, line 116:
```suggestion
	// if empty, then don't bother configuring but emit a log message - user might be expecting them to be configured
```

internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +103,41 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
...
+		log(ctx).Debug("no profile configuration. skipping PPROF setup")
+		return nil, nil
+	}
+
+	bufSizeB := DefaultDebugProfileDumpBufferSizeB
+
+	// look for matching services.  "*" signals all services for profiling
+	log(ctx).Info("configuring profile buffers")
+
+	// acquire global lock when performing operations with global side-effects
```
Shrekster, line 127: do you still need the global lock here since you are mostly doing in-memory work ? Maybe this comment is now obsolete ?

aaron-kasten: Its to prevent PPROF dump output from overlapping in the log output.  If they get mixed in the output its impossible to distinguish which base-64 encoded line belongs to which profile.

internal/pproflogging/pproflogging.go
```diff
@@ -97,10 +103,41 @@ var pprofProfileRates = map[ProfileName]pprofSetRate{
...
+type ProfileConfig struct {
+	flags []string
+	buf   *bytes.Buffer
+}
+
 // GetValue get the value of the named flag, `s`.  False will be returned
 // if the flag does not exist. True will be returned if flag exists without
 // a value.
-func (p ProfileConfig) GetValue(s string) (string, bool) {
+func (p *ProfileConfig) GetValue(s string) (string, bool) {
```
Shrekster, line 140: Shouldn't this be a non-pointer  for receiver methods which are read-only ?

internal/pproflogging/pproflogging.go
```diff
@@ -127,14 +164,19 @@ func parseProfileConfigs(bufSizeB int, ppconfigs string) map[ProfileName]*Profil
 		flagValue := ""
 
 		if len(profileFlagNameValuePairs) > 1 {
+			// only <key>=<value? allowed
```
Shrekster:
```suggestion
			// only <key>=<value> allowed
```

internal/pproflogging/pproflogging.go
aaron-kasten: Yes, it can be.  I wanted to stick with a "default" receiver type for the first implementation.  This code has never seen production, so I wanted to stick with some "least-common-denominator" implementations in the event that its not useful or needs to be changed.  I can make changes, if necessary, on a follow-on PR.

internal/pproflogging/pproflogging.go
```diff
@@ -281,7 +328,10 @@ func DumpPem(bs []byte, types string, wrt *os.File) error {
 		return fmt.Errorf("could not write PEM: %w", err2)
 	}
 
-	// did not get a read error.  file ends in newline
+	if err0 != nil {
```
Shrekster, line 331: What if `wrt.Write()` failed on this line above ?

```
for err1 == nil && err2 == nil {
    :
    :
    _, err2 = wrt.Write(ln)
```

The for loop will break out and we'll leave the above goroutine deadlocked ? Maybe I am not reading the code correctly... ?

internal/pproflogging/pproflogging.go
Shrekster: lets make this simple change if you agree:

```suggestion
func (p ProfileConfig) GetValue(s string) (string, bool) {
```

internal/pproflogging/pproflogging.go
Shrekster: @aaron-kasten I'd prefer that we either add the comment here or you can file an issue in Kopia and reference it here to track it, so we don't miss this. Otherwise, it'd be hard to track this comment.

Marking unresolved.

internal/pproflogging/pproflogging.go
Shrekster: got it, thanks for the clarification

internal/pproflogging/pproflogging.go
Shrekster: thank you for fixing

internal/pproflogging/pproflogging_test.go
```diff
@@ -124,3 +219,140 @@ func TestDebug_newProfileConfigs(t *testing.T) {
...
+
+func TestErrorWriter(t *testing.T) {
+	eww := &ErrorWriter{mx: 5, err: io.EOF}
+	n, err := eww.WriteString("Hello World")
+	require.ErrorIs(t, io.EOF, err)
+	require.Equal(t, 5, n)
+	require.Equal(t, "Hello", string(eww.bs))
+}
+
+// ErrorWriter allow injection of errors into the write stream.  There are a few
```
Shrekster, line 320:
```suggestion
// ErrorWriter allows injection of errors into the write stream.  There are a few
```

internal/pproflogging/pproflogging_test.go
Shrekster: @aaron-kasten I think this is fixed now. Can we mark resolved ? IS anything left to do here ?

aaron-kasten: @Shrekster I'm unsure.  @julio-lopez had a point in which local variables could be used for the save-and-restore.   I've made those changes but have not heard back from @julio-lopez

Shrekster: I think this is resolved. Julio is occupied with index compaction work. I'll mark this resolved here.

Shrekster (approved):
Perfect! Thanks for taking in all the feedback.

internal/pproflogging/pproflogging.go
```diff
@@ -63,23 +63,33 @@ const (
...
 }
 
 // ProfileConfigs configuration flags for all requested profiles.
 type ProfileConfigs struct {
 	mu sync.Mutex
-
+	// wrt represents the final destination for the PPROF PEM output.  Typically,
+	// this is attached to stderr or log output.  A custom writer is used because
+	// not all loggers support line oriented output through the io.Writer interface...
+	// support is often attached th a io.StringWriter.
```
julio-lopez, line 86: nit:
```suggestion
	// support is often attached to a io.StringWriter.
```

internal/pproflogging/pproflogging_test.go
```diff
@@ -124,3 +219,140 @@ func TestDebug_newProfileConfigs(t *testing.T) {
...
+
+	// append the bytes to the local buffer just
+	// in case someone wants to know.
+	p.bs = append(p.bs, bs[:n]...)
+	if n < len(bs) {
+		// here we assume that any less than len(bs)
+		// bytes written returns an error.  This
+		// allows setting ErrorWriter up once
+		// to produce an error after multiple
+		// writes
```
julio-lopez, line 348: It seems that there's a corner case here though when `len(bs) == 0`, is that the case?
That is, if an error was set and previously returned, a subsequent call to `p.Write([]byte{})` would return without an error? Is that intended behavior?

internal/pproflogging/pproflogging_test.go
```diff
@@ -124,3 +219,140 @@ func TestDebug_newProfileConfigs(t *testing.T) {
...
+
+//nolint:gocritic
+func saveLockEnv(t *testing.T) {
+	t.Helper()
+
+	oldEnv := os.Getenv(EnvVarKopiaDebugPprof)
+
+	t.Cleanup(func() {
+		// restore the old environment
+		t.Setenv(EnvVarKopiaDebugPprof, oldEnv)
```
julio-lopez, line 308: It seems that this has little or no effect.
`t.Setenv()` will internally setup a cleanup function to reset the value to whatever value it found when `t.Setenv()` was called.

In this case, `t.Setenv()` is being called as part of the cleanup, so this is after the test has executed. Then
- it will read the current value for the env var (`v1`), which may have been set during the test
- set the env var value to `oldEnv`
- setup a `t.Cleanup` function to reset the env var back to `v1`. It may be the case that this newly set cleanup function does not execute at all. I do not know what is the expected behavior or calling `t.Cleanup()` from within a `t.Cleanup()` callback.

https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/testing/testing.go;l=1299

The suggestion is to use `t.Setenv` directly in the tests, instead of the `saveLockEnv` implementation. 

Related, TIL, `t.Setenv` ensures that neither the test, nor any of is ancestors (in the case of a subtest) are executed in parallel, which is what is desired in this case.

julio-lopez (commented):
@aaron-kasten LG overall.

A few nits, see inline comments. This can be merged as is, and the comments can be addressed in a separate PR. PTAL though.

PR #3670, opened 2024-02-22
feat(general): Add alternate key derivation algorithms

bathina2:
This PR adds the ability to select a different key derivation algorithm instead of `scrypt`. We add the `pbkdf2` algorithm as an example. However, the default is still `scrypt` and we are not yet using the `pbkdf2` algorithm to generate keys yet.

### DIFF
--- a/internal/crypto/pbkdf.go
+++ b/internal/crypto/pbkdf.go
@@ -1,13 +1,16 @@
+//go:build !testing
+// +build !testing
+
 package crypto
 
 import (
 	"crypto/sha256"
-	"fmt"
 
+	"github.com/pkg/errors"
 	"golang.org/x/crypto/pbkdf2"
 )
 
-// The NIST recommended iterations for PBKDF2 with SHA256 hash is 600,000
+// The NIST recommended iterations for PBKDF2 with SHA256 hash is 600,000.
 const pbkdf2Sha256Iterations = 1<<20 - 1<<18 // 786,432
 
 // The NIST recommended minimum size for a salt for pbkdf2 is 16 bytes.
@@ -20,8 +23,9 @@ const minPbkdfSha256SaltSize = 32 // size in bytes == 256 bits
 func init() {
 	RegisterKeyDerivationFunc(Pbkdf2Algorithm, func(password string, salt []byte) ([]byte, error) {
 		if len(salt) < minPbkdfSha256SaltSize {
-			return nil, fmt.Errorf("required salt size is atleast %d bytes", minPbkdfSha256SaltSize)
+			return nil, errors.Errorf("required salt size is atleast %d bytes", minPbkdfSha256SaltSize)
 		}
+
 		return pbkdf2.Key([]byte(password), salt, pbkdf2Sha256Iterations, MasterKeyLength, sha256.New), nil
 	})
 }
--- a/internal/crypto/pbkdf.go
+++ b/internal/crypto/pbkdf.go
@@ -1,0 +1,16 @@
+package crypto
+
+import (
+	"crypto/sha256"
+
+	"golang.org/x/crypto/pbkdf2"
+)
+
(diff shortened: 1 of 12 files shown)


jkowalski:
this is interesting, I'm curious if there's any particular reason for this? is it to support memory-constrained environments? compliance?

bathina2:
@jkowalski FIPS compliance. Besides the key derivation, Kopia can mostly be run in a FIPS compliant manner (supports selecting encryption and hashing etc).

bathina2:
@jkowalski I've simplified the PR a bit. What do you think of this approach? Are there any other changes you would like to see?

internal/user/user_profile_hash_v1.go
```diff
@@ -6,54 +6,55 @@ import (
...
+	if len(hashedPassword) != v1SaltLength+crypto.MasterKeyLength {
 		return false
 	}
 
 	salt := hashedPassword[0:v1SaltLength]
 
-	h := computePasswordHashV1(password, salt)
+	h, err := computePasswordHashV1(password, salt)
+	if err != nil {
+		return false
```
Shrekster, line 56: Returning a `false` is still a valid response from `isValidPasswordV1()`. It seems like this function has a tri-state, including error state.

Would be good to log the error from `computePasswordHashV1` as well

if you believe we should not hit the error ever then we should panic() instead.

internal/crypto/pbkdf.go
```diff
@@ -0,0 +1,23 @@
...
+	"fmt"
+
+	"golang.org/x/crypto/pbkdf2"
+)
+
+// The NIST recommended iterations for PBKDF2 with SHA256 hash is 600,000
+const pbkdf2Sha256Iterations = 1<<20 - 1<<18 // 786,432
+
+// The NIST recommended minimum size for a salt to be used for pbkdf2
+const minPbkdfSaltSize = 16 // size in bytes == 128 bits
```
Shrekster, line 14: `A good rule of thumb is to use a salt that is the same size as the output of the hash function. For example, the output of SHA256 is 256 bits (32 bytes), so the salt should be at least 32 random bytes.`

REF: https://crackstation.net/hashing-security.htm

Should not this be 32 instead for both scrypt and pbkdf2 ?

internal/user/user_profile_hash_v1.go
bathina2: Hmm, it's very possible that we can hit an error here (lets say an invalid algorithm is passed). So would logging and retuning false be sufficient?

bathina2: I can `log.Fatal(err)` so that way we will quit with an error.

internal/crypto/pbkdf.go
bathina2: I believe throughout the code we pass in a 32 byte salt so I can change this to make that the minimum. I picked 16 because it was the minimum recommended for NIST. I will update the comments with this information.

Shrekster: Also, changing the salt size will break existing users, so it is required to be consistent with the existing code unless we have a key migration path forward.

internal/user/user_profile_hash_v1.go
Shrekster: `log.Fatal()` works, thanks!

Shrekster:
Screenshot 2024-03-12 at 5 17 31 PM

Looks like GitHub is misbehaving on this PR. @bathina2 and I attempted to push commits to this branch which are not showing up here. No further changes are visible, so we'll be reopening this PR.

Shrekster:
@jkowalski note that we had to duplicate the `MasterKeyLength` constant in each file with a `!testing` build flag. If you have a better approach let us know and we'll update it. Taking this in for now.

julio-lopez:
Superseded by:
- #3725
- #3770
- #3779
- #3799
- #3816

PR #3678, opened 2024-02-24
feat(repository): added metrics timeseries computation

jkowalski:
This code is not used anywhere yet, will be used for reporting over repository metrics.


internal/metrics/metrics_timeseries.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		if lastPoint.After(maxTime) {
+			maxTime = lastPoint
+		}
+
+		totalDuration := s.EndTime.Sub(s.StartTime)
+		pbt := ts[timeSeriesID]
+
+		// we know that between [StartTime, EndTime] te counter increased by `value`
```
Shrekster, line 100:
```suggestion
		// we know that between [StartTime, EndTime] the counter increased by `value`
```

PR #3695, opened 2024-02-29
fix(providers): Reduce memory allocation (gc thrash)

aaron-kasten:
This change causes Minio libraries to reduce memory allocations.  Specifically, buffer allocations for S3 upload.


jkowalski:
this looks good but the code is very tricky and needs unit tests before we can merge

aaron-kasten:
> this looks good but the code is very tricky and needs unit tests before we can merge

@jkowalski I'm unsure of how to unit test this.

aaron-kasten:
image

This is an example of the memory allocation savings for a 5GB snapshot - approximately 3x the memory of the snapshot size in this case.

redgoat650 (commented):
Looks good, just a suggestion for increasing test coverage:

I agree with @jkowalski's sentiment from his earlier review: "this code is very tricky". I manually inspected the logic and it _SEEMS_ fine, but it's very susceptible to off-by-one errors or regressions if not tested completely.

The unit test you added covers some corner+error cases, but seems to be missing some happy-path buffer filling and content-checking cases.

I've drawn up a diagram of some test cases that we might add for the `ReadAt` implementation. For each case, we could assert the expected number of bytes written, the correct `p []byte` resulting content (to ensure we're not starting from the wrong offset or leaving the output buffer incomplete), and whether it expects an error return.

The test cases are various permutations of buffer lengths and offsets aligning/overlapping in various ways, meant to catch any off-by-one errors that might get introduced.

I hope you find this sketch intelligible. I've included a "key" up top to explain what each proposed test case looks like. Each test case is comprised of a BLUE `WriteBuffer` which can consist of one or more `b.b.Slices`, a GREEN `p []bytes` target buffer, and an ORANGE `offset` value.

Feel free to incorporate any/all/none of the cases in the unit test; I'm just sharing some cases I would have considered adding, just to be extra sure of the implementation. Other than strengthening tests, as far as I can tell the change looks good!

image

aaron-kasten:
@redgoat650 Lets chat later.  Test coverage has been enlarged to deal with corner cases not for Kopia ReaderAt

aaron-kasten:
Im unclear on how the test cases you mention are specific to this implementation vs the Buffer implementation in general.  The thinking is that anything that applies to the original Buffer should be implemented (as its already been accepted).

aaron-kasten:
Are you proposing that additional coverage be introduced for the original Buffer implementation?

redgoat650:
Ah! I think I missed that all of the `ReaderAt` cases I mentioned are probably being tested by this line
```go
	require.NoError(t, iotest.TestReader(reader, buf))
```
Is that accurate, @aaron-kasten? If so, that's great. Perhaps we can add a comment to draw attention to it? It was easy to miss when reading through the test case.

aaron-kasten:
A majority, at least for compliance with the Go apis. The scatter-gather buffer in Kopia is a different beast though.
Im wondering if it needs additional tests ... which your comments imply ... then it might be worth while expanding them.
Its later on a friday for me.  Ill take a closer look at the code and your comments on Monday and see what can be done.

aaron-kasten:
Thank you for your detailed observations ... its a big help having something so specific to work from.

aaron-kasten:
> Ah! I think I missed that all of the `ReaderAt` cases I mentioned are probably being tested by this line
> 
> ```go
> 	require.NoError(t, iotest.TestReader(reader, buf))
> ```
> 
> Is that accurate, @aaron-kasten? If so, that's great. Perhaps we can add a comment to draw attention to it? It was easy to miss when reading through the test, and I thought we were missing the obvious cases 😄
> 
> [edit to add the `TestReader` description]:
> 
> ```go
> // TestReader tests that reading from r returns the expected file content.
> // It does reads of different sizes, until EOF.
> // If r implements [io.ReaderAt] or [io.Seeker], TestReader also checks
> // that those operations behave as they should.
> //
> // If TestReader finds any misbehaviors, it returns an error reporting them.
> // The error text may span multiple lines.
> ```

@redgoat650 In the future can you submit this as a change - ideally inline as a comment with a change or a change directly to the branch?  It saves on risking my interpretation of your intended meaning and saves steps (Its implied to be approved by you and does not need to go through another review cycle).

internal/gather/gather_bytes_test.go
```diff
@@ -173,6 +182,85 @@ func TestGatherBytesReadSeeker(t *testing.T) {
 	require.Error(t, err)
 }
 
+func TestGatherBytesReaderAt(t *testing.T) {
+	// assert some preconditions that the reader conforms to ReaderAt
+	contentBuf := make([]byte, 1234567)
```
redgoat650, line 187: As discussed with @Shrekster, we can probably make use of actual chunk size constants here. And we can iterate over a few key sizes, passing each into the iotest helper, just for an easy increase in coverage. I think @Shrekster will handle the quick enhancement, and I can approve.

jkowalski (approved):
LGTM, thanks for adding the test.

internal/gather/gather_bytes_test.go
```diff
@@ -173,6 +182,119 @@ func TestGatherBytesReadSeeker(t *testing.T) {
...
+			require.Equalf(t, defaultAllocator.chunkSize, preWrt.alloc.chunkSize,
+				"this test expects that the default-allocator will be used, but we are using: %#v", preWrt.alloc)
+
+			require.Lenf(t, buf, n, "unexpected size of data written, chunkSizeMultiplier: %f", chunkSizeMultiplier)
+
+			// get the reader out of the WriteBuffer so we can read what was written
+			preRdr := preWrt.inner.Reader()
+			_, ok := preRdr.(io.ReaderAt)
+			require.Truef(t, ok, "MUST implement io.ReaderAt, chunkSizeMultiplier: %f", chunkSizeMultiplier)
+			require.NoErrorf(t, iotest.TestReader(preRdr, buf),
```
redgoat650, line 292: nit: since this is THE important line, behind which all of the tests will run, I'd make it stand out a bit more, instead of cuddling it with the reader extraction above. Maybe give it its own newline and comment?
```suggestion
			require.Truef(t, ok, "MUST implement io.ReaderAt, chunkSizeMultiplier: %f", chunkSizeMultiplier)
			
			// execute standard ReadAt tests.
			require.NoErrorf(t, iotest.TestReader(preRdr, buf),
```

internal/gather/gather_bytes_test.go
```diff
@@ -173,6 +182,119 @@ func TestGatherBytesReadSeeker(t *testing.T) {
...
+	// times the size of the internal allocator byte slices.
+	//
+	// The idea here is to exercise the part of the buffer ReaderAt
+	// implementation where it has a longer buffer size than the size of the
+	// internal chunks of the buffer implementation. When we do this, the
+	// ReaderAt is forced to draw more data than it actually can from the
+	// first slice it found after searching for the current pointer in read
+	// cycle. Finally, it should increment the read index correctly.
+	//
+	// x.1 ... x.9
```
redgoat650, line 269: So IIUC, we're testing contentBufs with size `[3.1x, 3.2x, ..., 3.9x]` (x is the chunk size)? I was thinking we could try more interesting contentBuf sizes like `[0.5x, 1x, 2x, 2x-1, 2x+1,...]` that sort of thing (x is the chunk size). That would probe around the edges of the different regimes involved in the underlying buffer implementation. The current test just exercises the "multiple slices with a partial slice at the end" case.

redgoat650 (approved):
Some suggestions for improvement, otherwise looks good

Shrekster:
Added keep-open, label. I'd like to test this more thoroughly. Do not merge yet.

internal/gather/gather_bytes_test.go
Shrekster: done

internal/gather/gather_bytes_test.go
Shrekster: done

Shrekster:
> Added keep-open, label. I'd like to test this more thoroughly. Do not merge yet.

done with testing, should be good now; fixed some slice index increment bugs

internal/gather/gather_bytes.go
```diff
@@ -167,29 +167,30 @@ func (b *bytesReadSeekCloser) ReadAt(bs []byte, off int64) (int, error) {
 
 	n := copy(bs, curSlice[offset:])
 
-	// first chunk written, move on to the next
-	sliceNdx++
+	// if we were able to consume the entire slice then move to the next slice
```
Shrekster, line 170: I should add to the comment here to make an important behaviour slightly more evident here:

“if we have not read the full curSlice then later on we will simply return n, nil and not try to work on the next slice”.

internal/gather/gather_bytes_test.go
```diff
@@ -173,6 +182,157 @@ func TestGatherBytesReadSeeker(t *testing.T) {
...
+	for i := range contentBuf {
+		contentBuf[i] = uint8(i % math.MaxUint8)
+	}
+
+	type testCase struct {
+		name            string
+		inputBufferSize int
+	}
+	// Test some interesting input buffer sizes from a minimum three
+	// times the size of the internal allocator byte slices.
```
redgoat650, line 264: Comment can be updated I think
```suggestion
	// Test some interesting input buffer sizes from a 1-byte buffer to many multiples 
	// of the internal allocator chunk size.
```

internal/gather/gather_bytes_test.go
```diff
@@ -173,6 +182,157 @@ func TestGatherBytesReadSeeker(t *testing.T) {
...
+		{"2x-1", 2*defaultAllocator.chunkSize - 1},
+		{"2x", 2 * defaultAllocator.chunkSize},
+		{"2x+1", 2*defaultAllocator.chunkSize + 1},
+		{"2.5x", int(2.5 * float64(defaultAllocator.chunkSize))},
+
+		{"3x-1", 3*defaultAllocator.chunkSize - 1},
+		{"3x", 3 * defaultAllocator.chunkSize},
+		{"3x+1", 3*defaultAllocator.chunkSize + 1},
+		// The 3.5x is already tested later in this code
+		// {"3.5x", int(3.5 * float64(defaultAllocator.chunkSize))},
```
redgoat650, line 283: we can remove the case if it's unused
```suggestion
```

redgoat650 (approved):
The added test cases look great, thanks @Shrekster ! And nice catch on the 1-chunk bug 🚀

julio-lopez (commented):
@aaron-kasten Thanks for submitting the change, this is will result in significant memory usage reduction for users with repositories on S3 compatible storage. And kudos for profiling, diagnosing, tracking down the issue 🎖️ 🎩 and adding the tests.

And thanks to @redgoat650 and @Shrekster for the detailed review in this involved codepath, and for chipping in with the fixes.

Great job team.

Please feel free to merge once CI passes.

julio-lopez:
I also triggered the storage provider tests. Please keep an eye on those before merging.

https://github.com/kopia/kopia/actions/runs/8504413695/job/23291280771?pr=3695

aaron-kasten:
@julio-lopez (@Shrekster ) got an error in the code-coverage test: `EPIPE`.  Something happened when trying to upload coverage results (?).

Have either of you seen this before?  Is there a standard resolution?

```
gpg: directory '/home/runner/.gnupg' created
gpg: keybox '/home/runner/.gnupg/pubring.kbx' created
gpg: /home/runner/.gnupg/trustdb.gpg: trustdb created
gpg: key 806BB28AED779869: public key "Codecov Uploader (Codecov Uploader Verification Key) <security@codecov.io>" imported
gpg: Total number processed: 1
gpg:               imported: 1

Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:160:15)
    at writeGeneric (node:internal/stream_base_commons:151:3)
    at Socket._writeGeneric (node:net:952:11)
    at Socket._write (node:net:964:8)
    at writeOrBuffer (node:internal/streams/writable:447:12)
    at _write (node:internal/streams/writable:389:10)
    at Socket.Writable.end (node:internal/streams/writable:665:17)
    at Socket.end (node:net:722:31)
    at module.exports (/home/runner/work/_actions/codecov/codecov-action/54bcd8715eee62d40e33596ef5e8f0f48dbbccab/node_modules/gpg/lib/spawnGPG.js:50:1)
    at Object.call (/home/runner/work/_actions/codecov/codecov-action/54bcd8715eee62d40e33596ef5e8f0f48dbbccab/node_modules/gpg/lib/gpg.js:28:1)
```

aaron-kasten:
@Shrekster @jkowalski @julio-lopez I made some final changes. Please take a look.

internal/gather/gather_bytes.go
```diff
@@ -166,31 +166,28 @@ func (b *bytesReadSeekCloser) ReadAt(bs []byte, off int64) (int, error) {
...
 	n := copy(bs, curSlice[offset:])
+	m := n
 
 	// if we were able to consume the entire slice then move to the next slice
-	if int64(n) == (int64(len(curSlice)) - offset) {
-		// first chunk written, move on to the next
-		sliceNdx++
-	}
+	// first chunk written, move on to the next
+	sliceNdx++
```
julio-lopez: @aaron-kasten IIUC, @Shrekster modified these 3 lines because he ran into a bug. Could you folks please double check?

I'm going to re-trigger the providers test in the mean time.

internal/gather/gather_bytes.go
```diff
@@ -120,6 +124,78 @@ type bytesReadSeekCloser struct {
...
+	// accounting: keep track of total number of bytes written in n and
+	// number of bytes written from the current slice in m
+	n := m
+
+	// move on to next and then check if all slices were consumed
+	sliceNdx++
+
+	slicesN := len(slices)
+
+	for maxBsIndex > n && sliceNdx < slicesN {
```
Shrekster, line 178: what happens if `maxBsIndex = n` ? We probably do not want to enter this loop then, right ?

Shrekster: I think the cop() below is probably saving you from a panic here, but better to not even enter this loop when we have filled the input buffer

internal/gather/gather_bytes.go
```diff
@@ -120,6 +124,85 @@ type bytesReadSeekCloser struct {
...
+	n := m
+
+	// move on to next and then check if all slices were consumed
+	sliceNdx++
+
+	slicesN := len(slices)
+
+	// while there are more bytes to read (maxBsIndex > n) and there are more
+	// slices left (sliceNdx < slicesN)
+	for maxBsIndex > n && sliceNdx < slicesN {
```
redgoat650, line 182: perhaps we can make the code more self-describing in addition to comments?
```suggestion
	for moreRoomInOutput(n, maxBsIndex) && moreSlicesLeft(sliceNdx, slicesN) {
```

redgoat650 (approved):
Thanks for adding clarity. One optional suggestion, but otherwise looks good

julio-lopez:
> @julio-lopez (@Shrekster ) got an error in the code-coverage test: `EPIPE`. Something happened when trying to upload coverage results (?).

Likely an infra issue on their side?? dunno. 🤷🏼

jkowalski:
🎉

PR #3696, opened 2024-03-01
test(general): Increase timeout to check for generated TLS cert

chaitalisg:
The TLS certificate generation may not always happen within three minutes. This PR increases the retry period to 15 minutes.


tests/tools/kopiarunner/kopia_snapshotter.go
```diff
@@ -28,7 +28,7 @@ const (
 	noCheckForUpdatesFlag   = "--no-check-for-updates"
 	noProgressFlag          = "--no-progress"
 	parallelFlag            = "--parallel"
-	retryCount              = 180
+	retryCount              = 300
```
redgoat650, line 31:
```suggestion
	retryCount              = 900
```

How about we give it a generous bump to 15 minutes or so? That gives plenty of wiggle room to recover in case we're extra slow sometimes.

tests/tools/kopiarunner/kopia_snapshotter.go
```diff
@@ -395,7 +395,7 @@ func (ks *KopiaSnapshotter) ConnectOrCreateRepoWithServer(serverAddr string, arg
 	var cmdErr error
 
 	if cmd, cmdErr = ks.CreateServer(serverAddr, serverArgs...); cmdErr != nil {
-		return nil, "", cmdErr
+		return nil, "", errors.Wrap(cmdErr, "CreateServer failed")
 	}
 
 	if err := certKeyExist(context.TODO(), tlsCertFile, tlsKeyFile); err != nil {
```
redgoat650, line 401: I suggest we make our lives easier in the future, perhaps. How about we also leave a breadcrumb that maybe the repo open is taking too long?
```go
	if err := certKeyExist(context.TODO(), tlsCertFile, tlsKeyFile); err != nil {
		if buf, ok := cmd.Stderr.(*bytes.Buffer); ok 
			// If the STDERR buffer does not contain any obvious error output,
			// it is possible the async server creation above is taking a long time
			// to open the repository, and we timed out waiting for it to write the TLS certs.
			// Please see <add a link to some context here?> for more info.
			log.Print("failure in certificate generation:", buf.String())
		}
```

redgoat650 (approved):
Thanks @chaitalisg - I left some suggestions and approval

tests/tools/kopiarunner/kopia_snapshotter.go
chaitalisg: Good point. Done.

Shrekster (approved):

Shrekster:
This seems a flake: https://github.com/kopia/kopia/issues/3707

PR #3705, opened 2024-03-05
build(deps): upgrade Prometheus and reacted to breaking change

jkowalski:


julio-lopez (commented):
🥇Thanks for doing this.

PR #3710, opened 2024-03-06
fix(deps): upgrade google.golang.org/protobuf to v1.33.0

julio-lopez:
This change upgrades to google.golang.org/protobuf@v1.33.0 to mitigate a security vulnerability.

> Version v1.33.0 of the google.golang.org/protobuf module fixes a bug in the google.golang.org/protobuf/encoding/protojson package which could cause the Unmarshal function to enter an infinite loop when handling some invalid inputs. This condition could only occur when unmarshaling into a message which contains a google.protobuf.Any value, or when the UnmarshalOptions.UnmarshalUnknown option is set. Unmarshal now correctly returns an error when handling these inputs.

This is CVE-2024-24786.

— https://groups.google.com/g/golang-announce/c/ArQ6CDgtEjY

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,7 +67,7 @@ require (
 	golang.org/x/text v0.14.0
 	google.golang.org/api v0.166.0
 	google.golang.org/grpc v1.62.0
-	google.golang.org/protobuf v1.32.0
+	google.golang.org/protobuf v1.33.0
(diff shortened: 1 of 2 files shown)


julio-lopez:
Merged to unblock CI.

PR #3713, opened 2024-03-07
fix(server): Return internal server errors for Content Get errors

redgoat650:
The Content Get handler of the API server currently only returns `content not found` error responses. Other errors are swallowed and returned as successful, along with a `nil` data payload.

Fix by handling other errors as internal errors instead, equivalent to the behavior of e.g. `handleContentInfo`.


jkowalski:
LGTM. Btw this code path is no longer executed by default (we use grpc now). We should perhaps remove repository server APIs that use old http protocol completely since they were not in use for severa relapses now and have significant correctness issues.

jkowalski (approved):

jkowalski:
Filed #3716 to track the removal of the REST API.

PR #3715, opened 2024-03-09
chore(ci): install apple notarization key before build

### LINKED ISSUE #3711: macOS notarization frequently failing in CI
For example see https://github.com/kopia/kopia/actions/runs/8181303361/job/22370856752

Should it be moved out, so it is performed less frequently? perhaps the packages could be produced nightly, or with some other less frequent cadence.

Maybe related to #3699

Ref:
- #488
- #3202
- #3699

jkowalski:
This should recent regression after Electron upgrade.

Fixes #3711


jkowalski:
(this can only be tested post-merge)

PR #3721, opened 2024-03-12
refactor(general): add epoch.getKeyRange helper

julio-lopez:
Extracted from #3651.

Thanks to @plar and @redgoat650 for the suggestions.

Ref:

- #3603
- #3645
- #3638
- #3639

### DIFF
--- a/internal/epoch/epoch_utils.go
+++ b/internal/epoch/epoch_utils.go
@@ -6,6 +6,8 @@ import (
 	"time"
 	"unicode"
 
+	"github.com/pkg/errors"
+
 	"github.com/kopia/kopia/repo/blob"
 )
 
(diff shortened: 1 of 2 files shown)


internal/epoch/epoch_utils.go
```diff
@@ -91,3 +93,57 @@ func deletionWatermarkFromBlobID(blobID blob.ID) (time.Time, bool) {
...
+		}
+
+		count++
+	}
+
+	if count == 0 {
+		return intRange{}, nil
+	}
+
+	if uint(hi-lo) > count-1 {
```
redgoat650, line 144: Maybe we can add a quick note to remind ourselves of the implementation "trick" in the future - the fact that integer keys are necessarily unique (because the input is a map) means that this simple check tests for contiguity in the range.

redgoat650 (approved):
Looks good, just one comment suggestion

internal/epoch/epoch_utils.go
julio-lopez: OK, I'll merge it and address the comment in a sec.

internal/epoch/epoch_utils.go
```diff
@@ -91,3 +93,57 @@ func deletionWatermarkFromBlobID(blobID blob.ID) (time.Time, bool) {
...
+
+var errNonContiguousRange = errors.New("non-contiguous range")
+
+// constants from the standard math package.
+const (
+	//nolint:gomnd
+	intSize = 32 << (^uint(0) >> 63) // 32 or 64
+
+	maxInt = 1<<(intSize-1) - 1
+	minInt = -1 << (intSize - 1)
```
plar, line 117: Any reason to create consts here instead of using [math.MinMaxConsts](https://pkg.go.dev/math#pkg-constants)?

internal/epoch/epoch_utils.go
julio-lopez: @redgoat650 PTAL @ #3723

internal/epoch/epoch_utils.go
```diff
@@ -91,3 +93,57 @@ func deletionWatermarkFromBlobID(blobID blob.ID) (time.Time, bool) {
...
+)
+
+// Returns a continuous close-open epoch range for the keys, that is [lo, hi).
+// A range of the form [v,v) means the range is empty.
+// When the range is not continuous an error is returned.
+func getKeyRange[E any](m map[int]E) (intRange, error) {
+	var count uint
```
plar, line 124: Is there a specific reason to use `uint` here? The built-in `len` returns `int`.

internal/epoch/epoch_utils.go
julio-lopez: To avoid taking a dependency here on the math package just for importing 2 constants.
A test was included for sanity. The test does take a dependency on the math package.

plar: In kopia we already have `import "math"` in multiple places  and `epoch` package is under `internal` and couldn't be used as a standalone package outside the project. So we can save 10+ lines of the code with one import. :)

PR #3723, opened 2024-03-12
refactor(general): add clarification comment

julio-lopez:
Followup to #3721

### DIFF
--- a/internal/epoch/epoch_utils.go
+++ b/internal/epoch/epoch_utils.go
@@ -143,7 +143,7 @@ func getKeyRange[E any](m map[int]E) (intRange, error) {
 
 	// hi and lo are from unique map keys, so for the range to be continuous
 	// the difference between hi and lo cannot be larger than count -1.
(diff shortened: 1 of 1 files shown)


internal/epoch/epoch_utils.go
```diff
@@ -141,6 +141,9 @@ func getKeyRange[E any](m map[int]E) (intRange, error) {
 		return intRange{}, nil
 	}
 
+	// hi and lo are from unique map keys, so for the range to be continuous
+	// the difference between hi and lo cannot be larger than count -1.
+	// For example, if lo==2 & hi==4, then count must be 3 => {2, 3, 4}.
```
redgoat650, line 146: Just a slight rewording suggestion here, since `map[int]any{2, 4}` also has `lo==2 & hi==4`, but count is not 3 and it is not continuous. The `if...then...` makes it a bit confusing.
```suggestion
	// For example, if lo==2, hi==4, and count == 3, the range must be contiguous => {2, 3, 4}.
```

PR #3725, opened 2024-03-13
feat(general): Add alternate key derivation algorithms

bathina2:
Refer to https://github.com/kopia/kopia/pull/3670 for discussion

### DIFF
--- a/internal/crypto/key_derivation_nontest.go
+++ b/internal/crypto/key_derivation_nontest.go
@@ -5,22 +5,37 @@ package crypto
 
 import (
 	"github.com/pkg/errors"
-	"golang.org/x/crypto/scrypt"
+)
+
+const (
+	// MasterKeyLength describes the length of the master key.
+	MasterKeyLength = 32
+
+	// ScryptAlgorithm is the key for the scrypt algorithm.
+	ScryptAlgorithm = "scrypt-65536-8-1"
+	// Pbkdf2Algorithm is the key for the pbkdf algorithm.
+	Pbkdf2Algorithm = "pbkdf2"
 )
 
 // DefaultKeyDerivationAlgorithm is the key derivation algorithm for new configurations.
-const DefaultKeyDerivationAlgorithm = "scrypt-65536-8-1"
+const DefaultKeyDerivationAlgorithm = ScryptAlgorithm
+
+type keyDerivationFunc func(password string, salt []byte) ([]byte, error)
+
+//nolint:gochecknoglobals
+var keyDerivers = map[string]keyDerivationFunc{}
+
+// RegisterKeyDerivationFunc registers various key derivation functions.
+func RegisterKeyDerivationFunc(name string, keyDeriver keyDerivationFunc) {
+	keyDerivers[name] = keyDeriver
+}
 
 // DeriveKeyFromPassword derives encryption key using the provided password and per-repository unique ID.
 func DeriveKeyFromPassword(password string, salt []byte, algorithm string) ([]byte, error) {
-	const masterKeySize = 32
-
-	switch algorithm {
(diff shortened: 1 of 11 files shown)


Shrekster (approved):

internal/crypto/key_derivation_nontest.go
```diff
@@ -5,22 +5,37 @@ package crypto
...
+)
+
+const (
+	// MasterKeyLength describes the length of the master key.
+	MasterKeyLength = 32
+
+	// ScryptAlgorithm is the key for the scrypt algorithm.
+	ScryptAlgorithm = "scrypt-65536-8-1"
+	// Pbkdf2Algorithm is the key for the pbkdf algorithm.
+	Pbkdf2Algorithm = "pbkdf2"
```
julio-lopez, line 17: The `pbkdf2` "algorithm name" is too generic. This should be more specific and include what parameters are being used. For example, this should be something along the lines of `pbkdf2-sha256-<num-iterations>` or even more specific depending on what parameters we allow to overwrite.


(Granted, the "algorithm" infix does not make for a good name for these constants since it is too generic, and thus does not precisely convey what the values for these constants are)

internal/crypto/key_derivation_testing.go
```diff
@@ -12,6 +12,9 @@ import (
 // DefaultKeyDerivationAlgorithm is the key derivation algorithm for new configurations.
 const DefaultKeyDerivationAlgorithm = "testing-only-insecure"
 
+// MasterKeyLength describes the length of the master key.
+const MasterKeyLength = 32
```
julio-lopez, line 16: Is this used anywhere? or is it coming in a later PR?

julio-lopez (commented):
@bathina2 PTAL at the inline comments

internal/crypto/key_derivation_nontest.go
julio-lopez: Addressed in #3811

internal/crypto/pbkdf.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"crypto/sha256"
+
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/pbkdf2"
+)
+
+// The NIST recommended iterations for PBKDF2 with SHA256 hash is 600,000.
+const pbkdf2Sha256Iterations = 1<<20 - 1<<18 // 786,432
```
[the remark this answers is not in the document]
julio-lopez: #3810

internal/user/user_profile_hash_v1.go
```diff
@@ -6,54 +6,56 @@ import (
...
+	if len(hashedPassword) != v1SaltLength+crypto.MasterKeyLength {
 		return false
 	}
 
 	salt := hashedPassword[0:v1SaltLength]
 
-	h := computePasswordHashV1(password, salt)
+	h, err := computePasswordHashV1(password, salt, keyDerivationAlgorithm)
+	if err != nil {
+		panic(err)
```
julio-lopez, line 57: This should propagate the error instead of calling `panic`king.

Previously, computePasswordHashV1 could panic if `scrypt.Key` returned an error. `scrypt.Key` only returns an error when _"... the parameters do not satisfy the limits"_ https://pkg.go.dev/golang.org/x/crypto/scrypt#Key
The parameters specified to `scrypt.Key` were constants within the valid range, thus an error was expected to never be returned, and the panic was an "impossibility".

In this implementation, it is possible for `crypto.DeriveKeyFromPassword` to return an error for other conditions, such an invalid key derivation algorithm. This means, that the `panic` here is no longer an "impossibility", and instead the error should be handled or propagated.

This is addressed in #3907

PR #3727, opened 2024-03-13
refactor(general): add epoch.Manager.MaybeGenerateRangeCheckpoint

julio-lopez:
Ref:
- Subset and dependency of #3651
- #3638
- #3639

### DIFF
--- a/internal/epoch/epoch_manager.go
+++ b/internal/epoch/epoch_manager.go
@@ -593,6 +593,34 @@ func (e *Manager) loadSingleEpochCompactions(ctx context.Context, cs *CurrentSna
 	return nil
 }
 
+// MaybeGenerateRangeCheckpoint may create a new range index for all the
+// individual epochs covered by the new range. If there are not enough epochs
+// to create a new range, then a range index is not created.
+func (e *Manager) MaybeGenerateRangeCheckpoint(ctx context.Context) error {
+	p, err := e.getParameters(ctx)
+	if err != nil {
+		return err
+	}
+
+	cs, err := e.committedState(ctx, 0)
+	if err != nil {
+		return err
+	}
+
+	latestSettled, firstNonRangeCompacted, compact := getRangeToCompact(cs, *p)
+	if !compact {
+		e.log.Debug("not generating range checkpoint")
+
+		return nil
+	}
+
+	if err := e.generateRangeCheckpointFromCommittedState(ctx, cs, firstNonRangeCompacted, latestSettled); err != nil {
+		return errors.Wrap(err, "unable to generate full checkpoint, performance will be affected")
+	}
+
+	return nil
+}
+
 func (e *Manager) maybeGenerateNextRangeCheckpointAsync(ctx context.Context, cs CurrentSnapshot, p *Parameters) {
 	latestSettled, firstNonRangeCompacted, compact := getRangeToCompact(cs, *p)
 	if !compact {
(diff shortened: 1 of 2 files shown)


chaitalisg (approved):
Good test coverage!
LGTM!

internal/epoch/epoch_manager.go
```diff
@@ -593,19 +593,38 @@ func (e *Manager) loadSingleEpochCompactions(ctx context.Context, cs *CurrentSna
...
 }
 
-func (e *Manager) maybeGenerateNextRangeCheckpointAsync(ctx context.Context, cs CurrentSnapshot, p *Parameters) {
-	latestSettled := cs.WriteEpoch - numUnsettledEpochs
-	if latestSettled < 0 {
-		return
+// MaybeGenerateRangeCheckpoint may create a new range index for all the
+// individual epochs covered by the new range. If there are not enough epochs
+// to create a new range, then a range index is not created.
+func (e *Manager) MaybeGenerateRangeCheckpoint(ctx context.Context) error {
```
plar, line 599: I was wondering if extending the function's return value to include a `bool` indicating whether a range checkpoint was actually generated, alongside the existing `error`. This could enhance decision-making in caller functions. Thoughts?

```
func (e *Manager) MaybeGenerateRangeCheckpoint(ctx context.Context) (/*generated*/bool, error) {
...
```

plar (approved):

internal/epoch/epoch_manager.go
```diff
@@ -624,6 +643,24 @@ func (e *Manager) maybeGenerateNextRangeCheckpointAsync(ctx context.Context, cs
 	})
 }
 
+func getRangeToCompact(cs CurrentSnapshot, p Parameters) (low, high int, compactRange bool) {
+	latestSettled := cs.WriteEpoch - numUnsettledEpochs
```
redgoat650, line 647: (minor readability suggestion) I realize this is just a refactor of existing logic, but perhaps we can make a new helper
```go
(cs *CurrentSnapshot) getLastSettledEpochNumber() int {
   return cs.WriteEpoch - numUnsettledEpochs
}
```
and we can reuse that here, as well as in `cs.isSettledEpochNumber()`?
```go
func (cs *CurrentSnapshot) isSettledEpochNumber(epoch int) bool {
	return epoch <= cs.getLastSettledEpochNumber()
}
```
For me it's easier to interpret the significance of that subtraction operation if it's described by a helper function name 🤷

redgoat650 (approved):
Looks good to me

internal/epoch/epoch_manager.go
julio-lopez: Makes sense. It can be done in a separate PR.

internal/epoch/epoch_manager.go
julio-lopez: I am not (yet) sure it is needed as part of the actual functionality. The primary caller is "repo maintenance", and maintenance cares about whether or not there was an error in order to stop.

Or what use case did you have in mind?

Either way, we can iterate in a separate PR.

plar: I just don’t know where this function will be called, but I saw from the code that it either does something or doesn’t do it if the error is `nil`. I thought that this information might be useful, in any case it would be easier to ignore on the caller side than to change the function later.

julio-lopez: I think I understand what you are saying.

The `Maybe` prefixes are meant to indicate (at the call site) that the function may not perform the actual operation at this time, and that's "OK"; because either:
- the system is already in the desired state and no transition / change is needed; or 
- it is not necessary that the operation is performed at this time to continue execution.

In the current use case (in a following PR), the caller does not need to know whether or not the operation was executed to make a decision based on that. It DOES need to know if there was an error, because in that case it seems safer to abort.
If it becomes necessary (from the caller) to know whether or not the operation was performed, then yes, we'll add a return value as you suggest.

internal/epoch/epoch_manager.go
julio-lopez: - #4927

PR #3728, opened 2024-03-13
refactor(general): add epoch.Manager.MaybeCompactSingleEpoch

julio-lopez:
Subset and dependency of #3651
Depends on #3735

Ref:
- #3638
- #3639

### DIFF
--- a/internal/epoch/epoch_manager.go
+++ b/internal/epoch/epoch_manager.go
@@ -970,6 +970,57 @@ func (e *Manager) getCompleteIndexSetForCommittedState(ctx context.Context, cs C
 	return result, nil
 }
 
+// MaybeCompactSingleEpoch compacts the oldest epoch that is elegible for
+// compaction if there is one.
+func (e *Manager) MaybeCompactSingleEpoch(ctx context.Context) error {
+	cs, err := e.committedState(ctx, 0)
+	if err != nil {
+		return err
+	}
+
+	uncompacted, err := getUncompactedEpochRange(cs)
+	if err != nil {
+		return err
+	}
+
+	if uncompacted.isEmpty() {
+		e.log.Debug("empty uncompacted epoch range, not compacting")
+		return nil
+	}
+
+	compacted, err := getCompactedEpochRange(cs)
+	if err != nil {
+		return err
+	}
+
+	if uncompacted.lo < compacted.hi {
+		uncompacted.lo = compacted.hi // avoid compacting already compacted epoch
+	}
+
+	if uncompacted.isEmpty() || !cs.isSettledEpochNumber(uncompacted.lo) {
+		e.log.Debug("there are no uncompacted epochs eligible for compaction")
+
+		return nil
+	}
+
+	epochToCompact := uncompacted.lo
+
+	uncompactedBlobs, ok := cs.UncompactedEpochSets[epochToCompact]
+	if !ok {
+		e.log.Debugf("there are no uncompacted blobs for epoch %d, not compacting", epochToCompact)
+
+		return nil
+	}
+
+	e.log.Debugf("starting single-epoch compaction of %v")
+
+	if err := e.compact(ctx, blob.IDsFromMetadata(uncompactedBlobs), compactedEpochBlobPrefix(epochToCompact)); err != nil {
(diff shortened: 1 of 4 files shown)


internal/epoch/epoch_utils.go
```diff
@@ -168,3 +168,49 @@ func getUncompactedEpochRange(cs CurrentSnapshot) (closedIntRange, error) {
...
+	return closedIntRange{
+		lo: cs.LongestRangeCheckpointSets[0].MinEpoch,
+		hi: cs.LongestRangeCheckpointSets[rangeSetsLen-1].MaxEpoch,
+	}
+}
+
+func oldestUncompactedEpoch(cs CurrentSnapshot) (int, error) {
+	rangeCompacted := getRangeCompactedRange(cs)
+
+	var oldesUncompactedEpoch int
```
plar, line 186: typo: oldestUncompactedEpoch

plar (approved):
LGTM

internal/epoch/epoch_manager.go
```diff
@@ -970,6 +970,45 @@ func (e *Manager) getCompleteIndexSetForCommittedState(ctx context.Context, cs C
...
+
+	if !cs.isSettledEpochNumber(uncompacted) {
+		e.log.Debugw("there are no uncompacted epochs eligible for compaction", "oldestUncompactedEpoch", uncompacted)
+
+		return nil
+	}
+
+	uncompactedBlobs, ok := cs.UncompactedEpochSets[uncompacted]
+	if !ok {
+		// TODO: attempt to load blobs for this epoch
```
julio-lopez, line 994: argh, this is a stale left-over comment

internal/epoch/epoch_manager.go
```diff
@@ -970,6 +970,45 @@ func (e *Manager) getCompleteIndexSetForCommittedState(ctx context.Context, cs C
 	return result, nil
 }
 
+// MaybeCompactSingleEpoch compacts the oldest epoch that is elegible for
```
redgoat650, line 973:
```suggestion
// MaybeCompactSingleEpoch compacts the oldest epoch that is eligible for
```

internal/epoch/epoch_manager_test.go
```diff
@@ -800,6 +800,153 @@ func TestIndexEpochManager_RefreshContextCanceled(t *testing.T) {
...
+func TestMaybeCompactSingleEpoch_CompactionError(t *testing.T) {
+	t.Parallel()
+
+	te := newTestEnv(t)
+	ctx := testlogging.Context(t)
+
+	p, err := te.mgr.getParameters(ctx)
+	require.NoError(t, err)
+
+	// TODO: epochs, so there is an eligible epoch for compaction
```
redgoat650, line 839: Is this comment missing some words before "epochs,"? Seems like an incomplete thought

internal/epoch/epoch_manager.go
```diff
@@ -970,6 +970,45 @@ func (e *Manager) getCompleteIndexSetForCommittedState(ctx context.Context, cs C
...
+	if err != nil {
+		return err
+	}
+
+	uncompacted, err := oldestUncompactedEpoch(cs)
+	if err != nil {
+		return err
+	}
+
+	if !cs.isSettledEpochNumber(uncompacted) {
```
redgoat650, line 986: So if I'm not mistaken, this has to do with the ordering of compaction vs epoch advancement we discussed, right? I.e. if we only advance epoch after successful compaction we only ever expect compaction-eligible epochs to be within the unsettled range.

redgoat650 (approved):
Some comment nits and a question, but looks good

internal/epoch/epoch_manager.go
julio-lopez: @redgoat650 Yes, it's related. See the check right below.

In short, the current snapshot `cs` only loads the last 3 uncompacted epochs, that is, the current write epoch, and the previous 2. So, there may be some uncompacted epochs for which the blob metadata is not included in `cs`. The block below accounts for that and explicitly loads the list of the blobs for the "missing epochs".

internal/epoch/epoch_manager_test.go
julio-lopez: It's a stale comment, it needs to be removed. I'll remove it in a minute.

PR #3729, opened 2024-03-13
use different mac secret lenght

chaitalisg:


Shrekster:
@bathina2 is updating the min check in key-derivation logic in a separate PR. We'll keep the secret length as is until we are sure this won't break compatibility. Thanks Chaitali.

PR #3735, opened 2024-03-15
refactor(general): closedIntRange

julio-lopez:
Refactoring for the original implementation with `intRange` and `getKeyRange` from _closed-open_ ranges `[lo, hi)` to closed ranges: `[lo, hi]`. The primary motivation is for consistency with the implementation of `epoch.RangeMetadata` in the same package, and thus  avoid confusion and reduce cognitive load.

Changes:
- adds a `getContiguousKeyRange` wrapper that checks for contiguity.
- `getKeyRange` simply returns a range with minimum and maximum values for the keys in the map.
- changes the range implementation from _closed-open_ ranges `[lo, hi)` to closed ranges: `[lo, hi]` where both lo and hi are included in the range. Additional unit tests are included.
- renames `intRange` to `closedIntRange` to reflect new functionality.

Ref:
- Follow up #3721
- Needed for #3728
- #3638

### DIFF
--- a/internal/epoch/epoch_utils.go
+++ b/internal/epoch/epoch_utils.go
@@ -121,10 +121,11 @@ const (
 // A range of the form [v,v) means the range is empty.
 // When the range is not continuous an error is returned.
 func getKeyRange[E any](m map[int]E) (intRange, error) {
-	var count uint
+	if len(m) == 0 {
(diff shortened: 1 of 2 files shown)


redgoat650 (approved):
LGTM

PR #3761, opened 2024-04-01
refactor(deps): migrate from `go.uber.org/multierr` to standard `errors`

mmorel-35:
Since go1.20, standard errors package provides [Join function](https://pkg.go.dev/errors#Join) that can replace `go.uber.org/multierr`.

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/internal/providervalidation/providervalidation.go
+++ b/internal/providervalidation/providervalidation.go
@@ -5,7 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
-	std_errors "errors"
+	stderrors "errors"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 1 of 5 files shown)


jkowalski:
This is looking good, thanks for your contribution. Come join slack.kopia.io and say hi to other developers.

internal/providervalidation/providervalidation.go
```diff
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
+	std_errors "errors"
```
julio-lopez, line 8: use an alias without underscore (`_`), maybe `stderrors` for now?

julio-lopez (approved):

PR #3770, opened 2024-04-03
feat(general): User_profile_add_set cli changes

bathina2:
This PR allows users to set the key derivation algorithm in a profile. 
The previous control `user-password-hash-version` was in place to handle any variations, however it has only ever supported 1 version and is not quite accurate. It has been replace with `key-derivation-algorithm`.

There is support to handle older profiles that have the PasswordHashVersion set. It will default to using the `Scrypt` key derivation algorithm as it had in the past.

An example where a profile is created with the old changes and one that is created with these new ones-
```
./kopianew2 server user list --json
[
 {"username":"tes2@test","passwordHashVersion":1,"passwordHash":"CE/3mjeFF6PnFYfXw7H6Lxq6MWG0TKLzKZ1C2cH5YIQSnVtL9gP8KpHmlK88BiA0qhXKUwOvPHgiumCDQUFmTw=="},
 {"username":"test3@test","keyDerivationAlgorithm":"scrypt-65536-8-1","passwordHash":"ksfKeB2hg4XlTzOxqvsU7mog9F/7x1Ca5Huj5KnGN4kzcV/Fj9ws0PJGtYZ254scDerHuoAJR6+IMbymK6p5Gg=="},
 
]            
```

### DIFF
--- a/cli/command_user_add_set.go
+++ b/cli/command_user_add_set.go
@@ -13,11 +13,11 @@ import (
 )
 
 type commandServerUserAddSet struct {
-	userAskPassword            bool
-	userSetName                string
-	userSetPassword            string
-	userSetPasswordHashVersion int
-	userSetPasswordHash        string
+	userAskPassword        bool
+	userSetName            string
+	userSetPassword        string
+	keyDerivationAlgorithm string
+	userSetPasswordHash    string
 
 	isNew bool // true == 'add', false == 'update'
 	out   textOutput
@@ -37,7 +37,7 @@ func (c *commandServerUserAddSet) setup(svc appServices, parent commandParent, i
 	cmd.Flag("ask-password", "Ask for user password").BoolVar(&c.userAskPassword)
 	cmd.Flag("user-password", "Password").StringVar(&c.userSetPassword)
 	cmd.Flag("user-password-hash", "Password hash").StringVar(&c.userSetPasswordHash)
-	cmd.Flag("user-password-hash-version", "Password hash version").Default("1").IntVar(&c.userSetPasswordHashVersion)
+	cmd.Flag("key-derivation-algorithm", "Key derivation algorithm").Default(crypto.DefaultKeyDerivationAlgorithm).EnumVar(&c.keyDerivationAlgorithm, crypto.AllowedKeyDerivationAlgorithms()...)
 	cmd.Arg("username", "Username").Required().StringVar(&c.userSetName)
 	cmd.Action(svc.repositoryWriterAction(c.runServerUserAddSet))
 
@@ -54,7 +54,8 @@ func (c *commandServerUserAddSet) getExistingOrNewUserProfile(ctx context.Contex
 
 		case errors.Is(err, user.ErrUserNotFound):
 			return &user.Profile{
-				Username: username,
+				Username:               username,
+				KeyDerivationAlgorithm: c.keyDerivationAlgorithm,
 			}, nil
 		}
 	}
@@ -75,7 +76,7 @@ func (c *commandServerUserAddSet) runServerUserAddSet(ctx context.Context, rep r
 	if p := c.userSetPassword; p != "" {
 		changed = true
 
-		if err := up.SetPassword(p, crypto.DefaultKeyDerivationAlgorithm); err != nil {
+		if err := up.SetPassword(p); err != nil {
 			return errors.Wrap(err, "error setting password")
 		}
 	}
@@ -86,7 +87,6 @@ func (c *commandServerUserAddSet) runServerUserAddSet(ctx context.Context, rep r
 			return errors.Wrap(err, "invalid password hash, must be valid base64 string")
 		}
 
-		up.PasswordHashVersion = c.userSetPasswordHashVersion
 		up.PasswordHash = ph
 		changed = true
 	}
@@ -108,7 +108,7 @@ func (c *commandServerUserAddSet) runServerUserAddSet(ctx context.Context, rep r
 
 		changed = true
 
-		if err := up.SetPassword(pwd, crypto.DefaultKeyDerivationAlgorithm); err != nil {
+		if err := up.SetPassword(pwd); err != nil {
 			return errors.Wrap(err, "error setting password")
 		}
 	}
--- a/internal/auth/authn_repo_test.go
+++ b/internal/auth/authn_repo_test.go
@@ -43,6 +43,13 @@ func TestRepositoryAuthenticator(t *testing.T) {
 					},
 					password: "password3",
 				},
+				{
+					profile: &user.Profile{
+						Username:               "user4@host4",
+						KeyDerivationAlgorithm: crypto.Pbkdf2Algorithm,
+					},
+					password: "password4",
+				},
 			} {
 				tc.profile.SetPassword(tc.password)
 				err := user.SetUserProfile(ctx, w, tc.profile)
@@ -64,6 +71,9 @@ func TestRepositoryAuthenticator(t *testing.T) {
 
 	// Test for User with neither key derivation or PasswordHashVersion set
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user3@host3", "password3", false)
+
+	// Test for PBKDF2 key derivation
+	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user4@host4", "password4", true)
 }
 
 func verifyRepoAuthenticator(ctx context.Context, t *testing.T, a auth.Authenticator, r repo.Repository, username, password string, want bool) {
--- a/internal/auth/authn_repo_test.go
+++ b/internal/auth/authn_repo_test.go
@@ -7,7 +7,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/kopia/kopia/internal/auth"
-	"github.com/kopia/kopia/internal/crypto"
 	"github.com/kopia/kopia/internal/repotesting"
 	"github.com/kopia/kopia/internal/user"
 	"github.com/kopia/kopia/repo"
@@ -23,7 +22,7 @@ func TestRepositoryAuthenticator(t *testing.T) {
 				Username: "user1@host1",
 			}
 
-			p.SetPassword("password1", crypto.DefaultKeyDerivationAlgorithm)
+			p.SetPassword("password1")
 
 			return user.SetUserProfile(ctx, w, p)
 		}))
--- a/internal/auth/authn_repo_test.go
+++ b/internal/auth/authn_repo_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/kopia/kopia/internal/auth"
+	"github.com/kopia/kopia/internal/crypto"
 	"github.com/kopia/kopia/internal/repotesting"
 	"github.com/kopia/kopia/internal/user"
 	"github.com/kopia/kopia/repo"
@@ -18,13 +19,38 @@ func TestRepositoryAuthenticator(t *testing.T) {
 
 	require.NoError(t, repo.WriteSession(ctx, env.Repository, repo.WriteSessionOptions{},
 		func(ctx context.Context, w repo.RepositoryWriter) error {
-			p := &user.Profile{
-				Username: "user1@host1",
+			for _, tc := range []struct {
+				profile  *user.Profile
+				password string
+			}{
+				{
+					profile: &user.Profile{
+						Username:            "user1@host1",
+						PasswordHashVersion: crypto.HashVersion1,
+					},
+					password: "password1",
+				},
+				{
+					profile: &user.Profile{
+						Username:               "user2@host2",
+						KeyDerivationAlgorithm: crypto.ScryptAlgorithm,
+					},
+					password: "password2",
+				},
+				{
+					profile: &user.Profile{
+						Username: "user3@host3",
+					},
+					password: "password3",
+				},
+			} {
+				tc.profile.SetPassword(tc.password)
+				err := user.SetUserProfile(ctx, w, tc.profile)
+				if err != nil {
+					return err
+				}
 			}
-
-			p.SetPassword("password1")
-
-			return user.SetUserProfile(ctx, w, p)
+			return nil
 		}))
 
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1", "password1", true)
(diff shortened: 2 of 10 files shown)


internal/auth/authn_repo_test.go
```diff
@@ -19,20 +19,51 @@ func TestRepositoryAuthenticator(t *testing.T) {
...
+					profile: &user.Profile{
+						Username:            "user1@host1",
+						PasswordHashVersion: crypto.HashVersion1,
+					},
+					password: "password1",
+				},
+				{
+					profile: &user.Profile{
+						Username:               "user2@host2",
+						KeyDerivationAlgorithm: crypto.ScryptAlgorithm,
```
Shrekster, line 36: you'd be adding the `Pbkdf2Algorithm` test case here as well ?

internal/auth/authn_repo_test.go
```diff
@@ -19,20 +19,51 @@ func TestRepositoryAuthenticator(t *testing.T) {
...
+					},
+					password: "password2",
+				},
+				{
+					profile: &user.Profile{
+						Username: "user3@host3",
+					},
+					password: "password3",
+				},
```
Shrekster, line 45: Lets also add the new algorithm in the test cases here:

```suggestion
				},
				{
					profile: &user.Profile{
						Username:               "user4@host4",
						KeyDerivationAlgorithm: crypto.Pbkdf2Algorithm,
					},
					password: "password4",
				},
```

internal/auth/authn_repo_test.go
```diff
@@ -19,20 +19,51 @@ func TestRepositoryAuthenticator(t *testing.T) {
...
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1", "password2", false)
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1", "password11", false)
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1a", "password1", false)
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1a", "password1a", false)
+
+	// Test for password with KeyDerivationSet
+	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user2@host2", "password2", true)
+
+	// Test for User with neither key derivation or PasswordHashVersion set
+	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user3@host3", "password3", false)
```
Shrekster, line 66: Don't we want a `true` here instead of `false` ? The default algorithm should be scrypt or has that changed ?

internal/user/user_profile.go
```diff
@@ -1,38 +1,41 @@
...
 }
 
 // IsValidPassword determines whether the password is valid for a given user.
-func (p *Profile) IsValidPassword(password, keyDerivationAlgorithm string) bool {
+func (p *Profile) IsValidPassword(password string) bool {
 	if p == nil {
 		// if the user is invalid, return false but use the same amount of time as when we
 		// compare against valid user to avoid revealing whether the user account exists.
-		isValidPasswordV1(password, dummyV1HashThatNeverMatchesAnyPassword, keyDerivationAlgorithm)
+		isValidPassword(password, dummyV1HashThatNeverMatchesAnyPassword, crypto.DefaultKeyDerivationAlgorithm)
```
Shrekster, line 28: Well we are kind of out of options here... there is going to be some change in processing time because for an attacker "default algorithm" time means a 50% chance of being a valid user and non-default would mean a valid user ..., but better than no processing.

Ideally, the algorithm for the `nil` case should be picked up randomly to protect against timing attacks, and then you'll need to make sure that randomization is done each time (even if `p != nil`).

Shrekster: WDYT ?

internal/user/user_profile.go
```diff
@@ -1,38 +1,41 @@
...
 	}
-
-	switch p.PasswordHashVersion {
-	case hashVersion1:
-		return isValidPasswordV1(password, p.PasswordHash, keyDerivationAlgorithm)
-
-	default:
-		return false
+	// Legacy case where password hash version is set
+	if p.PasswordHashVersion == crypto.HashVersion1 {
```
Shrekster, line 33: nit; since you are at at, probably should use `subtle.ConstantTimeCompare()` for the timing attack reasons as above

internal/auth/authn_repo_test.go
Shrekster: nvm, this is the case where neither `PasswordHashVersion` and nor `KeyDerivationAlgorithm` are set

this makes sense now

internal/auth/authn_repo_test.go
```diff
@@ -19,20 +19,51 @@ func TestRepositoryAuthenticator(t *testing.T) {
...
 		}))
 
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1", "password1", true)
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1", "password2", false)
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1", "password11", false)
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1a", "password1", false)
 	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user1@host1a", "password1a", false)
+
+	// Test for password with KeyDerivationSet
+	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user2@host2", "password2", true)
```
Shrekster, line 63:
```suggestion
	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user2@host2", "password2", true)
	verifyRepoAuthenticator(ctx, t, a, env.Repository, "user4@host4", "password4", true)
```

internal/crypto/key_derivation_nontest.go
```diff
@@ -4,38 +4,59 @@
...
 //nolint:gochecknoglobals
-var keyDerivers = map[string]keyDerivationFunc{}
+var keyDerivers = map[string]KeyDeriver{}
 
-// RegisterKeyDerivationFunc registers various key derivation functions.
-func RegisterKeyDerivationFunc(name string, keyDeriver keyDerivationFunc) {
+// RegisterKeyDerivers registers various key derivation functions.
+func RegisterKeyDerivers(name string, keyDeriver KeyDeriver) {
+	if _, ok := keyDerivers[name]; ok {
+		panic(fmt.Sprintf("key (%s) deriver already register", name))
```
Shrekster, line 33:
```suggestion
		panic(fmt.Sprintf("key (%s) deriver already registered", name))
```

internal/crypto/key_derivation_nontest.go
```diff
@@ -4,38 +4,59 @@
...
 
 // DefaultKeyDerivationAlgorithm is the key derivation algorithm for new configurations.
 const DefaultKeyDerivationAlgorithm = ScryptAlgorithm
 
+type KeyDeriver interface {
+	DeriveKeyFromPassword(password string, salt []byte) ([]byte, error)
+	RecommendedSaltLength() int
+}
+
 type keyDerivationFunc func(password string, salt []byte) ([]byte, error)
```
Shrekster, line 25: we can drop this type now in favor of the interface, correct ?

```suggestion

```

internal/crypto/scrypt.go
```diff
@@ -16,15 +16,49 @@ import (
...
 	})
 }
+
+type scryptKeyDeriver struct {
+	// n scryptCostParameterN is scrypt's CPU/memory cost parameter.
+	n int
+	// r scryptCostParameterR is scrypt's work factor.
+	r int
+	// p scryptCostParameterP is scrypt's parallelization parameter.
+	p int
```
Shrekster, line 49: Thank you for adding these comments!

internal/user/user_profile_hash_v1.go
```diff
@@ -10,31 +10,35 @@ import (
...
+var dummyV1HashThatNeverMatchesAnyPassword = make([]byte, crypto.MasterKeyLength+crypto.V1SaltLength)
+
+func (p *Profile) setPassword(password string) error {
+	if len(p.KeyDerivationAlgorithm) == 0 {
+		if p.PasswordHashVersion == 0 {
+			return errors.New("key derivation algorithm and password hash version not set")
+		}
+		// Setup to handle legacy hashVersion.
+		if p.PasswordHashVersion == crypto.HashVersion1 {
+			p.KeyDerivationAlgorithm = crypto.ScryptAlgorithm
```
Shrekster, line 23: So, are you trying to use this as migration away from the old `PasswordHashVersion` ? (set new field and hope the caller saves the new profile)

I think the best way to do this migration is through an explicit "format upgrade" track / a command or something that customers have to run on their repository to move away from deprecated fields, otherwise we'll never be "sure" if this has already been done or not because we do not know whether everyone has reset their passwords using the new profile fields.

If that is the case, then do you think `p.KeyDerivationAlgorithm` should be set here ? If it SHOULD be set then why not also unset `p.PasswordHashVersion` ?

internal/user/user_profile_test.go
```diff
@@ -7,38 +7,82 @@ import (
...
+func TestLegacyUserProfile(t *testing.T) {
+	p := &user.Profile{
+		PasswordHashVersion: 1, // hashVersion1
+	}
+
+	if p.IsValidPassword("bar") {
+		t.Fatalf("password unexpectedly valid!")
+	}
+
+	p.SetPassword("foo")
```
Shrekster, line 19: nit; IFF you go down the route of changing the user-profile algorithm settings after SetPssword() then lets also assert that `p.PasswordHashVersion` is cleared and `KeyDerivationAlgorithm` is set correctly.

internal/user/user_profile_test.go
```diff
@@ -7,38 +7,82 @@ import (
...
+	}
+
+	// Setting the key derivation to scrypt and unsetting PasswordHashVersion
+	// Legacy profile should translate to scrypt
+	p.KeyDerivationAlgorithm = crypto.ScryptAlgorithm
+	p.PasswordHashVersion = 0
+	if !p.IsValidPassword("foo") {
+		t.Fatalf("password not valid!")
+	}
+
```
Shrekster, line 36:
```suggestion
```

internal/user/user_profile_test.go
```diff
@@ -7,38 +7,82 @@ import (
...
+	if p.IsValidPassword("bar") {
 		t.Fatalf("password unexpectedly valid!")
 	}
+
+	// Different key derivation algorithm besides the original should fail
+	p.KeyDerivationAlgorithm = crypto.Pbkdf2Algorithm
+	if p.IsValidPassword("foo") {
+		t.Fatalf("password unexpectedly valid!")
+	}
+
```
Shrekster, line 63:
```suggestion
```

Shrekster:
@bathina2 Sirish, changes look very good. Took first pass and left some comments, please take look. This is looking very close to done.

internal/user/user_profile.go
bathina2: Its comparing integers, not sure subtle time compare would work here.

Shrekster: yes, would be overkill if this is integers

internal/user/user_profile.go
```diff
@@ -23,10 +25,10 @@ func (p *Profile) SetPassword(password string) error {
 // IsValidPassword determines whether the password is valid for a given user.
 func (p *Profile) IsValidPassword(password string) bool {
 	if p == nil {
+		algorithms := crypto.AllowedKeyDerivationAlgorithms()
 		// if the user is invalid, return false but use the same amount of time as when we
 		// compare against valid user to avoid revealing whether the user account exists.
-		isValidPassword(password, dummyV1HashThatNeverMatchesAnyPassword, crypto.DefaultKeyDerivationAlgorithm)
-
+		isValidPassword(password, dummyV1HashThatNeverMatchesAnyPassword, algorithms[rand.Intn(len(algorithms))])
```
Shrekster, line 31: Thank you!

internal/user/user_profile.go
```diff
@@ -1,38 +1,42 @@
...
 // Profile describes information about a single user.
 type Profile struct {
 	ManifestID manifest.ID `json:"-"`
 
-	Username            string `json:"username"`
-	PasswordHashVersion int    `json:"passwordHashVersion"` // indicates how password is hashed
-	PasswordHash        []byte `json:"passwordHash"`
+	Username               string `json:"username"`
+	PasswordHashVersion    int    `json:"passwordHashVersion,omitempty"` // indicates how password is hashed, deprecated in favor of KeyDerivationAlgorithm
+	KeyDerivationAlgorithm string `json:"keyDerivationAlgorithm,omitempty"`
```
Shrekster, line 16: @julio-lopez @jkowalski could you please take a quick look if deprecating a field in favor of a better type and name in user profile struct is ok here ?

Shrekster (approved):
LGTM 👍 

Merging this. @julio-lopez / @jkowalski PTAL at my comment regarding user profiles post merge as well.

cli/command_user_add_set.go
```diff
@@ -37,7 +37,7 @@ func (c *commandServerUserAddSet) setup(svc appServices, parent commandParent, i
 	cmd.Flag("ask-password", "Ask for user password").BoolVar(&c.userAskPassword)
 	cmd.Flag("user-password", "Password").StringVar(&c.userSetPassword)
 	cmd.Flag("user-password-hash", "Password hash").StringVar(&c.userSetPasswordHash)
-	cmd.Flag("user-password-hash-version", "Password hash version").Default("1").IntVar(&c.userSetPasswordHashVersion)
+	cmd.Flag("key-derivation-algorithm", "Key derivation algorithm").Default(crypto.DefaultKeyDerivationAlgorithm).EnumVar(&c.keyDerivationAlgorithm, crypto.AllowedKeyDerivationAlgorithms()...)
```
bathina2, line 40: `user-password-hashing-algorithms`

julio-lopez, line 40: `user-password-hashing-algorithm`

julio-lopez: - Add `.Hidden()`
- For the description => `"[EXPERIMENTAL] Password hashing algorithm"`

internal/user/user_profile.go
julio-lopez: ADR:
- Reuse `PasswordHashVersion` and introduce a new version (`2`) for `pbkdf2-....-`
- Remove `KeyDerivationAlgorithm`

And then have a mapping from
- the string algo type to the version int to translate from the CLI to the struct
- the `PasswordHashVersion` int to the corresponding string for the algo, so its implementation can be looked up in the registered derivation implementations.

julio-lopez: Addressed in #3816

PR #3773, opened 2024-04-04
feat(cli): added decompression benchmark

jkowalski:


cli/command_benchmark_compression.go
```diff
@@ -155,11 +212,86 @@ func (c *commandBenchmarkCompression) run(ctx context.Context) error {
 			return compressedSize
 		}
 
+		outputBuffers := makeOutputBuffers(c.parallel, defaultCompressedDataByMethod)
```
julio-lopez, line 215: What prevents GC from running while benchmarking and affecting the results? Is GC turned off? would the last method to be benchmarked be likely to experience more memory pressure?

cli/command_benchmark_compression.go
```diff
@@ -106,13 +128,49 @@ func (c *commandBenchmarkCompression) run(ctx context.Context) error {
...
+		}
+	}
+
+	return nil
+}
+
+func (c *commandBenchmarkCompression) runCompression(ctx context.Context, data []byte, repeatCount int, algorithms map[compression.Name]compression.Compressor) error {
+	var results []compressionBechmarkResult
+
+	log(ctx).Infof("Compressing input file %q (%v) using all compression methods.", c.dataFile, units.BytesString(int64(len(data))))
```
julio-lopez, line 171: is it still the case that _all_ compression methods are used here?

cli/command_benchmark_compression.go
```diff
@@ -82,8 +88,26 @@ type compressionBechmarkResult struct {
...
+	if c.algorithms == "" {
+		if compression.IsDeprecated[name] && !c.deprecated {
+			return false
+		}
+
+		return true
+	}
+
+	for _, a := range strings.Split(c.algorithms, ",") {
+		if strings.HasPrefix(string(name), a) {
```
julio-lopez, line 101: for my understanding: why `HasPrefix`?

cli/command_benchmark_compression.go
```diff
@@ -106,13 +128,49 @@ func (c *commandBenchmarkCompression) run(ctx context.Context) error {
...
+	}
+
+	if benchmarkCompression {
+		if err := c.runCompression(ctx, data, repeatCount, algorithms); err != nil {
+			return err
 		}
+	}
+
+	if benchmarkDecompression {
+		if err := c.runDecompression(ctx, data, repeatCount, algorithms); err != nil {
```
julio-lopez, line 160: How does this work when multiple algorithms are specified and only decompression is benchmarked? essentially, what's the input for multiple decompressors when there is no prior compression.

julio-lopez (commented):
+1: LG 👍

A few comments and questions.
Feel free to merge, no need for an additional round of review here.

Thanks.

cli/command_benchmark_compression.go
jkowalski: so you can test all "zstd" in one shot

cli/command_benchmark_compression.go
jkowalski: we compress once per compressor, then decompress N times.

cli/command_benchmark_compression.go
jkowalski: we actually call `runtime.GC()` before each algorithm.

cli/command_benchmark.go
```diff
@@ -30,30 +31,58 @@ type cryptoBenchResult struct {
...
 			defer wg.Done()
 
-			run()
+			run(arg)
 		}()
 	}
 
 	// run one on the main goroutine and N-1 in parallel.
-	v := run()
+	v := run(args[len(args)-1])
```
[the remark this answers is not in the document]
jkowalski: premature optimization :)

cli/command_benchmark_compression.go
julio-lopez: I see. I missed that. Thanks.

cli/command_benchmark_compression.go
jkowalski: replaced "all" with a number.

julio-lopez (approved):
+1: LG 👍

PR #3797, opened 2024-04-11
refactor(general): avoid wrapping known-to-be-nil error

julio-lopez:
Ref:
- #1980


jkowalski:
FWIW wrapping nil is a no-op, and we use unconditional wrapping on possible errors in several other places.

jkowalski (approved):

julio-lopez:
> FWIW wrapping nil is a no-op, and we use unconditional wrapping on possible errors in several other places.

Understood. It may make sense to look for those, I'll put in the back of the queue.

In this case, is primarily to simplify the code, make it explicit that a nil error is returned, so it is easier to understand for a casual reader.

PR #3799, opened 2024-04-13
feat(general): key derivation algorithm for cache encryption

bathina2:
This PR provides the password key derivation algorithm  option when connecting to a kopia server.
When connecting to a kopia server the client uses the password to derive the cache encryption key.


repo/content/caching_options.go
```diff
@@ -29,6 +29,7 @@ type CachingOptions struct {
 	MinContentSweepAge          DurationSeconds `json:"minContentSweepAge,omitempty"`
 	MinIndexSweepAge            DurationSeconds `json:"minIndexSweepAge,omitempty"`
 	HMACSecret                  []byte          `json:"-"`
+	KeyDerivationAlgorithm      string          `json:"keyDerivationAlgorithm,omitempty"`
```
julio-lopez, line 32: This option is confusing. It conveys that the key derivation algorithm would apply when deriving the cache encryption key for all the different caches (use cases). However this is used **only** in a very specific scenario, that is for client caches when connecting to a kopia repository server (that is, via the repository server API).

julio-lopez (changes requested):

cli/command_repository_connect.go
```diff
@@ -79,6 +81,7 @@ func (c *connectOptions) setup(svc appServices, cmd *kingpin.CmdClause) {
 	cmd.Flag("enable-actions", "Allow snapshot actions").BoolVar(&c.connectEnableActions)
 	cmd.Flag("repository-format-cache-duration", "Duration of kopia.repository format blob cache").Hidden().DurationVar(&c.formatBlobCacheDuration)
 	cmd.Flag("disable-repository-format-cache", "Disable caching of kopia.repository format blob").Hidden().BoolVar(&c.disableFormatBlobCache)
+	cmd.Flag("cache-key-derivation-algorithm", "Cache key derivation algorithm").Default(crypto.DefaultKeyDerivationAlgorithm).EnumVar(&c.connectCacheKeyDerivationAlgorithm, crypto.AllowedKeyDerivationAlgorithms()...)
```
julio-lopez, line 84: This flag is specific to clients that connect to the repository via the kopia (API) server.

- Move this flag be moved to the `commandRepositoryConnectServer` struct in `command_repository_connect_server.go`?
- The flag name should also reflect that is specific for this connection type, let's use a more specific and descriptive flag name.
- Update and expand the flag description to clearly describe that this flag is for deriving the (content) cache encryption key for clients of the repository server.
- Add `.Hidden()` for this flag and include "[EXPERIMENTAL]" in the description.
- Is additional validation needed? or does `EnumVar` take care of the validation?

repo/content/caching_options.go
julio-lopez: @bathina2 What would be a good name for this field?

repo/caching.go
```diff
@@ -75,6 +76,11 @@ func setupCachingOptionsWithDefaults(ctx context.Context, configPath string, lc
 	lc.Caching.MinContentSweepAge = opt.MinContentSweepAge
 	lc.Caching.MinMetadataSweepAge = opt.MinMetadataSweepAge
 	lc.Caching.MinIndexSweepAge = opt.MinIndexSweepAge
+	lc.Caching.KeyDerivationAlgorithm = opt.KeyDerivationAlgorithm
+
+	if lc.Caching.KeyDerivationAlgorithm == "" {
+		lc.Caching.KeyDerivationAlgorithm = crypto.DefaultKeyDerivationAlgorithm
```
julio-lopez, line 82: - Add an appropriately named constant for the default algo in the `repo` package.
- Use the constant here and in the CLI flag.

julio-lopez (changes requested):

julio-lopez:
FYI @miquella

repo/content/caching_options.go
julio-lopez: I'm also wondering if this struct is the right place for this field. Thoughts?

julio-lopez (approved):
+1: LG 👍
🥇 @bathina2 Thanks for doing this.

We'll merge once CI passes

julio-lopez:
- #4255

PR #3810, opened 2024-04-19
refactor(general): change default number of iterations for PBKDF2

julio-lopez:
Use the value recommended by NIST

PR #3811, opened 2024-04-19
refactor(general):  include parameters in pbkdf2 instantiation

julio-lopez:
Use `pbkdf2-sha256-600000` as the name to be consistent with the scrypt instantiation.

The format is `pbkdf2-<hash_type>-<number_of_iterations>`

### DIFF
--- a/internal/crypto/pbkdf.go
+++ b/internal/crypto/pbkdf.go
@@ -23,7 +23,7 @@ const (
 	pbkdf2Sha256Iterations = 600_000
 
 	// Pbkdf2Algorithm is the key for the pbkdf algorithm.
-	Pbkdf2Algorithm = "pbkdf2"
+	Pbkdf2Algorithm = "pbkdf2-sha256-600000"
 )
 
 func init() {


julio-lopez:
FYI: @bathina2 @miquella @kidgilson

internal/crypto/pbkdf.go
```diff
@@ -20,10 +20,10 @@ const (
 	minPbkdfSha256SaltSize = 32 // size in bytes == 128 bits
 
 	// The NIST recommended iterations for PBKDF2 with SHA256 hash is 600,000.
-	pbkdf2Sha256Iterations = 1<<20 - 1<<18 // 786,432
+	pbkdf2Sha256Iterations = 600_000
```
kidgilson, line 23: Since this is coming from PR https://github.com/kopia/kopia/pull/3810 should this be here?

julio-lopez: It's coming from #3810 
The target branch should have been that one #3810 to show the actual diff. Now that #3810 has been merged and this one is updated, then that change will go away.

PR #3815, opened 2024-04-24
fix(cli): user add set user-password-hashing-algorithm cli option

bathina2:
This PR updates the `user-password-hashing-algorithm` option to be hidden and note that it is an experimental setting.


julio-lopez (approved):

julio-lopez:
FYI: @miquella

PR #3816, opened 2024-04-24
feat(general): User profile hashing version to algorithm translation

bathina2:
This PR reverts the behavior of storing the key derivation algorithm in the profile. 
We will continue to use the `PasswordHashVersion` and have internal tooling to translate to the key derivation algorithm (password hashing algorithm).


internal/crypto/key_derivation_testing.go
```diff
@@ -12,14 +12,18 @@ import (
...
 	MasterKeyLength = 32
 
-	V1SaltLength    = 32
-	HashVersion1    = 1 // this translates to Scrypt KeyDerivationAlgorithm
-	ScryptAlgorithm = "scrypt-65536-8-1"
-	Pbkdf2Algorithm = "pbkdf2"
+	V1SaltLength = 32
+
+	ScryptAlgorithm   = "scrypt-65536-8-1"
+	ScryptHashVersion = 1
```
julio-lopez, line 23: `ScryptHashVersion` & `Pbkdf2HashVersion` below should be defined in the `internal/user` package.

julio-lopez (commented):
The high level comment is that the password hashing versions, and thus the corresponding mapping to their string equivalents, are part of the user profile management, and belong in the `internal/user` package.

internal/crypto/key_derivation_testing.go
```diff
@@ -10,24 +10,25 @@ import (
...
-	HashVersion1    = 1 // this translates to Scrypt KeyDerivationAlgorithm
+	V1SaltLength = 32
+
 	ScryptAlgorithm = "scrypt-65536-8-1"
-	Pbkdf2Algorithm = "pbkdf2"
+
+	Pbkdf2Algorithm = "pbkdf2-sha256-600000"
+
+	// DefaultKeyDerivationAlgorithm is the key derivation algorithm for new configurations.
+	DefaultKeyDerivationAlgorithm = ScryptAlgorithm
```
julio-lopez, line 23: nit: it seems that the default should be "testing-only-insecure"

internal/user/user_profile_test.go
```diff
@@ -26,57 +25,27 @@ func TestLegacyUserProfile(t *testing.T) {
...
-	}
-
-	if p.IsValidPassword("bar") {
-		t.Fatalf("password unexpectedly valid!")
-	}
-
-	// Different key derivation algorithm besides the original should fail
-	p.KeyDerivationAlgorithm = crypto.Pbkdf2Algorithm
+	// Assume the key derivation algorithm is bad. This will cause
+	// a panic when validating
```
julio-lopez, line 42: nit: Is the comment accurate? does it panic?

internal/user/user_profile.go
```diff
@@ -24,23 +37,53 @@ func (p *Profile) SetPassword(password string) error {
...
+		passwordHashAlgorithm, err = GetPasswordHashAlgorithm(p.PasswordHashVersion)
+		if err != nil {
+			invalidProfile = true
+		}
+	}
+
+	if invalidProfile {
 		algorithms := crypto.AllowedKeyDerivationAlgorithms()
-		// if the user is invalid, return false but use the same amount of time as when we
+		// if the Username is invalid, return false but use the same amount of time as when we
```
julio-lopez, line 57: nit: 
```suggestion
		// if the user profile is invalid (either invalid user name or password hash version, return false but use the same amount of time as when we
```

julio-lopez (approved):
🥇Thanks @bathina2 for doing this.

LG. Couple of minor nits. Those can be addressed separately after merging.

Thanks again.

internal/crypto/key_derivation_testing.go
bathina2: This breaks tests, since we have separated the hash versions from the key derivation algorithms.
I would have to handle `testing-only-insecure` in the `internal/user` package but that would mean introducing testing code in production

julio-lopez: OK

PR #3779, opened 2024-04-04
feat(general): Repo key derivation algorithm

bathina2:
This PR adds support to set the key derivation algorithm when using repositories. It is a replacement for https://github.com/kopia/kopia/pull/3731

PR #3821, opened 2024-04-26
refactor(general): user password hashing and key derivation helpers

julio-lopez:
Code movement and simplification, no functional changes.

Objectives:
- Allow callers specifying the needed key (or hash) size, instead of hard-coding it in the registered PBK derivers. Conceptually, the caller needs to specify the key size, since that is a requirement of the (encryption) algorithm being used in the caller. Now, the code changes here do not result in any functional changes since the key size is always 32 bytes.
- Remove a global definition for the default PB key deriver to use. Instead, each of the 3 use case sets the default value.

Changes:
- `crypto.DeriveKeyFromPassword` now takes a key size.
- Adds new constants for the key sizes at the callers.
- Removes the global `crypto.MasterKeySize` const.
- Removes the global `crypto.DefaultKeyDerivationAlgorithm` const.
- Adds const for the default derivation algorithms for each use case.

Followups to:
- #3725
- #3770
- #3779
- #3799
- #3816

The individual commits show the code transformations to simplify the review of the changes.


internal/user/password_hashings_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/kopia/kopia/internal/crypto"
+)
+
+// The password hashing constants defined in this package are used as "lookup
+// keys" for the register password-based key derivers in the crypto package.
+// This trivial test is a change detector to ensure that the constants defined
+// in the user package match those defined in the crypto package.
+func TestPasswordHashingConstantMatchCryptoPackage(t *testing.T) {
```
julio-lopez, line 15: @bathina2 This is what I meant by _"adding a test to ensure the constants match"_.

julio-lopez:
@bathina2 @miquella FYI

PR #3832, opened 2024-04-30
fix(general): Check for valid key length before deriving key

bathina2:
This PR validates the key length when deriving a new key.


bathina2:
closing in favor of #3835

PR #3835, opened 2024-05-01
refactor(general): avoid using empty master keys

miquella:
Previously, empty master keys were passed to the underlying cryptographic primitives (HKDF, AEAD, etc.).

While this worked because the authentication mechanisms returned an error, it's best to avoid passing empty master keys to these primitives in the first place. This refactor avoids passing empty master keys and enforces this via an assertion in the key derivation function.

### DIFF
--- a/internal/crypto/key_derivation_test.go
+++ b/internal/crypto/key_derivation_test.go
@@ -4,8 +4,9 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/kopia/kopia/internal/crypto"
 	"github.com/stretchr/testify/require"
+
+	"github.com/kopia/kopia/internal/crypto"
 )
 
 var (
--- a/internal/crypto/key_derivation_test.go
+++ b/internal/crypto/key_derivation_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/kopia/kopia/internal/crypto"
+	"github.com/stretchr/testify/require"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


internal/repotesting/repotesting.go
```diff
@@ -68,7 +68,7 @@ func (e *Environment) setup(tb testing.TB, version format.Version, opts ...Optio
 			MutableParameters: format.MutableParameters{
 				Version: version,
 			},
-			HMACSecret:           []byte{},
+			HMACSecret:           []byte("test-hmac-secret"),
```
julio-lopez, line 71: nit: pad to 32 bytes and update the corresponding tests 🙏🏼 
Thanks.

internal/crypto/key_derivation_test.go
```diff
@@ -0,0 +1,46 @@
...
+		key := crypto.DeriveKeyFromMasterKey(TestMasterKey, TestSalt, TestPurpose, 32)
+
+		expected := "828769ee8969bc37f11dbaa32838f8db6c19daa6e3ae5f5eed2da2d94d8faddb"
+		got := fmt.Sprintf("%02x", key)
+		if got != expected {
+			t.Errorf("incorrect key\nexpected: %s\n     got: %s", expected, got)
+		}
+	})
+
+	t.Run("PanicsOnNilMasterKey", func(t *testing.T) {
```
julio-lopez, line 27: please use https://pkg.go.dev/github.com/stretchr/testify/require#Panics here and below

julio-lopez (commented):
couple of minor nits, LG otherwise

julio-lopez:
@bathina2 PTAL

internal/repotesting/repotesting.go
miquella: 👍 Updated.

internal/crypto/key_derivation_test.go
miquella: 👍 Updated.

internal/crypto/key_derivation_test.go
```diff
@@ -0,0 +1,37 @@
+package crypto_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/kopia/kopia/internal/crypto"
+	"github.com/stretchr/testify/require"
```
julio-lopez, line 8: this is going to fail linting.

julio-lopez, line 8:
```suggestion
	"github.com/stretchr/testify/require"

	"github.com/kopia/kopia/internal/crypto"
```

julio-lopez (approved):

Shrekster (commented):
LGTM, thank you!

PR #3839, opened 2024-05-02
test(server): Decompose base URL to URI to helper and add test

redgoat650:
Decompose the logic for parsing base URL into the URI for client dial.
Add a unit test for the new helper.

### DIFF
--- a/repo/grpc_repository_client.go
+++ b/repo/grpc_repository_client.go
@@ -829,7 +829,7 @@ func openGRPCAPIRepository(ctx context.Context, si *APIServerInfo, password stri
 		transportCreds = credentials.NewClientTLSFromCert(nil, "")
 	}
 
-	uri, err := BaseURLToURI(si.BaseURL)
+	uri, err := baseURLToURI(si.BaseURL)
 	if err != nil {
 		return nil, errors.Wrap(err, "parsing base URL")
 	}
@@ -860,9 +860,7 @@ func openGRPCAPIRepository(ctx context.Context, si *APIServerInfo, password stri
 	return rep, nil
 }
 
-// BaseURLToURI parses the provided base URL and returns the URI. The
-// input scheme must be either "https", "unix+https", or "kopia".
-func BaseURLToURI(baseURL string) (uri string, err error) {
+func baseURLToURI(baseURL string) (uri string, err error) {
 	u, err := url.Parse(baseURL)
 	if err != nil {
 		return "", errors.Wrap(err, "unable to parse server URL")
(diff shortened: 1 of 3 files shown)


repo/grpc_repository_client.go
```diff
@@ -869,6 +860,26 @@ func openGRPCAPIRepository(ctx context.Context, si *APIServerInfo, password stri
 	return rep, nil
 }
 
+// BaseURLToURI parses the provided base URL and returns the URI. The
+// input scheme must be either "https", "unix+https", or "kopia".
+func BaseURLToURI(baseURL string) (uri string, err error) {
```
julio-lopez, line 865: unexport this?

repo/grpc_repository_client_test.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/kopia/kopia/repo"
 	"github.com/kopia/kopia/repo/splitter"
+	"github.com/stretchr/testify/require"
```
julio-lopez, line 8: this fails linting, it needs to be in its own group before the kopia imports.

julio-lopez (approved):
LGTM  ✅
🚢it :shipit: 🚀

repo/grpc_repository_client_unit_test.go
```diff
@@ -0,0 +1,64 @@
...
+		expErrMsg string
+	}{
+		{
+			name:      "ipv4",
+			baseURL:   "https://1.2.3.4:5678",
+			expURI:    "1.2.3.4:5678",
+			expErrMsg: "",
+		},
+		{
+			name:      "ipv6",
```
Shrekster, line 23: nit; I don't think this parsing function is sensitive to link-local addressing (fe80::/10 namespace), but why not add a test case anyway since now you have it ?

Example link local addr: fe80::d6be:d9ff:feb0:562d

Disclaimer: I don't even know if rest of Kopia works with link-local addresses.

repo/grpc_repository_client_unit_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestBaseURLToURI(t *testing.T) {
+	for _, tc := range []struct {
+		name      string
+		baseURL   string
+		expURI    string
+		expErrMsg string
+	}{
+		{
+			name:      "ipv4",
+			baseURL:   "https://1.2.3.4:5678",
```
Shrekster, line 18: Shall we also test a URI with a `fragment`, `hostname & `username` pairs ?

Shrekster:
LGTM; we discussed the open comments and they should be tracked separately (as in Kopia should validate URI hostname/password being supplied and reject that setup explicitly). I'll file and link there issue here shortly.

PR #3846, opened 2024-05-06
docs(cli): cache hard limits flags

DarkArc:


julio-lopez (approved):

site/content/docs/Advanced/Caching/_index.md
```diff
@@ -57,6 +57,14 @@ $ kopia cache set --max-list-cache-duration=300s
 Note the cache sizes are not hard limits: cache is swept periodically (every few minutes) to bring
 the total usage below the defined limit by removing least-recently used cache items.
 
+A hard limit can be set if required via the corresponding `limit` flag:
```
Shrekster, line 60: Can we also add a side-note or a warning to this message that this setting hard limits can potentially increase the network activity ?

Shrekster: Not a problem, we can watch for community usage and add this note later if need be. Thank you for your contribution!

I understand your point about this being obvious to users, so the smaller the limit the greater the network utilization expectation. I raised the concern because the impact has been perceived to be significant in some case. For example this issue:

https://github.com/kopia/kopia/issues/3059

We fixed the above problem BTW by allowing separate cache limits and defaults for each type of cache (list, index, metadata & content). I don't have the data yet, but I think playing with all of these limits can surface some interesting network behavior.

DarkArc: Hmmm... That's definitely interesting, thanks for sharing. That does seem excessive. 

This comment in particular:

> But for now you really need to allocate a big cache, our is 300GB (260GB for metadata, 40GB for data), for 300 snapshots of about 10GB each.

That seems like a _strikingly large_ caching requirement.

PR #3879, opened 2024-05-23
test(general): add delete actions to robustness suite

chaitalisg:
The robustness tests do not cover delete actions at the moment. This PR adds the following actions to the robustness tests
- delete random subdirectories
- delete contents of directories


tests/tools/kopiarunner/kopia_snapshotter.go
```diff
@@ -171,7 +171,7 @@ func (ks *KopiaSnapshotter) DeleteSnapshot(snapID string) (err error) {
 
 // RunGC implements the Snapshotter interface, issues a gc command to the kopia repo.
 func (ks *KopiaSnapshotter) RunGC() (err error) {
-	_, _, err = ks.Runner.Run("maintenance", "run", "--full")
+	_, _, err = ks.Runner.Run("maintenance", "run", "--full", "--force")
```
chaitalisg, line 174: Add another action that runs forceful maintenance

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -51,6 +59,12 @@ func TestManySmallFiles(t *testing.T) {
 
 		_, err = eng.ExecAction(ctx, engine.RestoreSnapshotActionKey, snapOut)
 		require.NoError(t, err)
+
+		_, err = eng.ExecAction(ctx, engine.DeleteRandomSubdirectoryActionKey, deleteDirOpts)
+		require.NoError(t, err)
+
+		_, err = eng.ExecAction(ctx, engine.DeleteDirectoryContentsActionKey, deleteDirOpts)
+		require.NoError(t, err)
```
julio-lopez, line 67: @chaitalisg What is (a) the intent and (b) the effect of this change?

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -121,6 +140,12 @@ func TestManySmallFilesAcrossDirecoryTree(t *testing.T) {
...
```
julio-lopez, line 148: Same question as above: what is the effect of performing these operations here?

tests/robustness/multiclient_test/multiclient_test.go
chaitalisg: The robustness tests don't test the deleting directories or directory content at the moment. The intent is to provide the coverage for the missing actions. 
The current sequence of actions, deletion followed by data generation, reduces the snapshot size growth, thus ensuring the reduction in capacity required for long term testing.

tests/robustness/multiclient_test/multiclient_test.go
chaitalisg: The robustness tests don't test the deleting directories or directory content at the moment. The intent is to provide the coverage for the missing actions.
The current sequence of actions, deletion followed by data generation, reduces the snapshot size growth, thus ensuring the reduction in capacity required for long term testing.

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -132,19 +157,22 @@ func TestRandomizedSmall(t *testing.T) {
...
 		engine.WriteRandomFilesActionKey: map[string]string{
-			fiofilewriter.IOLimitPerWriteAction:    strconv.Itoa(512 * 1024 * 1024),
-			fiofilewriter.MaxNumFilesPerWriteField: strconv.Itoa(100),
-			fiofilewriter.MaxFileSizeField:         strconv.Itoa(64 * 1024 * 1024),
-			fiofilewriter.MaxDirDepthField:         strconv.Itoa(3),
+			fiofilewriter.IOLimitPerWriteAction:        strconv.Itoa(512 * 1024 * 1024),
+			fiofilewriter.MaxNumFilesPerWriteField:     strconv.Itoa(100),
+			fiofilewriter.MaxFileSizeField:             strconv.Itoa(64 * 1024 * 1024),
+			fiofilewriter.MaxDirDepthField:             strconv.Itoa(maxDirDepth),
+			fiofilewriter.DeletePercentOfContentsField: strconv.Itoa(deleteContentsPercentage),
```
julio-lopez, line 175: What is the effect of this line (175) ?

chaitalisg: The delete-directory-contents action deletes some of the contents of a random directory, based on its input options values. `DeletePercentOfContentsField` is one of the input options. It sets the percentage for deleting contents of the random directory. So, setting this field and executing `DeleteDirectoryContentsActionKey` action would result in deletion of 50% of contents of a random directory in the source directory.

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -132,19 +157,22 @@ func TestRandomizedSmall(t *testing.T) {
...
+
 	opts := engine.ActionOpts{
 		engine.ActionControlActionKey: map[string]string{
 			string(engine.SnapshotDirActionKey):              strconv.Itoa(2),
 			string(engine.RestoreSnapshotActionKey):          strconv.Itoa(2),
-			string(engine.DeleteRandomSnapshotActionKey):     strconv.Itoa(1),
+			string(engine.DeleteRandomSnapshotActionKey):     strconv.Itoa(4),
 			string(engine.WriteRandomFilesActionKey):         strconv.Itoa(8),
-			string(engine.DeleteRandomSubdirectoryActionKey): strconv.Itoa(1),
+			string(engine.DeleteRandomSubdirectoryActionKey): strconv.Itoa(maxDirDepth),
```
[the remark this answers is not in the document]
chaitalisg: DeleteRandomSubdirectory deletes a random directory up to a specified depth, based on its input option - MaxDirDepthField. The function uses the MaxDirDepthField when specified or uses a default depth of 20. The source directory may not always have a depth of 20. This may lead to the delete action being a no-op, which is reported as an error by the robustness test, and leads to a test failure. 
Specifying MaxDirDepthField will limit the depth for deletion. Thus, reducing the chances of hitting the no-op error.

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -132,19 +157,22 @@ func TestRandomizedSmall(t *testing.T) {
...
 	st := timetrack.StartTimer()
 
+	maxDirDepth := 3
+
 	opts := engine.ActionOpts{
 		engine.ActionControlActionKey: map[string]string{
 			string(engine.SnapshotDirActionKey):              strconv.Itoa(2),
 			string(engine.RestoreSnapshotActionKey):          strconv.Itoa(2),
-			string(engine.DeleteRandomSnapshotActionKey):     strconv.Itoa(1),
+			string(engine.DeleteRandomSnapshotActionKey):     strconv.Itoa(4),
```
julio-lopez, line 166: Why change this from 1 to 4?

tests/robustness/multiclient_test/multiclient_test.go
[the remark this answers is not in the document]
julio-lopez: 1. IIUC: In short, the intent in line 168 is to change a parameter for the _"Delete Random Subdirectory"_ action, is that the case?
2. What is the `ActionControlActionKey` map used for?

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: And where is `DeletePercentOfContentsField` being used? which "action" uses this parameter?

tests/robustness/multiclient_test/multiclient_test.go
chaitalisg: This will increase the weight for deleting random snapshots from the repository under test. Increasing the weight will result in deletion of more snapshots. This will maintain a smaller delta in the number of snapshots added in each test run. Thus reducing the filesystem capacity required to run the test long term.

tests/robustness/multiclient_test/multiclient_test.go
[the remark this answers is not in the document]
chaitalisg: 1. yes. 
2. The `ActionControlActionKey` map is used to provide a collection of engine actions for the `tryRandomAction` function to choose from, which is a part of `TestRandomizedSmall`.

tests/robustness/multiclient_test/multiclient_test.go
chaitalisg: Good point. I added the `DeleteDirectoryContentsActionKey` that can use `DeletePercentOfContentsField`

tests/robustness/multiclient_test/multiclient_test.go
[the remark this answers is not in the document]
julio-lopez: Why use `maxDirDepth` here and below (line 169)?

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: What happens when the weight for deleting snapshots is (much) larger than snapshot creation?

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -132,19 +157,23 @@ func TestRandomizedSmall(t *testing.T) {
...
 		engine.WriteRandomFilesActionKey: map[string]string{
-			fiofilewriter.IOLimitPerWriteAction:    strconv.Itoa(512 * 1024 * 1024),
-			fiofilewriter.MaxNumFilesPerWriteField: strconv.Itoa(100),
-			fiofilewriter.MaxFileSizeField:         strconv.Itoa(64 * 1024 * 1024),
-			fiofilewriter.MaxDirDepthField:         strconv.Itoa(3),
+			fiofilewriter.IOLimitPerWriteAction:        strconv.Itoa(512 * 1024 * 1024),
+			fiofilewriter.MaxNumFilesPerWriteField:     strconv.Itoa(100),
+			fiofilewriter.MaxFileSizeField:             strconv.Itoa(64 * 1024 * 1024),
+			fiofilewriter.MaxDirDepthField:             strconv.Itoa(maxDirDepth),
+			fiofilewriter.DeletePercentOfContentsField: strconv.Itoa(deleteContentsPercentage), // used in DeleteDirectoryContentsAction
```
julio-lopez, line 176: Does this have any effect? Will it be used?

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: Where?

julio-lopez (changes requested):
@chaitalisg Please put the changes for the `TestRandomizedSmall` (changes after line 155) in a separate PR. That way the other changes can be merged right away.

tests/robustness/multiclient_test/multiclient_test.go
chaitalisg: Line 177

tests/robustness/multiclient_test/multiclient_test.go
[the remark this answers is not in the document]
chaitalisg: good catch. Removed.

tests/robustness/multiclient_test/multiclient_test.go
chaitalisg: From the point of view of this specific test, it would run out of snapshots at some point. If we consider rest of tests in the suite, that may not happen. I have reverted the weight increase because it is not the best idea to rely on other test methods to create more snapshots than those are deleted here.

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: Good. I had not seen that before.

julio-lopez (approved):

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: True, other parts of the job may create additional snapshots.
Nevertheless, we should be aware of the implications of the change. It may be a good idea to do minor tweaks to the weights and monitor the number of snapshots over time.
We can discuss this in a separate PR.
Thanks.

PR #3883, opened 2024-05-26
chore(ci): upgraded linter to 1.59.0

jkowalski:


.golangci.yml
```diff
@@ -65,34 +65,26 @@ linters-settings:
 linters:
   enable-all: true
   disable:
-    - deadcode
```
julio-lopez, line 68: Does this enable all the linters that are removed from this section? Or are there linters that were removed from golanci-lint and are then implicitly disabled?

internal/epoch/epoch_manager_test.go
```diff
@@ -1416,8 +1416,8 @@ func TestCleanupMarkers_CleanUpManyMarkers(t *testing.T) {
 	require.Len(t, cs.EpochMarkerBlobs, 2) // at least 2 epoch markers are kept
 }
 
-func randomTime(min, max time.Duration) time.Duration {
-	return time.Duration(float64(max-min)*rand.Float64() + float64(min))
+func randomTime(minTime, maxTime time.Duration) time.Duration {
```
julio-lopez, line 1419: Was there a linter that prompted or suggested the change in variable names?

tools/cli2md/cli2md.go
```diff
@@ -42,8 +42,8 @@ func emitFlags(w io.Writer, flags []*kingpin.FlagModel) {
 		return
 	}
 
-	fmt.Fprintf(w, "| Flag | Short | Default | Help |\n")
-	fmt.Fprintf(w, "| ---- | ----- | --- | --- |\n")
+	fmt.Fprintf(w, "| Flag | Short | Default | Help |\n") //nolint:errcheck
```
julio-lopez, line 45: Wondering, given that there are many `fmt.Fprint*` and `fmt.Print*` where the errors are ignored, would it make sense to create a helper function that ignores the error and is used in cases were it makes sense to ignore the error, or maybe there's a separate helper that checks and panics?
WDYT?

julio-lopez (commented):
+1: LG 👍
🥇Thanks for doing this.

internal/epoch/epoch_manager_test.go
jkowalski: yes, min and max are now built-in functions in go

tools/cli2md/cli2md.go
jkowalski: honestly not worth doing it, if we can't write to stdout, panic is not something anybody's going to see anyway. I think in such case `SIGPIPE` will kill the process anyway.

julio-lopez: It depends, doesn't it?

- In the case of `fmt.Print*`, if `stdout` is closed (e.g., a pipe was closed), `stderr` _may_ still be available and by default `panic`'s output go to `stderr`.
- In the case of `fmt.Fprint*`, the output may be a file or something other than `stdout` (it could be `stdout` as well). The panic output would go to `stderr`.

.golangci.yml
julio-lopez: So, all of these were removed then?

julio-lopez (approved):

PR #3887, opened 2024-05-29
feat(snapshots): added support for per-directory splitter overrides

jkowalski:
This is useful when backing up directories that have giant files aligned at MiB boundary, such as VM disk backups, etc.


repo/object/object_manager.go
```diff
@@ -44,17 +44,29 @@ type contentManager interface {
...
 func (om *Manager) NewWriter(ctx context.Context, opt WriterOptions) Writer {
 	w, _ := om.writerPool.Get().(*objectWriter)
 	w.ctx = ctx
 	w.om = om
-	w.splitter = om.newSplitter()
+
+	var splitFactory splitter.Factory
+
+	if opt.Splitter != "" {
+		splitFactory = splitter.GetFactory(opt.Splitter)
```
julio-lopez, line 61: nit: if the specified splitter is not found, then this silently falls back to the default. This may be confusing since there's no error or warning. Is this logged anywhere?

cli/command_policy_set_splitter.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (c *policySplitterFlags) setup(cmd *kingpin.CmdClause) {
+	cmd.Flag("splitter", "Splitter algorithm override").EnumVar(&c.policySetSplitterAlgorithmOverride, supportedSplitterAlgorithms()...)
+}
+
+//nolint:unparam
+func (c *policySplitterFlags) setSplitterPolicyFromFlags(ctx context.Context, p *policy.SplitterPolicy, changeCount *int) error {
+	if v := c.policySetSplitterAlgorithmOverride; v != "" {
+		if v == inheritPolicyString {
```
julio-lopez, line 24: Q: IIUC, `inheritPolicyString` is used in order for resetting a policy setting when it was already set in a child, is that the case? Otherwise, what's the difference between "inheriting" and an unset value?

cli/command_policy_set_splitter_test.go
```diff
@@ -0,0 +1,42 @@
...
+	lines = compressSpaces(lines)
+	require.Contains(t, lines, " Algorithm override: (repository default) inherited from (global)")
+
+	e.RunAndExpectSuccess(t, "policy", "set", td, "--splitter=FIXED-4M")
+
+	lines = e.RunAndExpectSuccess(t, "policy", "show", td)
+	lines = compressSpaces(lines)
+	require.Contains(t, lines, " Algorithm override: FIXED-4M (defined for this target)")
+
+	e.RunAndExpectSuccess(t, "policy", "set", td, "--splitter=inherit")
```
julio-lopez, line 35: I guess this answers my previous (above) question

julio-lopez (approved):
LGTM  ✅

🥇Thanks for splitting the PR in easily reviewable commits. It made it so much easier. Thanks again!

Issue #3842, opened 2024-05-02
Snapshotting root of drive fails when Volume Shadow Copy enabled


Hakkin:
Snapshotting fails with `ERROR upload error: The parameter is incorrect.` when snapshotting the root of the drive with `--enable-volume-shadow-copy` set to `always`. Snapshotting a sub directory in the root works correctly. All commands were run with administrator privileges.

```shell
$ kopia --version
0.17.0 build: 89c8eb47af2e1d5c1d14fe299a0cf7eaac095abf from: kopia/kopia

$ kopia policy set --enable-volume-shadow-copy always C:\
Setting policy for Hakkin@desktop:C:\
 - setting "enable volume shadow copy" to always.

$ kopia snapshot create C:\
Snapshotting Hakkin@desktop:C:\ ...
creating volume shadow copy of C:\
new volume shadow copy id {8FB0579B-5364-4138-8CEA-A2CB2F553E42}
removing volume shadow copy id {8FB0579B-5364-4138-8CEA-A2CB2F553E42}
ERROR upload error: The parameter is incorrect.

$ kopia snapshot create C:\Users\
Snapshotting Hakkin@desktop:C:\Users ...
creating volume shadow copy of C:\
new volume shadow copy id {3FF5EB4F-8E6A-441A-87DD-A4975494BBBB}
 | 3 hashing, 1778 hashed (250 MB), 0 cached (0 B), uploaded 219.1 MB, estimating...
```

tekert:
Same case on windows server 2019, doesn't seem to happen on Windows 10 on my other pc.

Hakkin:
After running with debugging, I may know the issue.

```
DEBUG shadow copy root is \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy10
DEBUG snapshotted directory     {"path":".","error":"GetFileInformationByHandleEx \\\\?\\GLOBALROOT\\Device\\HarddiskVolumeShadowCopy10: The parameter is incorrect.","dur":"512.2µs"}
```

I believe I have ran into an issue like this before, it is caused by the shadow volume path not ending in `\`. For whatever reason Windows won't treat it as a folder without the ending `\`. You can observe the same behavior by trying to mount a Shadow Copy volume via mklink

```
$ mklink /j "vss" "\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy4"
Junction created for vss <<===>> \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy4

[...]

PR #3891, opened 2024-05-31
fix(snapshots): Append path separator to Shadow Copy root directory on Windows

Hakkin:
This is a fix for #3842.
It's a slightly modified version of the patch in https://github.com/kopia/kopia/issues/3842#issuecomment-2105616219, I separated the code out into a `maybe` function and added some tests. 
The root cause of this issue seems to be a regression in Go 1.22 (see https://github.com/kopia/kopia/issues/3842#issuecomment-2141250166), but as discussed with jkowalski in the issue, it seems better to just fix this up with a special case in Kopia.

### DIFF
--- a/fs/localfs/local_fs.go
+++ b/fs/localfs/local_fs.go
@@ -55,7 +55,7 @@ func (e *filesystemEntry) fullPath() string {
 }
 
 func (e *filesystemEntry) isWindowsVSSVolume() bool {
-	return runtime.GOOS == "windows" && //nolint:goconst
+	return runtime.GOOS == "windows" &&
 		e.prefix == `\\?\GLOBALROOT\Device\` &&
 		strings.HasPrefix(e.Name(), "HarddiskVolumeShadowCopy")
 }
--- a/fs/localfs/local_fs.go
+++ b/fs/localfs/local_fs.go
@@ -55,7 +55,7 @@ func (e *filesystemEntry) fullPath() string {
 }
 
 func (e *filesystemEntry) isWindowsVSSVolume() bool {
-	return runtime.GOOS == "windows" &&
+	return runtime.GOOS == "windows" && //nolint:goconst
 		e.prefix == `\\?\GLOBALROOT\Device\` &&
 		strings.HasPrefix(e.Name(), "HarddiskVolumeShadowCopy")
 }
--- a/fs/localfs/local_fs.go
+++ b/fs/localfs/local_fs.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"runtime"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -50,6 +52,12 @@ func (e *filesystemEntry) Sys() interface{} {
 
 func (e *filesystemEntry) fullPath() string {
 	return e.prefix + e.Name()
+}
+
+func (e *filesystemEntry) isWindowsVSSVolume() bool {
+	return runtime.GOOS == "windows" &&
+		e.prefix == `\\?\GLOBALROOT\Device\` &&
+		strings.HasPrefix(e.Name(), "HarddiskVolumeShadowCopy")
 }
 
 func (e *filesystemEntry) Owner() fs.OwnerInfo {
--- a/fs/localfs/local_fs_os.go
+++ b/fs/localfs/local_fs_os.go
@@ -66,7 +66,13 @@ func (it *filesystemDirectoryIterator) Close() {
 func (fsd *filesystemDirectory) Iterate(_ context.Context) (fs.DirectoryIterator, error) {
 	fullPath := fsd.fullPath()
 
-	f, direrr := os.Open(fullPath) //nolint:gosec
+	// Direct Windows volume paths (e.g. Shadow Copy) require a trailing \ or listing will fail
+	appendPath := ""
+	if fsd.isWindowsVSSVolume() {
+		appendPath = string(os.PathSeparator)
+	}
+
+	f, direrr := os.Open(fullPath + appendPath) //nolint:gosec
 	if direrr != nil {
 		return nil, errors.Wrap(direrr, "unable to read directory")
 	}


Hakkin:
Something additional to note, there is a similar check going on here in the NewEntry function:
https://github.com/kopia/kopia/blob/fcb8197f3f0773d70ad202c195c6ce88fcb01c2d/fs/localfs/local_fs_os.go#L112-L128

I wonder if it would be useful to change this to use the new `maybeAppendPathSeparatorForVSSOnWindows`  function as well.

Whissi:
I applied this PR on top of v0.17.0 tag which I confirmed to be showing the problem reported in #3842 using `go version go1.22.2 windows/amd64`. With this PR applied, I am now able again to snapshot an entire volume when using a volume shadow copy.

However, it's worth noting that appending a trailing backslash when detecting the `\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy` prefix is **not specific** to shadow copies but rather ensures proper handling of paths **within the Windows Object namespace**. `\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopyXX` represents a symbolic link within the Windows Object namespace (`\\GLOBALROOT` folder) pointing to `\Device\HarddiskVolumeShadowCopyXX` (which is a device and nothing within `\\GLOBALROOT` folder). For Windows, without a trailing backslash, it may not be clear whether we intend to operate on the symbolic link itself or on the actual resource, `\Device\HarddiskVolumeShadowCopyXX`. Thus, the workaround is essential for disambiguation. It might be beneficial to consider a more generalized approach by checking if a path starts with `\` and ensuring it ends with a trailing backslash. This would provide robustness in handling various path scenarios, especially in our case where we always expect to deal with iterable objects (directories). And if we wouldn't access the shadow copy through Windows Object namespace at all (for example we could mount the shadow copy somewhere) we wouldn't need to take care of this at all.

Hakkin:
Thanks for the extra info. 

> \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopyXX represents a symbolic link within the Windows Object namespace (\\GLOBALROOT folder) pointing to \Device\HarddiskVolumeShadowCopyXX (which is a device and nothing within \\GLOBALROOT folder).

I wonder if there's any Windows API to detect or resolve links like this in the Object namespace? For normal symlinks / reparse points in the filesystem this can be done with os.Lstat, but as noted in the code in my second comment, Lstat seems to fail on these kinds of paths unless you "manually" resolve them with the trailing backslash. If there's a way to detect them directly, it might have sense to implement them like normal symlinks are now, and there wouldn't be any need to handle shadow volumes explicitly.

> It might be beneficial to consider a more generalized approach by checking if a path starts with \ and ensuring it ends with a trailing backslash. This would provide robustness in handling various path scenarios, especially in our case where we always expect to deal with iterable objects (directories).

I considered doing something like this but I was worried about edge cases, which is why I went out of my way to make sure the workaround is only applied to the top-level root directory. Since "inside" the shadow volume device is just a normal filesystem, appending a backslash anywhere after the root might have different consequences, like resolving links. If we did something more generic, it would definitely need a lot more testing to make sure all the edge cases are handled correctly.

Hakkin:
I ended up reading a lot about Windows path handling, [this article](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) seems to contain a lot of the relevant information.

I believe the fundamental issue here is that Windows treats opening the path `\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy1` as attempting to open the raw block device for that volume, while it treats `\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy1\` as opening the root directory of the volume, and it passes it to the underlying filesystem driver.

Unfortunately, it seems non-triviable to handle this "correctly". It seems impossible to know whether a UNC path points to a device object without doing multiple Windows-specific syscalls for every path.

That being said, I think there is a way to handle this a bit more cleanly. For one, rather than the current `maybe` function, after thinking about it more, I think it's better to implement a method on `filesystemEntry` that checks the path parts. `filesystemEntry` already has the prefix and filename split out, so it simplifies the logic and we don't have to do any path handling in the checking function.

```go
func (e *filesystemEntry) isWindowsVSSVolume() bool {
	return runtime.GOOS == "windows" &&
		e.prefix == `\\?\GLOBALROOT\Device\` &&
		strings.HasPrefix(e.Name(), "HarddiskVolumeShadowCopy")
}
```

then we can just do this in `Iterate`:

```go
fullPath := fsd.fullPath()

appendPath := ""
if fsd.isWindowsVSSVolume() {
	appendPath = string(os.PathSeparator)
}

f, direrr := os.Open(fullPath + appendPath) //nolint:gosec
```

and that's all the logic needed.

If we ever want to make it a bit more generic in the future. the `isWindowsVSSVolume` function can be expanded to support other common volume paths like `\\?\Volume{...}` as well.

If there's no objections to this revised patch, I think I'll push another commit with it in a little bit since it's pretty much the same logic as the previous commit, just cleaner.

Hakkin:
Regarding the test coverage, should I add a test that covers this case? I think I could just mostly copy https://github.com/kopia/kopia/blob/master/tests/os_snapshot_test/os_snapshot_windows_test.go but enable `enable-volume-shadow-copy=always` and set it to snapshot the root of the C:\ drive instead, we can add an ignore policy so it doesn't actually snapshot any of the files, just tests creating the shadow copy, since that's where it would fail before.

xieve:
Tested on Windows 10 with a boot drive SSD and a data HDD, working just fine. Thank you for your contribution, I hope to see this upstream soon :)

wildyaboku:
I just tested this on the latest release of Windows 11 (24H2) and it's working as expected.

bparker98:
Now that we finally have a new Kopia release is there any chance this fix could get some love?

tekert:
Upgraded from 0.16 to 0.18.2 and vss broke, applied this path, now it works, not much more to say.

Janzert:
Are there any specific issues blocking this from getting merged?

This is the thing blocking me from upgrading off of 0.16. It'd be great to know if there is anything to help move this forward.

Lerrissirrel:
> Are there any specific issues blocking this from getting merged?
> 
> This is the thing blocking me from upgrading off of 0.16. It'd be great to know if there is anything to help move this forward.

Looks like it's (at least) awaiting a review.  I built my own 18.2 with this change included, was hoping it would make 19!

t3pco:
Also waiting for this fix, as it's a blocking usecase for me.
Would really appreciate to see finalization of the existing VSS and Windows-specific features connected to that (enabled via. UI, docs)

t3pco:
> @julio-lopez or @jkowalski would it be possible get this merged? Are there any blockers preventing this from going through?

+1, was not able to make a local working build with this, but user report it's working.
Using the very old client is not a valid option for me.

Lerrissirrel:
I was hoping the silence on this thread meant this wasn't a problem any more in .21 but it looks like it still is.  Fortunately the changes in this PR still work if you're able to build for yourself.

t3pco:
Had the same hope that the fix will be merged at least into this release.
Managed to build it locally now, but frustrated that a working and small fix did not make it again.

I understand that Windows is not top priority, but this bug makes it nearly unuseable for windows clients. Using 0.16 is a workaround, but not for such long time

Pushing again so we maybe have it for 0.22 then :-)

jkowalski:
i'll get this merged, i got a hold of a Windows box and will try it out and merge after that.

jkowalski (approved):

jkowalski:
ok, i think it works, i'm going to merge, please try it for your use case when a build is produced

t3pco:
> ok, i think it works, i'm going to merge, please try it for your use case when a build is produced

Thanks for your effort in this case. Tested recent master build 20250803.0.35640 installed via. KopiaUI-Setup-20250803.0.35640.exe. Confirmed working.

PR #3901, opened 2024-06-04
refactor(general): Index compaction enhancement

Lyndon-Li:
1. Fix the problem that `runTaskEpochMaintenanceQuick` never gets called in `runQuickMaintenance`
2. Add a hard limit to `epochAdvanceOnTotalSizeBytesThreshold` for `shouldAdvance`. Epoch is advanced when the total index size in the epoch exceeds the hard limit. In this way, we keep indexes in an epoch in a reasonable number/size, so that the index compaction for that epoch won't take extremely high CPU and memory

### DIFF
--- a/internal/epoch/epoch_advance.go
+++ b/internal/epoch/epoch_advance.go
@@ -13,7 +13,8 @@ import (
 //
 // - number of blobs in the epoch exceeds 'countThreshold'
 // - total size of blobs in the epoch exceeds 'totalSizeBytesThreshold'.
-func shouldAdvance(bms []blob.Metadata, minEpochDuration time.Duration, countThreshold int, totalSizeBytesThreshold int64) bool {
+// Or the total index blob size in the current epoch eceeds 'totalSizeBytesThresholdHard'.
+func shouldAdvance(bms []blob.Metadata, minEpochDuration time.Duration, countThreshold int, totalSizeBytesThreshold int64, totalSizeBytesThresholdHard int64) bool {
 	if len(bms) == 0 {
 		return false
 	}
@@ -34,6 +35,10 @@ func shouldAdvance(bms []blob.Metadata, minEpochDuration time.Duration, countThr
 		}
 
 		totalSize += bm.Length
+	}
+
+	if totalSize >= totalSizeBytesThresholdHard {
+		return true
 	}
 
 	// not enough time between first and last write in an epoch.
--- a/repo/maintenance/maintenance_run.go
+++ b/repo/maintenance/maintenance_run.go
@@ -252,19 +252,24 @@ func Run(ctx context.Context, runParams RunParameters, safety SafetyParameters) 
 }
 
 func runQuickMaintenance(ctx context.Context, runParams RunParameters, safety SafetyParameters) error {
+	s, err := GetSchedule(ctx, runParams.rep)
+	if err != nil {
+		return errors.Wrap(err, "unable to get schedule")
+	}
+
 	_, ok, emerr := runParams.rep.ContentManager().EpochManager(ctx)
 	if ok {
-		log(ctx).Debugf("quick maintenance not required for epoch manager")
+		log(ctx).Debugf("Epoch manager is available, run epoch maintenance only for quick maintenance")
+
+		if err := runTaskEpochMaintenanceQuick(ctx, runParams, s); err != nil {
+			return errors.Wrap(err, "error running quick epoch maintenance tasks")
+		}
+
 		return nil
 	}
 
 	if emerr != nil {
 		return errors.Wrap(emerr, "epoch manager")
-	}
-
-	s, err := GetSchedule(ctx, runParams.rep)
-	if err != nil {
-		return errors.Wrap(err, "unable to get schedule")
 	}
 
 	if shouldQuickRewriteContents(s, safety) {
@@ -297,10 +302,6 @@ func runQuickMaintenance(ctx context.Context, runParams RunParameters, safety Sa
 		}
 	} else {
 		notDeletingOrphanedBlobs(ctx, s, safety)
-	}
-
-	if err := runTaskEpochMaintenanceQuick(ctx, runParams, s); err != nil {
-		return errors.Wrap(err, "error running quick epoch maintenance tasks")
 	}
 
 	// consolidate many smaller indexes into fewer larger ones.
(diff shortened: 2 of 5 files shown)


internal/epoch/epoch_advance.go
```diff
@@ -36,6 +37,10 @@ func shouldAdvance(bms []blob.Metadata, minEpochDuration time.Duration, countThr
 		totalSize += bm.Length
 	}
 
+	if totalSize >= totalSizeBytesThresholdHard {
+		return true
+	}
+
```
julio-lopez, line 43: @Lyndon-Li Please remove this check from this PR. At the moment, we are not going to change the conditions for compaction.

julio-lopez, line 43: @Lyndon-Li Please remove this check from this PR (and the corresponding tests). At the moment, we are not going to change the conditions for compaction. Also, if it were to do it, then it should be separately in its own PR.

```suggestion
```

julio-lopez (changes requested):

internal/epoch/epoch_advance.go
Lyndon-Li: If we don't add this change, I am afraid the epoch maintenance in the quick maintenance doesn't make any difference --- since the `minEpochDuration` is 24 hour, even though quick maintenance launches the epoch maintenance every 1 hour, it actually does nothing since the epoch is not advanced.

julio-lopez: Running epoch maintenance in quick maintenance would make a difference. Specifically regarding epoch advancement, it would advance it within an hour of the reaching the minimum epoch duration age. Otherwise, in the worst case scenario the epoch would be advance close to 24 hours after being reaching the minimum epoch age.

repo/maintenance/maintenance_run.go
```diff
@@ -252,21 +252,26 @@ func Run(ctx context.Context, runParams RunParameters, safety SafetyParameters)
...
 func runQuickMaintenance(ctx context.Context, runParams RunParameters, safety SafetyParameters) error {
+	s, err := GetSchedule(ctx, runParams.rep)
+	if err != nil {
+		return errors.Wrap(err, "unable to get schedule")
+	}
+
 	_, ok, emerr := runParams.rep.ContentManager().EpochManager(ctx)
 	if ok {
-		log(ctx).Debugf("quick maintenance not required for epoch manager")
+		log(ctx).Debugf("Epoch manager is available, run epoch maintenance only for quick maintenance")
```
julio-lopez, line 262: Are you planning to add a test for this case?

Lyndon-Li: Yes, I can add some UT once we decide the scope of this PR.
At present, there is no UT for the entire function of runQuickMaintenance, let me see how to add it.

internal/epoch/epoch_advance.go
Lyndon-Li: Yes, but that is not enough, or the epoch advancing is not timely enough once a burst of index generation happens. The consequence is that the next index compaction takes huge memory.

julio-lopez: You can lower the minimum epoch duration for the repository.

Lyndon-Li: That doesn't work well for the case I mentioned:
1. We can not predict that the coming backups generate large number of indexes. Or in another word, this doesn't happen for every backup
2. This is not timely enough. E.g., say we set the duration to 4 hour, but a backup may complete in 1 hour and generates large number of indexes

Lyndon-Li: That doesn't work well for the case I mentioned:
1. We can not predict that the coming backups generate large number of indexes. Or in another word, this doesn't happen for every backup
2. This is not timely enough. E.g., say we set the duration to 4 hour, but a backup may complete in 1 hour and generates large number of indexes

Lyndon-Li: May I know the concern here? are we expecting the epochs to always be regular in terms of time?

repo/maintenance/maintenance_run.go
julio-lopez: Tests added in #4185

internal/epoch/epoch_advance.go
Lyndon-Li: @julio-lopez 

> Is it safe?
Sorry, I didn't see this comment.

Below is my thinking:

- The index compaction won't delete the uncompacted/superseded indexes immediately. After they are compacted in a quick maintenance, they will be deleted in the following FULL maintenance ONLY when they are old enough (controlled by `CleanupSafetyMargin`, by default 4 hours).
- Therefore, adding the hard limit ostensibly increase the risk of data lose, but risk is actually decided by `CleanupSafetyMargin`. If data safety is the concern, users should increase `CleanupSafetyMargin`.
- Additionally, at present, we support a customized duration time as shorter as 10min. So users should have already been using this experience --- advance the epoch in a short time.

Data safety is always with top priority, but scalability and performance is also important. I am not sure if the changes in this PR is 100% correct, but it brings a possible solution of considering the two aspects together. So appreciate your further comments.

julio-lopez: @Lyndon-Li 

You raise an important point about the safety of deleting compacted indexes. Indeed a safety margin is needed so clients that are in the process of reading an old index snapshot can read that snapshot.

The proposed change adds a new criterion for advancing the epoch (`totalSize >= totalSizeBytesThresholdHard`). The original protocol design uses 2 other criteria to reach consensus about:(a) when it is time to advance and epoch; and (b) when it safe to assume that the epoch will no longer be modified, so it can be compacted.

The new criterion appears to be violating one of the original assumptions. It _may_ be that the proposed change is safe, but it is not yet clear, at least not to me.

Lyndon-Li: OK, I got you points and understand your concern of data safety.

And it is also a fact that the current behavior of epoch advancing brings burst burden to repo maintenance (epoch index compaction), as index creations are never linearly streamed along with time, but influx with backups.
Let's see if this brings problems to other cases, then we can come back to discuss this.

julio-lopez:
Mostly superseded by #4185. Closing for now.

Credits to @Lyndon-Li for catching the issue and proposing the solution. Thanks again.

PR #3907, opened 2024-06-07
fix(general): avoid panic on computing password hash error

julio-lopez:
Rationale: this code path is primarily executed from the server. A potential error, say from a corrupt, unsupported or otherwise invalid user profile should not cause the server to panic (and crash).

It is possible for `computePasswordHash` to return an error, not just an impossibility.


julio-lopez:
@bathina2 FYI

PR #3919, opened 2024-06-15
fix(general): relax epoch compaction check for contiguous ranges

julio-lopez:
While the current implementation should not produce non-contiguous ranges, this invariant check may be too stringent, in particular for repositories created and mutated by older kopia (client) processes.

For backwards compatibility the check is relaxed to allow operations to proceed in older repositories.

Ref:
- #3735
- #3638
- #3728
- #3651

### DIFF
--- a/internal/epoch/epoch_utils_test.go
+++ b/internal/epoch/epoch_utils_test.go
@@ -347,6 +347,12 @@ func TestOldestUncompactedEpoch(t *testing.T) {
 		},
 		{
 			input: CurrentSnapshot{
+				SingleEpochCompactionSets: makeSingleCompactionEpochSets([]int{23}),
+			},
+			expectedEpoch: 0,
+		},
+		{
+			input: CurrentSnapshot{
 				// non-contiguous single epoch compaction set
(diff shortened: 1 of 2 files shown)


internal/epoch/epoch_utils_test.go
```diff
@@ -512,25 +538,205 @@ func TestOldestUncompactedEpoch(t *testing.T) {
...
+func makeSingleCompactionEpochSets(epochs []int) map[int][]blob.Metadata {
+	es := make(map[int][]blob.Metadata, len(epochs))
+	for _, e := range epochs {
+		es[e] = []blob.Metadata{{BlobID: compactedEpochBlobPrefix(e) + "foo_" + blob.ID(strconv.Itoa(e))}}
+	}
+
+	return es
+}
+
+func TestGetFirstContiguosKeyRange(t *testing.T) {
```
julio-lopez, line 635:
```suggestion
func TestGetFirstContiguousKeyRange(t *testing.T) {
```

jkowalski (approved):

PR #3927, opened 2024-06-18
test(general): run gc action robustness tests

chaitalisg:
The robustness tests do not cover maintenance action at the moment. This PR adds the full maintenance action to the robustness tests.


tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -150,6 +150,14 @@ func TestManySmallFilesAcrossDirecoryTree(t *testing.T) {
...
 	ctx := testlogging.ContextWithLevel(t, testlogging.LevelInfo)
 	th.RunN(ctx, t, numClients, f)
+
+	// Run gc action once
+	runGCFcn := func(ctx context.Context, t *testing.T) { //nolint:thelper
+		t.Log("run maintenance")
+		_, err := eng.ExecAction(ctx, engine.GCActionKey, nil)
+		require.NoError(t, err)
+	}
+	th.RunN(ctx, t, 1, runGCFcn)
```
redgoat650, line 160: Is there a reason to use `th.RunN` with 1 client vs `th.Run`? I'm pretty rusty in this code path 😅

redgoat650 (approved):
LGTM. I could have sworn we added maintenance runs in the past. Was it removed at some point or am I losing my mind? 😄

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: Related question: why run maintenance here (in  `TestManySmallFilesAcrossDirecoryTree`)?

chaitalisg: I had two choices either create a new test or add it to one of the existing tests that perform other operations which impacts maintenance. I can add a new test if that is preferable.

chaitalisg: > Is there a reason to use `th.RunN` with 1 client vs `th.Run`? I'm pretty rusty in this code path 😅

Yes, I tried `th.Run`, but the test errored out. I can take that as an enhancement/refactor item for later.

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -194,6 +194,19 @@ func TestRandomizedSmall(t *testing.T) {
 	th.RunN(ctx, t, numClients, f)
 }
 
+func TestMaintenanceAction(t *testing.T) {
+	ctx := testlogging.ContextWithLevel(t, testlogging.LevelInfo)
+
+	numClients := 1
+	// Run gc action onc
```
julio-lopez, line 201: This comment seems incomplete

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -194,6 +194,19 @@ func TestRandomizedSmall(t *testing.T) {
...
+	runGCFcn := func(ctx context.Context, t *testing.T) { //nolint:thelper
+		t.Log("run maintenance")
+		_, err := eng.ExecAction(ctx, engine.GCActionKey, nil)
+		require.NoError(t, err)
+	}
+	th.RunN(ctx, t, numClients, runGCFcn)
```
julio-lopez, line 207: Does it work? Is maintenance being triggered?

chaitalisg: Yes. Copied from the log:

`STDERR:
Running full maintenance...
Looking for active contents...
Looking for unreferenced contents...
GC found 0 unused contents (0 B)
GC found 0 unused contents that are too recent to delete (0 B)
GC found 10022 in-use contents (42.2 MB)
GC found 61 in-use system-contents (26.8 KB)
Rewriting contents from short packs...
Total bytes rewritten 873.7 KB
Found safe time to drop indexes: 2024-06-21 11:08:20.307877 -0700 PDT
Dropping contents deleted before 2024-06-21 11:08:20.307877 -0700 PDT
Skipping blob deletion because not enough time has passed yet (59m59s left).
Cleaned up 0 logs.
Cleaning up old index blobs which have already been compacted...
Finished full maintenance.`

tests/robustness/multiclient_test/multiclient_test.go
chaitalisg: addressed

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: Is this action being executed by a client?

chaitalisg: Yes. The test creates a single client, which connects to repo (`kopia repo connect server`), followed by the maintenance action (`kopia maintenance run --full`)

julio-lopez: I'm confused 😕. That does not seem to match the expected behavior.

chaitalisg: I see. Can you clarify the expected behavior? I read through the documentation, and was not able to find any reference to a client not being allowed to run maintenance. Is this the expectation you are referring to? Let me know if I am misinterpreted the comment.

chaitalisg: Here's the explanation from `snapshotter.go`

`// RunGC runs garbage collection on the server repository directly since clients
// are not authorized to do this.
func (mcs *MultiClientSnapshotter) RunGC(ctx context.Context, opts map[string]string) error {
	return mcs.server.RunGC(ctx, opts)
}`

julio-lopez: What's the alternative to make it clear and explicit?

chaitalisg: I could replace the function and the call to `RunN` with 
`	th.Engine().TestRepo.RunGC(ctx, nil)
`

This would run the maintenance command . However, it won't be counted in engine action stats.

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -190,6 +190,19 @@ func TestRandomizedSmall(t *testing.T) {
...
+	ctx := testlogging.ContextWithLevel(t, testlogging.LevelInfo)
+
+	numClients := 1
+	// Run gc action once for the repository under test.
+	runGCFcn := func(ctx context.Context, t *testing.T) { //nolint:thelper
+		t.Log("run maintenance")
+		_, err := eng.ExecAction(ctx, engine.GCActionKey, nil)
+		require.NoError(t, err)
+	}
+	th.RunN(ctx, t, numClients, runGCFcn)
```
julio-lopez, line 203: Would the following work?

```suggestion
	t.Log("running maintenance directly on the repository under test")

	// bypass the server to directly run maintenance on the repository
	// under test.
	// It launches a kopia process that directly accesses the repository
	// under test using the repo configuration for the server. The
	// server is concurrently running, since the framework starts
	// the server at the beginning of an execution of the framework.
	ctx := testlogging.ContextWithLevel(t, testlogging.LevelInfo)
	_, err := eng.ExecAction(ctx, engine.GCActionKey, nil)

	require.NoError(t, err)
```

chaitalisg: Verified that the suggestion works.

julio-lopez (approved):

PR #3945, opened 2024-06-27
test(general): set cache size limits for the repository under test used in robustness tests

chaitalisg:
The robustness tests set soft limits for content cache size and metadata cache size. This PR adds the hard limits for both the caches, thus reducing the memory required to run the robustness tests for long term.


tests/robustness/multiclient_test/framework/snapshotter.go
```diff
@@ -67,6 +67,15 @@ func (mcs *MultiClientSnapshotter) ConnectOrCreateRepo(repoPath string) error {
 	return err
 }
 
+// SetCacheSizeLimits sets the cache size limits for an existing repository
+// the repository server is connected to.
+func (mcs *MultiClientSnapshotter) SetCacheSizeLimits(cacheArgs string) error {
```
julio-lopez, line 72: What is `cacheArgs`? it seems that an arbitrary flag can be passed.

tests/robustness/multiclient_test/framework/harness.go
```diff
@@ -145,6 +145,13 @@ func (th *TestHarness) getSnapshotter() bool {
 		return false
 	}
 
+	contentCacheLimitMB := "--content-cache-size-limit-mb=500"
+	metadataCacheLimitMB := "--metadata-cache-size-limit-mb=500"
+	if err = s.SetCacheSizeLimits(contentCacheLimitMB + metadataCacheLimitMB); err != nil {
```
julio-lopez, line 150: Does this work? it seems to me that this will always return an error.

tests/robustness/multiclient_test/framework/snapshotter.go
chaitalisg: Added restrictions for allowed flags

tests/robustness/multiclient_test/framework/harness.go
chaitalisg: Updated the code

tests/robustness/multiclient_test/framework/harness.go
```diff
@@ -22,8 +22,12 @@ import (
 )
 
 const (
-	dataSubPath     = "robustness-data"
-	metadataSubPath = "robustness-metadata"
+	dataSubPath              = "robustness-data"
+	metadataSubPath          = "robustness-metadata"
+	contentCacheLimitMBFlag  = "--content-cache-size-limit-mb"
+	contentCacheLimitMB      = "500"
```
chaitalisg, line 28: Use int instead of string

chaitalisg: done

tests/robustness/multiclient_test/framework/snapshotter.go
```diff
@@ -67,6 +74,20 @@ func (mcs *MultiClientSnapshotter) ConnectOrCreateRepo(repoPath string) error {
...
+// SetCacheSizeLimits sets the cache size limits for an existing repository
+// the repository server is connected to.
+// Allowed flags: content-cache-size-limit-mb, metadata-cache-size-limit-mb.
+func (mcs *MultiClientSnapshotter) SetCacheSizeLimits(cacheFlag, cacheFlagValue string) error {
+	if !strings.Contains(cacheFlag, contentCacheLimitMBFlag) && !strings.Contains(cacheFlag, metadataCacheLimitMBFlag) {
+		return ErrUnsupportedFlagsForCacheSet
+	}
+
+	_, _, err := mcs.server.Run("cache", "set",
+		cacheFlag, cacheFlagValue)
```
julio-lopez, line 86: Convert `cacheSize` using `strconv.Itoa`

```suggestion
		cacheFlag, strconv.Itoa(cacheSize))
```

tests/robustness/multiclient_test/framework/harness.go
```diff
@@ -145,6 +149,16 @@ func (th *TestHarness) getSnapshotter() bool {
 		return false
 	}
 
+	// Set size limits for content cache and metadata cache for repository under test.
```
julio-lopez, line 152: Can the cache limits be set in `ConnectOrCreateRepo` ?

julio-lopez (commented):
There are various CI failures.

tests/robustness/multiclient_test/framework/harness.go
chaitalisg: I thought about it. Doing so will require changes to the function signature, i.e. passing the cache set flags. I decided not to do it in order to avoid fallout from the function being used in other places.

tests/robustness/multiclient_test/framework/snapshotter.go
```diff
@@ -67,6 +74,20 @@ func (mcs *MultiClientSnapshotter) ConnectOrCreateRepo(repoPath string) error {
 	return err
 }
 
+// SetCacheSizeLimits sets the cache size limits for an existing repository
+// the repository server is connected to.
+// Allowed flags: content-cache-size-limit-mb, metadata-cache-size-limit-mb.
+func (mcs *MultiClientSnapshotter) SetCacheSizeLimits(cacheFlag, cacheFlagValue string) error {
```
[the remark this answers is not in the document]
chaitalisg: Tried locally

chaitalisg:
> Can you rebase this PR on top of master?
> 
> Also, there are various CI failures.

The CI failures are unrelated to the PR. All of my open PRs are showing the same errors.

tests/robustness/multiclient_test/framework/snapshotter.go
```diff
@@ -67,6 +74,20 @@ func (mcs *MultiClientSnapshotter) ConnectOrCreateRepo(repoPath string) error {
 	return err
 }
 
+// SetCacheSizeLimits sets the cache size limits for an existing repository
+// the repository server is connected to.
+// Allowed flags: content-cache-size-limit-mb, metadata-cache-size-limit-mb.
+func (mcs *MultiClientSnapshotter) SetCacheSizeLimits(cacheFlag string, cacheSize int) error {
```
redgoat650, line 80: does this need to be exported? it seems to only be consumed by `package framework`?

tests/robustness/multiclient_test/framework/snapshotter.go
```diff
@@ -67,6 +74,20 @@ func (mcs *MultiClientSnapshotter) ConnectOrCreateRepo(repoPath string) error {
...
+	if !strings.Contains(cacheFlag, contentCacheLimitMBFlag) && !strings.Contains(cacheFlag, metadataCacheLimitMBFlag) {
+		return ErrUnsupportedFlagsForCacheSet
+	}
```
redgoat650, line 83: If we wanted to avoid the need for such validation, we could perhaps split it into two methods
- `SetContentCacheSizeLimit`
- `SetMetadataCacheSizeLimit`

that way we can easily see what we're trying to change without requiring the caller to pass in a particular string

redgoat650: by doing so, we could also move
```
contentCacheLimitMBFlag  = "--content-cache-size-limit-mb"
metadataCacheLimitMBFlag = "--metadata-cache-size-limit-mb"
```

to this file, closer to where they're being used

julio-lopez: That seems like a good suggestion.
Also, the approach in this PR is different than the one used for the metadata repo. Can they be unified for consistency?

chaitalisg: @redgoat650  good point! I will implement the feedback.

@julio-lopez  I did think about using the same approach in the beginning. However, the repos are created in two different ways. The repo under test is created/acted upon using kopia cli and the metadata repo is created/acted upon by directly calling the go functions underneath. Unifying the approach to setting cache size would mean unifying the approach to rest of the interactions and updating any other code that depends on the current interface. That would be a much larger task and out of scope for this PR. 
What do you think?

tests/robustness/multiclient_test/framework/snapshotter.go
chaitalisg: you are correct, updated.

tests/robustness/multiclient_test/framework/snapshotter.go
```diff
@@ -16,6 +17,16 @@ import (
...
+const (
+	contentCacheLimitMBFlag  = "--content-cache-size-limit-mb"
+	metadataCacheLimitMBFlag = "--metadata-cache-size-limit-mb"
+)
+
+// ErrUnsupportedFlagsForCacheSet is returned when an unsupported flag
+// is used with cache set command on the repository under test
+// in multiclient robustness test.
+var ErrUnsupportedFlagsForCacheSet = errors.New("unsupported flag used for cache set")
+
```
redgoat650, line 29: This error is unused

redgoat650 (approved):
lgtm, just need to remove unused error

tests/robustness/multiclient_test/framework/snapshotter.go
```diff
@@ -67,6 +72,23 @@ func (mcs *MultiClientSnapshotter) ConnectOrCreateRepo(repoPath string) error {
...
+	return err
+}
+
+// setMetadataCacheSizeLimit sets the metadata cache size limits for an existing repository, the repository server is connected to.
+func (mcs *MultiClientSnapshotter) setMetadataCacheSizeLimit(metadataCacheSize int) error {
+	_, _, err := mcs.server.Run("cache", "set",
+		metadataCacheLimitMBFlag, strconv.Itoa(metadataCacheSize))
+
+	return err
+}
```
redgoat650, line 90: Actually one more quick question: it looks like both flags can be passed simultaneously to the `cache set` call, is that right? Perhaps we can simplify this even more by merging these two functions into `setCacheSizeLimit(content, metadata int) error`? If a non-negative integer is passed for one or both arguments, we can add the appropriate flag(s) and only need to call a single `cache set` command. Thoughts?

tests/robustness/multiclient_test/framework/snapshotter.go
```diff
@@ -67,6 +72,23 @@ func (mcs *MultiClientSnapshotter) ConnectOrCreateRepo(repoPath string) error {
 	return err
 }
 
+// setCacheSizeLimits sets the content cache size limits for an existing repository
+// the repository server is connected to.
+func (mcs *MultiClientSnapshotter) setContentCacheSizeLimit(contentCacheSize int) error {
```
redgoat650, line 77: nit: perhaps we should rename these functions to include a hint on what the integer value argument represents: `setContentCacheSizeLimitMB` and/or rename the argument name to include the MB unit

tests/robustness/multiclient_test/framework/snapshotter.go
chaitalisg: yes, both flags can be passed simultaneously according to https://kopia.io/docs/reference/command-line/advanced/cache-set/
So, just one command that sets both the cache size limits by default? One could argue, two separate functions may promote reuse later as compared to a single function.

tests/robustness/multiclient_test/framework/snapshotter.go
chaitalisg: done

tests/robustness/multiclient_test/framework/snapshotter.go
redgoat650: I think there's some personal preference here; either way probably works fine. The proposal would accommodate setting one or both, e.g. `setCacheSizeLimit(100, -1)` would only set the content limit to 100 MB (hence the "non-negative" choice of words above). But some might argue that's less readable as well, and sometimes minor performance optimizations can be neglected in favor of better readability. Proceed as desired, my prior approval still stands!

julio-lopez (approved):

PR #3952, opened 2024-07-02
test(general): set cache size limits for metadata repo

chaitalisg:
The soft limits are set for content cache and metadata cache. This may cause the cache to bloat beyond expectations. Setting hard limits for the size would keep the cache bloat in check, thus reducing the memory needed to required to run the robustness tests over a long time.


tests/tools/kopiaclient/kopiaclient.go
```diff
@@ -73,6 +75,43 @@ func (kc *KopiaClient) CreateOrConnectRepo(ctx context.Context, repoDir, bucketN
...
+	}
+
+	if iErr := repo.Connect(ctx, kc.configPath, st, kc.pw, &repo.ConnectOptions{}); iErr != nil {
+		return errors.Wrap(iErr, "error connecting to repository")
+	}
+
+	return errors.Wrap(err, "unable to open repository")
+}
+
+// SetCacheLimits connects to an existing one if possible.
```
redgoat650, line 92: I think this comment needs updating

tests/tools/kopiaclient/kopiaclient.go
```diff
@@ -73,6 +75,43 @@ func (kc *KopiaClient) CreateOrConnectRepo(ctx context.Context, repoDir, bucketN
...
+
+	cacheOptsObtained, err := repo.GetCachingOptions(ctx, kc.configPath)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("content cache size" + strconv.FormatInt(cacheOptsObtained.ContentCacheSizeLimitBytes, 10))
+	log.Printf("metadata cache size " + strconv.FormatInt(cacheOptsObtained.MetadataCacheSizeLimitBytes, 10))
+
+	return errors.Wrap(err, "unable to set the cache repository")
```
redgoat650, line 112: It doesn't look like this `Wrap` does anything, since all non-nil errors are returned earlier. We can probably simplify to:
```suggestion
	return nil
```

tests/tools/kopiaclient/kopiaclient.go
```diff
@@ -73,6 +75,43 @@ func (kc *KopiaClient) CreateOrConnectRepo(ctx context.Context, repoDir, bucketN
...
+	st, err := kc.getStorage(ctx, repoDir, bucketName)
+	if err != nil {
+		return err
+	}
+
+	if iErr := repo.Connect(ctx, kc.configPath, st, kc.pw, &repo.ConnectOptions{}); iErr != nil {
+		return errors.Wrap(iErr, "error connecting to repository")
+	}
+
+	return errors.Wrap(err, "unable to open repository")
```
redgoat650, line 89: Same suggestion here - wrap is unnecessary
```suggestion
	return nil
```

tests/robustness/snapmeta/kopia_persister_light.go
```diff
@@ -45,6 +46,22 @@ func (kpl *KopiaPersisterLight) ConnectOrCreateRepo(repoPath string) error {
 	return kpl.kc.CreateOrConnectRepo(context.Background(), repoPath, bucketName)
 }
 
+// SetCacheLimits sets to an existing one if possible.
+func (kpl *KopiaPersisterLight) SetCacheLimits(repoPath string, cacheOpts *content.CachingOptions) error {
+	bucketName := os.Getenv(S3BucketNameEnvKey)
+	err := kpl.kc.ConnectRepo(context.Background(), repoPath, bucketName)
```
redgoat650, line 52: It looks like we're connecting to the repository three times in a row. Once in `getPersister`, then again here, then immediately after in `kc.SetCacheLimits`. Was that intended?

redgoat650 (commented):
Seems good, left a few suggestions

tests/robustness/snapmeta/kopia_persister_light.go
chaitalisg: good point.  I removed two redundant connects.

tests/tools/kopiaclient/kopiaclient.go
chaitalisg: done

tests/tools/kopiaclient/kopiaclient.go
chaitalisg: done

tests/tools/kopiaclient/kopiaclient.go
chaitalisg: done

redgoat650 (approved):

tests/tools/kopiaclient/kopiaclient.go
```diff
@@ -73,6 +75,24 @@ func (kc *KopiaClient) CreateOrConnectRepo(ctx context.Context, repoDir, bucketN
...
+		return err
+	}
+
+	cacheOptsObtained, err := repo.GetCachingOptions(ctx, kc.configPath)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("content cache size" + strconv.FormatInt(cacheOptsObtained.ContentCacheSizeLimitBytes, 10))
+	log.Printf("metadata cache size " + strconv.FormatInt(cacheOptsObtained.MetadataCacheSizeLimitBytes, 10))
```
julio-lopez, line 91: - The format string in `log.Printf` is missing a format specifier. However, `log.Print` would work here.
- The `strconv.FormatInt` call is not needed.

```suggestion
	log.Print("content cache size:", cacheOptsObtained.ContentCacheSizeLimitBytes)
	log.Printf("metadata cache size:", cacheOptsObtained.MetadataCacheSizeLimitBytes)
```

julio-lopez (commented):
It should be OK to merge after CI passes

tests/tools/kopiaclient/kopiaclient.go
```diff
@@ -11,6 +11,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
```
julio-lopez, line 14:
```suggestion
```

tests/tools/kopiaclient/kopiaclient.go
```diff
@@ -73,6 +75,24 @@ func (kc *KopiaClient) CreateOrConnectRepo(ctx context.Context, repoDir, bucketN
...
+		return err
+	}
+
+	cacheOptsObtained, err := repo.GetCachingOptions(ctx, kc.configPath)
+	if err != nil {
+		return err
+	}
+
+	log.Print("content cache size:", cacheOptsObtained.ContentCacheSizeLimitBytes)
+	log.Print("metadata cache size:", cacheOptsObtained.MetadataCacheSizeLimitBytes)
```
julio-lopez, line 91:
```suggestion
	log.Println("content cache size:", cacheOptsObtained.ContentCacheSizeLimitBytes)
	log.Println("metadata cache size:", cacheOptsObtained.MetadataCacheSizeLimitBytes)
```

julio-lopez (approved):

PR #3957, opened 2024-07-03
chore(deps): group observability dependencies

julio-lopez:
In the past, otel dependencies have introduced non-trivial depencency conflicts.


julio-lopez:
@jkowalski FYI

PR #3963, opened 2024-07-03
test(general): add delete random snapshot action to robustness tests

chaitalisg:
The robustness framework does not cover delete random snapshot action at the moment. 

With this change, the delete-random-snapshot action will result in no-op when there is one snapshots left.


tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -96,6 +96,8 @@ func TestOneLargeFile(t *testing.T) {
 
 		_, err = eng.ExecAction(ctx, engine.RestoreSnapshotActionKey, snapOut)
 		require.NoError(t, err)
+
+		tryDeleteAction(ctx, t, engine.DeleteRandomSnapshotActionKey, nil)
```
chaitalisg, line 100: Placed here so that the latest snapshot can be considered for deletion but the restore does not fail because of it.

redgoat650: I'm not sure I understand - why is the delete-snapshot being added to this test in particular? And why only this test?

tests/robustness/engine/action.go
```diff
@@ -327,6 +336,7 @@ func (e *Engine) getSnapIDOptOrRandLive(opts map[string]string) (snapID string,
 
 	snapIDList := e.Checker.GetLiveSnapIDs()
 	if len(snapIDList) == 0 {
+		log.Printf("No snapshots available for deletion\n")
```
julio-lopez, line 339: nit:
```suggestion
		log.Println("No snapshots available for deletion")
```

tests/robustness/engine/action.go
```diff
@@ -208,6 +208,15 @@ func restoreSnapshotAction(ctx context.Context, e *Engine, opts map[string]strin
 }
 
 func deleteRandomSnapshotAction(ctx context.Context, e *Engine, opts map[string]string, l *LogEntry) (out map[string]string, err error) {
+	// Do not delete snapshot when it is the only available snapshot.
+	// This will ensure that the repository under test in robustness tests
+	// grows over long term.
+	snapIDList := e.Checker.GetLiveSnapIDs()
+	if len(snapIDList) <= 1 {
+		log.Printf("No snapshots available for deletion\n")
```
julio-lopez, line 216: nit
```suggestion
		log.Println("No snapshots available for deletion")

```

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -227,20 +229,23 @@ func tryRandomAction(ctx context.Context, t *testing.T, opts engine.ActionOpts)
...
 	}
 	require.Contains(t, eligibleActionsList, action)
 
 	_, err := eng.ExecAction(ctx, action, actionOpts)
-	// Ignore the dir-not-found error, wrapped as no-op error.
+	// Ignore the dir-not-found error or no-snapshots-to-delete error
+	// wrapped as no-op error.
 	if errors.Is(err, robustness.ErrNoOp) {
-		t.Log("Delete action resulted in no-op")
+		t.Log("Delete action ", action, " resulted in no-op")
```
julio-lopez, line 248: nit: `Logf` is more appropriate in this case

```suggestion
		t.Logf("Delete action '%s' resulted in no-op", action)
```

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: Followup question: where else is the "DeleteRandomSnapshot" action executed?

chaitalisg: The reason for adding it to this test was to add some kind of delete action to this test, it did not execute any delete actions before this. 
On second thought, I will add the action to all the tests to follow the established pattern.

redgoat650: what are the implications of adding a delete action to every test? If we need to "start over" with a fresh repo-under-test, does that mean it'll only keep a single snapshot at steady state?

Asked another way: if every test creates one snapshot then deletes one snapshot (except when there's only one), we'll always be ending the test with only a single snapshot left in the repo, right?

chaitalisg: Yes, that was the reason behind adding the action to just one test.
I think the way to move forward here is to
1. call delete random sub dir and delete dir content in every test 
2. call delete snapshot action in a couple of tests, not all so that the repo under test can grow in size long term

julio-lopez: What about putting this action in its own separate test? would that work?

chaitalisg: done

tests/robustness/multiclient_test/multiclient_test.go
```diff
@@ -205,6 +205,17 @@ func TestMaintenanceAction(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestDeleteRandomSnapshotAction(t *testing.T) {
+	const numClients = 4
```
julio-lopez, line 209: Does this delete 4 snapshots at once in the common case? Isn't that too agressive?

chaitalisg: yes, reduced to 1.

tests/robustness/multiclient_test/multiclient_test.go
julio-lopez: @redgoat650 Does this work for you?
Merging now, we can iterate if needed.

julio-lopez (approved):

PR #3974, opened 2024-07-11
feat(cli): add server user set-password-hash command

julio-lopez:
Objectives:
- Facilitate the generation of valid password hashes that can be used with the `server user --user-password` CLI command.
- Encapsulate implementation details of password hashing in the `user` package.

Adds a new `server user hash-password` CLI command to generate the hash from a supplied password.

Modifies the `server user set/add --user-password-hash` CLI command to accept the password hash generated using the `hash-password` command.

Adds `GetNewProfile(ctx, rep, username)` helper to move implementation details to the `user` package.

Includes CLI and unit tests.

Cleans up and removes unused functions.

### DIFF
--- a/cli/command_user_hash_password.go
+++ b/cli/command_user_hash_password.go
@@ -1,0 +1,52 @@
+package cli
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/internal/user"
+	"github.com/kopia/kopia/repo"
+)
+
+type commandServerUserHashPassword struct {
+	password string
+
+	out textOutput
+}
+
+func (c *commandServerUserHashPassword) setup(svc appServices, parent commandParent) {
+	cmd := parent.Command("hash-password", "Hash a user password that can be passed to the 'server user add/set' command").Alias("hash")
+
+	cmd.Flag("user-password", "Password").StringVar(&c.password)
+
+	cmd.Action(svc.repositoryWriterAction(c.runServerUserHashPassword))
+
+	c.out.setup(svc)
+}
+
+// The current implementation does not require a connected repository, thus the
+// RepositoryWriter parameter is not used. Future implementations will need a
+// connected repository. To avoid a future incompatible change where the
+// 'hash-password' command stops working without a connected repository, then
+// the a connected repository is required now.
+func (c *commandServerUserHashPassword) runServerUserHashPassword(ctx context.Context, _ repo.RepositoryWriter) error {
+	if c.password == "" {
+		// when password hash is empty, ask for password
+		pwd, err := askConfirmPass(c.out.stdout(), "Enter password to hash: ")
+		if err != nil {
+			return errors.Wrap(err, "error getting password")
+		}
+
+		c.password = pwd
+	}
+
+	h, err := user.HashPassword(c.password)
+	if err != nil {
+		return errors.Wrap(err, "hashing password")
+	}
+
+	c.out.printStdout("%s\n", h)
+
+	return nil
+}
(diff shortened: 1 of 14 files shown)


cli/command_user_hash_password.go
```diff
@@ -0,0 +1,52 @@
...
+	cmd.Action(svc.repositoryWriterAction(c.runServerUserHashPassword))
+
+	c.out.setup(svc)
+}
+
+// The current implementation does not require a connected repository, thus the
+// RepositoryWriter parameter is not used. Future implementations will need a
+// connected repository. To avoid a future incompatible change where the
+// 'hash-password' command stops working without a connected repository, then
+// the a connected repository is required now.
```
redgoat650, line 32: nit: comment wording, perhaps: "...~then the~ a connected repository is required now" --> "...a connected repository is required now"?

internal/user/hash_password_test.go
```diff
@@ -40,3 +33,54 @@ func TestHashPassword_encoding(t *testing.T) {
...
+				PasswordHashVersion: defaultPasswordHashVersion,
+				// PasswordHash with invalid length
+				PasswordHash: []byte{'z', 'a'},
+			},
+			expectError: true,
+		},
+		{
+			ph: passwordHash{
+				PasswordHashVersion: defaultPasswordHashVersion,
+				// PasswordHash with invalid length
```
redgoat650, line 69: probable copy-pasted comment here: this test case has the correct hash length

internal/user/user_manager.go
```diff
@@ -21,6 +21,9 @@ const UsernameAtHostnameLabel = "username"
 // ErrUserNotFound is returned to indicate that a user was not found in the system.
 var ErrUserNotFound = errors.New("user not found")
 
+// ErrUserAlreadyExists indicates that a user already exist in the system when attempting to create a new one.
+var ErrUserAlreadyExists = errors.New("user already exists found")
```
redgoat650, line 25: Was this meant to be just `"user already exists"`?
(`"already exists found"` sounds kinda weird)

cli/command_user_hash_password_test.go
```diff
@@ -0,0 +1,76 @@
...
+	require.Len(t, out, 1)
+
+	passwordHash := out[0]
+	require.NotEmpty(t, passwordHash)
+
+	// attempt to create a user with a bad password hash
+	e.RunAndExpectFailure(t, "server", "users", "add", userFull, "--user-password-hash", "bad-base64")
+
+	// create a new user with and set the password using the password hash
+	e.RunAndExpectSuccess(t, "server", "users", "add", userFull, "--user-password-hash", passwordHash)
```
redgoat650, line 43: Do we want to test `server users set` also? Perhaps we can update the password hash and show fail/success when the same user connects with old/new passwords?

redgoat650 (commented):
A few nits, but looks good overall

internal/user/user_manager.go
julio-lopez: yeap, probably edit error.

cli/command_user_hash_password_test.go
julio-lopez: @redgoat650 PTAL.
I added a success case after changing the user password (via a hash).
I'll owe the failure cases, those surfaced a minor issue with a resource not being released when the client command fails to execute. It may be similar to the issue in #3980 but on the client side. 3980 is on the server side, it was surfaced when fixing a test.

redgoat650 (approved):
Thanks @julio-lopez, +1 from me

julio-lopez:
@bathina2 FYI

jkowalski:
:shipit:

jkowalski (approved):

Issue #2785, opened 2023-02-27
Import policy from json file


vibl:
From what I read in the issues and forum, it doesn't seem possible to import a policy from an exported json file.

Can you confirm that?

If so, is it a planned feature?

If not, could it be?

It does seem a basic feature to me for streamlining backup management in multiple machines, but maybe I'm missing something.

DerivativeOfLog7:
Yeah, the lack of a way to import a preexisting policy seems a big omission to me too

taltamir:
oddly enough, the docs _claim_ there is a policy export feature.
https://kopia.io/docs/reference/command-line/common/policy-export/

but this command does not actually exist.
`kopia: error: expected command but got "export", try --help`
the closest sub-command is `policy show`

```
$ kopia policy --help

usage: kopia policy <command> [<args> ...]
Commands to manipulate snapshotting policies.

Flags:
      --[no-]help             Show context-sensitive help (also try --help-long
                              and --help-man).
      --[no-]version          Show application version.
      --log-file=LOG-FILE     Override log file.
      --log-dir="/home/t/.cache/kopia"  
                              Directory where log files should be written.
                              ($KOPIA_LOG_DIR)
      --log-level=info        Console log level
      --file-log-level=debug  File log level
      --[no-]help-full        Show help for all commands, including hidden
      --config-file="repository.config"  
                              Specify the config file to use
                              ($KOPIA_CONFIG_PATH)
  -p, --password=PASSWORD     Repository password. ($KOPIA_PASSWORD)
      --[no-]persist-credentials  
                              Persist credentials
                              ($KOPIA_PERSIST_CREDENTIALS_ON_CONNECT)
      --[no-]use-keyring      Use Gnome Keyring for storing repository password.
                              ($KOPIA_USE_KEYRING)

Subcommands:
  policy
    edit [<flags>] [<target>...]
    list [<flags>]
    delete [<flags>] [<target>...]
    set [<flags>] [<target>...]
    show [<flags>] [<target>...]
```

derivativeoflog7:
> oddly enough, the docs _claim_ there is a policy export feature. https://kopia.io/docs/reference/command-line/common/policy-export/ `$ kopia policy export [target]` but this command does not actually exist. `kopia: error: expected command but got "export", try --help`
> 
> the closest sub-command is `policy show`
> 
> ```
> $ kopia policy --help
> 
> usage: kopia policy <command> [<args> ...]
> Commands to manipulate snapshotting policies.
[...]

PR #4020, opened 2024-08-02
feat(cli): add policy import/export commands to im-/export policies from/to json

### LINKED ISSUE #2432: Declarative configuration
Hello,

I'm trying to write a NixOS service for Kopia, and hoping I could get some advice here on it. At the moment, I'm thinking of storing the quick connect token with the password, and then having separate parameters for other connection options. Then the "activation script" (runs once on enabling) can execute these steps:
1. run `kopia repository disconnect` (clean up any prior state)
2. create config file at `SOME_PATH` with connection options
3. run `kopia repository connect --config-file=$SOME_PATH from-config --token-file=/run/secrets/kopia-token`

I haven't tried this, but from the docs this looks like it could work. For actually performing backups, some systemd services and timers with `kopia snapshot --config-file=$SOME_PATH` seem like all that's required.

One thing I haven't got any idea about is the policy. I see that you can get the current policy as JSON with `kopia policy show --json`, but I don't see any easy way to import a policy/set it declaratively? Advice here would be appreciated.

blenderfreaky:
This PR adds a `policy import` and `policy export` command to the cli.

The commands import and export policies from and to json files.

### Syntax

```
$ kopia policy export [--json-indent] [--to-file=file] [--overwrite] [--global] [<target>]
$ kopia policy export --json-indent | tee policies.json
$ kopia policy export --to-file policies.json "foo@bar:/home/foobar"
```
Exports policies as json, optionally pretty-printed.
Prints to stdout if no file is specified.
Outputs all policies if `--global` and `[<target>]` are unspecified.

```
$ kopia policy import [--from-file=file]
$ cat policies.json | kopia policy import
$ kopia policy import --from-file policy.json
```
Imports policies from json, reads from stdin if no file is specified.

In-/Output is in the format
```json
{
  "(global)": {
    "retention": {
      "keepLatest": 10,
      "keepHourly": 48,
      ...
    }
  },
  "foo@bar:/home/foobar": {
    ...
  }
}
```

Would close #2432 and #2785


tecosaur:
Bump @jkowalski @julio-lopez :slightly_smiling_face:

I (and it seems a bunch of other people), are eagerly looking forward to this functionality in order to make process on automated configuration.

If you could find the time to take a look at this work, that would be brilliant and much appreciated.

cli/command_policy_import.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		input = bufio.NewReader(file)
+	} else {
+		input = c.stdin
+	}
+
+	policies := make(map[string]*policy.Policy)
+	d := json.NewDecoder(input)
+	d.DisallowUnknownFields()
```
jkowalski, line 51: that should be flag-controlled, it may make sense to allow partial import ignoring unknown fields.

cli/command_policy_import.go
```diff
@@ -0,0 +1,91 @@
...
+			if _, ok := targetLimit[target]; !ok {
+				continue
+			}
+		}
+
+		if err := policy.SetPolicy(ctx, rep, target, newPolicy); err != nil {
+			return errors.Wrapf(err, "can't save policy for %v", target)
+		}
+	}
+
```
jkowalski, line 89: (optional) should we have an option to remove policies that can't be found in the input JSON (basically to replace all policies in a repo) ?

cli/command_policy_import.go
```diff
@@ -0,0 +1,91 @@
...
+			return err
+		}
+
+		for _, target := range targetLimitSrc {
+			targetLimit[target] = struct{}{}
+		}
+	}
+
+	for ts, newPolicy := range policies {
+		target, err := snapshot.ParseSourceInfo(ts, rep.ClientOptions().Hostname, rep.ClientOptions().Username)
```
jkowalski, line 74: not sure this works for `(global)`, in any case we need to add a test for it.

cli/command_policy_import.go
```diff
@@ -0,0 +1,91 @@
...
+
+	stdin io.Reader
+}
+
+func (c *commandPolicyImport) setup(svc appServices, parent commandParent) {
+	cmd := parent.Command("import", "Imports policies from a specified file, or stdin if no file is specified.")
+	cmd.Flag("from-file", "File path to import from").StringVar(&c.filePath)
+
+	c.policyTargetFlags.setup(cmd)
+	c.stdin = svc.stdin()
```
jkowalski, line 28: don't invoke method on `svc` in the `setup` instead of this, store `svc` in the struct and invoke `stdin()` in the `run()` method.

cli/command_policy_import.go
```diff
@@ -0,0 +1,91 @@
...
+	cmd.Action(svc.repositoryWriterAction(c.run))
+}
+
+func (c *commandPolicyImport) run(ctx context.Context, rep repo.RepositoryWriter) error {
+	var input io.Reader
+	var err error
+
+	if c.filePath != "" {
+		file, err := os.Open(c.filePath)
+
```
jkowalski, line 39: it's missing defer file.Close()

cli/command_policy_import.go
```diff
@@ -0,0 +1,91 @@
...
+	d.DisallowUnknownFields()
+
+	err = d.Decode(&policies)
+
+	if err != nil {
+		return errors.Wrap(err, "unable to decode policy file as valid json")
+	}
+
+	//var targetLimit []snapshot.SourceInfo
+	targetLimit := make(map[snapshot.SourceInfo]struct{})
```
jkowalski, line 60: this is slightly hard to read, how about removing this map and instead add `shouldImportSource(src)` method that is called inside the loop?

cli/command_policy_export.go
```diff
@@ -0,0 +1,127 @@
...
+			if os.IsExist(err) {
+				return errors.Wrap(err, "file already exists and overwrite flag is not set")
+			}
+		}
+
+		if err != nil {
+			return errors.Wrap(err, "error opening file to write to")
+		}
+
+		wr := bufio.NewWriter(file)
```
jkowalski, line 61: bufio is generally not needed, writing straight to `file` is fine. This avoids all the flush/sync below.

cli/command_policy_export.go
```diff
@@ -0,0 +1,127 @@
...
+		toWrite, err = json.MarshalIndent(policies, "", "  ")
+	} else {
+		toWrite, err = json.Marshal(policies)
+	}
+
+	if err != nil {
+		panic("error serializing JSON, that should not happen: " + err.Error())
+	}
+
+	fmt.Fprintf(output, "%s", toWrite) //nolint:errcheck
```
jkowalski, line 124: instead of ignoring the error, just return it

cli/command_policy_export.go
```diff
@@ -0,0 +1,127 @@
...
+	var toWrite []byte
+
+	if c.jsonIndent {
+		toWrite, err = json.MarshalIndent(policies, "", "  ")
+	} else {
+		toWrite, err = json.Marshal(policies)
+	}
+
+	if err != nil {
+		panic("error serializing JSON, that should not happen: " + err.Error())
```
jkowalski, line 121: `impossible.PanicOnError()` instead

jkowalski:
This functionality looks great, can you add a quick e2e test and address the comments I've put in, then we can merge.

blenderfreaky:
addressed most of your comments, gonna add tests and a remove/replace flag tomorrow

cli/command_policy_import.go
blenderfreaky: added an `--allow-unknown-fields` flag, let me know if this should be reversed (`--disallow-...` as flag)

cli/command_policy_import.go
blenderfreaky: implemented as `--delete-other-policies` flag

blenderfreaky:
added tests and (hopefully) addressed all the comments

jkowalski:
this is looking great, still have some linter errors, you can run `make lint` locally to see the linter errors.

blenderfreaky:
should be fixed now

I used `defer file.Close() //nolint:errcheck` in two places, LMK if that's acceptable

blenderfreaky:
it fails on windows only
I'm on linux so this may be tricky to debug

I assume it's due to the policy test target being specified as `"user@host:" + path` with the path returned by testutil using backslashes. If I'm right, this *should* fix it.

blenderfreaky:
gonna try to address the code coverage as well

blenderfreaky:
should be good now (hopefully)

jkowalski (approved):
LGTM! Thanks for the contribution!

blenderfreaky:
The windows test still seem to be failing, seems to be related to the paths

Gonna try to spin up a VM to debug

jkowalski:
policies in Windows use backslashes for path names `jarek@windowsbox:C:\Users\Jarek\somedir`

blenderfreaky:
it failed without `filepath.ToSlash` as well, so that can't be the only issue

jkowalski:
I have a Windows dev box, let me try it...

jkowalski:
removing ToSlash fixed the issue on Windows.

blenderfreaky:
Interesting that the two ToSlash calls in the import tests don't break anything

jkowalski:
i think they slash-separated policies will be imported/exported just fine, but will not really take effect on Windows boxes (since it uses backslashes). We should fix that perhaps at some point.

tecosaur:
It's brilliant to see this happen, thanks for getting this done blenderfreaky and Jarek!

julio-lopez:
- Fixes #2785

Issue #1636, opened 2022-01-01
Pull-based backups ?


nautik1:
Hello,

I would like to replace an old backup strategy with a tool like kopia, but I am unsure if my use case can be handled.

I have 2 servers and would like to use one of them as a "backup server" that connects to the other remote server and backup its files into a local repository.
In other words I would like to have a "pull-based" backup strategy while I see kopia's main use case is "push mode", where kopia gets local files and pushes them into a remote destination.

In my case, only the backup server can connect to the other server and **not the other way around**.

Looking at the documentation and playing with the cli, what I could see is:

- mount the remote filesystem using something like sshfs into the backup server, then run kopia to backup files "locally"; this is already possible I guess, but I have the feeling it might not be optimal as kopia is used to have the files to be locally accessible. Or is it?
- be able to specify a remote target when running `kopia snapshot create [source]`, where `source` includes the host to connect to?
- have a `repository sync-from <sftp|from-config>` command which would be the opposite of the existing `repository sync-to`; it would allow creating repository & snapshots on the backed up server, then pull the repository from the backup server.

Feedback about the above is welcome. Would the sshfs solution already be good enough, or would the other 2 ideas be relevant feature-request ideas?
Would anyone see another solution that I missed?
Would you see this topic relevant to address in the documentation (if I haven't missed it and it's already there)?

Have a good day :)

jkowalski:
In principle, any filesystem-like source can be backed up - we currently support snapshotting local files and snapshotting other snapshots (this is used during the migration from one repository to another).

Snapshotting SFTP sources (and possibly others, such as s3 buckets, etc.) should  be possible but will probably require major effort, so for now `sshfs` is recommended.

flokli:
It's not only sftp - it would be really nice if kopia could be used to back up remote destinations that don't provide a shell, without having to mount things locally.

Think of backing up files in a WebDAV endpoint, or a Google Drive. Could we use `rclone` to back up /from/ any rclone remote?

Mirodin:
I would love for a pull feature to be available some day. My server is the backup target for my whole family and most are not very tech savy so I do not trust their devices. This is why I would never give them write access to my server. Right now I have a homegrown rsync system that pulls backups via SSH from their machines. This works but is not very efficient spacewise since renamed files get synced again.

brechmos:
[...]

PR #4039, opened 2024-08-09
feat(repo): Add `sshfs` to docker container

jasonlessenich:
Hey, everyone :wave: 
I'm pretty new to Kopia & kind of have the same problem as described in https://github.com/kopia/kopia/issues/1636
sshfs would work, but it's a bit annoying getting this to work inside the docker container

This PR simply adds sshfs to the docker container, allowing the use of sshfs in before & after backup scripts

If my usecase is too niche, I can attempt building the container myself, but I think that there are some people, that would benefit from this


jasonlessenich:
Okay, this seems to be more complicated than just adding sshfs
Simply running sshfs results in 
```bash
fuse: device not found, try 'modprobe fuse' first
```

I'll take a look into this once I have some spare time

PR #4041, opened 2024-08-09
feat(repository): Stream loading of manifests

ashmrtn:
Rewrite the logic for loading manifests in the manifest manager to merge individual manifestEntry structs instead of loading all manifestEntry's and then deduping them. This can help reduce memory requirements for repositories that have many duplicate manifests (either due to deletions or concurrent compaction) by discarding entries as soon as possible

It may also require less memory because it can stream unzipped, partially deserialized manifestEntry's instead of needing to hold all manifestEntry's. However, it does still require enough memory to fully pull the content pieces being loaded into memory plus space for any unique manifests already loaded so far

Some benchmark results from the original code and a few variations of this new code. Benchmarks with `-Old-` are the original implementation, benchmarks with `-New-` hold a lock for the entire duration (e.x. lock/defer unlock) of the callback function that merges manifests, benchmarks with `-FGL-` hold a lock to find the previous version of the manifest and again when adding a manifest when executing the callback function. As the benchmark still deals with a relatively small manifest footprint (100 manifests of ~1KB) memory savings may grow as the number of manifests does

```
BenchmarkConcurrentCompactionLoading-New-10     30000     3134501 ns/op     1983854 B/op     7811 allocs/op
BenchmarkConcurrentCompactionLoading-FGL-10     30000     2976051 ns/op     1983797 B/op     7809 allocs/op
BenchmarkConcurrentCompactionLoading-Old-10     30000     3090741 ns/op     1995876 B/op     7863 allocs/op
```

### DIFF
--- a/repo/manifest/manifest_manager_test.go
+++ b/repo/manifest/manifest_manager_test.go
@@ -3,9 +3,11 @@ package manifest
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -311,8 +313,8 @@ type contentManagerOpts struct {
 	readOnly bool
 }
 
-func newContentManagerForTesting(ctx context.Context, t *testing.T, data blobtesting.DataMap, opts contentManagerOpts) contentManager {
-	t.Helper()
+func newContentManagerForTesting(ctx context.Context, tb testing.TB, data blobtesting.DataMap, opts contentManagerOpts) contentManager {
+	tb.Helper()
 
 	st := blobtesting.NewMapStorage(data, nil, nil)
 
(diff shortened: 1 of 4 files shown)


repo/manifest/manifest_manager_test.go
```diff
@@ -485,6 +488,108 @@ func TestManifestAutoCompactionWithReadOnly(t *testing.T) {
 	mgr, err = NewManager(ctx, bm, ManagerOptions{}, nil)
 	require.NoError(t, err, "getting other instance of manifest manager")
 
-	_, err = mgr.Find(ctx, map[string]string{"color": "red"})
+	entries, err := mgr.Find(ctx, map[string]string{"color": "red"})
 	require.NoError(t, err, "forcing reload of manifest manager")
+	assert.Len(t, entries, 100)
+}
+
+// BenchmarkConcurrentCompactionLoading benchmarks the runtime of loading
```
ashmrtn, line 496: I can remove this benchmark if folks would prefer not to keep benchmarks in the repo. Part of why it was committed was so I could cherry-pick it onto a branch with the original implementation for comparison

Issue #3581, opened 2024-01-24
Kopia failure during manifest compaction


pandeyabs:
Hello,

We use kopia as a storage engine for [corso](https://github.com/alcionai/corso). One of our customers is running into a backup failure, which points to a kopia error while building local index. Can someone please take a quick look and tell us what could be the underlying reason here? Any leads will help, we can carry forward the debug based off that. :) 

Just a quick note, we are running off a private [branch](github.com/alcionai/kopia@v0.12.2-0.20231205231702-863c24d6f8b1/) of kopia. It's not too far behind main, but just wanted to give a heads up.

`kopia backup: get policy tree: unable to get policies: unable to get parent policies: unable to find manifest for source corso@corso-host:base64-encoded-guid: error auto-compacting contents: unable to compact manifest contents: unable to write content: unable to write pack: error writing pack: error preparing data content: unable to build local index: maximum content length is too high: (packed 262228411, original 269897694, max 268435456)`

Stack

```

`github.com/kopia/kopia/repo/manifest.(*committedManifestManager).loadCommittedContentsLocked

/home/runner/go/pkg/mod/github.com/alcionai/kopia@v0.12.2-0.20231205231702-863c24d6f8b1/repo/manifest/committed_manifest_manager.go:184

github.com/kopia/kopia/repo/manifest.(*committedManifestManager).ensureInitializedLocked

/home/runner/go/pkg/mod/github.com/alcionai/kopia@v0.12.2-0.20231205231702-863c24d6f8b1/repo/manifest/committed_manifest_manager.go:311

github.com/kopia/kopia/repo/manifest.(*committedManifestManager).findCommittedEntries

/home/runner/go/pkg/mod/github.com/alcionai/kopia@v0.12.2-0.20231205231702-863c24d6f8b1/repo/manifest/committed_manifest_manager.go:75

github.com/kopia/kopia/repo/manifest.(*Manager).Find

/home/runner/go/pkg/mod/github.com/alcionai/kopia@v0.12.2-0.20231205231702-863c24d6f8b1/repo/manifest/manifest_manager.go:179

github.com/kopia/kopia/repo.(*directRepository).FindManifests

/home/runner/go/pkg/mod/github.com/alcionai/kopia@v0.12.2-0.20231205231702-863c24d6f8b1/repo/repository.go:225

github.com/kopia/kopia/snapshot/policy.GetPolicyHierarchy

[...]

Issue #3321, opened 2023-09-18
FR: Split manifest content from manifest metadata to reduce memory usage


ashmrtn:
#2956 helped reduce the memory usage when loading manifest data, but there's still cases where a repo that contains many backups (manifests) can lead to kopia using all memory on the host machine

To allow kopia to manage larger repos, it seems like it would be possible to push manifest content down a layer and have the manifest manager just load basic metadata about each manifest. Actual manifest content would be loaded on an as-needed basis

Internally, this could be implemented as storing manifest content data as separate kopia content pieces while keeping the manifest metadata (mostly) the way it is now. Manifest metadata would be expanded to contain a reference to the kopia content ID containing the manifest data

ashmrtn:
I have some stuff to finish off at the moment but I should be able to lend a hand on this in a bit if the proposal sounds reasonable

ashmrtn:
heh, well I guess my previous comment about working on this soon didn't age too well given that it's been almost a year 😅 At any rate, I do now actually have some time too look at this and I'd like to put some information down for discussion (also apologies in advance for the wall of text)

I wanted to call out some design decisions and potential problems that will need to be addressed for this feature. The things I’d like to call out explicitly for discussion are the following:

- what content prefix(es) to use for this new way of storing manifests
- garbage collection for orphaned or deleted manifest content pieces
- migration to the new way of storing manifests

The remainder of this comment summarizes the current manifest manager behavior (at least as I understand it) and then gives a sketch of a possible implementation of this feature along with some discussion of the above points

# Current implementation

Kopia currently stores manifest metadata (mod time, tags, ID, etc) and manifest data (additional information about snapshots or other JSON objects) together. All manifests are stored as content pieces with the `m` content prefix. When persisting manifests, multiple manifests can be written out in a single content piece by storing them in a slice. Only new or updated manifests are written out, so there may be multiple `m` prefixed content pieces in the kopia repository. When loading manifests, all `m` prefixed content pieces are retrieved and manifests are deduplicated based on their manifest ID and updated timestamp. All manifest metadata and manifest data is expected to reside in memory.

Deleting a manifest is accomplished by adding a tombstone marker to the set of manifests for the given manifest ID. The tombstone marker causes the logic for loading manifests to ignore the data for the manifest in question during GET requests.

[...]

PR #4062, opened 2024-08-20
feat(repository): Add level of indirection to manifest manager

ashmrtn:
This is a rough draft of #3321 meant to gather feedback about the approach in general before going more in-depth and fully writing out a solution

High-level points:
* adds a level of indirection to the manifest manager so that all manifest content is stored in content pieces separate from manifest metadata
* continues to use JSON/gzip for all persisted manifest data since it maintains the same serialization method as the previous implementation and is simple to get running
* groups manifest contents into larger chunks of data that are persisted as content pieces
* updated tests showing the proof-of-concept behaves as expected for simple read/write operations

This doesn't include the following as it's a proof-of-concept:
* optimizations to potentially reduce memory usage when reading large JSON arrays
* optimizations to make loading many manifests at the same time efficient (most things only seem to get manifest metadata instead of bulk-loading manifests)
* manifest content garbage collection and compaction code (i.e. compacting indirect manifest content pieces)
* robust error checking and recovery for things like content not found during lookups in indirect content pieces

### DIFF
--- a/repo/manifest/committed_manifest_manager.go
+++ b/repo/manifest/committed_manifest_manager.go
@@ -215,8 +215,13 @@ func (m *committedManifestManager) writeContentChunkLocked(
 		return errors.Wrapf(err, "writing manifest content chunk")
 	}
 
+	contentIDBytes, err := json.Marshal(contentID.String())
+	if err != nil {
+		return errors.Wrap(err, "serializing content ID")
+	}
+
 	for _, e := range entries {
-		e.ContentID = contentID.String()
+		e.Content = contentIDBytes
 	}
 
 	return nil
@@ -245,6 +250,8 @@ func (m *committedManifestManager) writeEntriesLockedV1(
 		// occurs as well as ensuring new entries are persisted since they piggyback
 		// on the v0 specifier.
 		if !e.Deleted && e.formatVersion == 0 {
+			e.Size = int32(len(e.Content))
+
 			staged = append(staged, e)
 
 			// TODO(ashmrtn): Pick some cutoff metric. This is a number out of a hat
@@ -280,15 +287,9 @@ func (m *committedManifestManager) writeEntriesLockedV1(
 		Version: 1,
 	}
 
-	// Now that all content pieces are safely persisted, go through and clear them
-	// so we don't persist the contents in the manifest metadata chunk. Deleted
-	// manifests still won't have a content ID associated with them, but that's
-	// alright since they're tombstones.
-	//
 	// Each manifestEntry that did have content written out should already have a
 	// content ID associated with it since we're working with pointers.
 	for _, e := range entries {
-		e.Content = json.RawMessage("")
 		man.Entries = append(man.Entries, e.manifestEntry)
 	}
 
--- a/repo/manifest/committed_manifest_manager.go
+++ b/repo/manifest/committed_manifest_manager.go
@@ -18,6 +18,8 @@ import (
 	"github.com/kopia/kopia/repo/content/index"
 )
 
+const contentChunkLimit = 10000
+
 // committedManifestManager manages committed manifest entries stored in 'm' contents.
 type committedManifestManager struct {
 	b contentManager
@@ -131,19 +133,18 @@ func (m *committedManifestManager) writeEntriesLocked(ctx context.Context, entri
 }
 
 // +checklocks:m.cmmu
-func (m *committedManifestManager) writeManifestLocked(
+func (m *committedManifestManager) writeContentLocked(
 	ctx context.Context,
-	man manifest,
+	prefix content.IDPrefix,
+	data interface{},
+	buf *gather.WriteBuffer,
 ) (content.ID, error) {
-	var buf gather.WriteBuffer
-	defer buf.Close()
-
-	gz := gzip.NewWriter(&buf)
-	mustSucceed(json.NewEncoder(gz).Encode(man))
+	gz := gzip.NewWriter(buf)
+	mustSucceed(json.NewEncoder(gz).Encode(data))
 	mustSucceed(gz.Flush())
 	mustSucceed(gz.Close())
 
-	contentID, err := m.b.WriteContent(ctx, buf.Bytes(), ContentPrefix, content.NoCompression)
+	contentID, err := m.b.WriteContent(ctx, buf.Bytes(), prefix, content.NoCompression)
 	if err != nil {
 		return content.EmptyID, errors.Wrap(err, "unable to write content")
 	}
@@ -164,7 +165,10 @@ func (m *committedManifestManager) writeEntriesLockedV0(
 		man.Entries = append(man.Entries, e.manifestEntry)
 	}
 
-	contentID, err := m.writeManifestLocked(ctx, man)
+	buf := &gather.WriteBuffer{}
+	defer buf.Close()
+
+	contentID, err := m.writeContentLocked(ctx, ContentPrefix, man, buf)
 	if err != nil {
 		return nil, err
 	}
@@ -181,15 +185,50 @@ func (m *committedManifestManager) writeEntriesLockedV0(
 }
 
 // +checklocks:m.cmmu
+func (m *committedManifestManager) writeContentChunkLocked(
+	ctx context.Context,
+	entries []*inMemManifestEntry,
+	buf *gather.WriteBuffer,
+) error {
+	contents := contentSet{
+		Version: 1,
+	}
+
+	for _, e := range entries {
+		contents.Contents = append(
+			contents.Contents,
+			&manifestContent{
+				ID:      e.ID,
+				Content: e.Content,
+			},
+		)
+	}
+
+	contentID, err := m.writeContentLocked(
+		ctx,
+		IndirectContentPrefix,
+		contents,
+		buf,
+	)
+	if err != nil {
+		return errors.Wrapf(err, "writing manifest content chunk")
+	}
+
+	for _, e := range entries {
+		e.ContentID = contentID.String()
+	}
+
+	return nil
+}
+
+// +checklocks:m.cmmu
 func (m *committedManifestManager) writeEntriesLockedV1(
 	ctx context.Context,
 	entries map[ID]*inMemManifestEntry,
 ) (map[content.ID]bool, error) {
 	var (
-		buf gather.WriteBuffer
-		man = manifest{
-			Version: 1,
-		}
+		staged []*inMemManifestEntry
+		buf    = &gather.WriteBuffer{}
 	)
 
 	defer buf.Close()
@@ -197,11 +236,6 @@ func (m *committedManifestManager) writeEntriesLockedV1(
 	// Write all manifest contents to the content manager and store the generated
 	// content IDs in the manifestEntries.
 	for _, e := range entries {
-		var (
-			contentID content.ID
-			err       error
-		)
-
 		// Deleted manifests don't need to be written out since they're just
 		// tombstones that exist until the next manifest index compaction.
 		//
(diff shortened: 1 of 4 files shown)


repo/manifest/manifest_manager_test.go
ashmrtn, line 1: A lot of the changes in this file are for indentation since the tests are now run for both v0/v1 manifest managers. Hiding whitespace makes viewing simpler

repo/manifest/serialized.go
```diff
@@ -11,14 +11,38 @@ import (
...
-	Content json.RawMessage   `json:"data"`
+	Content json.RawMessage   `json:"data,omitempty"`
+	// For index versions 1 and above, this is the content ID the manifest data
+	// resides in.
+	ContentID string `json:"contentID,omitempty"`
+}
+
+// inMemManifestEntry is the in-memory version of manifests. It adds a few
+// additional fields like a format version.
+type inMemManifestEntry struct {
```
ashmrtn, line 30: This struct is just to reduce the amount of serialized data that's stored. We could remove it if we stored the format version in every `manifestEntry` struct that's written out

repo/manifest/serialized.go
```diff
@@ -11,14 +11,38 @@ import (
...
+	*manifestEntry
+	// formatVersion is the version this manifest was serialized with. Separating
+	// this out to be for every in-memory struct but not persisted struct allows
+	// us to save some space when serializing content.
+	formatVersion int
+}
+
+type contentSet struct {
```
ashmrtn, line 38: This could technically re-use the manifestEntry struct and optionally leave some fields empty (e.x. modtime/labels) instead of making a new struct for persisting manifest content. I don't have a strong opinion one way or the other, I just picked something for a PoC

repo/manifest/committed_manifest_manager.go
```diff
@@ -132,6 +185,211 @@ func (m *committedManifestManager) writeEntriesLocked(ctx context.Context, entri
...
+		if !e.Deleted && e.formatVersion == 0 {
+			staged = append(staged, e)
+
+			// TODO(ashmrtn): Pick some cutoff metric. This is a number out of a hat
+			// based on an easily available metric, the number of contents we have in
+			// this group.
+			//
+			// If every snapshot manifest is ~1KB of data, then 10k of them is ~10MB
+			// of uncompressed content.
+			if len(staged) >= contentChunkLimit {
```
ashmrtn, line 256: I'm not sure if this is a good idea for an optimization, but it seems like we could open up the ability to dedupe these content chunks if we pick a (mostly) deterministic ordering. For example, placing manifests into chunks based on the modtime may be sufficient to generate chunks that could be dedupe if multiple clients performed the same operation, at least for older manifests

On the other hand, if the idea is to have a maintenance operation go through and compact these chunks, we should make sure that that operation isn't run concurrently, which means there's no real benefit to trying to ensure contents can be deduped

repo/manifest/serialized.go
```diff
@@ -11,14 +11,38 @@ import (
...
 type manifestEntry struct {
 	ID      ID                `json:"id"`
 	Labels  map[string]string `json:"labels"`
 	ModTime time.Time         `json:"modified"`
 	Deleted bool              `json:"deleted,omitempty"`
-	Content json.RawMessage   `json:"data"`
+	Content json.RawMessage   `json:"data,omitempty"`
+	// For index versions 1 and above, this is the content ID the manifest data
+	// resides in.
+	ContentID string `json:"contentID,omitempty"`
```
jkowalski, line 25: this will add 16 bytes to the structure, can we reuse 'Content' for this?

IIUC old-style Content will always be JSON object `{"field":"value", ....}`

Can we distinguish them with:

```
if (Content[0] == '{') {
  // treat as inline content
} else {
  // treat as ContentID
}
```

repo/manifest/committed_manifest_manager.go
```diff
@@ -30,17 +33,23 @@ type committedManifestManager struct {
...
 	// autoCompactionThreshold controls the threshold after which the manager auto-compacts
 	// manifest contents
 	// +checklocks:cmmu
 	autoCompactionThreshold int
+
+	// formatVersion is the serialization version that these manifests have.
+	// Version 0 stored all manifest content inline with metadata. Version 1 adds
+	// a level of indirection to store manifest content as separate content blobs
+	// and has the metadata point to the content blob.
+	formatVersion int
```
jkowalski, line 49: I presume not all repositories want this behavior, as it will create a lot more contents, which has its downsides. we would need to have a way to enabling this behavior per repository.

repo/manifest/committed_manifest_manager.go
```diff
@@ -98,31 +107,75 @@ func (m *committedManifestManager) commitEntries(ctx context.Context, entries ma
...
 // where the lock is already being held.
 // +checklocks:m.cmmu
-func (m *committedManifestManager) writeEntriesLocked(ctx context.Context, entries map[ID]*manifestEntry) (map[content.ID]bool, error) {
+func (m *committedManifestManager) writeEntriesLocked(ctx context.Context, entries map[ID]*inMemManifestEntry) (map[content.ID]bool, error) {
 	if len(entries) == 0 {
 		return nil, nil
 	}
 
-	man := manifest{}
+	switch m.formatVersion {
```
jkowalski, line 115: should there be some heuristics where for trivially small number of entries we always write them in V0 format and only compaction uses V1?

repo/manifest/committed_manifest_manager.go
```diff
@@ -98,31 +107,75 @@ func (m *committedManifestManager) commitEntries(ctx context.Context, entries ma
...
-	mustSucceed(json.NewEncoder(gz).Encode(man))
+// +checklocks:m.cmmu
+func (m *committedManifestManager) writeContentLocked(
+	ctx context.Context,
+	prefix content.IDPrefix,
+	data interface{},
+	buf *gather.WriteBuffer,
+) (content.ID, error) {
+	gz := gzip.NewWriter(buf)
+	mustSucceed(json.NewEncoder(gz).Encode(data))
```
jkowalski, line 144: now that I think about it, you will lose a lot of benefits of compression across manifests. Say you have 1MB manifest content with 10000 of similar manifests concatenated together (V0 format)- that might compress down to 150KB or so, but when you individually compress each of 10000 objects (100 bytes each), they will compress to 60 bytes each, so you'll end up with 600KB vs 150KB

repo/manifest/committed_manifest_manager.go
```diff
@@ -132,6 +185,211 @@ func (m *committedManifestManager) writeEntriesLocked(ctx context.Context, entri
...
+		IndirectContentPrefix,
+		contents,
+		buf,
+	)
+	if err != nil {
+		return errors.Wrapf(err, "writing manifest content chunk")
+	}
+
+	for _, e := range entries {
+		e.ContentID = contentID.String()
```
jkowalski, line 219: tip: contentID will be the same, so it makes sense to create the string once and set it to the same value - they will be properly aliased.

Also when parsing this manifest, we can take advantage of ContentID interning.

repo/manifest/serialized.go
ashmrtn: that's an excellent idea! I only thought about the serialized version of the data. It should be fairly trivial to check the format version and then switch the behavior of the Content field based on that. I think then we could avoid having to check the first character since deserialization of this field is deferred until a manifest is requested, at which point we have a format version associated with it

If we're looking to minimize memory usage, then re-arranging how we track manifests in-memory may also be useful. Right now we're wrapping the existing `manifestEntry` with a struct that associates a format version (4-8 bytes for an int) with every `manifestEntry`. We could instead keep the in-memory representation more like the serialized version and only store the format version once for each set of manifests

repo/manifest/committed_manifest_manager.go
ashmrtn: Will revisit interning for deserialization in a later PR as this one's getting a bit large

repo/manifest/serialized.go
ashmrtn: alright, I played around with trying to use a single version of the version specifier like `committedEntries = map[int]map[ID]*manifestEntry` and it complicated some of the logic quite a bit to expand functions to handle that. For now, I think I'll leave that out of this PR and we can revisit if we feel it's worthwhile

repo/manifest/committed_manifest_manager.go
ashmrtn: ah, I think I see why this would be a good idea. It seems like it would help reduce churn on the set of content pieces that the index is tracking. I've put in a quick update to do that, though like other things in the PR, the metrics I picked are sort of out of a hat. I'd be happy to take recommendations from folks if they have opinions

repo/manifest/committed_manifest_manager.go
Lyndon-Li: > allows users to set a flag and only upgrade to this way of storing manifests if requested

What is the upgrade like? For an existing repo, once the flag is set, will we read all the manifests, reorganize them and write them back? Will this be done through a dedicate CLI?

Lyndon-Li: Is it possible we allow the both formats coexist in one repo? E.g., the manifests are stored inline by default, and only if there are too much of them, we compact them into a content and store them indirectly through the repo maintenance?

This would avoid the repo upgrade, or any back compatibility issues, or the downsides involved by the increase of contents as mentioned above.

ashmrtn: > What is the upgrade like? For an existing repo, once the flag is set, will we read all the manifests, reorganize them and write them back? Will this be done through a dedicate CLI?

This hasn't been hashed out completely yet, but yes, likely read in all manifests (already done anyway when populating the manifest manager) and write them in the new format. This could technically either be a dedicated operation or happen during manifest compaction if the user has requested an upgrade. As the new format isn't completely compatible with the older one, doing it as a separate upgrade operation would be safer

> Is it possible we allow the both formats coexist in one repo? E.g., the manifests are stored inline by default, and only if there are too much of them, we compact them into a content and store them indirectly through the repo maintenance?

yes, this would be possible. One of the things jkowalski mentioned out of band is to continue to write all manifests in the old format and compact them into the newer format and/or only use the new format if we had more than a certain number of manifests total

ashmrtn:
Some stuff has come up for me so I'm actually going to close this out for now and we can revisit later if needed

OrhanTozan:
@ashmrtn Hey, does this close #3581? I have been encountering this error, and wonder if this fixes it.

ashmrtn:
> does this close #3581? I have been encountering this error, and wonder if this fixes it.

Hi, @OrhanTozan, a patch like this would make the error in #3581 less likely, but wouldn't solve the issue completely as it still makes the assumption that all of the top-level manifest references would fit in a single kopia content piece

A full breakdown of the error in #3581 is that kopia's content addressable block storage layer imposes a max size limit on the contents it can track (see [architecture diagram](https://kopia.io/docs/advanced/architecture/#architecture) for different layers). The manifest manager builds on top of the content addressable storage layer and attempts to store all manifests as a single content piece in the content storage layer (that's what leads to the error in #3581)

The level of indirection here would reduce the likelihood of seeing the error in the issue you referenced because it reduces the amount of data stored in any given content piece. Another workaround would be to allow kopia to compact manifests into multiple content pieces. There's different tradeoffs associated with each solution in terms of number of API calls to blob storage (e.x. s3) as well as required memory

OrhanTozan:
Thanks for your reply @ashmrtn. Got it.
> Another workaround would be to allow kopia to compact manifests into multiple content pieces.

Is this referring to the changes already made in the `multiple-contents` branch: https://github.com/kopia/kopia/compare/master...alcionai:kopia:multiple-contents?

I see that the `multiple-contents` has already been merged to this `corsostaging` branch: https://github.com/kopia/kopia/compare/master...alcionai:kopia:corsostaging.

Do you recommend using the latest version of the `corsostaging` branch for this error? Or is it a good idea to additionaly also merge your change (this PR) into it?

PR #4067, opened 2024-08-22
refactor(repo): change BLAKE3 implementation

szubersk:
Change BLAKE3 implementation from `github.com/zeebo/blake3` to `lukechampine.com/blake3` available in Debian. This is an attempt to include kopia to Debian.


jkowalski:
On the surface it looks reasonable. @julio-lopez what do you think about this?

The new one is MIT-licensed, the old one is CC0, they both similar star count.

I ran a quick benchmark comparing the performance, and it's a tiny bit worse on my MacBook Pro M1:

BEFORE:
  4. BLAKE3-256           2.1 GB / second
  5. BLAKE3-256-128       2.1 GB / second

AFTER:
  5. BLAKE3-256           1.9 GB / second
  6. BLAKE3-256-128       1.9 GB / second

From what I can tell, Intel results may be different.

julio-lopez:
@jkowalski I have not looked in detail at the differences in implementation.
Is there a compelling reason for the change?
It's not completely clear to me why this is needed or what the benefit would be.
I do not have context here, why does it matter whether github.com/zeebo/blake3 is available in Debian or not? Isn't kopia built from source and statically linked? Or is this different on Debian?

baryluk:
In Debian it depends. Some Go programs are statically linked, some other do link to libc (for some of the libc / nss services like username lookups via pam, etc).

And yes, all build dependencies must be packaged in Debian first. There are multiple good reasons for this. That includes `htmlui` build and runtime dependencies.  (but htmlui build can be probably skipped in the initial version with some simple patches).

I took a look and in Debian we have already: `https://github.com/dchest/blake2b`(in Go, so probably not as fast), `https://github.com/minio/blake2b-simd` (has simd but only for x86), and `https://github.com/lukechampine/blake3` (pure Go, so again probably not great). (I know the first two are blake2, just for reference).


As of `github.com/zeebo/blake3` , it would be nice to have it in Debian, but that also mean its dependency ( `github.com/klauspost/cpuid/` ) also needs to be package. Not too hard, but you would need to find some Go debian developer to help , and get a green light from ftp masters (mostly copyright checks, and some other minor things).

jkowalski:
So is it Debian's goal to ship a stripped-down version of Kopia? I'm curious what's going to be in it vs what will be left out. Is there some discussion we can follow/join?

I think we can make progress on this PR by introducing conditional compilation and have a build flag that selects lukechampine.com vs zeebo. 

On Debian an alternative implementation would register `BLAKE3-256` algorithm. This approach in general could be used to add/remove encryption/hashing/compression/splitters and would be good for making FIPS-compliant builds, for example.

This could also be useful if there are competing implementations of some algorithm and one works better on Intel and another works better on ARM, we could use build rules in a similar way.

cc: @baryluk @szubersk @julio-lopez

julio-lopez:
Selecting the implementation(s) at build time works (via build constraints/tags)

PR #4074, opened 2024-08-25
refactor(general): minor miscellaneous cleanups

julio-lopez:
Cleanups:

- use non-format variants of `Log`/`Print` with no additional args;
- fold in `Fprintf` call with no args into the following one;
- add missing arg placeholder in format strings;
- use `require.Positive` instead of `Greater(..., 0)`;
- rename function to `fillWithZeros` to avoid collision with builtin `clear`;
- define type for context key to avoid collisions;

See individual commits.

### DIFF
--- a/cli/json_output.go
+++ b/cli/json_output.go
@@ -115,8 +115,7 @@ func (l *jsonList) end() {
 }
 
 func (l *jsonList) emit(v interface{}) {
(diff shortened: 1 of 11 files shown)


julio-lopez:
@jkowalski FYI

PR #4084, opened 2024-08-29
refactor(repository): Reduce memory allocations during manifest compaction

ashmrtn:
Mostly aimed at reducing memory allocations during manifest compaction by not modifying the passed in set of items being persisted. Previously this set had persisted entries removed from it, which necessitated cloning the set of committed manifests during compaction. Assuming the number of manifests in the committed set is much larger than the number of dirty entries persisted during calls to `Flush()` this saves copying a map of pointers

Also makes a minor update to how committed contents are tracked since only the value `true` was stored in the map

Consolidated benchmark results with and without these changes where `XItems` represents the total number of manifest entries (benchmark included in PR for reference)
```
go test -v -run X -timeout 120m -bench . -benchmem -benchtime 100x ./repo/manifest/
goos: darwin
goarch: arm64
pkg: github.com/kopia/kopia/repo/manifest
BenchmarkLargeCompaction
BenchmarkLargeCompaction/10000Items
BenchmarkLargeCompaction/10000Items-10       100          20204307 ns/op         5823395 B/op      60463 allocs/op
BenchmarkLargeCompaction-PR/10000Items-10    100          18144909 ns/op         4806542 B/op      60180 allocs/op
BenchmarkLargeCompaction/100000Items
BenchmarkLargeCompaction/100000Items-10      100         241431648 ns/op        50366763 B/op     604206 allocs/op
BenchmarkLargeCompaction-PR/100000Items-10   100         213751905 ns/op        37580470 B/op     600200 allocs/op
BenchmarkLargeCompaction/1000000Items
BenchmarkLargeCompaction/1000000Items-10     100        2932598264 ns/op        613842307 B/op   6038552 allocs/op
BenchmarkLargeCompaction-PR/1000000Items-10  100        2498463415 ns/op        457776432 B/op   6000233 allocs/op

```

### DIFF
--- a/repo/manifest/committed_manifest_manager.go
+++ b/repo/manifest/committed_manifest_manager.go
@@ -32,7 +32,7 @@ type committedManifestManager struct {
 	// +checklocks:cmmu
 	committedEntries map[ID]*manifestEntry
 	// +checklocks:cmmu
-	committedContentIDs map[content.ID]bool
+	committedContentIDs map[content.ID]struct{}
 
 	// autoCompactionThreshold controls the threshold after which the manager auto-compacts
 	// manifest contents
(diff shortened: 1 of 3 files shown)


jkowalski (approved):

jkowalski:
This is huge, thanks for all the performance investigation efforts and work on fixing it.

PR #4100, opened 2024-09-09
feat(server): snapshot metrics

### LINKED ISSUE #609: [Feature Request] Prometheus metrics from kopia server
It would be great if kopia server exported Prometheus-format metrics, e.g. under "/metrics", for things such as:

- Snapshot age and size
- Maintenance interval for quick/full, plus time until the next scheduled run
- Recent maintenance results

That would allow monitoring via a dashboard and, more importantly, raising alerts if the time since the last snapshot grew larger than expected.

Probably that URL would be protected via the same user authentication as access to the web UI. If more advanced user management is implemented (#548), then maybe a "metrics" permission could be defined.

If this seems like a good idea I could suggest how those metrics could look.

alvistar:
This pull request addresses issue #609 by adding metrics to track information about the last successful snapshots for each source.
 These new metrics provide insights into the recency, size, and content of snapshots across different sources.

 Changes made:
 - Implemented new metrics in the sourceManager to track:
   - Timestamp of the last snapshot start time
   - Timestamp of the last snapshot end time
   - Size of the last snapshot in bytes
   - Number of files in the last snapshot
   - Number of directories in the last snapshot

 These additions will help users and administrators better monitor the details of their most recent backups, enabling quicker  assessment of backup coverage and content.

 Implementation details:
 - Metrics are implemented using the repository's metric registry
 - Each metric is tagged with the source's host, username, and path for precise tracking
 - Metrics are updated only for complete snapshots (those without an IncompleteReason)
 - --refresh-interval will affect the time interval in refreshing snapshot metrics

 Testing:
 - Manually verified the new metrics are correctly reported in the Prometheus output
 - Existing tests should cover the new functionality, as it's integrated into existing methods

 Please review and let me know if any further changes or clarifications are needed.

 Fixes #609

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -672,6 +672,7 @@ func (s *Server) syncSourcesLocked(ctx context.Context) error {
 	// stop source manager for sources no longer in the repo.
 	for _, sm := range oldSourceManagers {
 		sm.stop(ctx)
+		sm.removeMetrics() // Remove metrics for the source manager
 	}
 
 	for src, sm := range oldSourceManagers {
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -1006,6 +1006,13 @@ func (s *Server) getSchedulerItems(ctx context.Context, now time.Time) []schedul
 		NextTime:    nrt,
 	})
 
+	// add a scheduled item to refresh metrics
+	result = append(result, scheduler.Item{
+		Description: "refresh metrics",
+		Trigger:     s.refreshMetrics,
+		NextTime:    now.Add(1 * time.Minute), // Refresh metrics every minute
+	})
+
 	if s.maint != nil {
 		// If we have a direct repository, add an item to run maintenance.
 		// If we're the owner then nextMaintenanceTime will be zero.
@@ -1042,6 +1049,15 @@ func (s *Server) refreshScheduler(reason string) {
 	select {
 	case s.schedulerRefresh <- reason:
 	default:
+	}
+}
+
+func (s *Server) refreshMetrics() {
+	s.serverMutex.RLock()
+	defer s.serverMutex.RUnlock()
+
+	for _, sm := range s.sourceManagers {
+		sm.refreshStatus(s.rootctx)
 	}
 }
 
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -486,9 +486,6 @@ func handleShutdown(ctx context.Context, rc requestContext) (interface{}, *apiEr
 }
 
 func (s *Server) requestShutdown(ctx context.Context) {
-	// Workaround for passing tests
-	s.rep.Metrics().RemoveAllGauges()
-
 	if f := s.OnShutdown; f != nil {
 		go func() {
 			if err := f(ctx); err != nil {
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -1005,13 +1005,6 @@ func (s *Server) getSchedulerItems(ctx context.Context, now time.Time) []schedul
 		NextTime:    nrt,
 	})
 
-	// add a scheduled item to refresh metrics
-	result = append(result, scheduler.Item{
-		Description: "refresh metrics",
-		Trigger:     s.refreshMetrics,
-		NextTime:    now.Add(1 * time.Minute), // Refresh metrics every minute
-	})
-
 	if s.maint != nil {
 		// If we have a direct repository, add an item to run maintenance.
 		// If we're the owner then nextMaintenanceTime will be zero.
@@ -1048,15 +1041,6 @@ func (s *Server) refreshScheduler(reason string) {
 	select {
 	case s.schedulerRefresh <- reason:
 	default:
-	}
-}
-
-func (s *Server) refreshMetrics() {
-	s.serverMutex.RLock()
-	defer s.serverMutex.RUnlock()
-
-	for _, sm := range s.sourceManagers {
-		sm.refreshStatus(s.rootctx)
 	}
 }
 
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -1005,6 +1005,13 @@ func (s *Server) getSchedulerItems(ctx context.Context, now time.Time) []schedul
 		NextTime:    nrt,
 	})
 
+	// add a scheduled item to refresh metrics
+	result = append(result, scheduler.Item{
+		Description: "refresh metrics",
+		Trigger:     s.refreshMetrics,
+		NextTime:    now.Add(1 * time.Minute), // Refresh metrics every minute
+	})
+
 	if s.maint != nil {
 		// If we have a direct repository, add an item to run maintenance.
 		// If we're the owner then nextMaintenanceTime will be zero.
@@ -1041,6 +1048,15 @@ func (s *Server) refreshScheduler(reason string) {
 	select {
 	case s.schedulerRefresh <- reason:
 	default:
+	}
+}
+
+func (s *Server) refreshMetrics() {
+	s.serverMutex.RLock()
+	defer s.serverMutex.RUnlock()
+
+	for _, sm := range s.sourceManagers {
+		sm.refreshStatus(s.rootctx)
 	}
 }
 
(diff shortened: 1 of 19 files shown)


jkowalski:
At first glance this is looking good. We obviously need to add tests for this.
I think exposing this data should be opt-in, perhaps as a policy option.

This is because we would be exposing pathnames over `/metrics` endpoint, which is not secured and pathnames can be sensitive.

alvistar:
Let me know if any further action is required on my side.

jkowalski:
> Let me know if any further action is required on my side.

Can you:

1. Look at all faling checks in this PR, in particular linter failures and make changes to address those.
2. Add a way to include/exclude directories from metrics using policy options?
3. Add at least one end-to-end test that launches kopia server, runs one or two snapshots, fetches metrics over HTTP and ensures the metrics for that particular directory are what's expected.

alvistar:
> > Let me know if any further action is required on my side.
> 
> Can you:
> 
> 1. Look at all faling checks in this PR, in particular linter failures and make changes to address those.
> 2. Add a way to include/exclude directories from metrics using policy options?
> 3. Add at least one end-to-end test that launches kopia server, runs one or two snapshots, fetches metrics over HTTP and ensures the metrics for that particular directory are what's expected.

1) Lint is done
3) End to end tests done

Regarding 2, I was also thinking to a flag running the server, something like --include-snapshots-metrics. If I need to implement this or policy, it would speed up things if you point me toward files involved so I don't need to search all codebase.

alvistar:
Hi, I added metrics policy for expose metrics

snapshot/policy/policy_tree.go
```diff
@@ -69,6 +69,11 @@ var (
 		ParallelUploadAboveSize: newOptionalInt64(2 << 30), //nolint:mnd
 	}
 
+	// defaultMetricsPolicy is the default metrics policy.
+	defaultMetricsPolicy = MetricsPolicy{
+		ExposeMetrics: NewOptionalBool(true),
```
jkowalski, line 74: Not sure this should be true. @julio-lopez what do you think?

tests/end_to_end_test/server_metrics_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	wait, _ := e.RunAndProcessStderr(t, sp.ProcessOutput,
+		"server", "start",
+		"--ui",
+		"--address=localhost:0",
+		"--random-password",
+		"--random-server-control-password",
+		"--tls-generate-cert",
+		"--tls-generate-rsa-key-size=2048",      // use shorter key size to speed up generation
+		"--metrics-listen-addr=localhost:10000", // enable metrics
```
jkowalski, line 41: It's usually better to use port zero in tests to avoid any conflict with any process holding the particular port open.

internal/server/source_manager.go
```diff
@@ -455,6 +469,20 @@ func (s *sourceManager) refreshStatus(ctx context.Context) {
 	} else {
 		s.nextSnapshotTime = s.findClosestNextSnapshotTimeReadLocked()
 	}
+
+	// Do not expose metrics if the policy does not allow it.
+	if pol.MetricsPolicy.ExposeMetrics == nil || !*pol.MetricsPolicy.ExposeMetrics {
```
jkowalski, line 474: OrDefault(false) ?

internal/server/source_manager.go
```diff
@@ -295,6 +298,17 @@ func (s *sourceManager) stop(ctx context.Context) {
 	close(s.closed)
 }
 
+func (s *sourceManager) removeMetrics() {
+	if s.rep.Metrics() != nil {
+		registry := s.rep.Metrics()
+		registry.RemoveGauge(s.lastSnapshotStartTime)
```
jkowalski, line 304: That seems wrong-each of these calls would remove all gauges instead of just gauges for a particular source.

jkowalski (commented):
I took a quick pass...  it's looking really good. Thanks for the thorough test coverage.

tests/end_to_end_test/server_metrics_test.go
alvistar: I do agree, and indeed I did parsing of the random port,, but setting port to 0 it will be printed as 0 in server startup. I did not investigate further and set to 10000.

internal/server/source_manager.go
alvistar: Should be ok. Tested in e2e, only the gauge with the correct labels will be deleted.

internal/server/source_manager.go
alvistar: Sure. Will clean.

tests/end_to_end_test/server_metrics_test.go
jkowalski: you need to print the actual port number it binds to same as we do elsewhere.

alvistar:
@jkowalski  I have issues running complete e2e tests. Running individually they work, but they don't work running as sequence. It looks like that the prometheus registry is not cleared when restarting the server. Any clue with this?

internal/server/source_manager.go
alvistar: Done

tests/end_to_end_test/server_metrics_test.go
alvistar: Done

alvistar:
Hi,

> @jkowalski I have issues running complete e2e tests. Running individually they work, but they don't work running as sequence. It looks like that the prometheus registry is not cleared when restarting the server. Any clue with this?

As workaround I remove all gauges when the server is shutting down. It works, but this should not necessary beside the testing.

internal/server/server.go
```diff
@@ -482,6 +482,9 @@ func handleShutdown(ctx context.Context, rc requestContext) (interface{}, *apiEr
 }
 
 func (s *Server) requestShutdown(ctx context.Context) {
+	// Workaround for passing tests
+	s.rep.Metrics().RemoveAllGauges()
```
jkowalski, line 486: i think that's a bit too early as during shutdown we may emit some gauges.

Also, can this RemoveAllGauges be done in the test itself and not in the server?

alvistar: The worse thing can happen is prom will scrape /metrics that moment and think that these metrics have been removed.
As putting in testing, I think is not clean call an internal function while doing e2e.
Ideally this should be called after the metrics listener has been closed. BTW is this getting closed with server shutdown? I am not that sure

alvistar:
Regarding the linting, I do make lint before committing, but I see with GitHub action still fails. Why?

btw did you consider doing a pre-commit hook for linting?

alvistar:
Hi,
fixed checklocks, moved the removeAllGauges on registry close. Let me know if there is something else to address.

jkowalski:
can you rebase?

alvistar:
> can you rebase?

Done

v0ctor:
Is there anything left to integrate these changes? Thank you everyone!

alvistar:
> Is there anything left to integrate these changes? Thank you everyone!

Hi, as far as I know there is nothing to do left. I have been asked to rebase which I did, then nothing happen. 

@jkowalski  any input from your side?

andyundso:
@alvistar given from the state of the PR, maybe you can rebase your branch, fix the conflicts and check what is needed to make the linter pass. I assume once the CI is green, then your changes could be integrated.

ymettier:
Hello,

Will this PR be merged ?

If yes, could you synchronize so that @alvistar rebases (and resolves conflicts) and @jkowalski (or anyone) merges soon after rebase ? If both of you don't mind of course.

Many thanks for this PR

ymettier:
Hello,

I just forked the 2 repositories in order to rebase @alvistar 's PR on master.

The result is this branch on my own space : https://github.com/ymettier/kopia/tree/alvistar-snapshot-metrics-rebased

I will now build my own docker image and run it on my environment for few days.

If it works for me, what would be the next steps @jkowalski ? Should I make my own PR ? Would you merge it soon ? What are the plans ? Please tell me to ensure all this work is worth it.

Thanks @alvistar for the work (I juste rebased and fixes minor issues in the last commit. The whole work is yours) !

ymettier:
image

It's working for me. I'm excited to share this screnshot with you !!!

ymettier:
Hello,

I finally made my own PR #4685 . All credits are to @alvistar except last `git rebase` and rebase fixes.

Please @jkowalski have a look on PR #4685 and tell us if you would merge, or what is missing to merge.

PR #4107, opened 2024-09-12
improve docs: snapshot verify vs. content verify

ptandler:
I copied some parts of https://kopia.discourse.group/t/best-method-to-ensure-valid-snapshots-snapshot-verify-vs-snapshot-fix-invalid-files/1217/2 to site/content/docs/Advanced/Consistency/_index.md

and also extended the description of both verify commands a bit to
a) provide a hint that there are two different verify commands and
b) point to the docs to understand the details

I really miss more extensive explanations on the CLI doc pages like https://kopia.io/docs/reference/command-line/advanced/content-verify/ - I guess they are generated from the GO code?

what I actually did not understand yet: 

if you say that snapshot verify include content verify - why would one ever want to run content verify? @jkowalski

### DIFF
--- a/cli/command_content_verify.go
+++ b/cli/command_content_verify.go
@@ -26,7 +26,9 @@ type commandContentVerify struct {
 }
 
 func (c *commandContentVerify) setup(svc appServices, parent commandParent) {
-	cmd := parent.Command("verify", "Verify that each content is backed by a valid blob")
+	cmd := parent.Command("verify", "Verify that each content is backed by a valid blob. " +
+	    "Does **not** verify that all blobs for all existing snapshots do exist. " +
+	    "See also `kopia snapshot verify` and https://kopia.io/docs/advanced/consistency/ for details")
 
 	cmd.Flag("parallel", "Parallelism").Default("16").IntVar(&c.contentVerifyParallel)
 	cmd.Flag("full", "Full verification (including download)").BoolVar(&c.contentVerifyFull)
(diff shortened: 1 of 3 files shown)


ptandler:
Actually, when I started using kopia, for the CLI reference I missed more detailed descriptions and also example in general. Is there a way to add this - to the website and maybe also to the built-in CLI help?

jkowalski:
`content verify` is basically a low-level consistency check to make sure all contents have backing blobs and it's usually super quick and often sufficient.

`snapshot verify` is more expensive as it does a full tree walk of all directories in all snapshots to make sure all the referenced contents exist. It effecitvely verifies each content has a backing blob, so it's a superset of `content verify`

cli/command_content_verify.go
```diff
@@ -26,7 +26,9 @@ type commandContentVerify struct {
 }
 
 func (c *commandContentVerify) setup(svc appServices, parent commandParent) {
-	cmd := parent.Command("verify", "Verify that each content is backed by a valid blob")
+	cmd := parent.Command("verify", "Verify that each content is backed by a valid blob. " +
```
jkowalski, line 29: this is good - we have an example of such lengthy comment in a few places, but I would add an extra empty line after the first line summary:

```
Verify that each content is backed by a valid blob.

Does **not** verify that all blobs for all existing snapshots do exist. 
See also `kopia snapshot verify` and https://kopia.io/docs/advanced/consistency/ for details
```

PR #4120, opened 2024-09-20
refactor(providers): simplify GCS client construction

julio-lopez:
Based on the change originally put forward in #3564, which did not work.
The approach here passes the GCS provider tests.

Successful GCS provider job: https://github.com/kopia/kopia/actions/runs/11513537657/job/32050421531?pr=4120


repo/blob/gcs/gcs_storage.go
```diff
@@ -231,30 +209,14 @@ func tokenSourceFromCredentialsJSON(ctx context.Context, data json.RawMessage, s
...
-	}
-
+	var clientOption option.ClientOption
 	if sa := opt.ServiceAccountCredentialJSON; len(sa) > 0 {
-		ts, err = tokenSourceFromCredentialsJSON(ctx, sa, scope)
+		clientOption = option.WithCredentialsJSON(sa)
 	} else if sa := opt.ServiceAccountCredentialsFile; sa != "" {
-		ts, err = tokenSourceFromCredentialsFile(ctx, sa, scope)
-	} else {
-		ts, err = google.DefaultTokenSource(ctx, scope)
```
julio-lopez, line 248: This does not handle all the cases.

repo/blob/gcs/gcs_storage.go
```diff
@@ -231,30 +209,14 @@ func tokenSourceFromCredentialsJSON(ctx context.Context, data json.RawMessage, s
 func New(ctx context.Context, opt *Options, isCreate bool) (blob.Storage, error) {
 	_ = isCreate
 
-	var ts oauth2.TokenSource
-
-	var err error
-
-	scope := gcsclient.ScopeReadWrite
-	if opt.ReadOnly {
```
julio-lopez, line 239: This seems like a change in behavior as it appear to be dropping read-only scope on the service. I need clarity here.

repo/blob/gcs/gcs_storage.go
```diff
@@ -273,38 +251,28 @@ func tokenSourceFromCredentialsJSON(ctx context.Context, data json.RawMessage, s
...
-	}
+	clientOptions := []option.ClientOption{option.WithScopes(scope)}
 
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to initialize token source")
+	if j := opt.ServiceAccountCredentialJSON; len(j) > 0 {
+		clientOptions = append(clientOptions, option.WithCredentialsJSON(j))
+	} else if fn := opt.ServiceAccountCredentialsFile; fn != "" {
+		clientOptions = append(clientOptions, option.WithCredentialsFile(fn))
 	}
```
KastenMike, line 269: For completeness maybe to add:
```suggestion
	} else {
	        ts, err := google.DefaultTokenSource(ctx)
		if err != nil {
			return nil, errors.Wrap(err, "failed to get default token source")
		}

		clientOptions = append(clientOptions, option.WithTokenSource(ts))
	}
```
since before the 3 flows were supported

repo/blob/gcs/gcs_storage.go
```diff
@@ -273,38 +251,28 @@ func tokenSourceFromCredentialsJSON(ctx context.Context, data json.RawMessage, s
...
 	_ = isCreate
 
-	var ts oauth2.TokenSource
-
-	var err error
+	if opt.BucketName == "" {
+		return nil, errors.New("bucket name must be specified")
+	}
 
 	scope := gcsclient.ScopeReadWrite
```
KastenMike, line 258: to support GCS blob extensions via `runFullMaintenance` it requires `gcsclient.ScopeFullControl` here

julio-lopez: Already addressed in #4151

.github/workflows/provider-gcs.yml
julio-lopez, line 1: Not to be merged, only for testing these changes.

julio-lopez:
@jkowalski thoughts?

jkowalski (approved):

PR #4121, opened 2024-09-20
feat(cli): Allow snapshot ID arguments for command snapshot verify

redgoat650:
Snapshot verify can now accept a list of one or more snapshot IDs (manifest IDs) to be used as the subject(s) of the snapshot verify operation.

Added a unit test that ensures only the requested snapshots are verified when their IDs are provided.

Now in addition to the existing `--source`, `--file-id`, and `--directory-id` flags, which work the same, the user can provide snapshot IDs as args:

```
kopia snapshot verify <snapID> <snapID> <snapID>
```


Shrekster (approved):

julio-lopez (commented):
LGTM  ✅

PR #4132, opened 2024-09-24
refactor(providers): always clone default HTTP transport in S3 provider

julio-lopez:
The s3 storage provider had a different http transports for different cases:

- HTTPS without TLS verification: `&http.Transport{}` with default values;
- HTTPS with TLS verification: `http.DefaultTransport.Clone()`

This change uses `http.DefaultTransport` in all cases, instead of creating an empty (zero-value) `http.Transport` for consistency.

Author: @alexvbg


julio-lopez:
@alexvbg FYI

PR #4133, opened 2024-09-26
feat(repository): Google Immutability - Allow blob Put/ExtendBlobRete…

KastenMike:
Adds `ExtendBlobRetention` and removes the `ErrUnsupportedPutBlobOption` check from PutBlob. Also some slight refactorings to make it more like `azure_storage.go` and tweak to support getting a blob by its versionID.


KastenMike:
closing in favour of https://github.com/kopia/kopia/pull/4134

PR #4134, opened 2024-09-26
feat(providers): GCS immutability

KastenMike:
It seems not possible to create a PR train in this repo. I created 1 PR but the base only shows me master or providers. So instead it's broken down by commits. 

Put/Extend commit: Adds `ExtendBlobRetention` and removes the `ErrUnsupportedPutBlobOption` check from `PutBlob`. Also some slight refactorings to make it more like `azure_storage.go` and tweak to support getting a blob by its versionID. 

PITR commit: Code is largely identical to `azure_` and `s3_` equivalents. Based on @kreare's #3564 but more complete.

Tests commit:
`gcs_immu_test.go` and `gcs_versioned_test.go` are basically identical to `azure_` equivalents. Only `createBucket` is a bit different based on Google client. Tweaked `gcs_storage_test.go` slightly to make it more like the az/s3 ones with consts for env vars

Requires variables setting to create/use the new bucket

Ref:
- #3564

### DIFF
--- a/cli/storage_gcs.go
+++ b/cli/storage_gcs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"os"
+	"time"
 
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


julio-lopez (approved):
@KastenMike
🥇 Thanks for doing this.
👍 Looks good: ➕1

kreare:
thank you so much for this

kreare:
i think docs should be fixed, the first part say that GCS doesn't support the needed features but thats not true (they are not supported if using the S3 compatibility mode, but the native GCS supports everything we need)

Screenshot_2024-10-06-20-04-23-646_com android chrome-edit

KastenMike:
> i think docs should be fixed, the first part say that GCS doesn't support the needed features but thats not true anymore (they are still not supported if using the S3 compatibility mode, but the native GCS supports everything we need)

Ah I missed that part of the docs when I did a [docs tweak](https://github.com/kopia/kopia/pull/4135). But https://github.com/kopia/kopia/pull/4151 needs merging for it to fully work

PR #4135, opened 2024-09-26
docs(site): Add Google ransomware protection section

KastenMike:
Docs PR for https://github.com/kopia/kopia/pull/4134

image

### DIFF
--- a/site/content/docs/Advanced/Ransomware Protection/_index.md
+++ b/site/content/docs/Advanced/Ransomware Protection/_index.md
@@ -22,7 +22,7 @@ For the context of Kopia protection, ransomware refers to viruses, trojans or ot
  * Kopia's Google Cloud Services (GCS) engine provides neither restricted access key nor object-lock support.
     * Google's S3 compatibility layer does not provide sufficient access controls to use these features, and thus Kopia cannot use
       the ransomware mitigation discussed on this page with GCS at this time.
-* Kopia's Azure storage engine supports object-locks for ransomware protection.
+* Kopia's Azure & Google storage engines support object-locks for ransomware protection.
 
 ### Using application keys to protect your data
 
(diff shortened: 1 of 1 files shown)


julio-lopez (approved):

riedel:
Hi, just saw this PR and was wondering how the retention period works for "p blobs". Since blobs are deduplicated to me, it seems that the retention period is wrong in those cases (old file still used, or new file with same content as old one). AWS published a whole article on renewing retentions (https://aws.amazon.com/de/blogs/storage/maintaining-object-immutability-by-automatically-extending-amazon-s3-object-lock-retention-periods/) . How does kopia handle this correctly in GCS? From the GCS documentation versioning is independent and the retention period does not only start after the delete / overwrite, as I read it (which would be necessary for ransomware protection). Current files IMHO need period renewal of the retention period. Does Kopia really do this e.g. in maintenance (which at least would be needed)?

Sorry for high jacking this PR but there was no answer in the forums. I have the fear, that ransomware protection is not working as claimed without any better documentation . But I haven't had the chance to prove it.

Maybe the people involved in this PR can quickly comment on it (I am happy to add this to documentation if I am proven wrong).

KastenMike:
> Current files IMHO need period renewal of the retention period. Does Kopia really do this e.g. in maintenance (which at least would be needed)?

`--extend-object-locks` needs to be enabled as part of `kopia maintenance set`. I guess docs should updated to be explicit in this & [this section](https://kopia.io/docs/advanced/ransomware-protection/#some-notes-about-storage-providers) also needs updating to include GCS. Since [here](https://kopia.io/docs/advanced/ransomware-protection/#how-to-enable-object-locks) only talks about S3 but it's also required for Azure and GCS

But in order for it to work https://github.com/kopia/kopia/pull/4151 needs merging

PR #4136, opened 2024-09-27
fix(providers): GCS tests

julio-lopez:
The GCS provider tests broke in #4134

This is an attempt to fix the tests by using the credentials that are available in the providers validation job. Note that a credentials file is not available.

However, the tests still fail because:
- there isn't a versioned / immutable bucket for the tests; and
- the credentials available in the test do not allow creating buckets.

The GCS versioning / immutability tests are skipped in the mean time until the required bucket is created.


julio-lopez:
Sample job failure: https://github.com/kopia/kopia/actions/runs/11064672542/job/30742829604

PR #4138, opened 2024-09-27
test(providers): cleanup GCS tests

julio-lopez:
- Use `getEnvVarOrSkip()` helper
- Use creds from JSON env for test to actually run the test. A file with creds is not available in the testing environment.
- nit: relocate helpers


julio-lopez:
@KastenMike FYI

PR #4139, opened 2024-09-27
refactor(repository): Reduce memory usage for epoch index compaction

Lyndon-Li:
When there are large number of small files in the backup, epoch index compaction, as part of repository maintenance, consumes large size memory which is much larger than other part of repository maintenance.

Below is the result of my test case:
Backup 20M(20971520) files and each file is 4KB in size, so there will be 20M indexes in the backup
When the index compaction is due, it takes more than 15GB memory in the `Compacting an eligible uncompacted epoch` phase of repository maintenance:
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                              
 881389 root      20   0   19.0g  15.4g 859624 S 100.0  48.9  25:18.93 kopia 
```

This PR refactors the index builder, and could reduce the memory usage significantly. In my testing env, the memory reduces to 7.1GB for the same dataset:
```
   PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                              
 882647 root      20   0   11.1g   7.1g 860304 S 100.0  22.7  26:28.02 kopia 
```

Below are the primary reasons of large memory usage for the existing code:
1. There are 3 copies of the full indexes. The compaction has 3 stages --- index collection, shard, sort and build. The entire indexes are loaded and kept in memory in the 1st stage, the 2nd stage makes a full copy, the 3rd stage makes a third copy for each shard (2^24 indexes)
2. 1st/2nd stage uses `map[index.ID]Info` and 3rd stage uses `[][]Info` as the data struct. The Info items cannot be shared in the 3 structures even though they are identical. So the memory footprint for each index will be `(sizeof(Info) + sizeof(index.ID) + map/slice element overhead) * numOfCopies` while sizeof(Info) + sizeof(index.ID) + map/slice element overhead > 200 bytes
3. The existing index builder (in index_builder.go) is used by many scenarios. While for epoch index compaction specifically, the usage is simple --- it doesn't need to keep the full indexes all the time for further insertion or search, e.g., after shard, we don't need to keep the `map[index.ID]Info` generated in collection stage
4. There are redundancies for packBlobID(of type `blob.ID`). Multiple contents may be in one pack blob, so multiple indexes may share the same PackBlobID. However, there is one PackBlobID in each Info, while, each PackBlobID is more than 50 bytes
5. map[index.ID]Info has a redundancy of `index.ID`, between the key and value, since Info itself also containers the same `index.ID`. While `index.ID` takes 34 bytes each.

Below are the improvements in this PR:
1. Introduce a new builder for index compaction only
1. Use a new builder item struct (`InfoCompact`) which has `*blob.ID` instead of `blob.ID` for packBlobID
2. Use a map implementation that supports custom key compare function(built-in map doesn't), and use pointers to indexes as the map element (`map<*InfoCompact>`). In the PR, we use a rbtree implementation and don't need a key at all. In this way, there is always one piece of memory footprint for each index, no matter how they are copied/moved. The memory footprint for each index will be `sizeof(InfoCompact) + (8  + rbtree/slice element overhead) * numOfCopies`
3. Put the packBlobIDs pointers into a map of the same implementation (`map<*blob.ID>`), all `InfoCompact` reference packBlobIDs in the map, so as to remove the redundancies for packBlobID
4. Since the rbtree is auto sorted at insertion time, so no need to make a separate copy for sort
6. For shard stage, use `[][]*InfoCompact` as the output, index items are removed from `map<*InfoCompact>` and appended to  `[][]*InfoCompact` in the ascendant order. As a result, after shard stage, `map<*InfoCompact>` will be cleared to save memory 

This PR doesn't change any logics of the index compaction, e.g., how indexes are compared, sharded. It only changes the way to implement the logics. So the changes are safe.


Lyndon-Li:
Here are the inuse memory during the running:

Before
image

After
image

And the alloc memory:

Before
image

After
image


From the both comparison, we can see the memory footprint is significantly reduced.

repo/content/index/info.go
```diff
@@ -21,7 +21,119 @@ type Info struct {
 	EncryptionKeyID     byte                 `json:"encryptionKeyID,omitempty"`
 }
 
+type InfoCompact struct {
+	PackBlobID          *blob.ID
```
jkowalski, line 25: given that `blob.ID` is a `string` would it make sense to do interning instead of having a pointer here? IOW, can we avoid this structure altogether because it would be identical to the `Info`, we just need to be careful to share the string instead of having all these copies?

repo/content/index/index_builder.go
```diff
@@ -63,8 +67,8 @@ func init() {
 // sortedContents returns the list of []Info sorted lexicographically using bucket sort
 // sorting is optimized based on the format of content IDs (optional single-character
 // alphanumeric prefix (0-9a-z), followed by hexadecimal digits (0-9a-f).
-func (b Builder) sortedContents() []Info {
-	var buckets [36 * 16][]Info
+func (b Builder) sortedContents() []BuilderItem {
```
jkowalski, line 70: can we avoid `BuilderItem` by having `*Item` instead (assuming we can do `blob.ID` interning as suggested elsewhere)

Lyndon-Li: Yes, this comes down to whether we can avoid the introduction of `InfoCompact`, if we can use `Info` only, we will not need this `BuilderItem`

repo/content/index/info.go
Lyndon-Li: golang string copy assignment is always deep copy.
So if we use `blob.ID`, we will not able to share the memory kept in `packBlobIDs` of `oneUseBuilder`.

Moreover, whenever we copy `Info`, a new string with the whole new memory will be allocated. Even though we always be careful to dereference the source of the copy, the memory footprint also increases considering the delay of GC.

jkowalski: i'm pretty sure strings share the underlying data when assigning and even slicing.

To get a deep copy of strings, you need to use `strings.Clone()` https://pkg.go.dev/strings#Clone

Lyndon-Li: You are right. `StringData` is lazy copied only when the shared data is to be changed.
For the current case, `packBlobID` is never changed, so it is fine to use `string`.

And for the same reason, my previous change, which use *string for `packBlobID`, makes no help to the memory save, instead it increases memory footprint of string since it always allocs a new string for each `Info`.

Lyndon-Li: I changed the code accordingly and tested it again, now the peek memory usage reduces to 6GB:

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                              
1185499 root      20   0   10.7g   6.0g 860144 S 348.2  19.1  26:04.75 kopia

Lyndon-Li:
I am on holiday till 8th, will continue to work on this PR when I am back.

jkowalski:
> I am on holiday till 8th, will continue to work on this PR when I am back.

Awesome, I think we can approve it as soon as CI linter issues are fixed.

Lyndon-Li:
@jkowalski 
For this test failure https://github.com/kopia/kopia/actions/runs/11101846488/job/30884039243?pr=4139 on ARM64, I couldn't reproduce it in my x86 env.

Do you have any clue on why it happened?

repo/content/index/one_use_index_builder.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+}
+
+// Add adds a new entry to the builder or conditionally replaces it if the timestamp is greater.
+func (b *OneUseBuilder) Add(i Info) {
+	cid := i.ContentID
+
+	found := b.indexStore.Get(&Info{ContentID: cid})
+	if found == nil || contentInfoGreaterThanStruct(&i, found.(*Info)) { //nolint:forcetypeassert
+		_ = b.indexStore.ReplaceOrInsert(&Info{
```
jkowalski, line 38: wouldn't `&i` be sufficiient? The structure is passed by value.

repo/content/index/packindex_test.go
```diff
@@ -166,23 +168,37 @@ func testPackIndex(t *testing.T, version int) {
 		t.Fatalf("unable to build: %v", err)
 	}
 
+	if err := b4.BuildStable(&buf4, version); err != nil {
```
jkowalski, line 171: nit: `require.NoError(t, b4.BuildStable())` is usually sufficient in such cases. It will print the line number on failure.

jkowalski (approved):

repo/content/index/one_use_index_builder.go
Lyndon-Li: Yes, it seems this should have been modified together when removing `InfoCompact`.
On the other hand, this should make no difference to memory usage since the unnecessary variables are local variables on stack.

I submitted a PR #4167 for it, feel free to merge or close it.

julio-lopez:
> This is a very important improvement, thanks for your contribution!

@Lyndon-Li I second Jarek's comment here.

🚀 Great improvement!!!
🥇 Thanks for your contribution! Keep them coming.

Lyndon-Li:
@jkowalski @jkowalski @Shrekster 
Sure, Kopia is a great software with a perfect architecture, let's keep rolling to explore and extend it.

Agree with @Shrekster's judgment about the change rate of users' data, so it is important to support large scale of indexes with acceptable performance. And this is also essential to support of smaller chunk size for deduplication which could achieve much better deduplication ratio.

The index compaction memory usage is the most prominent issue I see with large number of indexes for now, though I must not have touched every corner. I will continue testing on that and let you know if I have new findings.

PR #4142, opened 2024-09-27
fix(cli): Deadline on s3 requests

arouene:
Hello,

Since last week one of the S3 providers has some issues on their Object Storage infrastructure preventing successful backups.

Specifically, after recent changes in their infrastructure, some requests fails to respond but do not close the connection, causing backups to stall indefinitely.

While this issue is on the provider's side, I took the opportunity to improve the resiliency of Kopia by implementing a "request timeout" using a context deadline, mitigating the issue we had. Requests that are stucks, are now failing in error and are simply retried by Minio.

This PR adds a new flag "--request-timeout" to the commands related to S3 repository (eg. "repository create s3" or "repository connect s3") and accept a value in seconds. By default the value is "0", which is "no timeout", so no breaking change of the actual behavior.

A context deadline is used on every requests but not on ListObjects, which already use a context.WithCancel.

What do you think ?

### DIFF
--- a/repo/blob/s3/s3_options.go
+++ b/repo/blob/s3/s3_options.go
@@ -23,6 +23,9 @@ type Options struct {
 	SecretAccessKey string `json:"secretAccessKey" kopia:"sensitive"`
 	SessionToken    string `json:"sessionToken"    kopia:"sensitive"`
 
+	// Timeout is used as a deadline context in Minio API calls.
+	RequestTimeout uint `json:"requestTimeout,omitempty"`
+
 	// Region is an optional region to pass in authorization header.
 	Region string `json:"region,omitempty"`
 
--- a/repo/blob/s3/s3_options.go
+++ b/repo/blob/s3/s3_options.go
@@ -24,7 +24,7 @@ type Options struct {
 	SessionToken    string `json:"sessionToken"    kopia:"sensitive"`
 
 	// Timeout is used as a deadline context in Minio API calls.
-	RequestTimeout uint `json:"requestTimeout,omitempty"`
+	RequestTimeout int64 `json:"requestTimeout,omitempty"`
 
 	// Region is an optional region to pass in authorization header.
 	Region string `json:"region,omitempty"`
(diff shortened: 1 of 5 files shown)


repo/blob/s3/s3_options.go
```diff
@@ -23,6 +23,9 @@ type Options struct {
 	SecretAccessKey string `json:"secretAccessKey" kopia:"sensitive"`
 	SessionToken    string `json:"sessionToken"    kopia:"sensitive"`
 
+	// Timeout is used as a deadline context in Minio API calls.
+	RequestTimeout int64 `json:"requestTimeout,omitempty"`
```
jkowalski, line 27: please rename the field and JSON field as RequestTimeoutSeconds

repo/blob/storage.go
```diff
@@ -412,3 +413,15 @@ func ReadBlobMap(ctx context.Context, br Reader) (map[ID]Metadata, error) {
 
 	return blobMap, nil
 }
+
+// WithRequestTimeout adds a deadline to the context if RequestTimeout is greater than 0.
+// Returns the updated context and a cancel function that should be called to free resources.
+func WithRequestTimeout(ctx context.Context, seconds int64) (context.Context, context.CancelFunc) {
+	if seconds > 0 {
+		deadline := clock.Now().Add(time.Duration(seconds) * time.Second)
+		return context.WithDeadline(ctx, deadline)
```
julio-lopez, line 422: Please use `context.WithTimeout()`
```suggestion
		return context.WithTimeout(ctx, seconds*time.Second)
```

repo/blob/s3/s3_storage.go
```diff
@@ -219,6 +228,9 @@ func (s *s3Storage) putBlob(ctx context.Context, b blob.ID, data blob.Bytes, opt
 }
 
 func (s *s3Storage) DeleteBlob(ctx context.Context, b blob.ID) error {
+	ctx, cancel := blob.WithRequestTimeout(ctx, s.RequestTimeout)
```
julio-lopez, line 231: Can this be done in a "wrapper" that implements the `blob.Storage` interface?

julio-lopez (changes requested):

repo/blob/s3/s3_options.go
arouene: Done

repo/blob/storage.go
arouene: Done

repo/blob/s3/s3_storage.go
arouene: Ok, it's done, I implemented a wrapper called StorageTimeout, can you take a look ?

arouene:
Hello @julio-lopez, can you take a look? I think the PR is good to go.

PR #4151, opened 2024-10-01
fix(repository): allow extending blob retentions

KastenMike:
Extending a blob requires full control. On the newer client styles (e.g. `option.WithCredentialsJSON(j)`), this is the default if no scope is specified. 

Before:
```
gcs_immu_test.go:90:
        	Error Trace:	/Users/admin/repo/kopia/repo/blob/gcs/gcs_immu_test.go:90
        	Error:      	Received unexpected error:
        	            	googleapi: Error 403: Access denied., forbidden
        	            	unable to extend retention period to 2024-10-01 20:02:11 +0000 UTC
```

After:
```
PASS
ok  	github.com/kopia/kopia/repo/blob/gcs	3.176s
```

### DIFF
--- a/repo/blob/gcs/gcs_storage.go
+++ b/repo/blob/gcs/gcs_storage.go
@@ -277,7 +277,7 @@ func New(ctx context.Context, opt *Options, isCreate bool) (blob.Storage, error)
 
 	var err error
 
-	scope := gcsclient.ScopeReadWrite
+	scope := gcsclient.ScopeFullControl
 	if opt.ReadOnly {
 		scope = gcsclient.ScopeReadOnly
 	}


julio-lopez (dismissed):
Hold on.

kreare:
any update?

KastenMike:
@jkowalski I see you added the generic readonly but I can't find the GCS specific PR at all (from "readonly", "read-only", "read only"), I guess it was casually added in amongst a bigger PR

What do you think about just removing the scope variation completely? It has 3 benefits:
1. unified approach amongst the providers (s3/az/gcs)
2. simplified client creation code in GCS
3. simplified [params](https://kopia.io/docs/reference/command-line/common/repository-connect-gcs/) for GCS, having 2 readonly flags is confusing imo

Then GCS would use the client default (ScopeFullControl) and the generic `readonly.NewWrapper` if the flag is set

julio-lopez (commented):
+1: LG 👍

We should get @jkowalski input before merging.

jkowalski (approved):
LGTM

jkowalski:
Thanks for the PR!

PR #4171, opened 2024-10-12
feat(server): added API to manipulate notification profiles in the UI

jkowalski:


internal/server/api_notification_profile.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return cfg, nil
+}
+
+func handleNotificationProfileDelete(ctx context.Context, rc requestContext) (any, *apiError) {
+	if err := repo.WriteSession(ctx, rc.rep, repo.WriteSessionOptions{
+		Purpose: "NotificationProfileDelete",
+	}, func(ctx context.Context, w repo.RepositoryWriter) error {
+		return notifyprofile.DeleteProfile(ctx, w, rc.muxVar("profileName"))
+	}); err != nil {
```
julio-lopez, line 48: Should this return "not found" in some cases? or does it silently succeed in those cases?

julio-lopez (approved):
+1: LG 👍

See question inline

Feel free to merge

PR #4182, opened 2024-10-17
feat(repository): role assumption for S3 storage

ashmrtn:
Adds support for AWS assume role credentials in the blob layer. This is only available if one uses the SDK, it doesn't wire up additional CLI options to access it. It does pull in a small amount of AWS-specific code


repo/blob/s3/s3_assume_role.go
```diff
@@ -0,0 +1,83 @@
+package s3
+
+import (
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	awscreds "github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sts"
```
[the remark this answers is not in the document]
julio-lopez: > to clarify, are you asking to remove just the v1 AWS SDK (as in the v2 one would be ok) or are you asking to remove all AWS SDKs (i.e. use minio's role assumption logic)?

Ideally use Minio's.
Why are tags needed when assuming the role? (for my understanding).

repo/blob/s3/s3_assume_role.go
```diff
@@ -0,0 +1,83 @@
...
+		if err != nil {
+			return &minioProvider{}, errors.Wrap(err, "NewSession")
+		}
+	}
+
+	sess, err := session.NewSession()
+	if err != nil {
+		return &minioProvider{}, errors.Wrap(err, "ParseDuration")
+	}
+
```
julio-lopez, line 43: The error messages are mixed up.

```suggestion
		roleDuration, err = time.ParseDuration(duration)
		if err != nil {
			return &minioProvider{}, errors.Wrap(err, "ParseDuration")
		}
	}

	sess, err := session.NewSession()
	if err != nil {
		return &minioProvider{}, errors.Wrap(err, "NewSession")
	}

```

repo/blob/s3/s3_options.go
```diff
@@ -19,9 +19,13 @@ type Options struct {
...
-	AccessKeyID     string `json:"accessKeyID"`
-	SecretAccessKey string `json:"secretAccessKey" kopia:"sensitive"`
-	SessionToken    string `json:"sessionToken"    kopia:"sensitive"`
+	AccessKeyID     string            `json:"accessKeyID"`
+	SecretAccessKey string            `json:"secretAccessKey" kopia:"sensitive"`
+	SessionToken    string            `json:"sessionToken" kopia:"sensitive"`
+	Tags            map[string]string `json:"tags" kopia:"sensitive"`
+	RoleARN         string            `json:"roleARN"`
+	SessionName     string            `json:"sessionName"`
+	RoleDuration    string            `json:"duration"`
```
julio-lopez, line 28: Currently trying out an alternative to make this a `time.Duration` here.
If it works, I'll push it in a separate PR.

ashmrtn: yea, I was just looking at that. I sort of just pulled this off the branch we had and pushed it. A `time.Duration` definitely makes more sense

repo/blob/s3/s3_assume_role.go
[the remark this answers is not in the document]
ashmrtn: mkay, let me poke at minio a bit (this is good learning for me too), beyond tags, it looks like it should hopefully work out of the box?

I believe tags were used to help with some metrics collection stuff. I was thinking that a good way to approach this would be to get assume role support merged and then add support for tags in a separate PR

repo/blob/s3/s3_options.go
julio-lopez: The issue is that JSON marshaling for `time.Duration` is not that usable.

For example:

```
d := 2*time.Hour + 3 + time.Minute

b, err := json.Marshal(d)

fmt.Printf("%s", b)
// output: 7260000000003
```


PTAL at #4194

repo/blob/s3/s3_assume_role.go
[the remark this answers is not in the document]
ashmrtn: I made some progress getting minio to work. Mostly it was just adjusting the parameters and calling minio's function. I don't have a role setup that I can assume right now, but I have gotten it to give me an error for one I don't have permissions to assume. I'm out for a few days but can circle back to this again when I get back

repo/blob/s3/s3_options.go
```diff
@@ -19,9 +19,16 @@ type Options struct {
...
-	AccessKeyID     string `json:"accessKeyID"`
-	SecretAccessKey string `json:"secretAccessKey" kopia:"sensitive"`
-	SessionToken    string `json:"sessionToken"    kopia:"sensitive"`
+	AccessKeyID     string            `json:"accessKeyID"`
+	SecretAccessKey string            `json:"secretAccessKey" kopia:"sensitive"`
+	SessionToken    string            `json:"sessionToken" kopia:"sensitive"`
+	Tags            map[string]string `json:"tags" kopia:"sensitive"`
+	RoleARN         string            `json:"roleARN"`
+	SessionName     string            `json:"sessionName"`
+	RoleDuration    time.Duration     `json:"duration"`
```
julio-lopez, line 28: This is probably OK, if it is not meant to be read or modified by humans.
The approach in #4194 would allow human-friendly marshaling.

ashmrtn: yea, we can pull in something like #4194 since it does make it more readable in logs and whatnot. I was mostly trying to get a minimal working set of code using just minio for the moment 😄

repo/blob/s3/s3_assume_role.go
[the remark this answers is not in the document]
ashmrtn: hmmm alright, I'm fairly sure this will work, though I haven't setup a role that I can actually try assuming with this

Another open question for me is whether the location for the region for the bucket can be the same as the assume role region. I lack deep knowledge of AWS, but it seems like there's the potential for them to be different which means there may need to be separate config options for them

jkowalski (approved):

ashmrtn:
@julio-lopez, did you have anything else that needed addressed in this PR?

repo/blob/s3/s3_options.go
```diff
@@ -21,7 +21,14 @@ type Options struct {
 
 	AccessKeyID     string `json:"accessKeyID"`
 	SecretAccessKey string `json:"secretAccessKey" kopia:"sensitive"`
-	SessionToken    string `json:"sessionToken"    kopia:"sensitive"`
+	SessionToken    string `json:"sessionToken" kopia:"sensitive"`
+
+	// Used for assume role authentication.
+	RoleARN      string        `json:"roleARN"`
+	SessionName  string        `json:"sessionName"`
+	RoleDuration time.Duration `json:"duration"`
```
julio-lopez, line 29: While this marshals correctly, it does not "roundtrip" (i.e., unmarshal back).
We can merge this, and iterate quickly to change the type to `jsonencoding.Duration`

repo/blob/s3/s3_assume_role.go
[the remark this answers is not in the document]
julio-lopez: 🤷🏼‍♂️

julio-lopez:
@Rohit-BM18 FYI

PR #4190, opened 2024-10-21
feat(snapshots): Fix for #2037 Add symlink support for .kopiaignore

### LINKED ISSUE #2037: [Feature Request]: Kopia should follow a .kopiaignore symlink.
Creating an issue to capture [this discussion](https://kopia.discourse.group/t/kopia-ignores-kopiaignore-symlink/1210) on the Kopia Discourse forum with @jkowalski. Per his reply:

> "It’s not intentional, but Kopia indeed expects this to be a file and not symlink today. Happy to review PR to fix it."

What follows is an embedding of the original post I wrote up for convenience:

---

### Behavior

Suppose I have a `.kopiaignore` file located in the root of my home directory, `~/.kopiaignore`.

In the service of keeping the exemplified use case simple, let's have the contents read:

```bash
.cache
```

If I then do:

```bash
kopia snapshot estimate $HOME
```

... we verify that the `~/.cache` dir was indeed excluded! In other words, this works as intended.

Now, let's suppose instead I have my "canonical" `.kopiaignore` located at `~/dotfiles/kopia/.kopiaignore` and I symlink that to my home directory, such that it looks like:

```bash
# ls -la ~
lrwxrwxrwx 1 user user 27 June 12 12:17 .kopiaignore -> dotfiles/kopia/.kopiaignore
```

If I again do:

```bash
kopia snapshot estimate $HOME
```

... we observe that none of the exclusions/ignores take effect!

### Commentary

While obviously the above is a bit of a contrived (simple) example, it definitely caught me off-guard in terms of a real world use case just now:

2022-06-12-122216_1745x91_scrot|690x35

Notice how, with the top snapshot, the number of files is roughly ~178K. Then, after making the above change, swapping a static file to a symlink, all of my exclusions were ignored for the subsequent snapshot (and we see a jump up of almost ~600K files and 17GB).

### Use Case

I have a `dotfiles` repository/dir (in the Git sense) that lives within my home directory, `~/dotfiles`. This directory is synced between machines that I use and/or work on regularly.

I use [GNU Stow](https://www.gnu.org/software/stow/) to "install" (symlink) these dotfiles onto machines.

Example:

```bash
pwd
# ~/dotfiles

ls -l
# alacritty
# git
# gnupg
# ...
# kopia
# ...

stow kopia
cd

ls -l
# .kopiaignore -> dotfiles/kopia/.kopiaignore
```

**TL;DR:** This symlinking approach allows me to keep my Kopia exclusions in-sync between machines without having to update/overwrite `N` different static `.kopiaignore` files.

For what it's worth, I use this pattern to great success with nearly all of my "daily driver" binaries!

mcamou:


mcamou:
The force-push was after rebasing from `master`.

mcamou:
@julio-lopez would you mind having a look at this? PRs have to be approved/merged in October.

mcamou:
@julio-lopez sorry to be a pest, but I'd really like to have this PR count towards Hacktoberfest.

mcamou:
@julio-lopez sorry to be such a pest, but the deadline for Hacktoberfest participation is tomorrow (the PR has to be approved by the end of October for it to count).

mcamou:
force-push after rebase from `master`

mcamou:
@julio-lopez Even though Hacktoberfest is over, I would really hate for this to go to waste.

fs/localfs/local_fs_test.go
```diff
@@ -22,6 +22,53 @@ type fileEnt struct {
...
+	entry, err := NewEntry(path)
+	if err != nil {
+		t.Errorf("Error when creating entry: %v", err)
+	}
+
+	if link, ok := entry.(fs.Symlink); !ok {
+		t.Errorf("entry is not a symlink: %s", path)
+	} else {
+		target, err := link.Resolve(ctx)
+		if err != nil {
```
jkowalski, line 54: please use: `require.NoError(t, err)` as is the convention

fs/ignorefs/ignorefs.go
```diff
@@ -283,8 +286,22 @@ func (d *ignoreDirectory) buildContext(ctx context.Context) (*ignoreContext, err
...
+				dotIgnoreFiles = append(dotIgnoreFiles, entry)
+
+			case fs.Symlink:
+				target, err := entry.Resolve(ctx)
+				if err != nil {
+					link, _ := entry.Readlink(ctx)
+					return nil, errors.Wrapf(err, "when resolving symlink %s of type %T, which points to %s", entry.Name(), entry, link)
+				}
+
+				if f, ok := target.(fs.File); !ok {
```
jkowalski, line 300: what about symlink->symlink->file, would it make sense to resolve symlinks in a loop until we encounter a `fs.File` ?

jkowalski:
sorry for late review, this looks very good, I left a couple comments, we can merge after those are addressed.

please also rebase to the latest master (the UI test failures should be fixed there)

mcamou:
@jkowalski Thanks for the review, I've addressed your comments. Please let me know if there's anything else.

jkowalski (approved):

jkowalski:

fs/ignorefs/ignorefs.go
```diff
@@ -271,6 +274,26 @@ func (d *ignoreDirectory) Child(ctx context.Context, name string) (fs.Entry, err
...
+		target, err := entry.Resolve(ctx)
+		if err != nil {
+			link, _ := entry.Readlink(ctx)
+			return nil, errors.Wrapf(err, "when resolving symlink %s of type %T, which points to %s", entry.Name(), entry, link)
+		}
+
+		switch t := target.(type) {
+		case fs.File:
+			return t, nil
+		case fs.Symlink:
```
julio-lopez, line 288: What prevents an infinite loop here?

fs/ignorefs/ignorefs_test.go
```diff
@@ -553,7 +615,7 @@ func walkTree(t *testing.T, dir fs.Directory) []string {
 
 			if subdir, ok := e.(fs.Directory); ok {
 				if err := walk(relPath, subdir); err != nil {
-					return err
+					t.Fatalf("%s not found in %s", relPath, subdir.Name())
```
julio-lopez, line 618: The if block can be replaced with `require.NoError(...)`

fs/localfs/local_fs_test.go
```diff
@@ -22,6 +22,50 @@ type fileEnt struct {
...
+	require.NoError(t, err)
+
+	if link, ok := entry.(fs.Symlink); !ok {
+		t.Errorf("entry is not a symlink: %s", path)
+	} else {
+		target, err := link.Resolve(ctx)
+		require.NoError(t, err)
+
+		if f, ok := target.(fs.File); !ok {
+			t.Errorf("Link does not resolve to a file: %s", path)
```
julio-lopez, line 55: ditto ... `require.True(...`

fs/localfs/local_fs_test.go
```diff
@@ -22,6 +22,50 @@ type fileEnt struct {
...
+func verifyLink(t *testing.T, path, expected string) {
+	t.Helper()
+
+	ctx := testlogging.Context(t)
+
+	entry, err := NewEntry(path)
+	require.NoError(t, err)
+
+	if link, ok := entry.(fs.Symlink); !ok {
+		t.Errorf("entry is not a symlink: %s", path)
```
julio-lopez, line 49: `require.True(t, ok, "entry is not a symlink: %s", path)`

julio-lopez (commented):
@mcamou thanks for your contribution.

See inline comments

fs/entry.go
```diff
@@ -72,16 +72,19 @@ type Directory interface {
 func IterateEntries(ctx context.Context, dir Directory, cb func(context.Context, Entry) error) error {
 	iter, err := dir.Iterate(ctx)
 	if err != nil {
-		return err //nolint:wrapcheck
+		return errors.Wrapf(err, "in fs.IterateEntries, creating iterator for directory %s", dir.Name())
```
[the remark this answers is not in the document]
mcamou: To give context regarding where the error came from (including the directory name)

fs/ignorefs/ignorefs.go
mcamou: I can't seem to find this code in current `master` (since the PR was merged - there is no `switch` in `fs/ignorefs/ignorefs.go`)

fs/localfs/local_fs_test.go
mcamou: I can't find this code in current `master` either

fs/localfs/local_fs_test.go
mcamou: Ditto, can't find this

julio-lopez:
- #2037

fs/ignorefs/ignorefs.go
julio-lopez: - Fixed #4413

PR #4194, opened 2024-10-23
feat(general): JSON marshaler helper for `time.Duration`

julio-lopez:
`time.Duration` does not have JSON (un)marshaling functions.

The helpers added here provide support for marshaling `time.Duration` as JSON. While the "ergonomics" are not the best, they seem to be workable enough.

### DIFF
--- a/internal/jsonext/jsonext.go
+++ b/internal/jsonext/jsonext.go
@@ -2,7 +2,9 @@
 package jsonext
 
 import (
+	"bytes"
 	"fmt"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


internal/jsonext/jsonext.go
```diff
@@ -0,0 +1,29 @@
+// Package jsonext defines common types with JSON marshalers.
+package jsonext
```
julio-lopez, line 2: @ashmrtn Not sure about the package name. Do you have a better suggestion?
If not, it will be easy to change later anyway.

internal/jsonext/jsonext_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func TestDurationJSONMarshaling(t *testing.T) {
+	ms := MyStruct{Timeout: jsonext.Duration{20*time.Minute + 10*time.Second}}
+
+	b, err := json.Marshal(ms)
+	require.NoError(t, err)
+	require.Equal(t, `{"timeout":"20m10s"}`, string(b))
+}
+
+func TestDurationJSONUnmarshaling(t *testing.T) {
```
jkowalski, line 25: can you also test that plain number parses correctly as nanoseconds (for compatibility with regular duration stored as a number)?

julio-lopez: Good catch. Addressed.
@jkowalski PTAL

internal/jsonext/jsonext.go
ashmrtn: ah, I'm really not too great at naming 😅 About the only other thing I can think of is `jsonmarshal`, but it's not really much more descriptive than what's already in this PR

julio-lopez: @ashmrtn `jsonmarshal` works, what about `jsonencoding` (it's a little long though), thoughts?

The stdlib's package is named `encoding/json`.

ashmrtn: `jsonencoding` is fine with me

Issue #4210, opened 2024-10-30
kopia benchmark encryption, "--repeat" multiplies memory use


Ember-ruby:
how to reproduce: 
- run `kopia benchmark encryption --repeat=2000000 --block-size=1KB`
- note kopia taking ~2GB of RAM
- 2000000KB ~= 2GB

i will note that the smaller the block size, and more repeats, the longer it takes, maybe the memory for each previous repeat is not being cleared until the program exits?

julio-lopez:
Pull requests welcome

PR #4213, opened 2024-10-30
fix(cli): avoid memory bloat in `benchmark encryption` command

julio-lopez:
Reuse the output buffer by resetting it between iterations.

- Fixes: #4210


julio-lopez:
This does NOT fix the issue.
The allocation is being done inside the encryptors, in a temporary write buffer.

PR #4218, opened 2024-10-31
feat(cli): introduce adaptive estimation mechanism

e-sumin:
Currently, to estimate amount of data to be processed, we need to iterate over all the data. This slows down hashing and also can take too much time when we have big amount of files.

This PR brings an improvement, now it's possible to use rough estimation, which is usable when we are snapshotting whole volume. In such case we can get filesystem stats (as `df` does) and use amount of data / number of files for estimation.

This new mechanism is available under the flag `./kopia snapshot create /some/path --progress-estimation-type=rough`

When using this flag, user needs to understand, that in such case any ignore rules are not taken into account and estimation could be a bit inaccurate.

In some cases it may worth to fallback to classic estimation to have more precise estimation of data to be uploaded.

This "adaptive" mechanism will also be available under the flag `kopia snapshot create /some-path --progress-estimation-type=adaptive --adaptive-estimation-threshold=10000`


cli/cli_progress.go
```diff
@@ -20,14 +21,19 @@ const (
...
+	enableProgress              bool
+	progressEstimationType      string
+	adaptiveEstimationThreshold int64
+	progressUpdateInterval      time.Duration
+	out                         textOutput
 }
 
 func (p *progressFlags) setup(svc appServices, app *kingpin.Application) {
 	app.Flag("progress", "Enable progress bar").Hidden().Default("true").BoolVar(&p.enableProgress)
+	app.Flag("progress-estimation-type", "Set type of estimation of the data to be snapshotted").Hidden().Default(snapshotfs.EstimationTypeClassic).
```
jkowalski, line 33: this and the other parameter should probably be policy-driven, so that each source can define per-source behavior. This way it will also work in the UI/Server mode where passing flags is impossible.

It is ok to have the CLI flags as ultimate overrides on top of policies. 

Having said that, this is self-consistent and we can add policies in separate PR(s).

jkowalski:
I like the approach, I think we can merge this PR as-is, but we should follow-up with a PR that adds policy-level settings for this.

In general there are few more policy-level settings which could control the estimation. Some form of initial delay (to avoid estimation altogether on trivially small directories), throttling (to slow down estimation to give hashing more priority), etc come to mind.

jkowalski (approved):

jkowalski:
@julio-lopez can you take a look and merge the PR if you're ok with that?

Shrekster (approved):
Agree with the thoughts here about making this policy driven in the future.

@e-sumin and I went over this PR internally as well, so just reposting my approval too.

Shrekster:
Also filed this future ticket based off the discussions surrounding this PR: https://github.com/kopia/kopia/issues/4228

PR #4231, opened 2024-11-05
test(kopiaui): reduce flaky failures HTML UI E2E test

julio-lopez:
Increase test (context) timeout from 2 minutes to 3 minutes.


jkowalski (approved):
pretty sure that's not the root cause, but we can give it a try

julio-lopez:
This does not help, it just takes longer to fail.
There may be a different timing issue in the UI test.

julio-lopez:
> pretty sure that's not the root cause, but we can give it a try

You are right. Mainly trying it out (shot in the dark)

PR #4234, opened 2024-11-06
feat(server): add logging of failed and successful login attempts

linust:
A minimal commit to add support for monitoring, troubleshooting, and to allow using tools like crowdsec or fail2ban. This to ease the comfort a bit on opening up access to a kopia server over internet.

When/if change is merged or a similar is created I can create instructions and templates for using said tools to protect a kopia server instance running om Linux.


linust:
Renamed PR to match regexp. (Missed this routine)

julio-lopez (approved):

Issue #4257, opened 2024-11-17
Can't connect from some machines after updating to v0.18.1 and also can't set a password


oderwat:
After upgrading to v0.18.1 I can't connect from 2 of 4 machines (that I checked so far).

The users that can't connect are relatively new and where created with the v0.17.0 server / clients.

When trying to re-set the password I get:

```
$ kopia server users set root@machine --ask-password
Enter new password for user root@machine:
Re-enter password for verification:
ERROR error setting password: unsupported hash version (0)
```

jkowalski:
can you try reconnecting on one of them?

oderwat:
I can not set the password of a user while being on the repository server itself. How should I even connect to the repository from another system to that server?

Deleting the user and adding it back did work. After that I could connect with that user from the machine that did not work anymore.

oderwat:
Just to add to that: As I downgraded to v0.17.0 the very same clients could still connect. Can it be that there is something similar to the v0.18.0 problem going on, with the PWs being encoded with a now unsupported hash?

jkowalski:
@julio-lopez this looks like an issue related to user/password hash management. Could you take a look?

bbccdd:
Same problem here. Using amd64 downgrading to v0.17.0 fixed it:

wget https://github.com/kopia/kopia/releases/download/v0.17.0/kopia_0.17.0_linux_amd64.deb
dpkg -i kopia_0.17.0_linux_amd64.deb

Thanks @oderwat for reporting and @julio-lopez for the quick fix!

julio-lopez:
@bbccdd Have you been able to confirm the fix works for you?
Thanks.

PR #4259, opened 2024-11-18
fix(server): handle unset password hash version in user profile

julio-lopez:
Fixes
- #4257

### DIFF
--- a/internal/auth/authn_repo_test.go
+++ b/internal/auth/authn_repo_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/kopia/kopia/internal/auth"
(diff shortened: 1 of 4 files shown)


jkowalski (approved):

internal/auth/authn_repo_test.go
```diff
@@ -86,7 +87,6 @@ func TestRepositoryAuthenticator(t *testing.T) {
 func verifyRepoAuthenticator(ctx context.Context, t *testing.T, a auth.Authenticator, r repo.Repository, username, password string, want bool) {
 	t.Helper()
 
-	if got := a.IsValid(ctx, r, username, password); got != want {
-		t.Errorf("invalid authenticator result for %v/%v: %v, want %v", username, password, got, want)
-	}
+	got := a.IsValid(ctx, r, username, password)
+	assert.Equal(t, want, got, "invalid authenticator result for %v/%v", username, password)
```
jkowalski, line 108: s/assert/require/

julio-lopez: No, `assert` makes sense here so other test cases are checked as well.
There is no reason to stop the test when this check fails.

PR #4358, opened 2025-01-22
fix(server): ensure server uploads "repodiag" blobs

julio-lopez:
The kopia server was not uploading any logs to the repository, because _"repodiag"_ blob uploads would always fail.

The cause was the following: When the (log) repodiag blob PUT operation was initiated, the `Context` used for this operation was already canceled.

The context used for blob uploads is passed to `repodiag.NewLogManager` when opening the repository. In the case of the kopia server, the repository is asynchronously opened in `server.Server.InitReposotoryAsync`. The context passed to `repo.Open` is canceled after the "open repository" server task completes. This issue was introduced in #1691

Change:
Use `context.WithoutCancel()` instead of the context passed when the repo "diagnoser" is created.

New tests are included to reproduce this failure and verify the fix.

Ref:
- #1691

Thanks to:
- 🎖️ @redgoat650 for reporting this issue;
- 🎖️ @aaron-kasten for the triaging, testing and reproduction of the issue.

### DIFF
--- a/internal/blobtesting/map.go
+++ b/internal/blobtesting/map.go
@@ -33,6 +33,10 @@ type mapStorage struct {
 }
 
 func (s *mapStorage) GetCapacity(ctx context.Context) (blob.Capacity, error) {
+	if err := ctx.Err(); err != nil {
+		return blob.Capacity{}, errors.Wrap(err, "get capacity failed")
+	}
+
 	if s.limit < 0 {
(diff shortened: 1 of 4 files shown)


redgoat650 (approved):
Looks good; thanks 🚀

julio-lopez:
FYI @jkowalski

Issue #4373, opened 2025-01-24
Kopia attempts to run maintenance jobs on read-only repository connection


ekauz:
Kopia v0.19.0

I have an SFTP backup repo that I'm connecting to with `"readonly": true` in the repository.config.
Kopia repeatedly tries to run maintenance tasks against it and it continuously fails because the repo is read only.

This is new to v0.19.0.

I could not find documentation for editing the config file to disable maintenance tasks.

julio-lopez:
Yeah, that's not a desired behavior and should be improved.

As a workaround, maintenance can be disabled.
The `kopia maintenance set ...` command can be used to explicitly disable both "quick" and "full" maintenance.

https://kopia.io/docs/reference/command-line/advanced/maintenance-set/

julio-lopez:
Related to: 
- #3174
- #3639

ekauz:
Thanks for the quick reply @julio-lopez I saw this command in the docs but I'm not sure how to only apply it to one repo. (I have 3 kopia repositories, only one of them is read-only)

maiswan:
I have a similar issue. I backup to a network share that isn't online 24/7. Kopia now spams notifications _literally_ every minute because it keeps trying to write to a location that isn't even read-only:

<img width="250" alt="Image" src="https://github.com/user-attachments/assets/fe1f4905-5208-4f2e-af69-eb9f448f148e" />

![Image](https://github.com/user-attachments/assets/f52f074c-477a-4ed4-93ef-2cc4d8fe5f2b)

**Suggestion**: it'll be wonderful if it is possible to configure a schedule for maintenance just like snapshotting.

PR #4408, opened 2025-02-13
fix(server): prevent maintenance task on read-only repos

julio-lopez:
Avoid starting a maintenance task on the server when the repository connection (configuration) is read-only.

- Fixes: #4373


redgoat650 (commented):
LGTM

PR #4421, opened 2025-02-21
fix(snapshots): fix accounting for symlinks in directory summary

redgoat650:
Fix propagation of symlink counts to parent directories. Add tests to check the directory summary has correct counts for symlinks and files.

Before the fix, symlink counts would not be passed to parent directories, resulting in summary data that do not reflect the count of symlinks in the directory tree.


julio-lopez (approved):
🥇 Thanks for doing this @redgoat650 
✅ LGTM

PR #4444, opened 2025-03-06
Compute stats during snapshot comparison using kopia diff

Rohit-BM18:
Added functionality to calculate aggregate statistics when comparing what's changed between snapshots using kopia diff 

Statistics collected during snapshot diff computation includes:
- files added/removed/modified
- dirs added/removed/modified
- files/dirs with metadata changes but same underlying content (OID)

Testing approach:
- Added a test for verifying stats collected when comparing two directories with the same objectID but metadata changes across snapshots (dir mode, dir mod time, dir owner, etc), expectation is all the appropriate dir stats fields are updated.
- Added another test for verifying stats collected when comparing two directories with similar file contents but the metadata for the files have changed between snapshots but content remains unchanged. Expectation is all the relevant file level stats fields are updated.
- Existing tests have been updated due to stats now being printed in addition to previous output.

### DIFF
--- a/cli/command_diff.go
+++ b/cli/command_diff.go
@@ -75,7 +75,7 @@ func (c *commandDiff) run(ctx context.Context, rep repo.Repository) error {
 			return errors.Wrap(err, "error marshaling computed snapshot diff stats")
 		}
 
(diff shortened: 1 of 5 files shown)


julio-lopez:
@jkowalski FYI

PR #4460, opened 2025-03-19
feat(snapshots): add hardlink/inode support

kinghrothgar:
I would like to add hardlink support to Kopia (https://github.com/kopia/kopia/issues/544). I believe the first step toward this is making Kopia aware of inodes, similar to what was attempted in this [closed PR](https://github.com/kopia/kopia/pull/1013). I have started by re-implementing that PR here in a (very) WIP, but I have some questions before moving forward with a full featured implementation.

The primary reason is to enable restoring snapshots with many hardlinks. Currently, all hardlinks result in file duplication during restore, requiring significantly more space than was originally used on the snapshotted FS. In the above inode PR, it was stated that storing knowledge of hardlinks in the repository is not needed:

> I don't understand the use of storing inodes in repository, we have a file identifier already oid which is globally unique. If we want to preserve hardlinks on restore, we can do the reverse process there - build a map of oid to inodes of restored files and if we find oid reuse we do a hard-link to a previously-written file.

If I understand the oid correctly, I believe this is in error. The oid for files with the same data would be identical, and thus they would be restored as hardlinks even if they were not originally hardlinks. ~~From reading the discussion on this [PR](https://github.com/kopia/kopia/pull/3643), I see that I will need to do this in a way that supports ""other non-core attributes which are currently unsupported."~~

A secondary reason to make Kopia hardlink-aware is that if it encounters a file that is a hardlink to a previously processed file, it could skip hashing that file—massively increasing snapshot speed on systems with many hardlinks. One caveat is that hardlinks would be detected by matching inodes (on POSIX systems) or File IDs (on Windows), which are not guaranteed to be unique across time. This is not an issue when backing up a static filesystem (e.g., an LVM, ZFS, or VSS snapshot), but we cannot assume that will always be the case. As such, if implemented, I believe this feature should be behind a flag with a warning. The time savings on initial snapshots of large heavily linked systems would still be huge (many days on my dataset).

Looking forward to feedback and if there's anything else for me to consider be for moving forward!

### DIFF
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -20,6 +20,7 @@ type Entry interface {
 	os.FileInfo
 	Owner() OwnerInfo
 	Device() DeviceInfo
+	HardLinkInfo() HardLinkInfo
 	LocalFilesystemPath() string // returns full local filesystem path or "" if not a local filesystem
 	Close()                      // closes or recycles any resources associated with the entry, must be idempotent
 }
@@ -34,6 +35,12 @@ type OwnerInfo struct {
 type DeviceInfo struct {
 	Dev  uint64 `json:"dev"`
 	Rdev uint64 `json:"rdev"`
+}
+
+// HardLinkInfo describes the hardlinks of a filesystem entry.
+type HardLinkInfo struct {
+	UniqId uint64 `json:"uniqid"`
+	NLink  uint64 `json:"nlink"`
 }
 
 // Reader allows reading from a file and retrieving its up-to-date file info.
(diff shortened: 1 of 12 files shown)


jkowalski:
thanks for doing this, I think we should implement inode-based optimization, I'll let CI run, please also add some tests

ckolumbus:
any progress on this one? Esp. windows support would be very helpful (though i saw the comment about the problems with implementing this)

kinghrothgar:
@ckolumbus this has been dormant for a while as you can see, but now with the help of AI agents, I plan to start tackling this again hopefully as building out robust test for this will be much more trivial now. Unfortunately, I don't think the first pass would support Windows, but I am planning on all the ground work that is done in the initial PRs being done in a way that Windows could in theory be implemented in a follow up.

PR #4472, opened 2025-03-25
Increase rclone startup timeout

### LINKED ISSUE #2573: unable to start rclone: timed out waiting for rclone to start
1/3 `snapshot create` execution I got `unable to start rclone: timed out waiting for rclone to start`
**Version**: kopia:20221110.0.161523
**Backend**:  rclone mouting gdrive.

command executed
```bash
docker run --rm --hostname $(hostname) \
-e KOPIA_PASSWORD=xxxxxxxxxxx \
    -v /mnt/user/appdata/kopia/config:/app/config \
    -v /mnt/user/appdata/kopia/cache:/app/cache \
    -v /mnt/user/appdata/kopia/logs:/app/logs \
    -v /mnt:/mnt \
    -v /boot/config/plugins/rclone/.rclone.conf:/app/rclone/rclone.conf \
kopia/kopia:20221110.0.161523 snapshot create \
/mnt/user/appdata
```

### LINKED ISSUE #4471: Configuring storage fails because the rclone timeout is too low
(Using KopiaUI 0.19.0 on Windows), when configuring with rclone, it failed with "Connect Error: INTERNAL: internal server error: unable to start rclone: timed out waiting for rclone to start". My rclone configuration takes several minutes to connect, and apparently the timeout is too low for this to work.

See #2573.

93Pd9s8Jt:
Increase rclone startup timeout from 15 seconds to 15 minutes
Should fix #2573


julio-lopez (changes requested):
Instead of changing the fixed value, make the value configurable,  persist it in the configuration and use the same default value as before.

93Pd9s8Jt:
> Instead of changing the fixed value, make the value configurable, persist it in the configuration and use the same default value as before.

To my understanding, it is already configurable:
https://github.com/kopia/kopia/blob/master/repo/blob/rclone/rclone_storage.go#L348-L351

However, this is useless on the UI builds, for which you can't modify the configuration (at least, I couldn't see how to).
In addition, a timeout should be a lower bound, and rclone configurations can easily exceed 15 seconds. It should not block me from using the app.

93Pd9s8Jt:
@julio-lopez , please could you have another look at this?

julio-lopez:
The rclone storage provider has a `StartupTimeout` option that is not exposed through the CLI. The right approach is to expose this option through the CLI. Once it is exposed through the CLI, then it could be exposed through the UI as well. Modifying the default value is not desirable.

93Pd9s8Jt:
@julio-lopez I would say modifying the default value is desirable, because it clearly blocks me and others from using kopia. If it is higher, I do not see any downsides major enough to warrant blocking people from using the app.
Whilst I agree it would be desirable to expose the option, my knowledge of go (i.e. none) isn't good enough to implement that. So this is therefore necessary as an intermeditary stopgap solution. If you disagree, please clo

repo/blob/rclone/rclone_storage.go
```diff
@@ -32,7 +32,7 @@ const (
 	defaultRCloneExe = "rclone"
 
 	// rcloneStartupTimeout is the time we wait for rclone to print the https address it's serving at.
-	rcloneStartupTimeout = 15 * time.Second
+	rcloneStartupTimeout = 15 * time.Minute
```
jkowalski, line 35: the default of 15 minutes is way too high, I would be ok with 1 minute

+1 to what Julio says about making it configurable. Configuration options like this are actually quite easy to expose in the CLI and UI.

93Pd9s8Jt: @jkowalski 
My current rclone config takes 3 minutes and 10 seconds to start up. (I realise this may not be normal, but nevertheless.) So I wouldn't mind 5 minutes, but 1 minute wouldn't help me (though might resolve others' problems). 

I believe you when you say it would be easy, but I'm not confident enough in my own programming knowledge to attempt this with no knowledge of go.

jkowalski: Adding a field in UI configuration is easily copy-pasteable:

You need to change:

https://github.com/kopia/htmlui/blob/main/src/components/SetupRepositoryRclone.jsx

Adding `OptionalNumberField` for a field named `startupTimeout` should do the trick.

https://github.com/kopia/htmlui/blob/1e35b648ac5a9444dce2d07fdbede9abb78daf83/src/components/SetupRepositorySFTP.jsx#L62C17-L62C103 is an example.

Instructions for testing UI changes are in the readme file.

93Pd9s8Jt: Okay, that does look easy enough to implement, but I can't get it to build on my system.

jkowalski: you need to install `npm` from NodeJS and then `npm install` in `htmlui` directory followed by `npm run build` should do the trick.

`npm run serve` should launch the UI for testing.

93Pd9s8Jt: So the readme says to run `go run . server start --insecure --without-password --disable-csrf-token-checks --log-level=debug  --config-file=$HOME/.config/kopia/disabled/repository-dev.config` to start the server (running in kopia directory), which works and serves a webpage on port 51515.
Then I ran npm etc. and it served on port 3000 and failed to load.

93Pd9s8Jt: It looks like running the server builds with the html from htmlui?

93Pd9s8Jt:
Am I correct in understanding (given #4543 was merged) that the change needs to be made to htmlui rather than this repository, and so this pull request can be closed?

Issue #4341, opened 2025-01-09
export command cleanup. missing newline on stdout output. non functional commands. mismatch in docs


taltamir:
#4020 added export command. there are some minor things leftover to resolve

1. when exporting a policy to stdout instead of a file, it is missing a newline at the end. which causes graphical glitches in bash as per image below:
![missing-newline](https://github.com/user-attachments/assets/93bc2607-50b1-4fd7-8349-0a6c630aae2e)

#4476 adds new line after output

---

2. The documentation (both online and in --help) is incorrect.

both say you can use 4 methods to specify a target:
a. `user@host` - does not work.
b. `@host` - does not work
c. `user@host:path` - works
d. `-global` (online docs) or `--global` (--help documentation) - the correct one is using two dashes.

----

3. the online docs at:
https://kopia.io/docs/reference/command-line/common/policy-export/
need to be modified with current info. as various things are outdated or incorrect there. it should just copy paste the info from --help after fixing it.

as of v0.18.2 --help outputs:
```
usage: kopia policy export [<flags>] [<target>...]
Exports the policy to the specified file, or to stdout if none is specified.

Flags:
      --[no-]help             Show context-sensitive help (also try --help-long and
                              --help-man).
      --[no-]version          Show application version.
      --log-file=LOG-FILE     Override log file.
      --log-dir="/home/t/.cache/kopia"  
                              Directory where log files should be written. ($KOPIA_LOG_DIR)
      --log-level=info        Console log level
      --file-log-level=debug  File log level
      --[no-]help-full        Show help for all commands, including hidden
      --config-file="repository.config"  
                              Specify the config file to use ($KOPIA_CONFIG_PATH)
  -p, --password=PASSWORD     Repository password. ($KOPIA_PASSWORD)
      --[no-]persist-credentials  
                              Persist credentials ($KOPIA_PERSIST_CREDENTIALS_ON_CONNECT)
      --[no-]use-keyring      Use Gnome Keyring for storing repository password.
                              ($KOPIA_USE_KEYRING)
      --to-file=TO-FILE       File path to export to
      --[no-]overwrite        Overwrite the file if it exists
      --[no-]global           Select the global policy.

Args:
  [<target>]  Select a particular policy ('user@host','@host','user@host:path' or a local
              path). Use --global to target the global policy.

```

#4479 partially addresses documentation issues mentioned in 2 and 3 by making it show double dash instead of em-dash. em-dash is incorrect.

taltamir:
[...]

PR #4479, opened 2025-03-31
fix(site): escape flags with backticks

blenderfreaky:
In the generated markdown docs, flags like `--foo` inside help texts currently get pretty-printed as `–foo` with an em-dash.

This PR applies backticks via a regex replacement (run by `cli2md`), so that they appear as `--foo` in the docs but remain as --foo (without backticks) in the CLI output.

See also #4341

### DIFF
--- a/tools/cli2md/cli2md.go
+++ b/tools/cli2md/cli2md.go
@@ -32,11 +32,12 @@ const (
 	dirMode = 0o750
 )
 
+// The group at the start is to avoid matching things like "in-place"
+var escapeFlagsRe = regexp.MustCompile(`(^|\s)(--?\w+[-\w]*)`)
+
 // escapeFlags escapes command-line flag references in help text by wrapping them in backticks
 func escapeFlags(text string) string {
-	// The group at the start is to avoid matching things like "in-place"
-	re := regexp.MustCompile(`(^|\s)(-[-\w]+)`)
-	return re.ReplaceAllString(text, "$1`$2`")
+	return escapeFlagsRe.ReplaceAllString(text, "$1`$2`")
 }
 
 //nolint:gochecknoglobals
--- a/tools/cli2md/cli2md.go
+++ b/tools/cli2md/cli2md.go
@@ -32,10 +32,10 @@ const (
 	dirMode = 0o750
 )
 
-// The group at the start is to avoid matching things like "in-place"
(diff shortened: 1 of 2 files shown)


tools/cli2md/cli2md.go
```diff
@@ -31,6 +32,13 @@ const (
 	dirMode = 0o750
 )
 
+// escapeFlags escapes command-line flag references in help text by wrapping them in backticks
+func escapeFlags(text string) string {
```
julio-lopez, line 36: Please add a test for `escapeFlags`

julio-lopez (changes requested):

julio-lopez (changes requested):

tools/cli2md/cli2md.go
```diff
@@ -31,6 +32,13 @@ const (
...
+	// The group at the start is to avoid matching things like "in-place"
+	re := regexp.MustCompile(`(^|\s)(-[-\w]+)`)
```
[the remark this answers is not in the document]
blenderfreaky: Good catch, it does infact match `---` (this actually comes up once in snapshot-copy-history)

Your regex looks good and all tests pass with it. Don't see any issues with it

tools/cli2md/cli2md.go
```diff
@@ -31,6 +32,14 @@ const (
 	dirMode = 0o750
 )
 
+// The group at the start is to avoid matching things like "in-place"
```
julio-lopez, line 35:
```suggestion
// The group at the start is to avoid matching things like "in-place".
```

tools/cli2md/cli2md.go
```diff
@@ -31,6 +32,14 @@ const (
...
+var escapeFlagsRe = regexp.MustCompile(`(^|\s)(--?\w+[-\w]*)`)
+
+// escapeFlags escapes command-line flag references in help text by wrapping them in backticks
```
julio-lopez, line 38:
```suggestion
// escapeFlags escapes command-line flag references in help text by wrapping them in backticks.
```

julio-lopez (changes requested):

julio-lopez (approved):

blenderfreaky:
Thanks for the merge! ^^

PR #4484, opened 2025-04-02
refactor(cli): misc cleanups in internal/diff

julio-lopez:
- remove spurious log message, duplicates what is being sent to the output;
- nit: update struct comment;
- consolidate code via a parameterized function;
- rename variable for clarity, it makes it easier to understand by avoiding negative condition check;
- remove unused return value;
- replace if blocks with switch.


julio-lopez:
@Rohit-BM18 PTAL

Rohit-BM18:
> @Rohit-BM18 PTAL

Looks good to me, thanks for refactoring the code. `compareMetadata` function reduces duplication of code and is easier to read. I will keep this pattern in mind for the future.

PR #4498, opened 2025-04-12
build(cli): set build info version and info when not set

julio-lopez:
Sets `repo.BuildInfo` and `repo.BuildVersion` when they are not specified via link flags.

The behavior for binaries built via CI and Make remains the same.

### DIFF
--- a/repo/buildinfo.go
+++ b/repo/buildinfo.go
@@ -1,0 +1,80 @@
+package repo
+
+import (
+	stdlib "log"
+	"runtime/debug"
+	"strings"
+)
+
+// Kopia's build information.
+//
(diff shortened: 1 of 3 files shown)


repo/buildinfo.go
```diff
@@ -0,0 +1,80 @@
...
+	"runtime/debug"
+	"strings"
+)
+
+// Kopia's build information.
+//
+//nolint:gochecknoglobals
+var (
+	BuildInfo       = ""
+	BuildVersion    = ""
```
julio-lopez, line 14: Now empty by default, that was not the case before.

julio-lopez:
@jkowalski FYI

PR #4500, opened 2025-04-13
chore(ci): bump golangci-lint to v2.1.2

mmorel-35:


jkowalski:
thanks for doing this!

.golangci.yml
```diff
@@ -1,87 +1,83 @@
...
+
+  default: all
   disable:
+    - gochecknoglobals
     - depguard
     - exhaustruct
-    - exportloopref
+    - exptostd
+    - forcetypeassert
+    - funcorder
```
jkowalski, line 78: can we keep "exposed" and "funcorder" as enabled? Could be next PR, but maintaining consistency would be great.

.golangci.yml
```diff
@@ -1,87 +1,81 @@
...
-    rules:
-      - (*go.uber.org/zap.SugaredLogger).With
-  maligned:
-    suggest-new: true
-  misspell:
-    locale: US
-
+version: "2"
 linters:
-  enable-all: true
```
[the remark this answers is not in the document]
ldez: There are 2 POV:
- using `default: all` (previously `enable-all`), but to use a fixed version of golangci-lint inside the CI to have something stable.
- using `default: none` (previously `disable-all`), but you have to dig yourself into the changelog to find useful new linters.

There is no right or wrong choice, it depends on the project, the team, etc.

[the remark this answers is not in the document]
ldez: I mainly use `default: all`, but I also have some projects that use `default: none` or `default: standard`.

This is not a recommendation, this is not an official statement, this is just a POV.

The choice depends on context. I don't have the context of this project, so I can't provide any suggestions or recommendations.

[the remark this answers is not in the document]
ldez: FYI, I sent you a DM on the Gopher Slack.

.golangci.yml
```diff
@@ -90,90 +82,123 @@ linters:
...
+      - text: "unnamedResult:"
+        linters:
+          - gocritic
+      - linters:
+          - cyclop
+        path: (.+)_test\.go
+      - linters:
+          - exptostd
+          - funcorder
+        path: (.+)\.go
```
julio-lopez, line 186: Please move these back to the `linters.disable` list.
This exclusion rule achieves the same as having these linters in the "disable" list, and makes it harder to spot that they are disabled.

.golangci.yml
julio-lopez: > I have enabled them but exluded them through `linters.exclusion.rules`.

Please revert that change, it makes it harder to track which linters are disabled.
The disabled linters should be included in this list.

IIUC, both `exptostd` and `funcorder` are new linters. It is OK to keep them disabled for now and enable them separately.

`funcorder` will require a large number of changes, and it is not clear (yet) how much benefit it will provide, if any.

julio-lopez (changes requested):

.golangci.yml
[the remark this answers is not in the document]
julio-lopez: The current approach in kopia is:
- to enable all linters by default;
- to disable the ones not in use (which makes it explicit and visible which ones are not being used);
- to keep the golangci-lint version fixed, to have CI stability.

The golangci-lint upgrades are explicit. When that causes CI breakages (due to new linters or changes in default behavior) then those are handled on a case by case basis.

That approach should be maintained.

julio-lopez:
@mmorel-35 To facilitate and speed up the review process, please push separate commits with the new changes when this PR, instead of squashing everything into a single commit.
All the commits will be squashed at the end before merging the PR.

.golangci.yml
```diff
@@ -90,90 +84,119 @@ linters:
...
+          - testifylint
+      - text: "fieldalignment:"
+        linters:
+          - govet
+      - text: "shadow:"
+        linters:
+          - govet
+      - text: "unnamedResult:"
+        linters:
+          - gocritic
```
julio-lopez, line 181: @mmorel-35 This is new, right?

.golangci.yml
```diff
@@ -90,90 +84,119 @@ linters:
...
+      - linters:
+          - cyclop
+        path: (.+)_test\.go
```
julio-lopez, line 184: this could probably be included in the list of linter for the rule with:
```yaml
- path: _test\.go|testing|tests|test_env|fshasher|fault
```

julio-lopez (approved):
🥇Thanks for doing this.

.golangci.yml
julio-lopez: Here's the current rationale, but it can definitely be revisited and refined.
- The paths mentioned in that rule are all packages (or files) only used in testing.
- `cyclop` (and `gocyclo`) end up being being noisy with test functions.

Now, given that this is only being applied to the narrower scope already and is not flagging anything, it makes sense to leave it here.

Thanks for the feedback 😉

PR #4506, opened 2025-04-17
chore(ci): UTC commit date

julio-lopez:
This is only used in unofficial regular builds and not in tagged releases.


julio-lopez:
@jkowalski FYI

PR #4507, opened 2025-04-17
chore(general): minor cleanups and other nits

julio-lopez:
- use `uint8` for clarity
- unexport `writeContentAsyncAndVerify`
- fix typo in test function names
- remove commented interface functions
- use `atomic.Int32`
- cleanups in socket server activation test
- leverage stdlib's `maps` and `slices` packages
- leverage `maps.Values`

PR #4509, opened 2025-04-17
chore(ci): enable exptostd linter

julio-lopez:
- #4500
- #4507

### DIFF
--- a/internal/metrics/metrics_thresholds.go
+++ b/internal/metrics/metrics_thresholds.go
@@ -2,12 +2,10 @@ package metrics
 
 import (
 	"time"
-
-	"golang.org/x/exp/constraints"
 )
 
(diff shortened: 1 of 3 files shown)


internal/metrics/metrics_thresholds.go
```diff
@@ -1,13 +1,12 @@
...
 import (
+	"cmp"
 	"time"
-
-	"golang.org/x/exp/constraints"
 )
 
 // Thresholds encapsulates a set of bucket thresholds used in Summary[T].
-type Thresholds[T constraints.Float | constraints.Integer] struct {
+type Thresholds[T cmp.Ordered] struct {
```
julio-lopez, line 9: Note: this is not equivalent, in also includes `string`.
It makes sense to revert it and add a `//nolint` annotation instead.
Ditto below.

PR #4519, opened 2025-04-24
refactor(general): cleanup a few tests

julio-lopez:
Cleanup robustness tests and `local_fs_test.go`

- Use `require` helpers
- Use `testing.T` helpers

"Mechanical" changes.

Note change in functionality: The use of `require` helpers stops tests once a check fails. Before, various checks were using `t.Error`, which fails the test but allows the test to continue its execution.


julio-lopez:
@Rohit-BM18 FYI

Rohit-BM18:
> @Rohit-BM18 FYI

Looks good to me

PR #4528, opened 2025-04-27
refactor(providers): change rclone's `StartupTimeout` type to Duration

julio-lopez:
This allows persistingand reading back the timeout as a duration.

~~The name reflects the unit used for the timeout.~~

There are be no backwards compatibility issues, since the field is
not currently set, so it is not persisted in the configuration file.

### DIFF
--- a/repo/blob/rclone/rclone_options.go
+++ b/repo/blob/rclone/rclone_options.go
@@ -1,21 +1,22 @@
 package rclone
 
 import (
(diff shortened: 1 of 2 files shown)


jkowalski (approved):

julio-lopez:
@jkowalski making it a duration is probably a better approach.

PTAL (again), thanks.

PR #4535, opened 2025-04-30
feat(providers): client certificate authentication for Azure Blob

julio-lopez:
Allow the use of a client certificate when authenticating to an Azure Blob storage provider.

Tests included.

Credit: @DeepikaDixit


julio-lopez:
FYI: @Rohit-BM18 @DeepikaDixit

PR #4536, opened 2025-04-30
fix(general): avoid removing error cause

julio-lopez:
Credit:  @alexvbg  Aleksandr Samarin <alexvbg@users.noreply.github.com>


julio-lopez:
FYI: @Rohit-BM18

PR #4546, opened 2025-05-01
fix(providers): human-friendly persistence of role durantion

julio-lopez:


redgoat650 (approved):
Looks good 👍

repo/blob/s3/s3_options.go
```diff
@@ -3,6 +3,7 @@ package s3
 import (
 	"time"
 
+	"github.com/kopia/kopia/internal/jsonencoding"
```
ashmrtn, line 6: will this cause problems for callers? The type is technically in a package under `internal`, so I'm not sure if callers will be able to name the type to cast to it when making the config. In other words, I don't know if the below will compile

```go
opts := Options{
  // I don't think external callers can import jsonencoding
  RoleDuration: jsonencoding.Duration(5 * time.Second),
}
```

julio-lopez: @ashmrtn I see what you mean.
It's not an issue if the caller is inside kopia, however it is a type in the "exposed" interface for the package.

The package could be moved to `repo/jsonencoding`, does that work?

ashmrtn: yea, that would work. Do you want me to spin up a PR moving the package? We could merge it after this one if that makes it easier

ashmrtn (approved):

PR #4559, opened 2025-05-05
feat(cli): helper functions to get preceding snapshots for `diff`

Rohit-BM18:
Added helper functions to find preceding snapshots.

Part of the planned enhancements for the `diff` command.

Added the helper methods to fetch preceding snapshots. These helpers will be consumed to enhance the CLI for the existing `diff` command implementation.
- Method to fetch the immediately preceding snapshot manifest given a snapshot ID
- Method to fetch the two latest snapshot manifests for a given source

Also added relevant unit tests for the above methods:

- In order to test `GetPrecedingSnapshot`, first the method is invoked with a wrong `snapshotID` and an error is expected. Next a snapshot is saved and the method is invoked again, since there is only a single snapshot in the repo, an error is expected. Subsequently more snapshots are saved and the method is invoked and the immediately preceding snapshot based on timestamp is expected
- In order to test `GetTwoLatestSnapshotsForASource`, a similar strategy is used. The method is invoked with insufficient snapshots stored in the repo and errors are expected. Eventually more snapshots are saved and when the method is invoked the two most recent manifests for snapshot based on start time are expected

### DIFF
--- a/internal/diff/diff_test.go
+++ b/internal/diff/diff_test.go
@@ -554,22 +554,22 @@ func TestGetTwoLatestSnapshots(t *testing.T) {
 	snapshotSrc := getSnapshotSource()
 	manifests := getManifests()
 
-	_, err := diff.GetTwoLatestSnapshotsForASource(ctx, env.RepositoryWriter, snapshotSrc)
-	require.Error(t, err)
+	_, _, err := diff.GetTwoLatestSnapshotsForASource(ctx, env.RepositoryWriter, snapshotSrc)
+	require.Error(t, err, "expected error as there aren't enough snapshots to get the two most recent snapshots")
 
 	initialSnapshotManifestID := mustSaveSnapshot(t, env.RepositoryWriter, manifests["initial_snapshot"])
-	_, err = diff.GetTwoLatestSnapshotsForASource(ctx, env.RepositoryWriter, snapshotSrc)
-	require.Error(t, err)
+	_, _, err = diff.GetTwoLatestSnapshotsForASource(ctx, env.RepositoryWriter, snapshotSrc)
+	require.Error(t, err, "expected error as there aren't enough snapshots to get the two most recent snapshots")
 
 	intermediateSnapshotManifestID := mustSaveSnapshot(t, env.RepositoryWriter, manifests["intermediate_snapshot"])
 
 	var expectedManifestIDs []manifest.ID
 	expectedManifestIDs = append(expectedManifestIDs, initialSnapshotManifestID, intermediateSnapshotManifestID)
 
-	gotManifests, err := diff.GetTwoLatestSnapshotsForASource(ctx, env.RepositoryWriter, snapshotSrc)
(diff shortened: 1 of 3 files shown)


internal/diff/diff_test.go
```diff
@@ -469,3 +474,144 @@ func TestCompareIdenticalDirectoriesWithDiffDirectoryMetadata(t *testing.T) {
...
+
+	return manifests
+}
+
+// Tests GetPrecedingSnapshot function
+//   - GetPrecedingSnapshot with an invalid snapshot id and expect an error;
+//   - Add a snapshot, expect an error from GetPreceedingSnapshot since there is
+//     only a single snapshot in the repo;
+//   - Subsequently add more snapshots and GetPreceedingSnapshot theimmediately
+//     preceding with no error.
```
julio-lopez, line 534: @Rohit-BM18 Added comments to this test and the one below. PTAL.

internal/diff/diff_test.go
```diff
@@ -469,3 +474,144 @@ func TestCompareIdenticalDirectoriesWithDiffDirectoryMetadata(t *testing.T) {
...
+
+	intermediateSnapshotManifestID := mustSaveSnapshot(t, env.RepositoryWriter, manifests["intermediate_snapshot"])
+	gotManID, err := diff.GetPrecedingSnapshot(ctx, env.RepositoryWriter, string(intermediateSnapshotManifestID))
+	require.NoError(t, err)
+	require.Equal(t, initialSnapshotManifestID, gotManID.ID)
+
+	latestSnapshotManifestID := mustSaveSnapshot(t, env.RepositoryWriter, manifests["latest_snapshot"])
+	gotManID2, err := diff.GetPrecedingSnapshot(ctx, env.RepositoryWriter, string(latestSnapshotManifestID))
+	require.NoError(t, err)
+	require.Equal(t, intermediateSnapshotManifestID, gotManID2.ID)
```
julio-lopez, line 554: What is the expected behavior when a single snapshot is added for a different _snapshot source_ (`{host,user,path}` tuple)? when:
- getting the 2 most recent snapshots for the source that has a single snapshot
- getting the 2 most recent snapshots for the source that has more than one snapshot

Can you create another PR with these test cases please?

Thanks.

internal/diff/diff_test.go
```diff
@@ -469,3 +474,144 @@ func TestCompareIdenticalDirectoriesWithDiffDirectoryMetadata(t *testing.T) {
...
+}
+
+// First call GetTwoLatestSnapshots with insufficient snapshots in the repo and
+// expect an error;
+// As snapshots are added, GetTwoLatestSnapshots is expected to return the
+// manifests for the the two most recent snapshots for a the given source.
```
julio-lopez, line 560:
```suggestion
// manifests for the two most recent snapshots for a the given source.
```

julio-lopez (approved):

internal/diff/diff_test.go
Rohit-BM18: When either of the methods are invoked on a source with insufficient snapshots, it should result in an error.
PR incoming with these tests.

Rohit-BM18: https://github.com/kastenhq/internal-kopia/pull/300

PR #4560, opened 2025-05-07
fix(cli): fix crash when doing `kopia policy show <path>`

ton31337:


jkowalski (approved):

jkowalski:
@ton31337 looks like there are linter failures, should be easy to fix.

ton31337:
Dzieki, should be fixed now :)

PR #4617, opened 2025-05-29
feat(cli): Add `--max-sources` for `kopia snapshot list`

ton31337:
We have around 70-100k snapshots with around 10k sources and this takes almost 8-10 minutes to list.

We noticed a couple times already that repository status reports everything is perfect, but doing snapshot list returns an error about missing blobs, etc.

With this new CLI knob, we could run it more regularly to catch those errors instead of abusing the system with unnecessary load (loading everything).

### DIFF
--- a/cli/command_snapshot_list.go
+++ b/cli/command_snapshot_list.go
@@ -33,6 +33,7 @@ type commandSnapshotList struct {
 	snapshotListShowIdentical        bool
 	snapshotListShowAll              bool
 	maxResultsPerPath                int
+	maxSnapshots                     int
 	snapshotListTags                 []string
 	storageStats                     bool
 	reverseSort                      bool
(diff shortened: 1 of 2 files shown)


cli/command_snapshot_list.go
```diff
@@ -56,6 +57,7 @@ func (c *commandSnapshotList) setup(svc appServices, parent commandParent) {
 	cmd.Flag("reverse", "Reverse sort order").BoolVar(&c.reverseSort)
 	cmd.Flag("all", "Show all snapshots (not just current username/host)").Short('a').BoolVar(&c.snapshotListShowAll)
 	cmd.Flag("max-results", "Maximum number of entries per source.").Short('n').IntVar(&c.maxResultsPerPath)
+	cmd.Flag("max-sources", "Maximum number of sources.").Short('s').IntVar(&c.maxSources)
```
jkowalski, line 60: that's not the right term, "source" generally refers to a source directory, what you're doing is limiting the number of snapshot manifests.

jkowalski: sure, let's do that, I'm wondering if it makes sense to plumb it all the way through the functions that list snapshots in the repository, but that's much bigger change.

jkowalski: also, this would need to have tests before we can merge

ton31337: Changed and added a test.

paumic: @jkowalski is there anything else we should add before we can merge it? Having this feature would make our lives a lot easier 😊

paumic:
@jkowalski I understand that you might be quite busy but if you could take a quick look to review and/or approve it would be highly appreciated.

stanynaite:
Hi @jkowalski! Would you be able to approve the pull request or let as know what needs to be adjusted? Thank you!

ton31337:
@julio-lopez, any chance to review?

PR #4622, opened 2025-05-31
refactor(general): guard access to objectWriter.contentWriteError

julio-lopez:
- add checklocks annotation for `contentWriteErrorMutex`
- guard access to `objectWriter.contentWriteError`

### DIFF
--- a/repo/object/object_manager.go
+++ b/repo/object/object_manager.go
@@ -86,7 +86,9 @@ func (om *Manager) NewWriter(ctx context.Context, opt WriterOptions) Writer {
 	}
 
 	w.buffer.Reset()
+	w.contentWriteErrorMutex.Lock()
 	w.contentWriteError = nil
+	w.contentWriteErrorMutex.Unlock()
 
 	return w
 }
--- a/repo/object/object_writer.go
+++ b/repo/object/object_writer.go
@@ -92,7 +92,8 @@ type objectWriter struct {
 	asyncWritesWG        sync.WaitGroup
 
 	contentWriteErrorMutex sync.Mutex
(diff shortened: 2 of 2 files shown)


repo/object/object_writer.go
```diff
@@ -289,8 +295,14 @@ func (w *objectWriter) checkpointLocked() (ID, error) {
 	// wait for any in-flight asynchronous writes to finish
 	w.asyncWritesWG.Wait()
 
-	if w.contentWriteError != nil {
-		return EmptyID, w.contentWriteError
+	var err error
+
+	w.contentWriteErrorMutex.Lock()
+	err = w.contentWriteError
```
julio-lopez, line 301: Guarding here the read of `w.contentWriteError` with the mutex because the error can be set outside of the `w.mu` mutex. This allows to reason about 'happened before / after'.

repo/object/object_manager.go
```diff
@@ -86,7 +86,7 @@ func (om *Manager) NewWriter(ctx context.Context, opt WriterOptions) Writer {
 	}
 
 	w.buffer.Reset()
-	w.contentWriteError = nil
+	w.setError(nil)
```
julio-lopez, line 89: It appears that resetting `contentWriteError` under `w.contentWriteErrorMutex` is not strictly necessary, given that at this point there _"should be no other references"_ to `w`. 🤷🏼

repo/object/object_writer.go
julio-lopez: @jkowalski maybe I'm missing something here, thoughts?

PR #4623, opened 2025-05-31
refactor(general): continue in loop and reduce indentation

julio-lopez:


internal/server/source_manager.go
```diff
@@ -190,23 +190,25 @@ func (s *sourceManager) runLocal(ctx context.Context) {
 		case <-s.snapshotRequests:
 			if s.isPaused() {
 				s.setStatus("PAUSED")
-			} else {
-				s.setStatus("PENDING")
 
-				log(ctx).Debugw("snapshotting", "source", s.src)
+				continue
```
julio-lopez, line 194: @jkowalski any concerns regarding this change?

PR #4666, opened 2025-06-14
refactor(general): small misc. cleanups

julio-lopez:


internal/stat/stat_test.go
```diff
@@ -13,30 +12,31 @@ import (
...
+func TestGetBlockSizeFromCurrentFS(t *testing.T) {
+	size, err := GetBlockSize(".")
+
+	require.NoError(t, err)
+	require.Positive(t, size)
 }
 
 func TestGetFileAllocSize(t *testing.T) {
-	const size = 4096
+	const expectedMinAllocSize = 512
```
julio-lopez, line 29: Minimum block size across platforms / file systems.
Most systems will use a multiple of this value, usually powers of 2 such as, 1K, 2K, 4K, 8K, 16K and so on.

PR #4678, opened 2025-06-19
refactor(general): migrate to `crypto/{hkdf,pbkdf2,sha3}`

julio-lopez:
Functional changes:
- Key derivation no longer panics, instead it returns an error on failed condition checks.
- Migration to `crypto/hkdf` required changes in various function signatures, adding an error return value for propagating key derivation errors up the calling chain. Tests were refactored to account for the signature change as well.

Non-functional changes:
- Added a key derivation helper to reduce function complexity.

### DIFF
--- a/repo/repository.go
+++ b/repo/repository.go
@@ -84,7 +84,7 @@ type DirectRepository interface {
 	AlsoLogToContentLog(ctx context.Context) context.Context
 	UniqueID() []byte
 	ConfigFilename() string
-	DeriveKey(purpose []byte, keyLength int) []byte
+	DeriveKey(purpose []byte, keyLength int) ([]byte, error)
 	Token(password string) (string, error)
 	Throttler() throttling.SettableThrottler
 	DisableIndexRefresh()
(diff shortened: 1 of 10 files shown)


julio-lopez:
FYI: @andrason @hairyhum

repo/repository.go
```diff
@@ -139,15 +139,25 @@ type directRepository struct {
 }
 
 // DeriveKey derives encryption key of the provided length from the master key.
-func (r *directRepository) DeriveKey(purpose []byte, keyLength int) []byte {
+func (r *directRepository) DeriveKey(purpose []byte, keyLength int) (derivedKey []byte, err error) {
 	if r.cmgr.ContentFormat().SupportsPasswordChange() {
-		return crypto.DeriveKeyFromMasterKey(r.cmgr.ContentFormat().GetMasterKey(), r.UniqueID(), purpose, keyLength)
```
jkowalski, line 144: not sure if we need this error handling, the only reason for it to fail is empty master key (where the panic is appropriate since it's a programmer error), all the other operations are never going to fail at runtime.

julio-lopez: @jkowalski 
Yes, specifying an empty master key was treated like a programmer error, and that makes sense. I have not checked the call paths to ensure a non-empty key is always specified though.

Now, the reason for the error propagation was primarily due to the change in the signature for `hkdf.Key`, which returns an error. IIUC, there are various cases where it could return an error, for conditions that can only be checked at runtime. It seems that we cannot _"ensure"_ that `hkdf.Key` will not return an error, and `panic` if it does. Maybe I'm missing something.

The error propagation change was not due to the change in the check for an empty key. Once the signature of the function changed to return an error, it seemed that the best approach would be to also return an error in the case of an empty key. We can revert that part and panic on an empty key, but the error propagation would still be done for the `hkdf.Key` error case.

Thoughts?

jkowalski: the way I understand it, HKDF has no reasons to fail provided that arguments are valid. Anyway, that's nitpicking. LGTM.

PR #4691, opened 2025-06-24
fix(general): migrate remnant piece to `crypto/hkdf`

julio-lopez:
- Followup to #4678


julio-lopez:
FYI: @kidgilson @hairyhum

I missed this change in #4678 😞

PR #4693, opened 2025-06-27
fix(cli): remove direct fmt.Printf calls to have a clean JSON log output

leonardoce:
This change replaces direct fmt.Printf calls with use of the logging infrastructure. These printf caused non-JSON lines to appear in the console output when using the --json-log-console option.

### DIFF
--- a/cli/command_server_tls.go
+++ b/cli/command_server_tls.go
@@ -10,7 +10,6 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/pem"
-	"fmt"
 	"net"
 	"net/http"
 	"os"
@@ -88,7 +87,7 @@ func (c *commandServerStart) maybeGenerateTLS(ctx context.Context) error {
 	}
 
 	fingerprint := sha256.Sum256(cert.Raw)
-	fmt.Fprintf(c.out.stderr(), "SERVER CERT SHA256: %v\n", hex.EncodeToString(fingerprint[:])) //nolint:errcheck
+	log(ctx).Infof("SERVER CERT SHA256: %v", hex.EncodeToString(fingerprint[:]))
 
 	log(ctx).Infof("writing TLS certificate to %v", c.serverStartTLSCertFile)
 
@@ -118,7 +117,7 @@ func (c *commandServerStart) startServerWithOptionalTLSAndListener(ctx context.C
 	switch {
 	case c.serverStartTLSCertFile != "" && c.serverStartTLSKeyFile != "":
 		// PEM files provided
-		fmt.Fprintf(c.out.stderr(), "SERVER ADDRESS: %shttps://%v\n", udsPfx, httpServer.Addr) //nolint:errcheck
+		log(ctx).Infof("SERVER ADDRESS: %shttps://%v", udsPfx, httpServer.Addr)
 		c.showServerUIPrompt(ctx)
 
 		return checkErrServerClosed(ctx, httpServer.ServeTLS(listener, c.serverStartTLSCertFile, c.serverStartTLSKeyFile), "error starting TLS server")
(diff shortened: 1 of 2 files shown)


cli/command_server_tls.go
```diff
@@ -88,7 +87,7 @@ func (c *commandServerStart) maybeGenerateTLS(ctx context.Context) error {
 	}
 
 	fingerprint := sha256.Sum256(cert.Raw)
-	fmt.Fprintf(c.out.stderr(), "SERVER CERT SHA256: %v\n", hex.EncodeToString(fingerprint[:])) //nolint:errcheck
```
jkowalski, line 91: this precise format is actually important the the functioning of KopiaUI which will parse it at startup, we should not change it without verifying that KopiaUI will still work ok. If you'd like to get rid of these messages completely, I suggest adding a flag to do so but by default have them printed out.

leonardoce: Thank you for spotting it. Are you referring to what is happening in `server.js`?

https://github.com/kopia/kopia/blob/a8fc80f47c9875076b1eaaa9ba0a2248a6b5c98d/app/public/server.js#L169-L172

In that case, we could avoid calling `fmt.Printf`only when the `json-log-console` is used, without adding a new flag.
I see that that option is not used by KopiaUI.

https://github.com/kopia/kopia/blob/a8fc80f47c9875076b1eaaa9ba0a2248a6b5c98d/app/public/server.js#L37-L51

This would make sense, because those `fmt.Printf` are disrupting the JSON format of the Kopia server output.
What do you think?

leonardoce:
Superseded by #4884

csenger:
I commented on #4884 why [I think this PR is the better change than adding another flag](https://github.com/kopia/kopia/pull/4884#issuecomment-3415455884). I now tested the electron app and it seems to work with the changes in this PR.

I checked out [this branch (leonardoce:stray-printfs)](https://github.com/leonardoce/kopia/tree/stray-printfs) and ran the electron app with `npm run dev`. It seems the changes from `fmt.Fprintf()` to `log(ctx).Infof()` do not change the output. The electron app works. As far as I can see only `/app/public/server.js` that @leonardoce found relys on the strings from kopia's stderr. `server.js` runs.

```
kopia server start --ui --tls-print-server-cert --tls-generate-cert-name=127.0.0.1 --random-password --random-server-control-password --tls-generate-cert --async-repo-connect --shutdown-on-stdin --address=127.0.0.1:0 --config-file /home/csenger/.config/kopia/repository.config
```

If I run that on the current master/HEAD this is the output: 
[kopia-server-start-master.log](https://github.com/user-attachments/files/22981799/kopia-server-start-master.log)

The output of the same command with kopia from "leonardoce:stray-printfs" is identical: 
[kopia-server-start-4693.log](https://github.com/user-attachments/files/22981798/kopia-server-start-4693.log)

I ran the electron app with `npm run dev`. I patched `server.js` to add some log output in `detectServerParams()`. It worked as I as I expected and I could create snapshots, see 
[main.log](https://github.com/user-attachments/files/22981825/main.log)

If I did not miss something, KopiaUI could only break with electron. `server.js` would need to start kopia with the json log flag. So I think the changes in this PR are safe.

It would even be good to change the Fprintln(s) for "SERVER PASSWORD" and "SERVER CONTROL PASSWORD" (csenger/kopia@ae5abf1a36f9c70fca55d7ff45cc603620fda5c9). This also works with the electron app. Other than that, I only get TLS handshake errors sometimes that won't log as json, but I can't find a source in kopia.

### Tests

Running the e2e tests in `/app` with `npm run e2e` results in 3 passing and 2 failing tests. But afaics these tests are not run in github workflows. Are they obsolete? See [npm-run-e2e.log](https://github.com/user-attachments/files/22981800/npm-run-e2e.log)

Maybe there should be some go tests to check for the output required by the electron app. I can only find test related code in `testutils/serverparameters.go`, but no go tests. I could look into that and try to write some. Tho I'm no go programmer. If it is possible to add an e2e test for the electron app to ensure the configuration is parseable I can look into it too.

### Visible json logging flags with full json logs

I know this is out of scope for this PR, but it would be nice to have full json logging like the zap production encoder. This is a generally useful feature. Full keys instead of "n", "m", "t" and caller, log level and timestamps by default. With visible flags that are named consistent with the current log related flag names, e.g. `--log-format=[text|json]` and `--file-log-format=[text|json]`. The `--json-log-*` and `--console-timestamp` flags could be kept for compatibility, if needed.

PR #4705, opened 2025-07-01
fix(general): use snapshot end time as tie breaker for equal start time

julio-lopez:
Refactor uploader to pass a prototype manifest used for creating the checkpoint manifests. This also removes the hack where the checkpoint start time is artificially 1 ns before the snapshot start time.

### DIFF
--- a/snapshot/manifest.go
+++ b/snapshot/manifest.go
@@ -237,11 +237,14 @@ func GroupBySource(manifests []*Manifest) [][]*Manifest {
 	return result
 }
 
-// SortByTime returns a slice of manifests sorted by start time.
+// SortByTime returns a slice of manifests sorted by start time, and end time as
+// the tie breaker when start time is the same.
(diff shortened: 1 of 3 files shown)


jkowalski (approved):
LGTM, it's a breaking change for folks using Uploader though, so let's mark it as such

julio-lopez:
> it's a breaking change for folks using Uploader though,

@jkowalski Happy to mark it as a breaking change.

I'm missing something here. Could you explain what the incompatibility is?

- AFAICT the signature for exported functions in `Uploader` was unchanged.
- As for behavior, it had changed in #4439, both in terms of ordering and the start time assigned to snapshots vs. checkpoints. This change reverts the "start time" change, while preserving the invariant of having "successful" snapshots be _after_ the checkpoints.

jkowalski:
n/m I misread  it, no breaking change here

PR #4716, opened 2025-07-11
fix(providers): add tags in `azure.Options`

julio-lopez:
Adds missing `` `kopia:"sensitive"` `` tags


julio-lopez:
FYI: @andrason @alisonb-veeam @ryanmt @kidgilson @DeepikaDixit

PR #4719, opened 2025-07-15
feat(general): check for clock skew on write session creation

julio-lopez:
Check for clock skew between the local clock and the storage clock when the session blob is written (on session creation). Tests included.

The check is currently disabled by default.


julio-lopez:
Thanks @abhishekm2610 for your contribution

PR #4682, opened 2025-06-19
fix(server): fixed snapshot scheduling in the event of machine goes to sleep

jkowalski:
Fixes #3512

PR #4732, opened 2025-07-22
fix(server): fixed scheduling bug

### LINKED ISSUE #4729: Since upgrade to 0.21 snapshots are taken every 20 to 30 seconds
Since upgrading .021 kopia is ignoring the snapshot frequency setting and is taking a snapshot every 20 to 30 seconds.

I have change the policy on the server and on the client, each time the policy is reported correctly but snapshots are still taken every 20 to seconds.

 if I click snapshots >> client name the next snapshot is correctly reported.

jkowalski:
This was a very stupid bug introduced in
https://github.com/kopia/kopia/pull/4682

Fixes #4729

### DIFF
--- a/internal/sleepable/sleepable_timer_test.go
+++ b/internal/sleepable/sleepable_timer_test.go
@@ -40,9 +40,9 @@ func TestNewTimer(t *testing.T) {
 			expected: 20 * time.Millisecond,
 		},
 		{
-			name:     "long duration capped at maxSleepTime",
+			name:     "long duration should wait full duration",
 			duration: 1 * time.Second,
-			expected: testMaxSleepTime,
+			expected: 1 * time.Second,
 		},
 		{
 			name:     "exactly maxSleepTime",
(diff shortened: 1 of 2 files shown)


internal/sleepable/sleepable_timer_test.go
```diff
@@ -76,14 +76,14 @@ func TestNewTimer(t *testing.T) {
...
-				if elapsed < testMaxSleepTime-5*time.Millisecond || elapsed > testMaxSleepTime+50*time.Millisecond {
-					t.Errorf("long timer triggered at wrong time: expected ~%v, got %v", testMaxSleepTime, elapsed)
-				}
+			// Allow some tolerance for timing variations
+			tolerance := 50 * time.Millisecond
+			if tt.duration < 100*time.Millisecond {
+				tolerance = 20 * time.Millisecond
+			}
+
+			if elapsed < tt.duration-tolerance || elapsed > tt.duration+tolerance {
```
julio-lopez, line 85: is this going to lead to flaky tests?

jkowalski: we can adjust it if ends up being flaky

internal/sleepable/sleepable_timer_test.go
```diff
@@ -175,15 +175,16 @@ func TestTimerEdgeCases(t *testing.T) {
 
 	t.Run("very long duration", func(t *testing.T) {
 		start := clock.Now()
-		target := start.Add(24 * time.Hour)
+		target := start.Add(100 * time.Millisecond) // Use a shorter duration for testing
```
julio-lopez, line 178: The subtest name no longer describes the case being tested with the new parameters.

internal/sleepable/sleepable_timer_test.go
```diff
@@ -175,15 +175,16 @@ func TestTimerEdgeCases(t *testing.T) {
...
 		timer := NewTimer(clock.Now, target)
 		select {
 		case <-timer.C:
 			elapsed := clock.Now().Sub(start)
-			if elapsed < testMaxSleepTime-5*time.Millisecond || elapsed > testMaxSleepTime+50*time.Millisecond {
-				t.Errorf("very long timer triggered at wrong time: expected ~%v, got %v", testMaxSleepTime, elapsed)
+			// Allow tolerance for timing variations
+			if elapsed < 100*time.Millisecond-20*time.Millisecond || elapsed > 100*time.Millisecond+50*time.Millisecond {
```
julio-lopez, line 184: It'd be good to check the tolerance in terms of the _"target duration"_, that is, what is being added to `start.Add` above.

julio-lopez:
Thanks @jkowalski! 🏆

PR #4767, opened 2025-08-17
fix(cli): additional safeguard dangerous commands

julio-lopez:


jkowalski:
Thanks for doing this!

PR #4774, opened 2025-08-20
refactor(general): miscellaneous cleanups

julio-lopez:
Nits and cleanups:
- clarify log message to indicate the effect of advancing the deletion watermark;
- add `omitzero` JSON tag to appropriate fields in `snapshot.Manifest` struct;
- use `maps.Clone` instead of explicit loop;
- rename function to `IterateUnreferencedPacks` for clarity;
- use `atomic.Int32` type;
- move a `continue` check to the beginning of the loop, no actual work / side effects were performed before the check;
- reduce type requirement in `blob.ReadBlobMap`

### DIFF
--- a/cli/command_index_inspect.go
+++ b/cli/command_index_inspect.go
@@ -109,16 +109,16 @@ func (c *commandIndexInspect) inspectAllBlobs(ctx context.Context, rep repo.Dire
 
 func (c *commandIndexInspect) dumpIndexBlobEntries(entries chan indexBlobPlusContentInfo) {
 	for ent := range entries {
+		if !c.shouldInclude(ent.contentInfo) {
+			continue
+		}
+
 		ci := ent.contentInfo
 		bm := ent.indexBlob
 
 		state := "created"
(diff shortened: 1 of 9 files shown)


julio-lopez:
@Rohit-BM18 @redgoat650 @andrason the PR is broken down into small, self-contained commits to facilitate reviewing the changes. Thanks.

Rohit-BM18 (commented):
LGTM

cli/command_index_inspect.go
```diff
@@ -109,6 +109,10 @@ func (c *commandIndexInspect) inspectAllBlobs(ctx context.Context, rep repo.Dire
 
 func (c *commandIndexInspect) dumpIndexBlobEntries(entries chan indexBlobPlusContentInfo) {
 	for ent := range entries {
+		if !c.shouldInclude(ent.contentInfo) {
```
julio-lopez, line 112: AFAICT this check can be performed at the beginning of the loop (for clarity).

PR #4777, opened 2025-08-21
feat(providers): add user-agent string in Azure storage provider

julio-lopez:
- Adds `kopia` as the application ID.
- Adds a unit test verifying that the header sent to Azure contains the application ID in the user agent string.
- Adds an option to use an insecure connection to Azure, primarily for testing.

Authored by: @andrason

### DIFF
--- a/repo/blob/azure/azure_storage_test.go
+++ b/repo/blob/azure/azure_storage_test.go
@@ -3,7 +3,10 @@ package azure_test
 import (
 	"context"
 	"crypto/rand"
+	"encoding/base64"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


repo/blob/azure/azure_storage_test.go
```diff
@@ -359,3 +362,41 @@ func getBlobCount(ctx context.Context, t *testing.T, st blob.Storage, prefix blo
 
 	return count
 }
+
+func TestUserAgent(t *testing.T) {
```
julio-lopez, line 366: This cannot be merged yet.
This test is causing the "providers" tests to time out.

See https://github.com/kopia/kopia/actions/runs/17118799732/job/48554976801?pr=4777

julio-lopez:
Merging as is. Test fix in a separate PR.

PR #4781, opened 2025-08-22
test(providers): make `azure.TestUserAgent` more robust

julio-lopez:
Leverage `require.Eventually` to prevent indefinite test hangs / timeouts, removes blocking receive op on a channel

Additional cleanups:
- rename test file
- remove unused const
- consistently use `testloggin.Context(t)`
- consistently leverage assertions from the `testify` package

Refs:
- #4780
- #4777


julio-lopez:
FYI: @andrason

PR #4825, opened 2025-09-16
feat(snapshots): Flush after restoring each file

Lyndon-Li:
The file restore is using buffered IO, so after the restore completes, file data are actually written to the system cache, there is no guarantee all data have been flushed to disk. The flush may happen some time later according to the system cache's own data management policy.
On the other hand, users may restore data to a "removable disk", e.g., a virtual disk or a USB disk.

As a result, if users remove the disk immediately after the restore (e.g., detach the virtual disk from the hypervisor), the restore data may be lost.

This PR adds a flush call (`f.Sync`) for each file after the restore.

The flush may take some time because it waits for the kernel to completes the flush IO. So for large amount of files, the throughput may get affected.
However, I think for a restore tool, it is always correct to make sure the data is unconditionally restored to the disk. So the overhead is something ought to spend.
So I don't make this as a conditional call (e.g., adding a user option for it). Let me know if this affects other modules or scenarios.

### DIFF
--- a/snapshot/restore/local_fs_output.go
+++ b/snapshot/restore/local_fs_output.go
@@ -109,6 +109,12 @@ type FilesystemOutput struct {
 	// copier is the StreamCopier to use for copying the actual bit stream to output.
 	// It is assigned at runtime based on the target filesystem and restore options.
 	copier streamCopier `json:"-"`
+
+	// Indicate whether or not flush files after restore.
+	// Varying from OS, the copier may write the file data to the system cache,
+	// so the data may not be written to disk when the restore to the file completes.
+	// This flag guarantees the file data is flushed to disk.
+	FlushFiles bool `json:"flushFiles"`
 }
 
 // Init initializes the internal members of the filesystem writer output.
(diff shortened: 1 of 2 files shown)


snapshot/restore/local_fs_output.go
```diff
@@ -392,6 +392,10 @@ func write(targetPath string, r fs.Reader, size int64, c streamCopier) error {
 		return errors.Wrapf(err, "cannot write data to file %q", f.Name())
 	}
 
+	if err := f.Sync(); err != nil {
```
julio-lopez: How much longer would it take to restore a backup make of 100K small files (< 1KB each)?

Lyndon-Li: I don't have opportunity to benchmark it yet, so I don't have accurate data.
But I would say it would definitely increase the write time and it depends on the file system and storage system. In some cases the time may be remarkable.

On the other hand, I think this time should be included into the restore time --- literally we need to make sure the data is written to the storage before we say that restore completes. In this concept, this is an existing gap, this is why I didn't make it optional.

Here we can make an analogy:

- When restoring a volume from a CSI snapshot, we create a PVC with its `dataSource` pointing to the VolumeSnapshot, that is all
- But we cannot say the restore is done since the storage needs to clone a volume from it which would take a remarkable time

Lyndon-Li: I don't insist that this should be the default behavior. But we need to make a choice:
If we don't do this, or we make it optional, it is impossible to guide users do the right thing beforehand.
Most probably, users report issue after a failed restore, then we troubleshoot it and guide users to use the option and do the restore again.

julio-lopez: It needs to be optional

Lyndon-Li: OK. Will add an option for it.

Lyndon-Li: Done. Added the option and made it `false` as default.

PR #4848, opened 2025-09-28
feat(general): Increase maintenance observability

Lyndon-Li:
Deliver a message for each maintenance task which could be displayed by maintenance info command


repo/maintenance/blob_gc.go
```diff
@@ -27,7 +27,7 @@ type DeleteUnreferencedBlobsOptions struct {
 // DeleteUnreferencedBlobs deletes o was created after maintenance startederenced by index entries.
 //
 //nolint:gocyclo,funlen
-func DeleteUnreferencedBlobs(ctx context.Context, rep repo.DirectRepositoryWriter, opt DeleteUnreferencedBlobsOptions, safety SafetyParameters) (int, error) {
+func DeleteUnreferencedBlobs(ctx context.Context, rep repo.DirectRepositoryWriter, opt DeleteUnreferencedBlobsOptions, safety SafetyParameters) (int, int64, error) {
```
jkowalski: I like the idea,

Since you're doing this, I suggest doing this in a structural way, so we can programmatically analyze these stats over time:

Basically instead of returning individual 1-2 numbers, I would Introduce JSON-serializable structure for each step that has key stats:

type DeleteUnreferencedBlobStats struct {
  Count int32 `json:"count"`
  TotalSize int64 `json:"totalSize"
  PreservedBelowMinAge int32 `json:"preservedBelowMinAge"`
  PreservedAfterCutoffTime int32 `json:"preservedAfterCutoffTime"`
}

and so on for:

RewriteContentStats
IndexCompactionStats
BlobGCStats
SnapshotGCStats

At the end, log the entire structure as JSON (via contentlog) and also add a field "Stats any" in RunInfo where you store it.

Each structure can implement `func MaintenanceSummary() string` which would be used to generate user-facing message (which only has user-interesting stats).

Lyndon-Li: Sure, will make the changes accordingly

Lyndon-Li: @jkowalski Changes are completed. You can start the review now. I am trying to add UT by the meanwhile.

repo/maintenance/content_rewrite.go
```diff
@@ -38,18 +40,35 @@ type contentInfoOrError struct {
...
 
+type RewriteContentsStats struct {
+	RewrittenCount uint32 `json:"rewrittenCount"`
+	RewrittenSize  int64  `json:"rewrittenSize"`
+	PreservedCount uint32 `json:"preservedCount"`
+	PreservedSize  int64  `json:"preservedSize"`
+}
+
+func (rs *RewriteContentsStats) WriteValueTo(jw *contentlog.JSONWriter) {
+	if bytes, err := json.Marshal(rs); err == nil {
```
jkowalski: I know it's not performance-critical and memory allocations don't matter, but can we follow the pattern we use everywhere:

```
jw.BeginObjectField("rewriteContentsStats")
jw.Int32Field("rewrittenCount", rs.RewrittenCount)
...
jw.EndObject()
```

Lyndon-Li: Done

cli/command_content_rewrite.go
```diff
@@ -49,7 +49,7 @@ func (c *commandContentRewrite) runContentRewriteCommand(ctx context.Context, re
 	}
 
 	//nolint:wrapcheck
```
julio-lopez: is this annotation still needed?

```suggestion
```

repo/content/content_manager_test.go
```diff
@@ -522,7 +523,7 @@ func (s *contentManagerSuite) TestContentManagerConcurrency(t *testing.T) {
 
 	validateIndexCount(t, data, 4, 0)
 
-	if err := bm4.CompactIndexes(ctx, indexblob.CompactOptions{MaxSmallBlobs: 1}); err != nil {
+	if _, err := bm4.CompactIndexes(ctx, indexblob.CompactOptions{MaxSmallBlobs: 1}); err != nil {
 		t.Errorf("compaction error: %v", err)
 	}
```
julio-lopez: nit: while at it ... and the in other instances below

```suggestion
	_, err := bm4.CompactIndexes(ctx, indexblob.CompactOptions{MaxSmallBlobs: 1})
	require.NoError(t, err, "compaction error")
```

repo/maintenance/cleanup_logs.go
```diff
@@ -88,10 +89,19 @@ func CleanupLogs(ctx context.Context, rep repo.DirectRepositoryWriter, opt LogRe
...
+	var unusedSize int64
+	for _, bm := range toDelete {
+		unusedSize += bm.Length
+	}
+
+	result := &maintenancestats.CleanupLogsStats{
+		PreservedCount: uint32(deletePosition),
+		PreservedSize:  preservedSize,
+		UnusedCount:    uint32(max(len(allLogBlobs)-deletePosition-1, 0)),
+		UnusedSize:     unusedSize,
```
julio-lopez: If the other two fields are the stats for the preserved or retained logs, then these should be the stats for the "deleted" (or discarded / reclaimed) logs.

repo/maintenance/content_rewrite.go
```diff
@@ -61,9 +62,12 @@ func RewriteContents(ctx context.Context, rep repo.DirectRepositoryWriter, opt *
...
 
 	var (
-		mu          sync.Mutex
-		totalBytes  int64
-		failedCount int
+		mu             sync.Mutex
+		totalBytes     int64
+		totalCount     uint32
+		preservedBytes int64
+		preservedCount uint32
```
julio-lopez:
```suggestion
		retainedBytes int64
		retainedCount uint32
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/kopia/kopia/internal/contentlog"
+)
+
+// RawStats holds the raw data for maintenance statistics
+type RawStats struct {
+	Type string          `json:"type,omitempty"`
```
julio-lopez:
```suggestion
	Kind string          `json:"kind,omitempty"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+	Raw  json.RawMessage `json:"raw,omitempty"`
+}
+
+// Stats defines the methods for maintenance statistics
+type Stats interface {
+	Type() string
```
julio-lopez:
```suggestion
	Kind() string
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+
+// Type returns the type name for CompactStats
+func (cs *CompactStats) Type() string {
+	return compactStatsType
+}
+
+const deleteUnreferencedBlobsStatsType = "deleteUnreferencedBlobsStats"
+
+// DeleteUnreferencedBlobsStats delivers the statistics for DeleteUnreferencedBlobs
+type DeleteUnreferencedBlobsStats struct {
```
julio-lopez:
```suggestion
type DeleteUnreferencedPacksStats struct {
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+func (cs *CompactStats) Summary() string {
+	return fmt.Sprintf("Dropped indexes before %v", cs.DroppedBefore)
+}
+
+// Type returns the type name for CompactStats
+func (cs *CompactStats) Type() string {
+	return compactStatsType
+}
+
+const deleteUnreferencedBlobsStatsType = "deleteUnreferencedBlobsStats"
```
julio-lopez:
```suggestion
const deleteUnreferencedPacksStatsType = "deleteUnreferencedPacksStats"
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+
+// DeleteUnreferencedBlobsStats delivers the statistics for DeleteUnreferencedBlobs
+type DeleteUnreferencedBlobsStats struct {
+	UnusedCount    uint32 `json:"unusedCount"`
+	UnusedSize     int64  `json:"unusedSize"`
+	DeletedCount   uint32 `json:"deletedCount"`
+	DeletedSize    int64  `json:"deletedSize"`
+	PreservedCount uint32 `json:"PreservedCount"`
+	PreservedSize  int64  `json:"PreservedSize"`
```
julio-lopez:
```suggestion
	RetainedCount uint32 `json:"retainedCount"`
	RetainedSize  int64  `json:"retainedSize"`
```

repo/maintenancestats/types.go
julio-lopez, line 1: - This should be split into multiple files, with a Go file per type.
- Why are these in a separate `maintenancestats` package?

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+// Type returns the type name for GenerateRangeCheckpointStats
+func (gs *GenerateRangeCheckpointStats) Type() string {
+	return generateRangeCheckpointStatsType
+}
+
+const advanceEpochStatsType = "advanceEpochStats"
+
+// AdvanceEpochStats delivers the statistics for MaybeAdvanceWriteEpoch
+type AdvanceEpochStats struct {
+	CurEpoch uint32 `json:"curEpoch"`
```
julio-lopez:
```suggestion
	CurrentEpoch uint32 `json:"currentEpoch"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+// Type returns the type name for AdvanceEpochStats
+func (as *AdvanceEpochStats) Type() string {
+	return advanceEpochStatsType
+}
+
+const compactSingleEpochStatsType = "compactSingleEpochStats"
+
+// CompactSingleEpochStats delivers the statistics for MaybeCompactSingleEpoch
```
julio-lopez: Instead of referring to the function that returns this struct, the comment should refer to the corresponding operation and what the stats mean.

For example,

```suggestion
// CompactSingleEpochStats are the stats for compacting an index epoch
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+// WriteValueTo writes CompactSingleEpochStats to JSONWriter
+func (cs *CompactSingleEpochStats) WriteValueTo(jw *contentlog.JSONWriter) {
+	jw.BeginObjectField(cs.Type())
+	jw.UInt32Field("blobCount", cs.BlobCount)
+	jw.Int64Field("blobSize", cs.BlobSize)
+	jw.UInt32Field("epoch", cs.Epoch)
+	jw.EndObject()
+}
+
+// Summary generates readable summary for CompactSingleEpochStats
```
julio-lopez: No need to repeat the type in the comment, it is implied from the receiver type.
```suggestion
// Summary generates a human readable summary for the stats
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+func (cs *CompactSingleEpochStats) Summary() string {
+	return fmt.Sprintf("Compacted %v(%v) index blobs for epoch %v", cs.BlobCount, cs.BlobSize, cs.Epoch)
+}
+
```
julio-lopez, line 215: Same here and for other stats types and functions.

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+func (as *AdvanceEpochStats) Type() string {
+	return advanceEpochStatsType
+}
+
+const compactSingleEpochStatsType = "compactSingleEpochStats"
+
+// CompactSingleEpochStats delivers the statistics for MaybeCompactSingleEpoch
+type CompactSingleEpochStats struct {
+	BlobCount uint32 `json:"blobCount"`
+	BlobSize  int64  `json:"blobSize"`
```
julio-lopez: We need to improve the names for these fields so it is clear exactly what they mean or refer to, and avoid ambiguity.

Is `blobCount` the number of index blobs that was compacted or the resulting number of blobs after compaction? (although it is a single "index set", it can comprise more than 1 index blob)

Is `blobSize` the sum of the size of the "input" blobs that were compacted, or the size of the resulting blobs after compaction?

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+	return cleanupMarkersStatsType
+}
+
+const cleanupSupersededIndexesStatsType = "cleanupSupersededIndexesStats"
+
+// CleanupSupersededIndexesStats delivers the statistics for CleanupSupersededIndexes
+type CleanupSupersededIndexesStats struct {
+	MaxReplacementTime time.Time `json:"maxReplacementTime"`
+	DeletedBlobs       uint32    `json:"deletedBlobs"`
+	DeletedSize        int64     `json:"deletedSize"`
```
julio-lopez: Should this be:
```suggestion
	DeletedTotalSize        int64     `json:"deletedTotalSize"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+func (cs *CleanupMarkersStats) Type() string {
+	return cleanupMarkersStatsType
+}
+
+const cleanupSupersededIndexesStatsType = "cleanupSupersededIndexesStats"
+
+// CleanupSupersededIndexesStats delivers the statistics for CleanupSupersededIndexes
+type CleanupSupersededIndexesStats struct {
+	MaxReplacementTime time.Time `json:"maxReplacementTime"`
+	DeletedBlobs       uint32    `json:"deletedBlobs"`
```
julio-lopez:
```suggestion
	DeletedBlobCount       uint32    `json:"deletedBlobCount"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+			return nil, err
+		}
+
+		return &ss, nil
+	default:
+		return nil, errors.New("unsupported stats type")
+	}
+}
+
+const cleanupMarkersStatsType = "cleanupMarkersStats"
```
julio-lopez: Why not use the names for the maintenance phases?

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+
+// CleanupMarkersStats delivers the statistics for CleanupMarkers
+type CleanupMarkersStats struct {
+	EpochMarkers       uint32 `json:"epochMarkers"`
+	DeletionWaterMarks uint32 `json:"deletionWaterMarks"`
+}
+
+// WriteValueTo writes CleanupMarkersStats to JSONWriter
```
julio-lopez: Simplify this comment, it's redundant and does not provide any additional information that cannot be derived from the the function signature.

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+const cleanupLogsStatsType = "cleanupLogsStats"
+
+// CleanupLogsStats delivers the statistics for CleanupLogs
+type CleanupLogsStats struct {
+	UnusedCount    uint32 `json:"unusedCount"`
+	UnusedSize     int64  `json:"unusedSize"`
+	PreservedCount uint32 `json:"preservedCount"`
+	PreservedSize  int64  `json:"preservedSize"`
```
julio-lopez:
```suggestion
	DeletedCount    uint32 `json:"deletedCount"`
	DeletedSize     int64  `json:"deletedSize"`
	RetainedCount uint32 `json:"retainedCount"`
	RetainedSize  int64  `json:"retainedSize"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+const rewriteContentsStatsType = "rewriteContentsStats"
+
+// RewriteContentsStats delivers the statistics for RewriteContents
+type RewriteContentsStats struct {
+	RewrittenCount uint32 `json:"rewrittenCount"`
+	RewrittenSize  int64  `json:"rewrittenSize"`
+	PreservedCount uint32 `json:"preservedCount"`
+	PreservedSize  int64  `json:"preservedSize"`
```
julio-lopez:
```suggestion
	RetainedCount uint32 `json:"retainedCount"`
	RetainedSize  int64  `json:"retainedSize"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+const snapshotGCStatsType = "snapshotGCStats"
+
+// SnapshotGCStats delivers the statistics for SnapshotGC
+type SnapshotGCStats struct {
+	UnusedCount          uint32 `json:"unusedCount"`
+	UnusedSize           int64  `json:"unusedSize"`
+	TooRecentUnusedCount uint32 `json:"tooRecentUnusedCount"`
+	TooRecentUnusedSize  int64  `json:"tooRecentUnusedSize"`
+	InUseCount           uint32 `json:"inUseCount"`
+	IntUseSize           int64  `json:"intUseSize"`
```
julio-lopez, line 374:
```suggestion
	InUseSize           int64  `json:"inUseSize"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+	InUseSystemCount     uint32 `json:"inUseSystemCount"`
+	IntUseSystemSize     int64  `json:"intUseSystemSize"`
```
julio-lopez:
```suggestion
	InUseSystemSize     int64  `json:"inUseSystemSize"`
```

repo/maintenance/cleanup_logs.go
julio-lopez: If the other two fields are the stats for the preserved or retained logs, then these should be the stats for the "deleted" (or discarded / reclaimed) logs.

```suggestion
		RetainedCount: uint32(deletePosition),
		RetainedSize:  preservedSize,
		DeletedCount:    uint32(max(len(allLogBlobs)-deletePosition-1, 0)),
		DeletedSize:     unusedSize,
```

repo/maintenance/cleanup_logs.go
```diff
@@ -88,10 +89,19 @@ func CleanupLogs(ctx context.Context, rep repo.DirectRepositoryWriter, opt LogRe
 
 	toDelete := allLogBlobs[deletePosition:]
 
-	contentlog.Log2(ctx, log,
-		"Keeping logs",
-		logparam.Int("count", deletePosition),
-		logparam.Int64("bytes", totalSize))
+	var unusedSize int64
```
julio-lopez, line 92:
```suggestion
	var deletedSize int64
```

repo/maintenance/cleanup_logs.go
```diff
@@ -63,14 +64,14 @@ func CleanupLogs(ctx context.Context, rep repo.DirectRepositoryWriter, opt LogRe
 		return allLogBlobs[i].Timestamp.After(allLogBlobs[j].Timestamp)
 	})
 
-	var totalSize int64
+	var preservedSize int64
```
julio-lopez:
```suggestion
	var retainedSize int64
```

Lyndon-Li:
> This PR has become too large to review. @Lyndon-Li please split it into multiple, smaller PRs. A good strategy would be a PR per affected maintenance phases / stat struct.
> 
> Please include the reason for making this change in the PR description with an answer to the question: _"why is this change needed?"_
> 
> The naming for a the stat structs and their fields needs to be refined. See inline comments with a subset of the requested changes, we will iterate in the individual PRs to ensure that the names make sense and correspond to the stat being reported. Note, that while we are making an effort upfront to choose good names for these structs, it does not imply that any form of backwards or forward compatibility will be maintained for these fields (whether implicitly or explicitly).
> 
> The function comments need to be simplified as well.
> 
> Please address the requested changes, and run run the corresponding workflows either locally or in your repository to reduce the number of CI job failures in the changes submitted to the pull requests.
> 
> Overall, this is going in the right direction, and the changes should prove useful in the context of maintenance runs and in diagnosing the reported repository corruption cases.
> 
> Thank you.

Sure, will need some time to complete these and will let you know.

repo/maintenance/cleanup_logs.go
Lyndon-Li: The reason I didn't use the word "deleted" is that there is a dryRun option, when it is true, nothing will be deleted.
So to better cover the two cases, I just picked "unused".

Let me know your thoughts on this.

repo/maintenancestats/types.go
Lyndon-Li: If we have it in the existing maintenance package, there is will be conflict of loop include.
And if we have each stat definition together with the function doing the task, it will be too deep for `BuildRaw` or `BuildFromRaw`.

repo/maintenancestats/types.go
Lyndon-Li: I think the type name should be separate without coupling with the maintenance tasks, after all, the functions are not only called by maintenance but also called by some CLIs.

repo/maintenancestats/types.go
Lyndon-Li: Same reason with [BuildFromRaw](https://github.com/kopia/kopia/pull/4848#discussion_r2438107688) for the word of `deleted` vs. `unused`

repo/maintenance/cleanup_logs.go
Lyndon-Li: Same reason with https://github.com/kopia/kopia/pull/4848#discussion_r2438107688 for `deleted` vs. `unused`.

repo/maintenance/cleanup_logs.go
Lyndon-Li: Same reason with https://github.com/kopia/kopia/pull/4848#discussion_r2438107688 for deleted vs. unused.

Lyndon-Li:
@julio-lopez I replied some of your comments in place and keep them unresolved for which I think we need some further discussion, please check them and share your opinions.

repo/maintenance/cleanup_logs.go
Lyndon-Li: Done for `RetainedCount ` and `RetainedSize`

repo/maintenancestats/types.go
Lyndon-Li: Done for `RetainedCount ` and `RetainedSize`

repo/maintenance/cleanup_logs.go
julio-lopez: I see, thanks for pointing that out.
For now, let's go with "deleted". Not sure there is a word that would work for both 'to-be deleted' and 'was actually deleted'.

repo/maintenancestats/types.go
julio-lopez: got it, separate package it is. We can revisit down the road and do a refactor-only change to keep things self contained.

julio-lopez: please split the types into separate files though.

repo/maintenancestats/types.go
julio-lopez: OK, let's keep them separate then.

internal/epoch/epoch_manager.go
```diff
@@ -717,21 +740,29 @@ func (e *Manager) refreshAttemptLocked(ctx context.Context) error {
...
-		return err
+		return nil, err
 	}
 
 	e.mu.Lock()
 	cs := e.lastKnownState
 	e.mu.Unlock()
 
+	result := &maintenancestats.AdvanceEpochStats{
+		CurrentEpoch: uint32(cs.WriteEpoch),
```
julio-lopez: For consistency, `CurrentEpoch` should **always** have the epoch that is current when this function returns, otherwise, it is confusing. This means `CurrentEpoch` should be updated below if advancement takes place.

The `WasAdvanced` field then indicates whether the current epoch is a "new" one or whether it had been previously advanced.

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+	"fmt"
+	"time"
+
+	"github.com/kopia/kopia/internal/contentlog"
+)
+
+// RawStats holds the raw data for maintenance statistics
+type RawStats struct {
+	Kind string          `json:"kind,omitempty"`
+	Raw  json.RawMessage `json:"raw,omitempty"`
```
julio-lopez, line 17: rename this struct to `Extra` and the field below to `Data`

```suggestion
type Extra struct {
	Kind string          `json:"kind,omitempty"`
	Data json.RawMessage `json:"data,omitempty"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+// Stats defines the methods for maintenance statistics
+type Stats interface {
```
julio-lopez, line 21:
```suggestion
type ExtraSummarizer interface {
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,449 @@
...
+func (gs *GenerateRangeCheckpointStats) Kind() string {
+	return generateRangeCheckpointStatsKind
+}
+
+const advanceEpochStatsKind = "advanceEpochStats"
+
+// AdvanceEpochStats are the stats for advancing write epoch
+type AdvanceEpochStats struct {
+	CurrentEpoch uint32 `json:"currentEpoch"`
+	Advanced     bool   `json:"advanced"`
```
julio-lopez, line 165:
```suggestion
	CurrentEpoch int    `json:"currentEpoch"`
	WasAdvanced  bool   `json:"wasAdvanced"`
```

repo/maintenance/cleanup_logs.go
Lyndon-Li: @julio-lopez 
Could you share your concerns on "unusedxxx"? I think it right works for both 'to-be deleted' and 'was actually deleted', doesn't it?

repo/maintenancestats/types.go
Lyndon-Li: Sure, will create some separate PRs. I am just trying to solve most the comments in the current PR and then split with less efforts.

Lyndon-Li:
One split PR #4890 is created.
I will keep the current PR open until all the opening comments are resolved.

repo/maintenance/maintenance_schedule.go
```diff
@@ -32,10 +33,11 @@ const maxRetainedRunInfoPerRunType = 50
...
 type RunInfo struct {
-	Start   time.Time `json:"start"`
-	End     time.Time `json:"end"`
-	Success bool      `json:"success,omitempty"`
-	Error   string    `json:"error,omitempty"`
+	Start   time.Time                  `json:"start"`
+	End     time.Time                  `json:"end"`
+	Success bool                       `json:"success,omitempty"`
+	Error   string                     `json:"error,omitempty"`
+	Stats   *maintenancestats.RawStats `json:"stats,omitempty"`
```
julio-lopez, line 40:
```suggestion
	Extra   []maintenancestats.Extra `json:"extra,omitempty"`
```

julio-lopez (changes requested):
See additional comment about renaming the field to `Extra` in the maintenance schedule and making it a slice instead of a pointer.

And then renaming `Raw` to `Data` in the new 'Extra' struct.

Please break it down into a PR per data struct so we can look at the use cases for each struct and ensure the names are appropriate. Thanks.

repo/maintenance/cleanup_logs.go
julio-lopez: How does "unusedxxx" reflect what the values are about?

Please split the PR into a PR per struct and its corresponding use case (that is, the maintenance phase where it is used). Then we can see what each of the individual fields mean.

At the moment, this PR is _too large to review_ as it is.

repo/maintenancestats/types.go
julio-lopez: What are the values for these fields? ^^^

julio-lopez:
And please add unit tests that (un)marshal the new structs, to ensure they are what is expected.

Lyndon-Li:
> See additional comment about renaming the field to Extra in the maintenance schedule and making it a slice instead of a pointer.

@julio-lopez Could you explain why we want to use a slice? I think there is always one piece of `Extra` for each run, so isn't a pointer enough?

julio-lopez:
> > See additional comment about renaming the field to Extra in the maintenance schedule and making it a slice instead of a pointer.
>
> @julio-lopez Could you explain why we want to use a slice? I think there is always one piece of `Extra` for each run, so isn't a pointer enough?

Yes, for now there will be at most one entry per task run. Please make it a slice nevertheless. Thanks

internal/epoch/epoch_manager.go
Lyndon-Li: Sure, will consider this in the split PR.

repo/maintenancestats/types.go
Lyndon-Li: Will consider this in the split PR.

repo/maintenancestats/types.go
Lyndon-Li: Will modify it in the split PR.

repo/maintenance/cleanup_logs.go
Lyndon-Li: Sure, let me consider this and continue this discussion in the split PR.

Lyndon-Li:
I am closing this PR as all the changes has been included in the splited PRs.

julio-lopez:
Thanks for your contribution!

PR #4859, opened 2025-10-02
test(general): fix endurance test failures

julio-lopez:
Prevent running "auto-maintenance" on snapshot create.

Ref:
- #4851

Context: the test fails because there are concurrent "endurance" runners and each of these advances the clock, some of them significantly (`action{Small,Medium,Large}ClockJump`), which causes the clock skewness check to fail when (auto-)maintenance runs. The test maintenance action does not run into this issue because it runs exclusively on its own, that is, other actions have to wait until maintenance completes.


julio-lopez:
Successful run: https://github.com/kopia/kopia/actions/runs/18180048576/job/51754051221

Issue #4831, opened 2025-09-18
`failed to open repository: too many open files` when accessing repository (index cache size exceeds default limits)


samm81:
### problem

running any `kopia` command that requires accessing the repository (e.g. `kopia snapshot list`) results in an error like:

```
ERROR open repository: unable to open repository: unable to create shared content manager: error loading indexes: unable to open pack index "xn84_6260a73609a1d24fef712646899085f4-s71751e05a7275847138-c1": unable to open file despite retries: open /home/maynard/.cache/kopia/bfc036a6bbb0d4c1/indexes/xn84_6260a73609a1d24fef712646899085f4-s71751e05a7275847138-c1.sndx: too many open files
```

this has been reported multiple times already:

* https://kopia.discourse.group/t/cant-check-snapshot-list-in-client/2095
* https://kopia.discourse.group/t/too-many-open-files/3293
* https://kopia.discourse.group/t/average-epoch-time/3295
* https://kopia.discourse.group/t/error-failed-to-open-repository-too-many-open-files/1286/4
* https://github.com/kopia/kopia/issues/2633
* https://github.com/kopia/kopia/issues/3070

---

### current workaround

the usual recommendation is to raise the open-files limit (`ulimit -n`).

for example, my defaults are:

```bash
$ ulimit -Hn
4096
```

i can make `kopia` work temporarily with:

```bash
sudo prlimit --pid "$$" --nofile=8192:8192
```

but this raises two questions:

1. will I need to keep increasing this number as the repository grows?
2. should `kopia` really require thousands of simultaneously open files?

---

### observations

* the error involves cached index files at:
  `.cache/kopia/${HASH}/indexes/${INDEX_ID}`

* my cache directory contains over 4000 files:

```bash
ls -l ~/.cache/kopia/bfc036a6bbb0d4c1/indexes | wc -l
4148
```

* this slightly exceeds my default limit (`4096`), explaining the error.

* clearing the cache (`kopia cache clear` or deleting `~/.cache/kopia`) doesn’t help — the cache quickly grows back to the same scale.

* it looks like `kopia` attempts to open all these files simultaneously rather than reading and closing them, which leads directly to hitting the system limit.

---

### question

would it be possible for `kopia` to avoid holding all index files open at once? for example, by:

* loading them into memory and closing the file descriptors, or
* batching access to a smaller working set

this would make `kopia` behave better on systems with default limits, without requiring system-wide changes.

---

### thanks 🙏

[...]

PR #4866, opened 2025-10-04
feat(content): close FD after mmap on Unix to reduce open FDs

samm81:
On Unix-like platforms, close the file descriptor immediately after a successful mmap.Map of .sndx index cache files. This keeps the mapping valid until Unmap (per POSIX semantics) and significantly reduces steady-state FD usage when many indexes are open, helping avoid EMFILE ("too many open files").

- Split mmapOpenWithRetry into platform-specific implementations:
  - committed_content_index_disk_cache_unix.go (!windows):
    - Map RDONLY, close FD immediately.
    - Return closer that only unmaps.
  - committed_content_index_disk_cache_windows.go (windows):
    - Keep FD open until Unmap.
    - Return closer that unmaps and closes FD.
- Remove old mmapOpenWithRetry and mmap import from repo/content/committed_content_index_disk_cache.go.
- Add Linux-only unit test verifying FD count does not grow proportionally: repo/content/committed_content_index_fd_linux_test.go
  - Creates N small indexes, opens them all, checks /proc/self/fd delta stays low.

Notes:
- Behavior unchanged on Windows due to OS semantics.
- Mapping failures close the FD to avoid leaks.
- Unlink semantics remain correct; mappings stay valid until Unmap.

### DIFF
--- a/repo/content/committed_content_index_disk_cache_unix.go
+++ b/repo/content/committed_content_index_disk_cache_unix.go
@@ -1,0 +1,70 @@
+//go:build !windows
+
+package content
+
+import (
+	"os"
+	"time"
+
+	"github.com/edsrzf/mmap-go"
+	"github.com/pkg/errors"
+)
+
+// mmapOpenWithRetry attempts mmap.Open() with exponential back-off to work around rare issue
+// specific to Windows where we can't open the file right after it has been written.
+//
+// Unix semantics: Close the file descriptor immediately after a successful mmap so the
+// process does not retain FDs for all mapped index files. The mapping remains valid until
+// Unmap is called.
+func (c *diskCommittedContentIndexCache) mmapOpenWithRetry(path string) (mmap.MMap, func() error, error) {
+	const (
+		maxRetries    = 8
+		startingDelay = 10 * time.Millisecond
+	)
+
+	// retry milliseconds: 10, 20, 40, 80, 160, 320, 640, 1280, total ~2.5s
+	f, err := os.Open(path) //nolint:gosec
+	nextDelay := startingDelay
+
+	retryCount := 0
+	for err != nil && retryCount < maxRetries {
+		retryCount++
+		c.log.Debugf("retry #%v unable to mmap.Open(): %v", retryCount, err)
+		time.Sleep(nextDelay)
+		nextDelay *= 2
+
+		f, err = os.Open(path) //nolint:gosec
+	}
+
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "unable to open file despite retries")
+	}
+
+	mm, err := mmap.Map(f, mmap.RDONLY, 0)
+	if err != nil {
+		_ = f.Close()
+		return nil, nil, errors.Wrap(err, "mmap error")
+	}
+
+	// On Unix, it's safe to close the FD now; the mapping remains valid.
+	if err := f.Close(); err != nil {
+		// If close fails, still return mapping, but report error on closer to surface the issue later.
+		closeErr := errors.Wrapf(err, "error closing index %v after mmap", path)
+
+		return mm, func() error {
+			if err2 := mm.Unmap(); err2 != nil {
+				return errors.Wrapf(err2, "error unmapping index %v (also had close error: %v)", path, closeErr)
+			}
+
+			return closeErr
+		}, nil
+	}
+
+	return mm, func() error {
+		if err2 := mm.Unmap(); err2 != nil {
+			return errors.Wrapf(err2, "error unmapping index %v", path)
+		}
+
+		return nil
+	}, nil
+}
--- a/repo/content/committed_content_index_fd_linux_test.go
+++ b/repo/content/committed_content_index_fd_linux_test.go
@@ -1,0 +1,73 @@
+//go:build linux
+
+package content
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/kopia/kopia/internal/faketime"
+	"github.com/kopia/kopia/internal/testlogging"
+	"github.com/kopia/kopia/internal/testutil"
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/content/index"
+)
+
+// countFDsLinux returns the number of open file descriptors for the current process on Linux.
+func countFDsLinux(t *testing.T) int {
+	t.Helper()
+
+	entries, err := os.ReadDir("/proc/self/fd")
+	if err != nil {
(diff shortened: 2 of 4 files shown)


samm81:
trying to solve my own issue #4831 . works ok on my machine locally so I figured I'd push it up if it would help anyone else.

jkowalski:
please rebase and fix merge conflict, we should test this on Mac and Linux and then we can merge

repo/content/committed_content_index_disk_cache_unix.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/edsrzf/mmap-go"
+	"github.com/pkg/errors"
+
+	"github.com/kopia/kopia/internal/contentlog"
+	"github.com/kopia/kopia/internal/contentlog/logparam"
+)
+
+// mmapOpenWithRetry attempts mmap.Open() with exponential back-off to work around rare issue
+// specific to Windows where we can't open the file right after it has been written.
```
julio-lopez: this comment is stale, it applies to the windows implementation and should be moved to that implementation.

repo/content/committed_content_index_disk_cache_unix.go
```diff
@@ -0,0 +1,76 @@
...
+	if err := f.Close(); err != nil {
+		// If close fails, still return mapping, but report error on closer to surface the issue later.
+		closeErr := errors.Wrapf(err, "error closing index %v after mmap", path)
+
+		return mm, func() error {
+			if err2 := mm.Unmap(); err2 != nil {
+				return errors.Wrapf(err2, "error unmapping index %v (also had close error: %v)", path, closeErr)
+			}
+
+			return closeErr
```
julio-lopez, line 38: Would it make sense to _retry_ the close when `closeErr != nil` ?

repo/content/committed_content_index_disk_cache_unix.go
```diff
@@ -0,0 +1,76 @@
...
+	for err != nil && retryCount < maxRetries {
+		retryCount++
+		contentlog.Log2(ctx, c.log, "retry unable to mmap.Open()",
+			logparam.Int("retryCount", retryCount),
+			logparam.Error("err", err))
+		time.Sleep(nextDelay)
+		nextDelay *= 2
+
+		f, err = os.Open(path) //nolint:gosec
+	}
```
julio-lopez: IIUC, the retry functionality is only required for Windows, and could be removed. @jkowalski can you confirm please?

```suggestion
```

repo/content/committed_content_index_disk_cache_unix.go
samm81: retrying `f.Close()` won’t help: Go’s `os.File.Close` marks the descriptor as closed before issuing the syscall, so on any error it can’t be retried (a second call just yields `ErrInvalid`).

julio-lopez: got it

repo/content/committed_content_index_disk_cache_unix.go
julio-lopez: This can be removed (confirmed by @jkowalski)

repo/content/committed_content_index_fd_linux_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// countFDsLinux returns the number of open file descriptors for the current process on Linux.
+func countFDsLinux(t *testing.T) int {
+	t.Helper()
+
+	entries, err := os.ReadDir("/proc/self/fd")
+	if err != nil {
+		t.Fatalf("unable to read /proc/self/fd: %v", err)
+	}
```
julio-lopez:
```suggestion
	require.NoError(t, err, "unable to read /proc/self/fd")
```

repo/content/committed_content_index_fd_linux_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return len(entries)
+}
+
+// Test that opening many indexes on Linux does not retain a file descriptor per index.
+func TestCommittedContentIndexCache_Disk_FDsNotGrowingOnOpen_Linux(t *testing.T) {
+	// Do not run in parallel to avoid fd count noise from other tests.
+	ctx := testlogging.Context(t)
+	ft := faketime.NewClockTimeWithOffset(0)
+	cache := &diskCommittedContentIndexCache{testutil.TempDirectory(t), ft.NowFunc(), func() int { return 3 }, testlogging.Printf(t.Logf, ""), DefaultIndexCacheSweepAge}
```
julio-lopez: @samm81 This failed to compile in a previous CI job and it is going to fail again due to a type mismatch for the "log" type.

See https://github.com/kopia/kopia/actions/runs/18611420883/job/53293937511?pr=4866

repo/content/committed_content_index_fd_linux_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	after := countFDsLinux(t)
+
+	// Despite keeping many mappings alive, the FD count should not grow proportionally.
+	// Allow some slack for incidental FDs opened by runtime or test harness.
+	const maxDelta = 32
+	if delta := after - before; delta > maxDelta {
+		t.Fatalf("fd count grew too much after opening %d indexes: before=%d after=%d delta=%d (max allowed %d)", N, before, after, delta, maxDelta)
+	}
```
julio-lopez: this can be simplified as:

```suggestion

	require.LessOrEqualf(t, after - before, maxDelta "fd count grew too much after opening %d indexes:", N)
```

The delta (that is `after - before`) and `maxDelta` will be already included in the message when the assertion fails.

repo/content/committed_content_index_fd_linux_test.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Test that opening many indexes on Linux does not retain a file descriptor per index.
+func TestCommittedContentIndexCache_Disk_FDsNotGrowingOnOpen_Linux(t *testing.T) {
+	// Do not run in parallel to avoid fd count noise from other tests.
+	ctx := testlogging.Context(t)
+	ft := faketime.NewClockTimeWithOffset(0)
+	cache := &diskCommittedContentIndexCache{testutil.TempDirectory(t), ft.NowFunc(), func() int { return 3 }, testlogging.Printf(t.Logf, ""), DefaultIndexCacheSweepAge}
+
+	const N = 200
```
julio-lopez: nit:
```suggestion
	const indexCount = 200
```

julio-lopez:
@samm81 Thanks for your digging into this issue and for the corresponding improvement.

PR #4884, opened 2025-10-15
feat(server): "--kopiaui-info" option

leonardoce:
This patch introduces an option to suppress the printing of KopiaUI interface options to stderr by using
"--kopiaui-info=false", with "--kopiaui-info" defaulting to "true" to preserve the original behavior.

This feature is especially beneficial when used alongside the "--json-log-console" option, as it enables a clean, parsable JSON logging stream.

### DIFF
--- a/cli/command_server_start.go
+++ b/cli/command_server_start.go
@@ -70,8 +70,10 @@ type commandServerStart struct {
 	debugScheduler                      bool
 	minMaintenanceInterval              time.Duration
 
-	shutdownGracePeriod  time.Duration
+	shutdownGracePeriod time.Duration
+
 	kopiauiNotifications bool
+	kopiauiInfo          bool
 
 	logServerRequests bool
 
@@ -127,6 +129,8 @@ func (c *commandServerStart) setup(svc advancedAppServices, parent commandParent
 	cmd.Flag("shutdown-grace-period", "Grace period for shutting down the server").Default("5s").DurationVar(&c.shutdownGracePeriod)
 
 	cmd.Flag("kopiaui-notifications", "Enable notifications to be printed to stdout for KopiaUI").BoolVar(&c.kopiauiNotifications)
+
+	cmd.Flag("kopiaui-info", "Outputs connection info for KopiaUI integration").Default("true").Hidden().BoolVar(&c.kopiauiInfo)
 
 	c.sf.setup(svc, cmd)
 	c.co.setup(svc, cmd)
--- a/cli/command_server_tls.go
+++ b/cli/command_server_tls.go
@@ -88,7 +88,9 @@ func (c *commandServerStart) maybeGenerateTLS(ctx context.Context) error {
 	}
 
 	fingerprint := sha256.Sum256(cert.Raw)
-	fmt.Fprintf(c.out.stderr(), "SERVER CERT SHA256: %v\n", hex.EncodeToString(fingerprint[:])) //nolint:errcheck
+	if c.kopiauiInfo {
+		fmt.Fprintf(c.out.stderr(), "SERVER CERT SHA256: %v\n", hex.EncodeToString(fingerprint[:])) //nolint:errcheck
+	}
 
 	log(ctx).Infof("writing TLS certificate to %v", c.serverStartTLSCertFile)
 
@@ -118,7 +120,10 @@ func (c *commandServerStart) startServerWithOptionalTLSAndListener(ctx context.C
 	switch {
 	case c.serverStartTLSCertFile != "" && c.serverStartTLSKeyFile != "":
 		// PEM files provided
-		fmt.Fprintf(c.out.stderr(), "SERVER ADDRESS: %shttps://%v\n", udsPfx, httpServer.Addr) //nolint:errcheck
+		if c.kopiauiInfo {
+			fmt.Fprintf(c.out.stderr(), "SERVER ADDRESS: %shttps://%v\n", udsPfx, httpServer.Addr) //nolint:errcheck
+		}
+
 		c.showServerUIPrompt(ctx)
 
 		return checkErrServerClosed(ctx, httpServer.ServeTLS(listener, c.serverStartTLSCertFile, c.serverStartTLSKeyFile), "error starting TLS server")
@@ -141,7 +146,9 @@ func (c *commandServerStart) startServerWithOptionalTLSAndListener(ctx context.C
 		}
 
 		fingerprint := sha256.Sum256(cert.Raw)
-		fmt.Fprintf(c.out.stderr(), "SERVER CERT SHA256: %v\n", hex.EncodeToString(fingerprint[:])) //nolint:errcheck
+		if c.kopiauiInfo {
+			fmt.Fprintf(c.out.stderr(), "SERVER CERT SHA256: %v\n", hex.EncodeToString(fingerprint[:])) //nolint:errcheck
+		}
 
 		if c.serverStartTLSPrintFullServerCert {
 			// dump PEM-encoded server cert, only used by KopiaUI to securely connect.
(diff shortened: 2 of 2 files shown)


csenger:
Kopia and the Repository Server have a lot of command line parameters already. This makes using Kopia and also writing good documentation more complex. I try to give the Repository Server documentation a [slight makeover](https://github.com/csenger/kopia/tree/docs/repository-server). Printing the cert fingerprint is one of the things that already have an unnecessary condition attached (printed the fingerprint only the one time the cert is generated). It's on my list of suggestions to always print the fingerprint. 

Even so the default is the current behavior, it would be good to not expand the list of flags if not necessary.  If the goal is to have a parseable json stream without plain text in between, turning it into json if  `--json-log-console` is enabled is the better option IMO. In the end this change makes `--kopiaui-info=true` still incompatible with  `--json-log-console`.

I read #4693. I think it is the cleaner solution (without knowing much about kopia's logging setup). If the problem is just keeping KopiaUI compatible, I'd be willing to take the change from #4693 and add the necessary changes to KopiaUI.

julio-lopez:
> to have a parseable json stream without plain text in between, turning it into json if `--json-log-console` is enabled is the better option

That's a good idea.
It results in well (or better) -defined semantics for JSON logging.

Also, an orthogonal change is to print the cert on `stdout` instead of `stderr`, however that is an incompatible change that also requires changes to the UI (driver) and tests.

csenger:
> > to have a parseable json stream without plain text in between, turning it into json if `--json-log-console` is enabled is the better option
> 
> That's a good idea. It results in well (or better) -defined semantics for JSON logging.

@julio-lopez I think that was @leonardoce's idea for the change in #4693. There was a [comment](https://github.com/kopia/kopia/pull/4693#discussion_r2172264799) from @jkowalski to make sure kopia-ui keeps on working. I think that's what motivated @leonardoce  to keep them as `Fprintf` and instead add a new command line flag.

But I just tested kopia-ui with @leonardoce's branch leonardoce:stray-printfs from #4693, with the electron app using `npm run dev`. kopia-ui runs as expected. The output of `kopia server start` that is spawned in server.js is still what kopia-ui expects. `detectServerParams()` picks up the config values. I'll add a comment to #4693 later with some more fmt.Fprintf()s that should be changed to log(ctx).

cli/command_server_start.go
```diff
@@ -128,6 +130,8 @@ func (c *commandServerStart) setup(svc advancedAppServices, parent commandParent
 
 	cmd.Flag("kopiaui-notifications", "Enable notifications to be printed to stdout for KopiaUI").BoolVar(&c.kopiauiNotifications)
 
+	cmd.Flag("kopiaui-info", "Outputs connection info for KopiaUI integration").Default("true").Hidden().BoolVar(&c.kopiauiInfo)
```
jkowalski, line 133: it probably makes sense to make this default to false and in the same PR pass this flag from KopiaUI to override to true.

leonardoce:
> > > to have a parseable json stream without plain text in between, turning it into json if `--json-log-console` is enabled is the better option
> > 
> > 
> > That's a good idea. It results in well (or better) -defined semantics for JSON logging.
> 
> @julio-lopez I think that was @leonardoce's idea for the change in #4693. There was a [comment](https://github.com/kopia/kopia/pull/4693#discussion_r2172264799) from @jkowalski to make sure kopia-ui keeps on working. I think that's what motivated @leonardoce to keep them as `Fprintf` and instead add a new command line flag.
> 
> But I just tested kopia-ui with @leonardoce's branch leonardoce:stray-printfs from #4693, with the electron app using `npm run dev`. kopia-ui runs as expected. The output of `kopia server start` that is spawned in server.js is still what kopia-ui expects. `detectServerParams()` picks up the config values. I'll add a comment to #4693 later with some more fmt.Fprintf()s that should be changed to log(ctx).

Yes exactly, my idea was to address that comment, and it would be definitely nice to avoid expanding the list of flags.
I'm happy to do all the changes needed once we agree on the approach.
Should we:

1. use `--json-log-console` instead of `--kopiaui-info`
2. make `--kopiaui-info` default to `false`
3. resume #4693 and do the changed needed to keep KopiaUI compatible

or do something different?
Thanks!

csenger:
I [tested the electron app  and commented on #4693](https://github.com/kopia/kopia/pull/4693#issuecomment-3417737389). The electron app worked for me with @leonardoce's changes, so I think #4693 adds no regression. Somebody knowing the internals may take a look if there is more to test.

julio-lopez:
> > > to have a parseable json stream without plain text in between, turning it into json if `--json-log-console` is enabled is the better option
> > 
> > 
> > That's a good idea. It results in well (or better) -defined semantics for JSON logging.
> 
> @julio-lopez I think that was @leonardoce's idea for the change in #4693. There was a [comment](https://github.com/kopia/kopia/pull/4693#discussion_r2172264799) from @jkowalski to make sure kopia-ui keeps on working. I think that's what motivated @leonardoce to keep them as `Fprintf` and instead add a new command line flag.

> 
> But I just tested kopia-ui with @leonardoce's branch leonardoce:stray-printfs from #4693, with the electron app using `npm run dev`. kopia-ui runs as expected. The output of `kopia server start` that is spawned in server.js is still what kopia-ui expects. `detectServerParams()` picks up the config values. I'll add a comment to #4693 later with some more fmt.Fprintf()s that should be changed to log(ctx).

@csenger thanks for pointing that out.

IIUC:

- the change in #4693 would (already) print out the server info in JSON when `--json-log-console` was specified (and would produce a clean JSON logging stream;

- with the change in #4693, starting the server in the UI **could** break if other logging flags were to be specified, such as: `--console-timestamps`, `--json-log-console`, `--log-level=warning` (suppressing the "info" messages).

- however, the "UI" would continue to work with the default logging output, that is, when no flags that affect the default logging output are specified, and currently the UI app does not specify any of those flags when launching the server.

Is this what you are saying? Does that match your understanding?

julio-lopez:
> If the goal is to have a parseable json stream without plain text in between ...

It seems that there are 2 (semi-conflicting) objectives here:
- ensure that starting the server from the UI does not break; and
- having a (clean) parseable JSON stream (without "spurious" stuff).

Is that your understanding as well?

julio-lopez:
> I'm happy to do all the changes needed once we agree on the approach. Should we:
> 
> 1. use `--json-log-console` instead of `--kopiaui-info`

IIUC, this is "almost" the same as the initial change in #4693, with maybe a minor difference that the server info would be marshaled to JSON and directly sent to `stderr` instead of "logging" it (sending it through the Rube Goldberg logging machinery).
However, this change alone would still be _incompatible_ with the UI and other uses as [noted above](#issuecomment-3444995493). As such, we should probably discard it.

---

> 2. make `--kopiaui-info` default to `false`

While the server info is needed for the UI, that is not the only use case. This information is also useful when launching a stand-alone server, for instance, to act as a repository server. And, the server address is used in those other contexts to get the port number or interface where the server is listening.

If we choose this route, the flag should have a different name, such as: `--server-info`, `--output-server-info` or similar.

There is already a `--tls-print-server-cert` flag to control whether or not the sever (public) cert is dumped. That only affects `SERVER CERTIFICATE:` line.

As @csenger mentioned, there are also a few other cases that print non-JSON output to `stderr`, such as `SERVER PASSWORD`, `SERVER CONTROL PASSWORD` and `NOTIFICATION`.

---

> 3. resume [fix(cli): remove direct fmt.Printf calls to have a clean JSON log output #4693](https://github.com/kopia/kopia/pull/4693) and do the changed needed to keep KopiaUI compatible

I assume you mean "_replace_ direct `fmt.Printf` calls with logging calls ...", is that correct?

The "advantage" of this approach is that it is "compatible by default";
but the downside is that it breaks when other logging flags are specified.

Maybe a _warning_ log message could be added notifying users that when the console log format changes, then this would break consumers of the server info, including the UI.

---

> 4. or do something different? Thanks!

Perhaps print the server info on `stdout` instead of `stderr`, and change the UI and tests accordingly. Note: this could be a breaking change for downstream dependents.

---

I'm inclined to say, go with approach (3), including the warning. And delay (4) or an alternate approach, unless folks have other ideas (or objections).

@jkowalski thoughts?

PR #4886, opened 2025-10-16
feat(server): mTLS authentication

leonardoce:
This update introduces a `--tls-ca-file` command line option for `kopia server start`, enabling users to specify a CA certificate.

With this option specified:

* A client-side certificate is requested during the TLS handshake and
  it will be validated using the CA certificate specified.

* A new authenticator is added, allowing clients to authenticate by
  supplying a valid certificate with its CN field matching the requested
  username.

Additionally, two new options have been added to `kopia repository connect server`: `--client-certificate` and `--client-key`. These options enable users to present a client certificate during the TLS handshake, allowing
authentication with it.

### DIFF
--- a/cli/command_repository_connect_server.go
+++ b/cli/command_repository_connect_server.go
@@ -16,6 +16,8 @@ type commandRepositoryConnectServer struct {
 	connectAPIServerURL                              string
 	connectAPIServerCertFingerprint                  string
 	connectAPIServerLocalCacheKeyDerivationAlgorithm string
+	connectAPIServerClientCertificate                string
+	connectAPIServerClientPrivateKey                 string
 
 	svc advancedAppServices
 	out textOutput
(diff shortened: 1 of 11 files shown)


jkowalski:
This looks great, can you add some unit tests for the new authentication mode?

leonardoce:
> This looks great, can you add some unit tests for the new authentication mode?

I added a unit test as you suggested and fixed the issues raised by the linters.
Thanks!

leonardoce:
Hi @jkowalski. Do we need anything else to get this patch reviewed?
I added the unit tests as you requested.
Thanks!

yonas:
> A client-side certificate is requested during the TLS handshake, though the server does not enforce that the client provides one.

Would calling this mTLS be misleading. The presence and validity of a client certificate should be enforced.

gbartolini:
> Would calling this mTLS be misleading. The presence and validity of a client certificate should be enforced.

I guess the reason is for back compatibility.

gbartolini:
This feature is really awesome! Do you think there’s a chance we’ll see it merged soon? Thank you!

yonas:
I still think there should atleast be a flag to enforce client presence and validation.

gbartolini:
> I still think there should atleast be a flag to enforce client presence and validation.

Could you please elaborate on that? I believe that if the feature were actually enforced, it would disrupt existing deployments. Clearly, I am missing something, so I would appreciate it if you could provide an example. Thank you!

yonas:
> > I still think there should atleast be a flag to enforce client presence and validation.
> 
> Could you please elaborate on that? I believe that if the feature were actually enforced, it would disrupt existing deployments. Clearly, I am missing something, so I would appreciate it if you could provide an example. Thank you!

If there's a new flag that enables full mTLS (enforces client certificates), then it wouldn't disrupt existing deployments.

But that's just one method.

## Recommend Method

Create two differently named features.

- "mTLS" would enforce client certificates
- "mTLS-opt-in" would not expect a client certificate to be present. However, if it's presented, it would require the client certificate to be valid.

This more properly preserves the original meaning of mTLS.

gbartolini:
> If there's a new flag that enables full mTLS (enforces client certificates), then it wouldn't disrupt existing deployments.

My interpretation is that if you want to enforce it, you supply the certificates for the authentication.

I would not require an additional option, but that's just me.

gbartolini:
@jkowalski thoughts about merging this PR? Thanks!

yonas:
> My interpretation is that if you want to enforce it, you supply the certificates for the authentication.

"A client-side certificate is requested during the TLS handshake, **though the server does not enforce that the client provides one**."

The problem is that this PR does not enforce that the client even provide a client certificate, while calling the feature mutual TLS.

It's not mutual TLS until both the server and client provide and verify each other's certificates.

> [!CAUTION]
**This should not be committed as-is.**
>
> I recommend two features, one requires a client certificate, and the other does not.

leonardoce:
Before changing the CLI interface to enable strict validation of client-side certificates - and consequently break older clients - I'd like to have feedback from the PMC.
This PR has been open for approximately two months and I'd like to allocate time for it only when the direction is clear.

gbartolini:
@jkowalski and @julio-lopez, as Project Management Committee members, could you please provide your feedback here? Thanks.

Shrekster:
Thanks for this contribution, @leonardoce. @gbartolini — we've reviewed this internally and have the following feedback:

**1. Change `VerifyClientCertIfGiven` to `RequireAndVerifyClientCert`**

The current implementation uses `tls.VerifyClientCertIfGiven`, which means when `--tls-ca-file` is set, a client without a certificate can still connect and fall through to password authentication. This doesn't align with mTLS semantics — when a server operator explicitly configures a CA file for client authentication, the expectation is that client certificates are **required**, not optional.

Please change this to `tls.RequireAndVerifyClientCert` so that when `--tls-ca-file` is provided, the server enforces that all clients must present a valid certificate signed by the specified CA. This makes the behavior explicit: no `--tls-ca-file` = standard TLS with password auth, with `--tls-ca-file` = full mTLS enforced.

The expected connection matrix with this change:

| Server Config | Client Config | Result | Behavior |
|---|---|---|---|
| No `--tls-ca-file` | No client cert | SUCCESS | Standard TLS, password auth required |
| No `--tls-ca-file` | With client cert | SUCCESS | Client cert ignored, password auth required |
| With `--tls-ca-file` | Valid client cert | SUCCESS | Full mTLS, client cert CN mapped to username |
| With `--tls-ca-file` | Invalid client cert | FAILURE | TLS handshake rejected, cert fails CA validation |
| With `--tls-ca-file` | Cert from different CA | FAILURE | TLS handshake rejected, cert not trusted by server CA |
| With `--tls-ca-file` | No client cert | FAILURE | TLS handshake rejected, server requires client cert |

**2. Add unit tests for each connection combination**

It would be great to add test coverage for each row of the matrix above, verifying both successful and failed connection scenarios to ensure the outlined behavior is correct.

cc @carlbraganza @redgoat650 @Rohit-BM18

gbartolini:
Thanks @Shrekster for the feedback, and @leonardoce for the updated contribution!

leonardoce:
I'm working on the unit tests now. I'll keep you posted.

leonardoce:
Thanks for the feedback!
I added the unit tests for each row of the proposed matrix and also for some other complementary scenarios.
Can you take a look now?

cc: @Shrekster

gbartolini:
Hi! Any news on merging this PR? @jkowalski @Shrekster

gbartolini:
Hi, is there any chance this patch will land on the trunk? Thanks!

PR #4890, opened 2025-10-20
feat(general): add stats to maintenance run

Lyndon-Li:

1. It is helpful for troubleshooting. Any data change to the repository is critical, the observability info helps to understand what happened during the maintenance and why that happened
2. It is helpful for users to understand/predict the repo's behavior. The repo data may be stored in a public cloud for which costs are sensitive to scale/duration of data stored. On the other hand, repository has its own policy to manage the data, so the data is not deleted until it is safe enough according to the policy. The observability info helps users to understand how much data is in-use, how much data is out of use and when it is deleted

The current PR adds the interfaces, structures and workflows to save, retrieve and display the statistics info for maintenance tasks.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	cloud.google.com/go/storage v1.57.0
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.19.1
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.0
-	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.2
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.3
 	github.com/alecthomas/kingpin/v2 v2.4.0
 	github.com/alecthomas/units v0.0.0-20240927000941-0f3dac36c52b
 	github.com/chmduquesne/rollinghash v4.0.0+incompatible
@@ -27,7 +27,7 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/hanwen/go-fuse/v2 v2.8.0
 	github.com/hashicorp/cronexpr v1.1.3
-	github.com/klauspost/compress v1.18.0
+	github.com/klauspost/compress v1.18.1
 	github.com/klauspost/pgzip v1.2.6
 	github.com/klauspost/reedsolomon v1.12.5
 	github.com/kopia/htmluibuild v0.0.1-0.20250916054243-fc9411799322
--- a/go.sum
+++ b/go.sum
@@ -32,8 +32,8 @@ github.com/Azure/azure-sdk-for-go/sdk/internal v1.11.2 h1:9iefClla7iYpfYWdzPCRDo
 github.com/Azure/azure-sdk-for-go/sdk/internal v1.11.2/go.mod h1:XtLgD3ZD34DAaVIIAyG3objl5DynM3CQ/vMcbBNJZGI=
 github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage v1.8.1 h1:/Zt+cDPnpC3OVDm/JKLOs7M2DKmLRIIp3XIx9pHHiig=
 github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage v1.8.1/go.mod h1:Ng3urmn6dYe8gnbCMoHHVl5APYz2txho3koEkV2o2HA=
-github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.2 h1:FwladfywkNirM+FZYLBR2kBz5C8Tg0fw5w5Y7meRXWI=
-github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.2/go.mod h1:vv5Ad0RrIoT1lJFdWBZwt4mB1+j+V8DUroixmKDTCdk=
+github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.3 h1:ZJJNFaQ86GVKQ9ehwqyAFE6pIfyicpuJ8IkVaPBc6/4=
+github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.3/go.mod h1:URuDvhmATVKqHBH9/0nOiNKk0+YcwfQ3WkK5PqHKxc8=
 github.com/AzureAD/microsoft-authentication-extensions-for-go/cache v0.1.1 h1:WJTmL004Abzc5wDB5VtZG2PJk5ndYDgVacGqfirKxjM=
 github.com/AzureAD/microsoft-authentication-extensions-for-go/cache v0.1.1/go.mod h1:tCcJZ0uHAmvjsVYzEFivsRTN00oz5BEsRgQHu5JZ9WE=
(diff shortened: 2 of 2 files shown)


repo/maintenance/maintenance_schedule.go
```diff
@@ -32,10 +33,11 @@ const maxRetainedRunInfoPerRunType = 50
...
 type RunInfo struct {
-	Start   time.Time `json:"start"`
-	End     time.Time `json:"end"`
-	Success bool      `json:"success,omitempty"`
-	Error   string    `json:"error,omitempty"`
+	Start   time.Time                  `json:"start"`
+	End     time.Time                  `json:"end"`
+	Success bool                       `json:"success,omitempty"`
+	Error   string                     `json:"error,omitempty"`
+	Stats   *maintenancestats.RawStats `json:"stats,omitempty"`
```
julio-lopez:
```suggestion
	Extra   []maintenancestats.Extra `json:"extra,omitempty"`
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"encoding/json"
+)
+
+// RawStats holds the raw data for maintenance statistics.
+type RawStats struct {
+	Kind string          `json:"kind,omitempty"`
+	Raw  json.RawMessage `json:"raw,omitempty"`
+}
```
julio-lopez:
```suggestion
type Extra struct {
	Kind string          `json:"kind,omitempty"`
	Data  json.RawMessage `json:"data,omitempty"`
}
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,28 @@
...
+
+// Stats defines the methods for maintenance statistics.
+type Stats interface {
```
julio-lopez: Either `Summarizer` or `ExtraSumarizer`
```suggestion
type Summarizer interface {
```

julio-lopez:
As soon as feedback is addressed, CI can run and this PR can be merged afterwards

- Ref: #4848

Lyndon-Li:
@julio-lopez 
I have addressed all the comments.

However, for "Add marshaling unit tests", I cannot add UT for the current PR.
The testing code are more like end to end function test instead of UT. So for example, for `TestMaintenanceInfoSimple`:

- I cannot control the the retrieved maintenance info, so cannot make sure the test covers every path of the changed code
- It is in a separate package `cli_test`, I cannot make a UT to test a private function `getMessageFromRun` in `cli` package

Which way do you prefer? And can you give some advice?

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,42 @@
...
+	Data json.RawMessage `json:"data,omitempty"`
+}
+
+// Summarizer defines the methods for summarizing a maintenance statistics.
+type Summarizer interface {
+	Summary() string
+}
+
+// Kind defines the methods for detecting kind of a maintenance statistics.
+type Kind interface {
```
julio-lopez, line 22: No change requested, only a comment: it seems that these did not need to be separated, what's the rationale for separating them?
Anyway, we can merge as is and roll the ball forward.

julio-lopez:
> However, for "Add marshaling unit tests", I cannot add UT for the current PR. The testing code are more like end to end function test instead of UT. So for example, for `TestMaintenanceInfoSimple`:
> 
> * I cannot control the the retrieved maintenance info, so cannot make sure the test covers every path of the changed code
> * It is in a separate package `cli_test`, I cannot make a UT to test a private function `getMessageFromRun` in `cli` package
> 
> Which way do you prefer? And can you give some advice?

The suggestion is to have simple tests that ensure the struct marshaling/unmarshaling "round-trips", that is:
- instantiate the struct;
- marshal it to JSON, it should not error out;
- then unmarshal the resulting JSON back into a separate instance of the struct;
- ensure that the values in the unmarshaled struct match the expected values.

These tests are simple, they **DO NOT** require a complex setup such to run maintenance, forcing it to create a schedule, etc.

cli/command_maintenance_info.go
```diff
@@ -113,3 +115,24 @@ func (c *commandMaintenanceInfo) displayCycleInfo(cp *maintenance.CycleParams, t
...
 }
+
+func getMessageFromRun(extra []maintenancestats.Extra) string {
+	succeed := "SUCCESS"
+
+	if len(extra) == 0 {
+		return succeed
+	}
+
+	extraStr := ""
```
julio-lopez, line 126:
```suggestion
	extraStr := ""

```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	return Extra{
+		Kind: stats.Kind(),
+		Data: bytes,
+	}, nil
+}
+
+// BuildFromExtra builds maintenance statistics from an Extra and returns a Summarizer.
+func BuildFromExtra(extra Extra) (Summarizer, error) {
```
julio-lopez, line 40:
```suggestion
func BuildFromExtra(_ Extra) (Summarizer, error) {
```

repo/maintenancestats/types.go
```diff
@@ -0,0 +1,42 @@
...
+// Kind defines the methods for detecting kind of a maintenance statistics.
+type Kind interface {
+	Kind() string
+}
+
+// BuildExtra builds an Extra from maintenance statistics.
+func BuildExtra(stats Kind) (Extra, error) {
+	bytes, err := json.Marshal(stats)
+	if err != nil {
+		return Extra{}, errors.Wrapf(err, "error marshalling stats %v", stats)
```
julio-lopez, line 30:
```suggestion
		return Extra{}, errors.Wrapf(err, "error marshaling stats %v", stats)
```

julio-lopez:
Merging as is, I'll take care of the linter issues in a separate PR.
- See #4897

julio-lopez:
For the PRs to follow, please run the linter and other CI jobs in your computer, or in your repository so you catch any test and linter errors early. This reduces the number of back-and-forth and overall speeds up the PR review process and thus, merging the changes.

repo/maintenancestats/types.go
Lyndon-Li: > it seems that these did not need to be separated, what's the rationale for separating them?

I just feel that they don't need to be from the same implementation --- the task functions who generate the statistics structs don't need to know what `Summary ` mean, this `Summarizer ` should not appear in these functions' signature (return value).

Lyndon-Li:
> > However, for "Add marshaling unit tests", I cannot add UT for the current PR. The testing code are more like end to end function test instead of UT. So for example, for `TestMaintenanceInfoSimple`:
> > 
> > * I cannot control the the retrieved maintenance info, so cannot make sure the test covers every path of the changed code
> > * It is in a separate package `cli_test`, I cannot make a UT to test a private function `getMessageFromRun` in `cli` package
> > 
> > Which way do you prefer? And can you give some advice?
> 
> The suggestion is to have simple tests that ensure the struct marshaling/unmarshaling "round-trips", that is:
> 
> * instantiate the struct;
> * marshal it to JSON, it should not error out;
> * then unmarshal the resulting JSON back into a separate instance of the struct;
> * ensure that the values in the unmarshaled struct match the expected values.
> 
> These tests are simple, they **DO NOT** require a complex setup such to run maintenance, forcing it to create a schedule, etc.

Sure, so it is more like a UT. Will add it in the next PR.

PR #4900, opened 2025-10-23
feat(general): add stats to maintenance run - CleanupMarkers

Lyndon-Li:
Maintenance is critical for healthy of the repository.
On the other hand, Maintenance is complex, because it runs multiple sub tasks each may generate different results according to the maintenance policy. The results may include deleting/combining/adding data/metadata to the repository.

It is worthy to add more observability for these tasks for below reasons:

1. It is helpful for troubleshooting. Any data change to the repository is critical, the observability info helps to understand what happened during the maintenance and why that happened
2. It is helpful for users to understand/predict the repo's behavior. The repo data may be stored in a public cloud for which costs are sensitive to scale/duration of data stored. On the other hand, repository has its own policy to manage the data, so the data is not deleted until it is safe enough according to the policy. The observability info helps users to understand how much data is in-use, how much data is out of use and when it is deleted

There will be a serial of PRs to add observability info for each sub task.

The current PR add the stats info for CleanupMarkers sub task.

PR #4911, opened 2025-10-24
fix(general): field and variable names

julio-lopez:
Fixes field names in `CleanupMarkersStats` persistent struct.
Also renames variables for consistency.

- Followup fix for #4900
- #4848


julio-lopez:
FYI @Lyndon-Li

PR #4916, opened 2025-10-27
feat(general): add stats to maintenance run - CleanupSupersededIndexes

Lyndon-Li:

The current PR add the stats info for CleanupSupersededIndexes sub task.

### DIFF
--- a/internal/epoch/epoch_manager.go
+++ b/internal/epoch/epoch_manager.go
@@ -391,15 +391,15 @@ func (e *Manager) cleanupWatermarks(ctx context.Context, cs CurrentSnapshot, p *
 }
 
 // CleanupSupersededIndexes cleans up the indexes which have been superseded by compacted ones.
-func (e *Manager) CleanupSupersededIndexes(ctx context.Context) error {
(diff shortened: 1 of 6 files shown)


internal/epoch/epoch_manager.go
```diff
@@ -422,24 +422,31 @@ func (e *Manager) CleanupSupersededIndexes(ctx context.Context) error {
 	// that was written sufficiently long ago.
 	blobs, err := blob.ListAllBlobs(ctx, e.st, UncompactedIndexBlobPrefix)
 	if err != nil {
-		return errors.Wrap(err, "error listing uncompacted blobs")
+		return nil, errors.Wrap(err, "error listing uncompacted blobs")
 	}
 
 	var toDelete []blob.ID
 
+	var deleltedTotalSize int64
```
julio-lopez, line 430:
```suggestion
	var deletedTotalSize int64
```

andrason, line 430: Typo, should be `deletedTotalSize`.

julio-lopez:
- #4848
- #4920

PR #4925, opened 2025-10-28
refactor(general): rename GenerateRangeCheckpointStats fields

julio-lopez:
Rename `GenerateRangeCheckpointStats` fields to be more consistent with respect to `epoch.RangeMetadata`.

```go
type GenerateRangeCheckpointStats struct {
	RangeMinEpoch int `json:"rangeMinEpoch"`
	RangeMaxEpoch int `json:"rangeMaxEpoch"`
}
```

- Follow-up to #4924


andrason (approved):
LGTM

PR #4931, opened 2025-10-29
chore(ci): upgrade to golangci-lint v2.5.0

NathanBaulch:
I recently tried linting this project to check something I was working on and golangci-lint v2.1.2 paniced because I'm using Go 1.25 and apparently they don't play nice together. So I figured I'd upgrade to v2.5.0 (latest).

I've pushed fixes for each linter in separate commits since they're generally quite opinionated and each will likely need to be reviewed in isolation.

Notes:
* `wsl_v5` (successor to `wsl`) - **disabled** - created a lot of churn.
* `embeddedstructfieldcheck` - **disabled** - not against this but it created false positives related to comments I think.
* `noinlineerr` - **disabled** - personal preference, feel free to overrule.
* ~`gofumpt` - removed naked returns, somewhat controversial in the community.~
* `noctx` - used `context.Background()` in a couple of places, maybe should be `context.TODO()`.
* `godot` - some packages have a `doc.go`, most don't. I went with a light touch where possible.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -101,6 +101,7 @@ linters:
     - musttag
     - nilnil
     - nlreturn
+    - noctx
     - noinlineerr # inline error handling is a common Go idiom used in this codebase
     - nonamedreturns
     - paralleltest
--- a/internal/apiclient/apiclient.go
+++ b/internal/apiclient/apiclient.go
@@ -224,8 +224,8 @@ func NewKopiaAPIClient(options Options) (*KopiaAPIClient, error) {
 		tp, _ := transport.(*http.Transport)
 		transport = tp.Clone()
 		tp, _ = transport.(*http.Transport)
-		tp.DialContext = func(ctx context.Context, _, _ string) (net.Conn, error) {
-			dial, err := (&net.Dialer{}).DialContext(ctx, "unix", u.Path)
+		tp.DialContext = func(_ context.Context, _, _ string) (net.Conn, error) {
+			dial, err := net.Dial("unix", u.Path)
 			return dial, errors.Wrap(err, "Failed to connect to socket: "+options.BaseURL)
 		}
 	}
--- a/cli/command_server_tls.go
+++ b/cli/command_server_tls.go
@@ -48,9 +48,9 @@ func (c *commandServerStart) startServerWithOptionalTLS(ctx context.Context, htt
 	switch len(listeners) {
 	case 0:
 		if after, ok := strings.CutPrefix(httpServer.Addr, "unix:"); ok {
-			l, err = (&net.ListenConfig{}).Listen(ctx, "unix", after)
+			l, err = net.Listen("unix", after)
 		} else {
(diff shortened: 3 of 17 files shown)


jkowalski:
can we keep wsl_v5 but override its config to be the same as the old one?

jkowalski:
btw, we'll first cut v0.22 release and then merge this change along with upgrading to golang 1.25.

NathanBaulch:
Possibly @jkowalski, I didn't spend too much time on `wsl_v5` in case it wasn't wanted.

NathanBaulch:
OK, `wsl_v5` has been enabled, along with ~30 whitespace fixes to get it passing.
I've disabled the 4 rules (`assign`, `expr`, `err`, `return`) that caused > 10 issues each since that's probably a good indication that this project isn't following those ones.

.golangci.yml
```diff
@@ -80,9 +80,16 @@ linters:
         - (*go.uber.org/zap.SugaredLogger).With
     misspell:
       locale: US
+    wsl_v5:
+      disable:
+        - assign
+        - err
+        - expr
+        - return
```
julio-lopez, line 89:
```suggestion
    wsl_v5:
      allow-whole-block: true
      disable:
        - assign
        - err
        - expr
        - return
```

internal/apiclient/apiclient.go
```diff
@@ -224,8 +224,8 @@ func NewKopiaAPIClient(options Options) (*KopiaAPIClient, error) {
 		tp, _ := transport.(*http.Transport)
 		transport = tp.Clone()
 		tp, _ = transport.(*http.Transport)
-		tp.DialContext = func(_ context.Context, _, _ string) (net.Conn, error) {
-			dial, err := net.Dial("unix", u.Path)
+		tp.DialContext = func(ctx context.Context, _, _ string) (net.Conn, error) {
```
julio-lopez: It is not completely clear (to me) what the side effect of the `ctx` changes might be w.r.t context cancellation.

Can you put the `ctx` changes in a separate PR please? maybe by disabling the `noctx` linter in the config for now.

That way, all the other changes in this PR, including the linter upgrade. Then the `ctx` changes can be merged after the release.

.golangci.yml
```diff
@@ -93,6 +100,7 @@ linters:
     - musttag
     - nilnil
     - nlreturn
+    - noinlineerr
```
julio-lopez:
```suggestion
    - noinlineerr # inline error handling is a common Go idiom used in this codebase
```

.golangci.yml
NathanBaulch: This doesn't appear to make any difference but sure thing. :+1:

NathanBaulch: Correction, it did make a difference in 5 places where variables are defined just prior to a goroutine. I'll revert those empty lines.

julio-lopez:
Thanks!

internal/apiclient/apiclient.go
julio-lopez: - #4972

PR #4948, opened 2025-11-03
feat(general): add stats to maintenance run - CompactIndexesStats

Lyndon-Li:

The current PR add the stats info for CompactIndexesStats sub task.

### DIFF
--- a/repo/content/indexblob/index_blob.go
+++ b/repo/content/indexblob/index_blob.go
@@ -8,13 +8,14 @@ import (
 	"github.com/kopia/kopia/internal/gather"
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/content/index"
+	"github.com/kopia/kopia/repo/maintenancestats"
 )
 
 // Manager is the API of index blob manager as used by content manager.
 type Manager interface {
 	WriteIndexBlobs(ctx context.Context, data []gather.Bytes, suffix blob.ID) ([]blob.Metadata, error)
 	ListActiveIndexBlobs(ctx context.Context) ([]Metadata, time.Time, error)
-	Compact(ctx context.Context, opts CompactOptions) error
+	Compact(ctx context.Context, opts CompactOptions) (*maintenancestats.CompactIndexesStats, error)
 	Invalidate()
 }
 
--- a/repo/content/indexblob/index_blob_manager_v0.go
+++ b/repo/content/indexblob/index_blob_manager_v0.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/content/index"
 	"github.com/kopia/kopia/repo/format"
+	"github.com/kopia/kopia/repo/maintenancestats"
 )
 
 // V0IndexBlobPrefix is the prefix for all legacy (v0) index blobs.
@@ -148,28 +149,35 @@ func (m *ManagerV0) Invalidate() {
 
 // Compact performs compaction of index blobs by merging smaller ones into larger
 // and registering compaction and cleanup blobs in the repository.
-func (m *ManagerV0) Compact(ctx context.Context, opt CompactOptions) error {
+func (m *ManagerV0) Compact(ctx context.Context, opt CompactOptions) (*maintenancestats.CompactIndexesStats, error) {
 	indexBlobs, _, err := m.ListActiveIndexBlobs(ctx)
 	if err != nil {
-		return errors.Wrap(err, "error listing active index blobs")
+		return nil, errors.Wrap(err, "error listing active index blobs")
 	}
 
 	mp, mperr := m.formattingOptions.GetMutableParameters(ctx)
 	if mperr != nil {
-		return errors.Wrap(mperr, "mutable parameters")
+		return nil, errors.Wrap(mperr, "mutable parameters")
 	}
 
 	blobsToCompact := m.getBlobsToCompact(ctx, indexBlobs, opt, mp)
 
-	if err := m.compactIndexBlobs(ctx, blobsToCompact, opt); err != nil {
-		return errors.Wrap(err, "error performing compaction")
+	compacted, err := m.compactIndexBlobs(ctx, blobsToCompact, opt)
+	if err != nil {
+		return nil, errors.Wrap(err, "error performing compaction")
 	}
 
 	if err := m.cleanup(ctx, opt.maxEventualConsistencySettleTime()); err != nil {
-		return errors.Wrap(err, "error cleaning up index blobs")
+		return nil, errors.Wrap(err, "error cleaning up index blobs")
 	}
 
-	return nil
+	if compacted {
+		return &maintenancestats.CompactIndexesStats{
+			DroppedContentsDeletedBefore: opt.DropDeletedBefore,
+		}, nil
+	}
+
+	return nil, nil
 }
 
 func (m *ManagerV0) registerCompaction(ctx context.Context, inputs, outputs []blob.Metadata, maxEventualConsistencySettleTime time.Duration) error {
@@ -489,14 +497,14 @@ func (m *ManagerV0) getBlobsToCompact(ctx context.Context, indexBlobs []Metadata
 	return nonCompactedBlobs
 }
 
-func (m *ManagerV0) compactIndexBlobs(ctx context.Context, indexBlobs []Metadata, opt CompactOptions) error {
+func (m *ManagerV0) compactIndexBlobs(ctx context.Context, indexBlobs []Metadata, opt CompactOptions) (bool, error) {
 	if len(indexBlobs) <= 1 && opt.DropDeletedBefore.IsZero() && len(opt.DropContents) == 0 {
-		return nil
+		return false, nil
 	}
 
 	mp, mperr := m.formattingOptions.GetMutableParameters(ctx)
 	if mperr != nil {
-		return errors.Wrap(mperr, "mutable parameters")
+		return false, errors.Wrap(mperr, "mutable parameters")
 	}
 
 	bld := make(index.Builder)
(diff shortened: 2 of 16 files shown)


Lyndon-Li:
This task performs quite differently for V0 and V1 format, and is much simpler for V1 format.
We could include more info in the stats for V0 format, but since V0 is legacy, I just ignored them and only include the info that is shared by V0 and V1.

Let me know if we want the stats to be more informative for V0 format.

julio-lopez:
- #4848

repo/content/indexblob/index_blob.go
```diff
@@ -8,13 +8,14 @@ import (
...
 	"github.com/kopia/kopia/repo/content/index"
+	"github.com/kopia/kopia/repo/maintenancestats"
 )
 
 // Manager is the API of index blob manager as used by content manager.
 type Manager interface {
 	WriteIndexBlobs(ctx context.Context, data []gather.Bytes, suffix blob.ID) ([]blob.Metadata, error)
 	ListActiveIndexBlobs(ctx context.Context) ([]Metadata, time.Time, error)
-	Compact(ctx context.Context, opts CompactOptions) error
+	Compact(ctx context.Context, opts CompactOptions) (*maintenancestats.DropDeletedIndexesStats, error)
```
julio-lopez: `indexblob.Manager.Compact` is used for two conceptually distinct and separate tasks:
- dropping deleted contents;
- legacy index compaction (V0 indexes only)

The `Compact` function has different implementations depending on the index manager, albeit the same function signature.

Instead of returning stats from the `Compact` function, it should be enough to return the stats from `runTaskDropDeletedContentsFull`
That would simplify this change.

repo/maintenancestats/stats_compact_indexes.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"time"
+
+	"github.com/kopia/kopia/internal/contentlog"
+)
+
+const dropDeletedIndexesStatsKind = "dropDeletedIndexesStats"
+
+// DropDeletedIndexesStats are the stats for dropping deleted indexes.
+type DropDeletedIndexesStats struct {
```
julio-lopez: The current struct name is misleading, and thus confusing.

There are two conceptually distinct and separate maintenance tasks:
- dropping deleted contents;
- legacy index compaction (V0 indexes only)

In both of these operations what is "being dropped" is _contents_ that were deleted before a given time, and not indexes before a given time.

The `indexblob.Manager.Compact` function is called in both operations. The `Compact` function has different implementations depending on the index manager, albeit the same function signature.

---

Instead of returning stats from the `indexblob.Manager.Compact` function, either one of the following approaches can be taken.

1. Put the stats into separate structs that are associated with the maintenance operations, and not the specific implementation. the stats should be instantiated and returned from `runTaskDropDeletedContentsFull` ~and `runTaskIndexCompactionQuick`,~ that is:
    - `runTaskDropDeletedContentsFull` would return a `DropDeletedContentStats`; (as in #4949)
    - ~`runTaskIndexCompactionQuick` would return an `CompactIndexStats`.~ Edit: the `dropBefore` parameter is always 0 while compacting legacy indexes during quick maintenance. Attempting to surface other stats from V0 index compaction at this point is likely not needed.

~2. Use the same stats struct type for conceptually separate operations. The struct name would be changed to `CompactIndexStats`. In this case, the stats can be instantiated (a) preferably in `runTaskDropDeletedContentsFull` and `runTaskIndexCompactionQuick`; or (b) in `content.SharedManager.CompactIndexes` (less desirable option).~

In either case, neither the signature of the `indexblob.Manager.Compact`  function, nor its implementations need to change.

~Going with approach 2.a for now is simpler, and allows changing to approach 1. down the road if there is a need for it.~

repo/maintenancestats/stats_compact_indexes.go
```diff
@@ -0,0 +1,32 @@
...
+	DroppedBefore time.Time `json:"droppedBefore"`
```
julio-lopez:
```suggestion
	DroppedContentsDeletedBefore time.Time `json:"droppedContentsDeletedBefore"`
```

repo/maintenancestats/stats_compact_indexes.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// WriteValueTo writes the stats to JSONWriter.
+func (ds *DropDeletedIndexesStats) WriteValueTo(jw *contentlog.JSONWriter) {
+	jw.BeginObjectField(ds.Kind())
+	jw.TimeField("droppedBefore", ds.DroppedBefore)
```
julio-lopez:
```suggestion
	jw.TimeField("droppedContentsDeletedBefore", ds.DroppedContentsDeletedBefore)
```

repo/maintenancestats/stats_compact_indexes.go
```diff
@@ -0,0 +1,32 @@
...
+	jw.EndObject()
+}
+
+// Summary generates a human readable summary for the stats.
+func (ds *DropDeletedIndexesStats) Summary() string {
+	return fmt.Sprintf("Dropped deleted indexes before %v", ds.DroppedBefore)
```
julio-lopez:
```suggestion
	return fmt.Sprintf("Dropped contents deleted before %v", ds.DroppedBefore)
```

repo/maintenancestats/stats_compact_indexes.go
julio-lopez:
```suggestion
const compactIndexesStatsKind = "compactIndexesStats"

// DropDeletedIndexesStats are the stats for dropping deleted indexes.
type CompactIndexesStats struct {
```

julio-lopez:
> but since V0 is legacy, I just ignored them and only include the info that is shared by V0 and V1.

That's OK. There is NO need to include details about the V0 implementation at the moment.

What is desirable is to clearly call out the 2 conceptually separate maintenance tasks.

repo/maintenancestats/stats_compact_indexes.go
Lyndon-Li: If we don't go deep into `indexblob.Manager.Compact`, we will not be able to report the stats accurately.
The `Compact` function may give up on specific conditions and make the task actually do nothing.
At present, these conditions are all trivial, e.g.:
```
	if opt.DropDeletedBefore.IsZero() {
		return nil, nil
	}
``` 
and
```
	if ts.Before(cs.DeletionWatermark) {
		contentlog.Log2(ctx, e.log,
			"ignoring attempt to move deletion watermark time backwards",
			logparam.Time("ts", ts),
			logparam.Time("deletionWatermark", cs.DeletionWatermark),
		)

		return nil
	}
```
However, once we have more of these conditions in future, this may be a problem.

Conclusively, without going into `indexblob.Manager.Compact`, we report `Dropped deleted contents before %v` for every maintenance, but it may not be the truth.

julio-lopez: I see. Thanks for pointing that out.

That also means that `epoch.Manager.AdvanceDeletionWatermark` should return the stats.

Then, if the same "stats struct" is used for both cases ("dropping deleted contents" and "compacting indexes"), how should the stats struct be named then? `CompactIndexesStats` ?
And in that case, its only field should be named `DroppedContentsDeletedBefore`.

Does that work?

Lyndon-Li: > That also means that epoch.Manager.AdvanceDeletionWatermark should return the stats

Yes, otherwise, `AdvanceDeletionWatermark` should at least return one more boolean value which indicates deletionWatermark is moved. If we think this stats is passed too deep, we can use the boolean value.

Which one do you prefer?

julio-lopez: Yeah, an additional return `bool` works. It can be changed down the road if needed.

Lyndon-Li: > Then, if the same "stats struct" is used for both cases ("dropping deleted contents" and "compacting indexes"), how should the stats struct be named then? CompactIndexesStats ?

`CompactIndexesStats` is somehow abstract to me, but since it is not exposed to users and the existing function and code also use this name, let's use it until we find some better name to cover the both cases. Then we can come back to modify it.

Lyndon-Li: > And in that case, its only field should be named DroppedContentsDeletedBefore.

Yes, let me change to it.

Lyndon-Li: > Yeah, an additional return bool works. It can be changed down the road if needed.

OK. Will change accordingly.

julio-lopez: > `CompactIndexesStats` is somehow abstract to me, but since it is not exposed to users and the existing function and code also use this name, let's use it until we find some better name to cover the both cases. Then we can come back to modify it.

The "kind" name (`"compactIndexesStats"`) would be persisted in the maintenance info / schedule. That becomes harder to change without dropping stats, so it'd be good to avoid having to change it. But yes, it could be changed later if required.

repo/content/indexblob/index_blob.go
Lyndon-Li: See discussions in https://github.com/kopia/kopia/pull/4948#discussion_r2487692809

repo/maintenancestats/stats_compact_indexes.go
Lyndon-Li: Done all changes.

repo/maintenancestats/stats_compact_indexes.go
Lyndon-Li: Done

repo/maintenancestats/stats_compact_indexes.go
Lyndon-Li: Done

repo/maintenancestats/stats_compact_indexes.go
Lyndon-Li: Done

repo/maintenancestats/stats_compact_indexes.go
Lyndon-Li: Done

repo/content/indexblob/index_blob_manager_v0.go
```diff
@@ -148,28 +149,35 @@ func (m *ManagerV0) Invalidate() {
...
-		return errors.Wrap(err, "error cleaning up index blobs")
+		return nil, errors.Wrap(err, "error cleaning up index blobs")
 	}
 
-	return nil
+	if compacted {
+		return &maintenancestats.CompactIndexesStats{
+			DroppedContentsDeletedBefore: opt.DropDeletedBefore,
+		}, nil
+	}
```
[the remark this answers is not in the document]
julio-lopez: Watermark advancement is what "drops deleted contents" in V1.

julio-lopez:
Thanks!

PR #4955, opened 2025-11-05
feat(general): add stats to maintenance run - DeleteUnreferencedPacks

Lyndon-Li:

The current PR add the stats info for DeleteUnreferencedPacks sub task.

### DIFF
--- a/cli/command_blob_gc.go
+++ b/cli/command_blob_gc.go
@@ -39,12 +39,12 @@ func (c *commandBlobGC) run(ctx context.Context, rep repo.DirectRepositoryWriter
 		Prefix:   blob.ID(c.prefix),
 	}
 
-	n, err := maintenance.DeleteUnreferencedBlobs(ctx, rep, opts, c.safety)
+	stats, err := maintenance.DeleteUnreferencedBlobs(ctx, rep, opts, c.safety)
 	if err != nil {
 		return errors.Wrap(err, "error deleting unreferenced blobs")
 	}
 
-	if opts.DryRun && n > 0 {
+	if opts.DryRun && stats.UnreferencedPackCount > 0 {
 		log(ctx).Info("Pass --delete=yes to delete.")
 	}
 
--- a/repo/maintenance/blob_gc.go
+++ b/repo/maintenance/blob_gc.go
@@ -14,6 +14,7 @@ import (
 	"github.com/kopia/kopia/repo"
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/content"
+	"github.com/kopia/kopia/repo/maintenancestats"
 )
 
 // DeleteUnreferencedBlobsOptions provides option for blob garbage collection algorithm.
@@ -24,10 +25,10 @@ type DeleteUnreferencedBlobsOptions struct {
 	NotAfterTime time.Time
 }
 
-// DeleteUnreferencedBlobs deletes o was created after maintenance startederenced by index entries.
+// DeleteUnreferencedBlobs deletes blobs that are unreferenced by index entries.
 //
 //nolint:gocyclo,funlen
-func DeleteUnreferencedBlobs(ctx context.Context, rep repo.DirectRepositoryWriter, opt DeleteUnreferencedBlobsOptions, safety SafetyParameters) (int, error) {
+func DeleteUnreferencedBlobs(ctx context.Context, rep repo.DirectRepositoryWriter, opt DeleteUnreferencedBlobsOptions, safety SafetyParameters) (*maintenancestats.DeleteUnreferencedPacksStats, error) {
 	ctx = contentlog.WithParams(ctx,
 		logparam.String("span:blob-gc", contentlog.RandomSpanID()))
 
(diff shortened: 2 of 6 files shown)


cli/command_blob_gc.go
```diff
@@ -39,12 +39,12 @@ func (c *commandBlobGC) run(ctx context.Context, rep repo.DirectRepositoryWriter
...
 	}
 
-	n, err := maintenance.DeleteUnreferencedBlobs(ctx, rep, opts, c.safety)
+	stats, err := maintenance.DeleteUnreferencedBlobs(ctx, rep, opts, c.safety)
 	if err != nil {
 		return errors.Wrap(err, "error deleting unreferenced blobs")
 	}
 
-	if opts.DryRun && n > 0 {
+	if opts.DryRun && stats.UnreferencedBlobCount > 0 {
```
Lyndon-Li: Previously `DeleteUnreferencedBlobs` returned `int(del)` which is always 0 when `opts.DryRun` is true. So changed to check `stats.UnreferencedBlobCount`.

julio-lopez: Makes sense. That was a subtle bug.

Aside: the `blob gc` command is deprecated and dangerous (it has no safeguards), and will be removed in the future.

repo/maintenancestats/stats_delete_unreferenced_packs.go
```diff
@@ -0,0 +1,41 @@
...
+const deleteUnreferencedPacksStatsKind = "deleteUnreferencedPacksStats"
+
+// DeleteUnreferencedPacksStats are the stats for deleting unreferenced packs.
+type DeleteUnreferencedPacksStats struct {
+	UnreferencedBlobCount uint32 `json:"unreferencedBlobCount"`
+	UnreferencedSize      int64  `json:"unreferencedSize"`
+	DeletedBlobCount      uint32 `json:"deletedBlobCount"`
+	DeletedSize           int64  `json:"deletedSize"`
+	RetainedBlobCount     uint32 `json:"retainedBlobCount"`
+	RetainedSize          int64  `json:"retainedSize"`
```
julio-lopez: It makes sense to use here the term "pack" for all the fields.

And `TotalSize` for the cumulative pack sizes for consistency with the field names in other stat structs, such as `CleanupSupersededIndexesStats` and `CompactSingleEpochStats`

```suggestion
	UnreferencedPackCount uint32 `json:"unreferencedPackCount"`
	UnreferencedTotalSize int64  `json:"unreferencedTotalSize"`
	DeletedPackCount      uint32 `json:"deletedPackCount"`
	DeletedTotalSize      int64  `json:"deletedTotalSize"`
	RetainedPackCount     uint32 `json:"retainedPackCount"`
	RetainedTotalSize     int64  `json:"retainedTotalSize"`
```

Lyndon-Li: Done

repo/maintenance/blob_gc.go
```diff
@@ -131,32 +138,37 @@ func DeleteUnreferencedBlobs(ctx context.Context, rep repo.DirectRepositoryWrite
...
-		"Found blobs to delete",
-		logparam.UInt32("count", unreferencedCount),
-		logparam.Int64("bytes", unreferencedSize))
+	retainedCount, retainedSize := retained.Approximate()
+
+	result := &maintenancestats.DeleteUnreferencedPacksStats{
+		UnreferencedBlobCount: unreferencedCount,
+		UnreferencedSize:      unreferencedSize,
+		RetainedBlobCount:     retainedCount,
+		RetainedSize:          retainedSize,
```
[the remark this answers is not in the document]
Lyndon-Li: Done

repo/maintenance/blob_gc.go
```diff
@@ -27,7 +28,7 @@ type DeleteUnreferencedBlobsOptions struct {
 // DeleteUnreferencedBlobs deletes o was created after maintenance startederenced by index entries.
```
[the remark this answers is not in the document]
Lyndon-Li: Done

repo/maintenancestats/stats_delete_unreferenced_packs.go
```diff
@@ -0,0 +1,41 @@
...
+	jw.UInt32Field("deletedBlobCount", ds.DeletedBlobCount)
+	jw.Int64Field("deletedSize", ds.DeletedSize)
+	jw.UInt32Field("retainedBlobCount", ds.RetainedBlobCount)
+	jw.Int64Field("retainedSize", ds.RetainedSize)
+	jw.EndObject()
+}
+
+// Summary generates a human readable summary for the stats.
+func (ds *DeleteUnreferencedPacksStats) Summary() string {
+	return fmt.Sprintf("Found %v(%v) unreferenced pack blobs, deleted %v(%v) and retained %v(%v).", ds.UnreferencedBlobCount, ds.UnreferencedSize, ds.DeletedBlobCount, ds.DeletedSize, ds.RetainedBlobCount, ds.RetainedSize)
```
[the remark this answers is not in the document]
Lyndon-Li: Not all values could be aligned to KB because the value may be smaller than 1KB; or if the value is several KBs, aligning it to KB would lose significant accuracy.

This is consistent with other stats. So if we want to do this, let's change for all stats in a separate PR.

julio-lopez:
- #4848

julio-lopez:
Thanks!

PR #4959, opened 2025-11-06
feat(general): add stats to maintenance run - CleanupLogs

Lyndon-Li:

The current PR add the stats info for CleanupLogs sub task.

### DIFF
--- a/repo/maintenance/cleanup_logs.go
+++ b/repo/maintenance/cleanup_logs.go
@@ -12,6 +12,7 @@ import (
 	"github.com/kopia/kopia/internal/contentlog/logparam"
 	"github.com/kopia/kopia/repo"
 	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/maintenancestats"
 )
 
 // LogRetentionOptions provides options for logs retention.
@@ -43,7 +44,7 @@ func defaultLogRetention() LogRetentionOptions {
 }
 
 // CleanupLogs deletes old logs blobs beyond certain age, total size or count.
-func CleanupLogs(ctx context.Context, rep repo.DirectRepositoryWriter, opt LogRetentionOptions) ([]blob.Metadata, error) {
+func CleanupLogs(ctx context.Context, rep repo.DirectRepositoryWriter, opt LogRetentionOptions) (*maintenancestats.CleanupLogsStats, error) {
 	ctx = contentlog.WithParams(ctx,
 		logparam.String("span:cleanup-logs", contentlog.RandomSpanID()))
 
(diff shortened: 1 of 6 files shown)


repo/maintenancestats/stats_clean_up_log.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/kopia/kopia/internal/contentlog"
+)
+
+const cleanupLogsStatsKind = "cleanupLogsStats"
+
+// CleanupLogsStats are the stats for cleanning up logs.
+type CleanupLogsStats struct {
+	ToDeleteBlobCount int   `json:"toDeleteBlobCount"`
+	ToDeleteBlobSize  int64 `json:"toDeleteBlobSize"`
+	DeletedBlobCount  int   `json:"deletedBlobCount"`
```
Lyndon-Li, line 15: At present, `Deletedxxx` always equals to `ToDeletexxx` if dryRun is FALSE. But it hasn't to be in the case of partially successes in deletion.
To support future enhancement and consider the consistent with other stats, I am adding both of them using them appropriately in different scenarios.

repo/maintenance/cleanup_logs.go
```diff
@@ -88,18 +89,32 @@ func CleanupLogs(ctx context.Context, rep repo.DirectRepositoryWriter, opt LogRe
 
 	toDelete := allLogBlobs[deletePosition:]
 
-	contentlog.Log2(ctx, log,
-		"Keeping logs",
-		logparam.Int("count", deletePosition),
-		logparam.Int64("bytes", totalSize))
+	var toDeleteSize int64
+	for _, bm := range toDelete {
```
Lyndon-Li, line 93: This adds one more iterations.

There could be an optimization:
1. Collect the blobID list in the current iteration
2. Call `blob.DeleteMultiple` when `DryRun` is FALSE

I believe this would further beneficial to the performance of deletion.

Let me know if we want this in the current PR, or we can do it in a separate PR when necessary.

repo/maintenancestats/stats_clean_up_log.go
```diff
@@ -0,0 +1,41 @@
...
+	DeletedBlobSize   int64 `json:"deletedBlobSize"`
+	ActiveBlobCount   int   `json:"activeBlobCount"`
+	ActiveBlobSize    int64 `json:"activeBlobSize"`
```
julio-lopez:
```suggestion
	RetainedBlobCount   int   `json:"retainedBlobCount"`
	RetainedBlobSize    int64 `json:"retainedBlobSize"`
```

repo/maintenance/cleanup_logs.go
julio-lopez: This seems OK as it is.

repo/maintenancestats/stats_clean_up_log.go
Lyndon-Li: Done

julio-lopez:
- #4848

julio-lopez:
Thanks!

PR #4961, opened 2025-11-06
feat(general): add stats to maintenance run - RewriteContents

Lyndon-Li:

The current PR add the stats info for RewriteContents sub task.

### DIFF
--- a/repo/maintenance/content_rewrite.go
+++ b/repo/maintenance/content_rewrite.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kopia/kopia/repo"
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/content"
+	"github.com/kopia/kopia/repo/maintenancestats"
 )
 
 const parallelContentRewritesCPUMultiplier = 2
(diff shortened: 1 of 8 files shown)


repo/maintenancestats/stats_rewrite_contents.go
```diff
@@ -0,0 +1,42 @@
...
+	jw.IntField("rewrittenContentCount", rs.RewrittenContentCount)
+	jw.Int64Field("rewrittenContentSize", rs.RewrittenContentSize)
+	jw.IntField("retainedContentCount", rs.RetainedContentCount)
+	jw.Int64Field("retainedContentSize", rs.RetainedContentSize)
+	jw.EndObject()
+}
+
+// Summary generates a human readable summary for the stats.
+func (rs *RewriteContentsStats) Summary() string {
+	return fmt.Sprintf("Found %v(%v) contents to rewrite, and rewritten %v(%v). Retained %v(%v) contents from rewrite",
```
julio-lopez, line 35:
```suggestion
	return fmt.Sprintf("Found %v(%v) contents to rewrite, and rewrote %v(%v). Retained %v(%v) contents from rewrite",
```

repo/maintenance/content_rewrite.go
```diff
@@ -61,9 +62,14 @@ func RewriteContents(ctx context.Context, rep repo.DirectRepositoryWriter, opt *
...
-		mu          sync.Mutex
-		totalBytes  int64
-		failedCount int
+		mu             sync.Mutex
+		toRewriteBytes int64
+		toRewriteCount int
+		retainedBytes  int64
+		retainedCount  int
+		rewrittenBytes int64
+		rewrittenCount int
```
julio-lopez, line 71: nit: `stats.CountSum` could be used here, will address in a separate PR.

julio-lopez:
Thanks!

PR #4964, opened 2025-11-07
feat(general): add stats to maintenance run - SnapshotGC

Lyndon-Li:

The current PR add the stats info for SnapshotGC sub task.

### DIFF
--- a/repo/maintenancestats/stats_snapshot_gc.go
+++ b/repo/maintenancestats/stats_snapshot_gc.go
@@ -10,18 +10,18 @@ const snapshotGCStatsKind = "snapshotGCStats"
 
 // SnapshotGCStats delivers are the stats for snapshot GC.
 type SnapshotGCStats struct {
-	UnreferencedContentCount uint32 `json:"unreferencedContentCount"`
-	UnreferencedContentSize  int64  `json:"unreferencedContentSize"`
-	DeletedContentCount      uint32 `json:"deletedContentCount"`
-	DeletedContentSize       int64  `json:"deletedContentSize"`
-	RetainedContentCount     uint32 `json:"retainedContentCount"`
-	RetainedContentSize      int64  `json:"retainedContentSize"`
-	InUseContentCount        uint32 `json:"inUseContentCount"`
-	InUseContentSize         int64  `json:"inUseContentSize"`
-	InUseSystemContentCount  uint32 `json:"inUseSystemContentCount"`
-	InUseSystemContentSize   int64  `json:"inUseSystemContentSize"`
-	RecoveredContentCount    uint32 `json:"recoveredContentCount"`
-	RecoveredContentSize     int64  `json:"recoveredContentSize"`
+	UnreferencedContentCount       uint32 `json:"unreferencedContentCount"`
+	UnreferencedContentSize        int64  `json:"unreferencedContentSize"`
+	DeletedContentCount            uint32 `json:"deletedContentCount"`
+	DeletedContentSize             int64  `json:"deletedContentSize"`
+	UnreferencedRecentContentCount uint32 `json:"unreferencedRecentContentCount"`
+	UnreferencedRecentContentSize  int64  `json:"unreferencedRecentContentSize"`
+	InUseContentCount              uint32 `json:"inUseContentCount"`
+	InUseContentSize               int64  `json:"inUseContentSize"`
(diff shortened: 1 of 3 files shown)


repo/maintenancestats/stats_snapshot_gc.go
```diff
@@ -0,0 +1,55 @@
...
+	jw.UInt32Field("inUseSystemContentCount", ss.InUseSystemContentCount)
+	jw.Int64Field("inUseSystemContentSize", ss.InUseSystemContentSize)
+	jw.UInt32Field("recoveredContentCount", ss.RecoveredContentCount)
+	jw.Int64Field("recoveredContentSize", ss.RecoveredContentSize)
+	jw.EndObject()
+}
+
+// Summary generates a human readable summary for the stats.
+func (ss *SnapshotGCStats) Summary() string {
+	return fmt.Sprintf("Found %v(%v) unreferenced contents and marked %v(%v) for deletion. Found %v(%v) inused contents and %v(%v) inused system contents. Retained %v(%v) unused contents. Recovered %v(%v) contents",
```
julio-lopez: nit: this can be addressed later

```suggestion
	return fmt.Sprintf("Found %v(%v) unreferenced contents and marked %v(%v) for deletion. Found %v(%v) in-use contents and %v(%v) in-use system contents. Retained %v(%v) recent contents. Recovered %v(%v) contents",
```

repo/maintenancestats/stats_snapshot_gc.go
```diff
@@ -0,0 +1,55 @@
...
+const snapshotGCStatsKind = "snapshotGCStats"
+
+// SnapshotGCStats delivers are the stats for snapshot GC.
+type SnapshotGCStats struct {
+	UnreferencedContentCount uint32 `json:"unreferencedContentCount"`
+	UnreferencedContentSize  int64  `json:"unreferencedContentSize"`
+	DeletedContentCount      uint32 `json:"deletedContentCount"`
+	DeletedContentSize       int64  `json:"deletedContentSize"`
+	RetainedContentCount     uint32 `json:"retainedContentCount"`
+	RetainedContentSize      int64  `json:"retainedContentSize"`
```
julio-lopez: In order to preserve the original meaning, should this be `UnreferencedRecentContentCount` and `UnreferencedRecentContentSize`?

Or maybe add a comment to these fields explaining what `Retained` means, that is, what is the reason for retaining those contents. Thoughts?

snapshot/snapshotgc/gc.go
```diff
@@ -184,50 +184,52 @@ func findUnreferencedAndRepairRereferenced(
...
-		logparam.UInt32("count", systemCount),
-		logparam.Int64("bytes", systemBytes))
-	contentlog.Log2(ctx, log,
-		"GC undeleted contents",
-		logparam.UInt32("count", undeletedCount),
-		logparam.Int64("bytes", undeletedBytes))
+	result := buildGCResult(&unused, &inUse, &system, &tooRecent, &undeleted, &deleted)
+
+	userLog(ctx).Infof("Snapshot GC statistics: %v", result)
+	contentlog.Log1(ctx, log, "Snapshot GC", result)
```
julio-lopez, line 196: This is sending the `results` only to the `contentlog`.
Before, there was also an informational message, with human-readable units in the `userLog`. It may make sense to keep the message in `userLog`.

julio-lopez (approved):
LG overall.

See inline:
- question about the name for a couple of the fields https://github.com/kopia/kopia/pull/4964/files#r2508641996 ; and
- comment about the user-facing logs that were there before, with human-readable units.

repo/maintenancestats/stats_snapshot_gc.go
Lyndon-Li: Done

repo/maintenancestats/stats_snapshot_gc.go
Lyndon-Li: Yes, changed them to `UnreferencedRecentContentxxx`

snapshot/snapshotgc/gc.go
Lyndon-Li: Added the human-readable userlog back.

julio-lopez (approved):

julio-lopez:
- #4848

julio-lopez:
Thanks!

PR #4973, opened 2025-11-11
chore(ci): upgrade to golangci-lint 2.6.1

NathanBaulch:
Keeping up the momentum of #4931. A new minor version of golangci-lint was released after that PR was opened. One commit per linter as before.

### DIFF
--- a/snapshot/policy/actions_policy.go
+++ b/snapshot/policy/actions_policy.go
@@ -15,8 +15,8 @@ type ActionsPolicy struct {
 
 // ActionsPolicyDefinition specifies which policy definition provided the value of a particular field.
 type ActionsPolicyDefinition struct {
-	BeforeSnapshotRoot snapshot.SourceInfo `json:"beforeSnapshotRoot"`
-	AfterSnapshotRoot  snapshot.SourceInfo `json:"afterSnapshotRoot"`
+	BeforeSnapshotRoot snapshot.SourceInfo `json:"beforeSnapshotRoot,omitempty"`
+	AfterSnapshotRoot  snapshot.SourceInfo `json:"afterSnapshotRoot,omitempty"`
 }
 
 // ActionCommand configures a action command.
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -80,6 +80,9 @@ linters:
         - (*go.uber.org/zap.SugaredLogger).With
     misspell:
       locale: US
+    modernize:
+      disable:
+        - omitzero
     wsl_v5:
       allow-whole-block: true
       disable:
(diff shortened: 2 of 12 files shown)


internal/uitask/uitask_manager.go
```diff
@@ -106,13 +106,7 @@ func (m *Manager) WaitForTask(ctx context.Context, taskID string, maxWaitTime ti
...
 	deadline := clock.Now().Add(maxWaitTime)
 
-	sleepInterval := min(
-		//nolint:mnd
-		maxWaitTime/10, maxWaitInterval)
-
-	if sleepInterval < minWaitInterval {
-		sleepInterval = minWaitInterval
-	}
+	sleepInterval := max(min(maxWaitTime/10, maxWaitInterval), minWaitInterval) //nolint:mnd
```
julio-lopez, line 109: should this be split for clarity?

```suggestion
	sleepInterval := min(maxWaitTime/10, maxWaitInterval) //nolint:mnd
	sleepInterval = max(sleepInterval, minWaitInterval) 
```

snapshot/policy/actions_policy.go
```diff
@@ -15,8 +15,8 @@ type ActionsPolicy struct {
 
 // ActionsPolicyDefinition specifies which policy definition provided the value of a particular field.
 type ActionsPolicyDefinition struct {
-	BeforeSnapshotRoot snapshot.SourceInfo `json:"beforeSnapshotRoot,omitempty"`
-	AfterSnapshotRoot  snapshot.SourceInfo `json:"afterSnapshotRoot,omitempty"`
+	BeforeSnapshotRoot snapshot.SourceInfo `json:"beforeSnapshotRoot"`
```
julio-lopez: The `omitempty` changes (in the last commit) probably need to be reverted for now. Even though `omitempty` is ineffective on structs, there is a test that checks for the presence of these tags, it really checks for an equivalence between the "policy definition source" and the structs containing the actual values.

julio-lopez (changes requested):
🥇Thanks for doing this.

And thanks for splitting it into self contained, easily reviewable commits. It simplifies and speeds up the review process.

The `omitempty` changes (in the last commit) probably need to be reverted for now. Even though `omitempty` is ineffective on structs, there is a test that checks for the presence of these tags, it really checks for an equivalence between the "policy definition source" and the structs containing the actual values.

internal/uitask/uitask_manager.go
NathanBaulch: I'm accustomed to seeing `min`/`max` composed together to "clamp" a value, but happy to split if you prefer.

snapshot/policy/actions_policy.go
NathanBaulch: Done, including disabling that rule. You may want to add a comment in `.golangci.yml` explaining why.

PR #4976, opened 2025-11-12
refactor(general): preallocate slices when final size known

NathanBaulch:
Preallocate slices when the final size is definitely known. In other words, slices that are populated by a simple for loop that contains no control flow statements (`continue`, `break`, `return`).

This is the low hanging fruit on the way to enabling the `prealloc` linter.

### DIFF
--- a/internal/completeset/complete_set_test.go
+++ b/internal/completeset/complete_set_test.go
@@ -19,7 +19,7 @@ func TestFindFirstAndAll(t *testing.T) {
 		{
 			input:                 []blob.ID{},
 			wantFirst:             []blob.ID{},
-			wantAll:               [][]blob.ID(nil),
+			wantAll:               [][]blob.ID{},
 			wantExcludeIncomplete: []blob.ID{},
 		},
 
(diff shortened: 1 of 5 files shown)


jkowalski:
I'm not a big fan of such optimizations, since memory allocations only matter in tight loops and doing this everywhere creates a perception that performance is critical when it's not. From what I'm seeing, none of the changes are in performance-sensitive code.

Golang recommendation is to keep it simple - start with nil slices and append to them causing them to grow exponentially and only optimize when performance really matters. 

I would recommend not doing this and closing this PR, @julio-lopez thoughts?

julio-lopez:
> doing this everywhere creates a perception that performance is critical when it's not. From what I'm seeing, none of the changes are in performance-sensitive code.
> 
> Golang recommendation is to keep it simple - start with nil slices and append to them causing them to grow exponentially and only optimize when performance really matters.
> 
> I would recommend not doing this and closing this PR, @julio-lopez thoughts?

Makes sense. Most, if not all, of these changes are in codepaths where it makes no difference, for example in tests among others.

Aside: this change was too large to easily review, and the expected benefit of the change appears to be low.

PR #4990, opened 2025-11-14
feat(general): add sensitive data encryption to config file.

wastis:
## Problem Description

Sensitive connection details (e.g., password) were stored in plain text within Kopia's `repository.config` file, managed by the `LocalConfig` struct. This exposed sensitive parameters at rest, despite repository data being encrypted.

## Solution

To protect sensitive fields in `LocalConfig`, we encrypt and decrypt them before writing to and after reading from `repository.config`.

**Key Components:**

1.  **Helper Functions:** `EncryptSensitiveFields` and `DecryptSensitiveFields` (in `kopia/repo/blob/sensitive_field_encryption.go`) use Go reflection to process fields.
2.  **Tag-Based Filtering:** Fields marked with `kopia:"sensitive"` are targeted for encryption/decryption. This applies to `string` and `json.RawMessage` types.
3.  **Encryption Method:** Uses AES-GCM with a key derived from the repository password. `rclone` configurations are excluded.
4.  **Backward Compatibility:** Encrypted values are prefixed with `enc:` for identification and seamless handling of older, unencrypted configurations.
5.  **Integration:**
    *   `kopia/repo/connect.go`: Calls `EncryptSensitiveFields` before writing `LocalConfig`.
    *   `kopia/repo/open.go`: Calls `DecryptSensitiveFields` after loading `LocalConfig` and before `blob.NewStorage`.

## Files Involved:

*   **New:** `kopia/repo/blob/sensitive_field_encryption.go`
*   **Changed:** `kopia/repo/open.go`, `kopia/repo/connect.go`

### DIFF
--- a/repo/blob/sensitive_field_encryption.go
+++ b/repo/blob/sensitive_field_encryption.go
@@ -1,0 +1,259 @@
+package blob
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/kopia/kopia/internal/crypto"
+)
+
+var (
+	errConfigMustBePointerToStruct = errors.New("config must be a pointer to a struct")
+	errErrorGeneratingSalt         = errors.New("error generating salt")
+	errInvalidEncryptedData        = errors.New("invalid encrypted data")
+)
+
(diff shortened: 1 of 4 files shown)


wastis:
Apologies, it seems this breaks the kopia repository connect from-config --file=path, I am troubleshooting this.

PR #4993, opened 2025-11-16
fix(cli): potential overflow in integer conversion

julio-lopez:


cli/throttle_set.go
```diff
@@ -109,7 +109,7 @@ func (c *commonThrottleSet) setThrottleInt(ctx context.Context, desc string, val
 		return nil
 	}
 
-	v, err := strconv.ParseInt(str, 10, 64)
+	v, err := strconv.Atoi(str)
```
julio-lopez, line 112: Atoi already returns a platform-dependent `int`, it calls `strconv.ParseInt(str, 10, 0)`

PR #5003, opened 2025-11-18
fix(general): prevent writing outside of target dir

julio-lopez:
Enhances security by preventing path traversal vulnerabilities during installation of development tools.


julio-lopez:
@jkowalski FYI

PR #5026, opened 2025-11-23
refactor(cli): profile flags

julio-lopez:
Refactors flags for profiling and specifying the location of diagnostics output data.

### Enhancements and behavior changes
- New `--diagnostics-output-directory` flag. Both metrics and profiling
  data are stored in this directory when the kopia process exits.
- New `--metrics-store-on-exit` controls whether the metrics data is
  saved on exit.
- Multiple profile types can be enabled at once, before only
  a single type profiling could be done during a process execution.
- The new `--profiles-store-on-exit` enables memory, mutex, blocking and goroutine
  profiles, but not CPU profiling, which needs to be explicitly enabled.
  Profiling output is now stored in a `profiles` subdirectory in the directory
  specified with the `--diagnostics-output-directory` flag.
- Profile flags now have effect for all `kopia` commands, including
  `server start`. Before these flags did not have any effect
  in a few commands.
- Profiling parameters can now be set via new flags. This allows setting
  the profile parameters for the pprof endpoint, as well as when saving
  profiles to files on exit.

### The following flags have been removed:
- `--profile-dir`: superseded by the `--diagnostics-output-directory` flag
- `--profile-blocking`: the `--profile-store-on-exit` flag enables blocking
  profiling. Use `--profile-blocking-rate=0` to explicitly disable it.
- `--profile-memory`: the `--profile-store-on-exit` flag enables memory
  profiling. Use `--profile-memory-rate=0` to explicitly disable it.
- `--profile-mutex`: the `--profile-store-on-exit` flag enables mutex
  profiling. Use `--profile-mutex-fraction=0` to explicitly disable it.

Removes dependency on an unmaintained package previously used for
saving the pprof data.

### DIFF
--- a/cli/profile.go
+++ b/cli/profile.go
@@ -11,6 +11,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const profDirName = "profiles"
+
 type profileFlags struct {
 	profileGCBeforeSaving bool
 	profileCPU            bool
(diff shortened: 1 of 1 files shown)


julio-lopez:
FYI @Kgrah

Kgrah (approved):

Issue #5006, opened 2025-11-18
Starting kopiaui only prints the version


tobiasBora:
My sister (macos) has this weird interface when we open a connected repository:

<img width="1212" height="700" alt="Image" src="https://github.com/user-attachments/assets/ade75023-4e64-48d8-adb6-e95c5b7ba9c9" />

Any idea what's going on here? We can't do anything now… Kopia 0.22.0

basldfalksjdf:
These two issues may be related to yours:

https://kopia.discourse.group/t/gui-blank-with-0-22-update/8703

https://github.com/kopia/kopia/issues/5007

tobiasBora:
Thanks, the first one is indeed the same issue, the second one just misused the gui.

kanbock:
A similar problem just started happening for me with version **0.22** on **Fedora 43**. I’m using the [official RPM](https://kopia.io/docs/installation/#linux-installation-using-rpm-redhat-centos-fedora) package. No tray icon appears, and the `kopia-ui` process seems to hang. The server doesn’t start, and the process doesn’t respond to **SIGTERM** (though **SIGKILL** works). The [Flatpak version](https://github.com/flathub/io.kopia.KopiaUI) shows the tray icon, but the server still doesn’t start and the application exits after about a minute.

However, starting the server manually from the command line works fine:

```
kopia server start \
  --config-file=/home/user/.config/kopia/repository.config \
  --address=localhost:51515 \
  --ui \
  --insecure \
  --server-username="user" \
  --server-password="user"
```

readr00m:
Can this issue be made high priority? As it currently is KopiaUI is unusable with existing repositories.

basldfalksjdf:
> Can this issue be made high priority? As it currently is KopiaUI is unusable with existing repositories.

pinging the two main devs on this to see if they can help @jkowalski @julio-lopez 

if anyone else can drop a PR, that would be great as well

xxxliqu1dxxx:
Same issue here... 0.22 is unusable in its state.

cgspeck:
Same issue here running on Arch with version 0.22.0.

The server will start, you can browse to it and log in, but it either:
1. displays only the version, or
2. displays the first page of snapshots (if any) and if you attempt to change anything then only shows the version

It is unusable now.

richieg1000:
Same issue with 0.22 in zimaos in multiple docker containers.

milindpatel63:
Similar issue here on fedora with latest version.
[...]

PR #5027, opened 2025-11-23
fix(server): nil DirSummary handling in snapshot manifests

LukaHummel:
Handle cases where DirSummary is nil to prevent frontend errors. This change ensures that a default empty summary is provided, allowing the frontend to function correctly without encountering undefined errors. A test case verifies that snapshots with nil DirSummary are processed without issues.
Fix for https://github.com/kopia/kopia/issues/5006

### DIFF
--- a/internal/server/api_snapshots.go
+++ b/internal/server/api_snapshots.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/kopia/kopia/internal/serverapi"
+	"github.com/kopia/kopia/fs"
 	"github.com/kopia/kopia/repo"
 	"github.com/kopia/kopia/repo/manifest"
 	"github.com/kopia/kopia/snapshot"
--- a/internal/server/api_snapshots_test.go
+++ b/internal/server/api_snapshots_test.go
@@ -265,8 +265,6 @@ func TestSnapshotWithNilDirSummary(t *testing.T) {
 
 	si := env.LocalPathSourceInfo("/test/path")
 
-	var snapshotID manifest.ID
-
 	// Create a snapshot manifest with nil DirSummary
 	require.NoError(t, repo.WriteSession(ctx, env.Repository, repo.WriteSessionOptions{Purpose: "Test"}, func(ctx context.Context, w repo.RepositoryWriter) error {
 		// Create a minimal snapshot without proper DirSummary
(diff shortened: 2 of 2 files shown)


internal/server/api_snapshots.go
```diff
@@ -260,8 +261,12 @@ func convertSnapshotManifest(m *snapshot.Manifest) *serverapi.Snapshot {
 		Pins:             append([]string{}, m.Pins...),
 	}
 
-	if re := m.RootEntry; re != nil {
+	if re := m.RootEntry; re != nil && re.DirSummary != nil {
```
jkowalski, line 264: I think this could be because of omitzero linter changes, looks safe to merge.

jkowalski (approved):

jkowalski:
@LukaHummel can you fix the linter issues?

internal/server/api_snapshots.go
julio-lopez: @jkowalski shouldn't this be handled in the UI instead?

Yes, the issue may be related to the omitzero linter changes. AFAICT, this particular struct / field was not changed.

@LukaHummel could you share your findings / log messages with the error you are seeing related to this change?

Thanks.

internal/server/api_snapshots.go
```diff
@@ -260,8 +261,12 @@ func convertSnapshotManifest(m *snapshot.Manifest) *serverapi.Snapshot {
...
 		e.Summary = re.DirSummary
+	} else {
+		// Provide a default empty summary when RootEntry or DirSummary is nil
+		// This prevents undefined errors in the frontend when accessing totalSize
```
julio-lopez, line 268: This should be fixed in the frontend instead.

internal/server/api_snapshots.go
```diff
@@ -260,8 +261,12 @@ func convertSnapshotManifest(m *snapshot.Manifest) *serverapi.Snapshot {
...
+		e.Summary = &fs.DirectorySummary{}
```
julio-lopez, line 269: The error reported in https://github.com/kopia/kopia/issues/5006#issuecomment-3567627991 suggest that the issue is elsewhere.

internal/server/api_snapshots.go
jkowalski: handling in UI makes sense, but it's technically a breaking change - the frontend was relying on this field being set.

Can we revert all `omitzero` annotations to be safe?

julio-lopez: > the frontend was relying on this field being set.

That's not the case AFAICT. The server has been explicitly serializing this field as `nil` in prior releases.

julio-lopez:
The UI issues appear to be solved, closing this for now.

PR #5037, opened 2025-11-25
chore(ci): remove ineffective `omitempty` tags

julio-lopez:


repo/object/indirect.go
```diff
@@ -4,7 +4,7 @@ package object
 type IndirectObjectEntry struct {
 	Start  int64 `json:"s,omitempty"`
 	Length int64 `json:"l,omitempty"`
-	Object ID    `json:"o,omitempty"`
+	Object ID    `json:"o"`
```
julio-lopez, line 7: - `object.ID` as currently defined is a `struct`,
- `object.ID` is expected to never have an "empty" or "zero" value.

PR #5038, opened 2025-11-25
fix(providers): allow S3 versioned tests cleanup to succeed

julio-lopez:


julio-lopez:
Successful run https://github.com/kopia/kopia/actions/runs/19658873417/job/56301185428?pr=5038

Issue #5045, opened 2025-11-28
Inaccessible mount point causes parent directory to be skipped


geoffreybennett:
When backing up /home as root, I have found that:
- if a user has an sshfs mount within their home directory (which is by default inaccessible to root), then kopia skips the entirety of the containing directory (i.e. the user's home directory) and I get the message "Ignored 1 error(s) while snapshotting root@hostname:/home"
- if there is an sshfs mount within /home (e.g. at /home/anyuser), then nothing gets backed up (no snapshot) and there's a misleading error message: "upload error: permission denied"

It was not possible to fix this error and get a complete backup by setting "Scan one filesystem only", "Ignore file/directory read errors", or by putting the problematic directories into .kopiaignore, because the attempted stat on the sshfs directory causes its parent directory scan to be aborted before the filesystem/error/ignore checks are done.

I will submit a PR which fixes these issues shortly.

PR #5046, opened 2025-11-28
fix(localfs): return ErrorEntry for permission denied instead of aborting the directory scan

geoffreybennett:
When lstat fails with permission denied during directory iteration, return an ErrorEntry instead of aborting iteration. This allows the error to be handled according to policy and iteration to continue.

Fixes: #5045

### DIFF
--- a/fs/localfs/local_fs_os.go
+++ b/fs/localfs/local_fs_os.go
@@ -102,6 +102,21 @@ func toDirEntryOrNil(dirEntry os.DirEntry, prefix string, options *Options) (fs.
 			return nil, nil
 		}
 
+		// For permission denied errors, return an ErrorEntry instead of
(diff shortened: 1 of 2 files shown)


jkowalski:
incidentally https://github.com/kopia/kopia/pull/5044 was merged very recently, we can combine your approach with mine - I think returning ErrorEntry is a better fix than just skipping unreadable directories. A separate question is whether this should be driven by options.

@julio-lopez do you think we should 

a) merge #5046 as-is
b) merge #5046 but make this option-controlled (options introduced in #5044)
c) revert #5044 and only merge #5046

julio-lopez:
@jkowalski => d) revert #5044 and iterate on #5046 to:

  - differentiate directory vs. non-dir unreadable entries; and
  - fix the tests

- Superseded by #5217

PR #5056, opened 2025-12-02
refactor(general): Switching keyring implementation

qpanda:
Switching keyring implementation from [github.com/zalando/go-keyring](https://github.com/zalando/go-keyring) to [github.com/byteness/keyring](https://github.com/byteness/keyring) (formerly [github.com/99designs/keyring](https://github.com/99designs/keyring)) to better protect the Kopia repository password stored in Keychain on macOS. With this change, the Kopia repository password in Keychain on macOS is accessible only by the `kopia` binary instead of the `security` command (which can be used to extract the password).

The issue with [github.com/zalando/go-keyring](https://github.com/zalando/go-keyring) is that it [uses the `security` command on macOS](https://github.com/zalando/go-keyring?tab=readme-ov-file#os-x) to store / retrieve secrets in / from Keychain. This makes it trivial for any process running as the same user to extract the password from Keychain. The [github.com/byteness/keyring](https://github.com/byteness/keyring) library uses the same [github.com/danieljoos/wincred](https://github.com/danieljoos/wincred) and [github.com/godbus/dbus](https://github.com/godbus/dbus) libraries on Windows and Linux respectively, but relies on [github.com/byteness/go-keychain](https://github.com/byteness/go-keychain) to store / retrieve secrets in / from Keychain on macOS. [github.com/byteness/go-keychain](https://github.com/byteness/go-keychain) uses the native macOS API which makes it possible to [configure Access Control](https://support.apple.com/en-sg/guide/mac-help/kychn002/mac) to allow only `kopia` to store / retrieve the password.

[github.com/byteness/keyring](https://github.com/byteness/keyring) also supports other backends (such as Pass and 1Password), but they have been disabled in this change to keep the feature set as similar as possible.


julio-lopez:
Won't be merged at this time.

qpanda:
> Won't be merged at this time.

Could you please provide more details / reason as to why this change is not being considered?

I find it surprising that Kopia has implemented sophisticated [Ransomware Protection](https://kopia.io/docs/advanced/ransomware-protection/) protection for the repository content itself, but does little to protect ransomware / malware from extracting Kopia password and repository credentials on the local device. This is definitely an area where Kopia could provide better protection and this pull request was an attempt to move in this direction.

What suggestions do you have to protect the Kopia password on the local device from malware / ransomware running as the same user on macOS?

qpanda:
Another option to securely pass the Kopia repository password to the `kopia` binary would to allow the password to be read from `stdin`. This would make it possible to write a wrapper that securely extracts the password from Keychain on macOS and passes it via `stdin` to `kopia`.

There are two ways in which this could be implemented. The `--password` parameter could support `@` syntax to specify a filename to read the password from and if the filename is set to `-` it would read from `stdin`. This is how `curl` has implemented it, but could potentially break backwards compatibility for users who have a password starting with `@`.

`--password @<filename>` would read password from a file or unix pipe
`--password @-` would read password from stdin

Another option could be to introduce a second `--password-stdin` (or similar) command line parameter that can be used instead of `--password`.

Would you consider a pull request that introduces these changes?

See https://smallstep.com/blog/command-line-secrets/ for more details on securely passing credentials / secrets on the command line.

qpanda:
Would appreciate if you could respond to above comments @julio-lopez.

jkowalski:
Let's not discard this idea just yet. Sure it complicates the build on Mac but there are lots of benefits and we should try to get this in. Can we do conditional compilation that works with either library based on build flags?

qpanda:
> Let's not discard this idea just yet. Sure it complicates the build on Mac but there are lots of benefits and we should try to get this in. Can we do conditional compilation that works with either library based on build flags?

Thank you for your response!

[github.com/ByteNess/keyring](https://github.com/byteness/keyring) is already using build tags (see https://github.com/ByteNess/keyring/blob/main/keychain.go#L1) if that's what you mean.

Could you please elaborate how exactly conditional compilation should be implemented? Do you mean a separate [passwordpersist](https://github.com/kopia/kopia/tree/master/internal/passwordpersist) implementation that is compiled and added to [multiple](https://github.com/kopia/kopia/blob/master/cli/app.go#L240) only on macOS? 

Is the main concern that [github.com/byteness/keyring](https://github.com/byteness/keyring) relies on `cgo`? Before submitting this pull request, I found [github.com/lstoll/keychain](https://github.com/lstoll/keychain) - which is based on [github.com/ebitengine/purego](https://github.com/ebitengine/purego) - but it is in early development and I am not sure if it is functional yet nor whether it is maintained.

Issue #3022, opened 2023-05-06
Error Opening Repository after Updating to v0.13.0


PreciselyPixels:
After updating to v0.13.0, Kopia throws the following error while trying to connect to the repository. 

`Connect Error: INTERNAL: internal server error: connect error: error opening repository: error connecting: unable to create shared content manager: error loading indexes: error downloading indexes: error loading index blob xn15_f._022f18e6d28f59eb4bf1789720d1b75-s41669725541c852e11c-c1: decrypt blob: unable to get index blob IV: invalid blob ID: xn15_f._022f18e6d28f59eb4bf1789720d1b75`

Before rebooting my machine, I had issues with backing up. Unfortunately, I did not get a snippet of this other error.

patrsc:
I have the same issue with an external hard drive (formatted with ExFAT) being used as repository location on macOS. In my case the problem is that macOS creates files starting with `._` on external drives to store filesystem metadata (see also [Why are dot underscore ._ files created, and how can I avoid them?](https://apple.stackexchange.com/questions/14980/why-are-dot-underscore-files-created-and-how-can-i-avoid-them)). After running the command
```
dot_clean /path/to/kopia-repository
```
I was able to connect to the repository again. However, this solution is not permanent, because macOS will create these files on subsequent write operations to the repository. Is it possible that Kopia ignores these files by default?

mrtorrent:
I just ran into this issue on KopiaUI 0.15.0 I also have an external drive formatted ExFAT. I was pulling my hair out trying to figure out how to get Kopia to connect to my repository after it was created. Thanks to ChrisA on the forum and @patrsc, now I know what the problem is and how to work around it, but as @patrsc points out, it will keep happening when macOS recreates the files. Can't Kopia recognize these files and handle them gracefully?

bverkron:
Just ran into this as a first time user. This is a pretty glaring issue for any macOS users. It basically breaks users ability to save a repository on a network share from macOS, which is crazy considering that's the first option presented in the GUI and likely a common use-case for backing up clients.

How is it that Kopia breaks so badly just due to the presence of . files? So far the only "fix" I see is to run independent scripts to delete all the . files within the repository. Which is a pretty terrible experience, especially for inexperienced users.

jhou322:
I too have just run into this issue as a macOS user with KopiaUI v0.19.0. I'm assuming there still hasn't been a permanent fix?

fivestones:
[...]

Issue #4418, opened 2025-02-18
Repository immediately becomes brokes upon creation on an external USB drive (macOS)


wetfloo:
Whenever I try to create a repository on an USB storage, I get hit with the following:

```
[I] > kopia repository create filesystem --path /Volumes/USB/mc_backups
Enter password to create new repository:
Re-enter password for verification:
Initializing repository with:
  block hash:          BLAKE2B-256-128
  encryption:          AES256-GCM-HMAC-SHA256
  key derivation:      scrypt-65536-8-1
  splitter:            DYNAMIC-4M-BUZHASH
Connected to repository.

NOTICE: Kopia will check for updates on GitHub every 7 days, starting 24 hours after first use.
To disable this behavior, set environment variable KOPIA_CHECK_FOR_UPDATES=false
Alternatively you can remove the file "/Users/alice/Library/Application Support/kopia/repository.config.update-info.json".

Retention:
  Annual snapshots:                 3   (defined for this target)
  Monthly snapshots:               24   (defined for this target)
  Weekly snapshots:                 4   (defined for this target)
  Daily snapshots:                  7   (defined for this target)
  Hourly snapshots:                48   (defined for this target)
  Latest snapshots:                10   (defined for this target)
  Ignore identical snapshots:   false   (defined for this target)
Compression disabled.

To find more information about default policy run 'kopia policy get'.
To change the policy use 'kopia policy set' command.

NOTE: Kopia will perform quick maintenance of the repository automatically every 1h0m0s
and full maintenance every 24h0m0s when running as alice@guardian.

See https://kopia.io/docs/advanced/maintenance/ for more information.

^[kNOTE: To validate that your provider is compatible with Kopia, please run:

$ kopia repository validate-provider


~/Documents/mc_backups
[N] > kopia repository connect filesystem --path /Volumes/USB/mc_backups
Enter password to open repository:

ERROR failed to open repository: unable to create shared content manager: error loading indexes: error downloading indexes: error loading index blob xn0_._f92dfcead34a564cb36daecfc8344e65-s620f2067af9d95d7132-c1: decrypt blob: error decrypting BLOB xn0_._f92dfcead34a564cb36daecfc8344e65-s620f2067af9d95d7132-c1: unable to decrypt content: cipher: message authentication failed
deleted repository password for /Users/alice/Library/Application Support/kopia/repository.config.
[...]

PR #5061, opened 2025-12-03
fix: Add macOS AppleDouble file skipping to ListBlobs

ReadyPlayerEmma:
In progress fix for #4418

Still needs testing.


OblackatO:
I can help with testing, do you have any specific use-cases in mind ?

sky-dragn:
This would also most likely fix #3022 .
My primary comment (having no experience with this codebase and not really familiar with the repo format, to be clear) is that it would probably be helpful have a regex of file patterns to ignore, in order to not trip up on other common trash files such as `.DS_Store` and `desktop.ini` and such. As it is now, it seems like these files would also be collected by ListBlobs and potentially cause similar issues...

PR #5062, opened 2025-12-04
feat: support migrating tags in `kopia snapshot migrate`

leonardoce:
This PR adds a new command line option to `kopia snapshot migrate` named `--tags`. This new flag allows the user to specify a list of tags to be migrated at the same some as the snapshots.

If no tag is specified no tag will be migrated, preserving the current behaviour.


jkowalski:
this is looking good, but needs tests, should not be difficult to add in the existing framework we have

leonardoce:
> this is looking good, but needs tests, should not be difficult to add in the existing framework we have

Yes, I added them now @jkowalski.
Thanks!

Shrekster (approved):
Discussed with @Kgrah . Concept makes sense.

Issue #1738, opened 2022-02-08
FR - Native way for kopia to recover from underlying storage running out of space


xxxliqu1dxxx:
FR - Native way for kopia to recover from underlying storage running out of space (Spoke with Jarek on slack and he recommended I log this feature request)

Problem
As of 0.10.4, when storage runs out of space, errors like this are thrown: DEBUG got error cannot create temporary file: cannot create directory: unexpected error creating directory: mkdir f:\newkopia\xs4\_21: There is not enough space on the disk. when PutBlobInPath:f:\newkopia/xs4/_21/69ab5f009aefb557979503cedf8f16-s27e2e49f1c49d18a-c1.f (#1), sleeping for 150ms before retrying

Impact
When any command is triggered, either snapshot delete, maintenance, or anything really, it seems like Kopia expects some storage to be available on the repo to do temp files. Problem is, when there's no more storage, one would depend on snapshot delete --delete to actually "free" some space but it seems that there's no way for Kopia to natively recover from that situation and is unable to mark snapshots to be deleted and allow maintenance to runs to "clean things up", which would actually free some space.

Repro steps
Use external sd, i.e. 4TB external harddrive for this test
using KopiaUI or CLI, create repo on destination
create snapshots until out of space
reach out of space state

Expected result
I would expect to be able to delete a snapshot then run maintenance to "clean things up"

Actual result
Every command run from kopia CLI or UI is returning very similar errors as below, which was from maintenance run

ERROR error updating maintenance schedule: unable to complete PutBlobInPath:f:\newkopia/kopia.maintenance.f despite 10 retries, last error: cannot create temporary file: open f:\newkopia/kopia.maintenance.f.tmp.6b5e03a0660772e8: There is not enough space on the disk.

Conclusion
Therefore, there does not seem to be a native way to gracefully recover from this state. The below is not a workaround by any means, but the "only way" I found to actually get things going again was to arbitrarily delete an actual file from the hdd, i.e. df000a557ce9d9967853879c814-s7b3fa8a5996b047e109.f, understanding that this "content" gets deleted "forever", allowing subsequent snapshots delete as well as subsequent maintenance to run and clean things up.

adowair:
This is a very valid problem that is also affecting us. A prerequisite for this would be to figure out how much "free space" kopia really has available (#1520).

[...]

PR #5077, opened 2025-12-18
feat(repository): Implement native recovery for full storage

ke-ma-fi:
# Kopia Storage Recovery Project

## Overview
This project aims to implement a native recovery mechanism for Kopia when the underlying storage repository runs out of space. 

## The Problem
Currently (as of v0.22.3), when Kopia's storage repository becomes full, the application enters an unrecoverable state where:
- All operations fail due to inability to create temporary files
- `snapshot delete` commands cannot execute
- `maintenance` operations cannot run to clean up and free space
- Users are forced to manually delete repository files, risking data corruption

## Current Behavior
When storage is exhausted, Kopia throws errors like:
```
ERROR error updating maintenance schedule: unable to complete PutBlobInPath despite 10 retries, 
last error: cannot create temporary file: There is not enough space on the disk. 
```

This affects all operations including:
- Snapshot deletion
- Maintenance runs
- Any command requiring temporary file creation

## Expected Behavior
Kopia should be able to gracefully handle out-of-space conditions by:
- Allowing snapshot deletion operations to proceed without requiring temp space
- Enabling maintenance operations to clean up and reclaim storage
- Providing a native recovery path that doesn't require manual file deletion

## Goal
Implement a recovery mechanism that allows Kopia to: 
1. Delete snapshots even when storage is full
2. Execute maintenance operations to free up space
3. Recover gracefully without manual intervention or data loss

## Reference
- Original Issue: [kopia/kopia#1738](https://github.com/kopia/kopia/issues/1738)
- Discussed with Jarek on Slack prior to issue creation

### DIFF
--- a/cli/command_snapshot_delete.go
+++ b/cli/command_snapshot_delete.go
@@ -3,6 +3,8 @@ package cli
 import (
 	"context"
 	"fmt"
+	"strings"
+	"syscall"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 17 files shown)


adhaussy:
For those like me who are running backups continuously during day/night, as Kopia server is not aware how backup are scheduled, isn’t there a risk of backup(s) firing during the emergency phase and/or filling up the space again ?

I read on the website that we should ensure that no backups run during a maintenance with safety off, 
if maintenance is automatically triggered, how do we ensure that ?

Maybe the emergency mode could also suspend uploads when setting the “LowSpace” flag ?
I may be missing something, i’m just trying to think worst case scenario.

ke-ma-fi:
@adhaussy My implementation is blocking any snapshots or storage filling tasks, when the system cannot ensure the storage reserve (requirement: reserve is written to disk and you have at least 10MB left). So when entering emergency mode, the reserve gets deleted and therefore snapshots are not possible until there is enough space again to write the reserve back to disk and recover from emergency mode. That's also why it's safe to use safety=none as default for recovery since parallel snapshots are not possible when the disk is full. Is that what you asked? 

For those who know how much space they need left for a snapshot, we could also add an environment variable that determines the threshold to which a reserve is considered "ensured" eg. changing it from 10MB to 10GB. In that way we could prevent the likelihood that a snapshot cannot finish and breaks while running because of insufficient storage. Example: 11GB left on disk, snapshot will start, ensures the reserve and finishes successfully with 2 GB left on disk. The next snapshot will be blocked and triggers emergency mode until gc recovers the repo to at least the 10GB + 500MB reserve. The Reserve Blob is always being deleted to make sure that gc or snapshot delete has enough space for temps regardless of the emergency mode getting triggered with 2GB or 0MB left on disk.

adhaussy:
@ke-ma-fi Indeed i can see now that my eyes are open that storagereserve.Ensure is called on snapshot creation. 
Snapshots will fail unless theres 500mb + 10% of total storage. 
Tunable threshold sounds good, thanks for clarifying.

cli/command_snapshot_delete.go
```diff
@@ -117,3 +130,52 @@ func (c *commandSnapshotDelete) deleteSnapshotsByRootObjectID(ctx context.Contex
...
+
+		return nil
+	}
+
+	return errors.Wrap(err, "error ensuring storage reserve")
+}
+
+// isNoSpaceError is a helper to detect out-of-space conditions consistently.
+// This matches the implementation in repo/maintenance/maintenance_run.go.
+func isNoSpaceError(err error) bool {
```
kerryland: Shouldn't this code be shared rather than duplicated?

PS: I'm just a nosy busy-body. Feel free to ignore me :-)

PR #5168, opened 2026-02-19
fix(providers): perform maximum of 2 attempt in `PutBlobInPath`

julio-lopez:
Adjusts tests accordingly.
Also, fixes error injection for `MkdirAll` in tests.

### DIFF
--- a/repo/blob/filesystem/filesystem_storage_test.go
+++ b/repo/blob/filesystem/filesystem_storage_test.go
@@ -241,21 +241,45 @@ func TestFileStorage_PutBlob_RetriesOnErrors(t *testing.T) {
 
 	ctx := testlogging.Context(t)
 
-	errorCases := []func(*mockOS){
-		func(osi *mockOS) { osi.createNewFileRemainingErrors.Store(1) },
-		func(osi *mockOS) { osi.mkdirRemainingErrors.Store(1) },
-		func(osi *mockOS) { osi.writeFileRemainingErrors.Store(1) },
-		func(osi *mockOS) { osi.writeFileCloseRemainingErrors.Store(1) },
-		func(osi *mockOS) { osi.renameRemainingErrors.Store(1) },
-		func(osi *mockOS) { osi.chownRemainingErrors.Store(2) }, // these are ignored
-		func(osi *mockOS) { osi.chtimesRemainingErrors.Store(1) },
+	cases := []struct {
+		desc        string
+		injectError func(*mockOS)
+	}{
+		{
+			desc:        "CreateNewFile",
+			injectError: func(osi *mockOS) { osi.createNewFileRemainingErrors.Store(1) },
+		},
+		{
+			desc:        "Mkdir",
+			injectError: func(osi *mockOS) { osi.mkdirRemainingErrors.Store(1) },
+		},
+		{
+			desc:        "Write",
+			injectError: func(osi *mockOS) { osi.writeFileRemainingErrors.Store(1) },
+		},
+		{
+			desc:        "Close",
+			injectError: func(osi *mockOS) { osi.writeFileCloseRemainingErrors.Store(1) },
+		},
+		{
+			desc:        "Rename",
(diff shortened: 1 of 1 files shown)


repo/blob/filesystem/osinterface_mock_test.go
```diff
@@ -141,13 +142,21 @@ func (osi *mockOS) CreateNewFile(fname string, perm os.FileMode) (osWriteFile, e
...
 func (osi *mockOS) Mkdir(fname string, mode os.FileMode) error {
-	if osi.mkdirAllRemainingErrors.Add(-1) >= 0 {
+	if osi.mkdirRemainingErrors.Add(-1) >= 0 {
 		return &os.PathError{Op: "mkdir", Err: errors.New("injected mkdir error")}
 	}
 
 	return osi.osInterface.Mkdir(fname, mode)
 }
 
+func (osi *mockOS) MkdirAll(fname string, mode os.FileMode) error {
```
julio-lopez, line 152: Injected errors (`mkdirAllRemainingErrors`) applied to `Mkdir` and not `MkdirAll`.

This change injects errors in `MkdirAll` when `mkdirAllRemainingErrors > 0`; and
adds a ``mkdirRemainingErrors` field to control the error injection behavior for `Mkdir`.

Tests are updated accordingly.

repo/blob/filesystem/filesystem_storage_test.go
```diff
@@ -249,52 +249,60 @@ func TestFileStorage_PutBlob_RetriesOnErrors(t *testing.T) {
...
-		FileUID: &fileUID,
-		FileGID: &fileGID,
-		Options: sharded.Options{
-			DirectoryShards: []int{5, 2},
-		},
-		osInterfaceOverride: osi,
-	}, true)
-	require.NoError(t, err)
+	for _, ec := range errorCases {
+		t.Run("", func(t *testing.T) {
```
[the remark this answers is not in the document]
julio-lopez: Go's testing framework adds a `#<sequence>` identifier by default.
Yes, adding a more descriptive name would be useful. To be added if other changes are needed in this PR.

repo/blob/filesystem/filesystem_storage_test.go
```diff
@@ -249,52 +249,60 @@ func TestFileStorage_PutBlob_RetriesOnErrors(t *testing.T) {
...
-	osi.chtimesRemainingErrors.Store(3)
+	errorCases := []func(*mockOS){
+		func(osi *mockOS) { osi.createNewFileRemainingErrors.Store(1) },
+		func(osi *mockOS) { osi.mkdirRemainingErrors.Store(1) },
+		func(osi *mockOS) { osi.writeFileRemainingErrors.Store(1) },
+		func(osi *mockOS) { osi.writeFileCloseRemainingErrors.Store(1) },
+		func(osi *mockOS) { osi.renameRemainingErrors.Store(1) },
+		func(osi *mockOS) { osi.chownRemainingErrors.Store(2) }, // these are ignored
+		func(osi *mockOS) { osi.chtimesRemainingErrors.Store(1) },
+	}
```
redgoat650: suggestion/question (non-blocking): we seem to cover the "happy" retry path, but do we also cover the "unhappy" path? I.e. do we want to assert that if we exceed `maxAttempts`, we actually do throw an error? If someone tried changing `const maxAttempts` to `100`, do we want a test to fail? If someone tries changing one of the error returns to a `log` (like `chown` is handled here), we'd want a test to fail, right?

Just an observation in passing; your PR seems to be covering equivalent behavior to before, so it's not a blocker.

redgoat650 (approved):
Looks good, just one question

repo/blob/filesystem/filesystem_storage_test.go
julio-lopez: Good point. Let me add a test to ensure the limit is not exceeded. Thanks for the review!

julio-lopez: @redgoat650 done!

PR #5216, opened 2026-03-11
fix(providers): clean up empty shard directories after blob deletion (#2600)

### LINKED ISSUE #2600: Empty directories not deleted from repository
After paths/snapshots are deleted, KopiaUI seems to keep around empty directories in the repository (directories like `q3e/7ce`). It would be great if those could get auto deleted, especially for repositories that are limited by item count rather than storage space (Google Shared Drives)

perfectra1n:
When blobs get deleted from sharded repositories, the parent shard directories (e.g., `q3e/7ce/`) are left behind, completely empty. Some users in GitHub Issues have reported 70,000+ empty directories piling up over time, which tanks performance and runs into Google Shared Drive item count limits.

The root cause is pretty straightforward: `DeleteBlobInPath` only removes the blob file itself. The `dirPath` parameter is passed in but explicitly ignored (`_ = dirPath`).

I went with Go's optional interface pattern, backends opt into cleanup by implementing a `DirRemover` interface. The sweep logic lives centrally in `sharded.Storage.DeleteBlob`, so individual backends only need to add a single method. I'd love feedback if this isn't the pattern you'd expect here.

Here's the output of the test in our branch before the changes:
image

Then that same test after our changes:
image

Here's how the pieces fit together:

### Normalize `RootPath` in `sharded.New()`

`repo/blob/sharded/sharded.go` (line 281)

Changed `RootPath: rootPath` to `RootPath: path.Clean(rootPath)`. Without this, a trailing-slash path like `/data/repo/` breaks the `dir != s.RootPath` guard in the sweep, `path.Dir` returns cleaned paths, so the comparison target needs to be clean too. Otherwise we'd try to remove directories _above_ the repo root, which would be bad.

### The `DirRemover` interface and sweep logic

`repo/blob/sharded/sharded.go`

```go
// DirRemover is optionally implemented by Impl to enable
// cleanup of empty shard directories after blob deletion.
type DirRemover interface {
    RemoveDirInPath(ctx context.Context, dirPath string) error
}

```

After `DeleteBlobInPath` succeeds, `DeleteBlob` checks if the backend implements `DirRemover`. If it does, we walk up the directory tree removing empty dirs until we hit `RootPath` or a removal fails (meaning the directory isn't empty):

```go
func (s *Storage) DeleteBlob(ctx context.Context, blobID blob.ID) error {
    dirPath, filePath, err := s.GetShardedPathAndFilePath(ctx, blobID)
    if err != nil {
        return errors.Wrap(err, "error determining sharded path")
    }
    if err := s.Impl.DeleteBlobInPath(ctx, dirPath, filePath); err != nil {
        return err //nolint:wrapcheck
    }
    if dr, ok := s.Impl.(DirRemover); ok {
        s.sweepEmptyDirs(ctx, dr, dirPath)
    }
    return nil
}

func (s *Storage) sweepEmptyDirs(ctx context.Context, dr DirRemover, dir string) {
    for dir != s.RootPath && dir != "" && dir != "." && dir != "/" {
        if err := dr.RemoveDirInPath(ctx, dir); err != nil {
            return
        }
        log(ctx).Debugf("removed empty shard directory %v", dir)
        dir = path.Dir(dir)
    }
}

```

### Making `ListBlobs` tolerate disappearing directories

`repo/blob/sharded/sharded.go` (in `walkDir`, ~line 89)

There's a subtle race here: the concurrent sweep can remove a directory between `ListBlobs` discovering it (via parent's `ReadDir`) and actually walking into it. Without this fix, that causes a hard failure. The fix is simple, if the directory is gone, just skip it:

```go
entries, err := s.Impl.ReadDir(ctx, directory)
if err != nil {
    if os.IsNotExist(err) {
        return nil // directory removed by concurrent sweep, skip
    }
    return errors.Wrap(err, "error reading directory")
}

```

### Backend implementations

Filesystem (`repo/blob/filesystem/filesystem_storage.go`), just delegates to `os.Remove`, which fails with `ENOTEMPTY` on non-empty dirs, exactly what we want:

```go
func (fs *fsImpl) RemoveDirInPath(_ context.Context, dirPath string) error {
    return fs.osi.Remove(dirPath)
}

```

SFTP (`repo/blob/sftp/sftp_storage.go`), uses `SSH_FXP_RMDIR` under the hood, which also only removes empty directories:

```go
func (s *sftpImpl) RemoveDirInPath(ctx context.Context, dirPath string) error {
    return s.rec.UsingConnectionNoResult(ctx, "RemoveDirInPath", func(conn connection.Connection) error {
        return sftpClientFromConnection(conn).RemoveDirectory(dirPath)
    })
}

```

I did **not** implement `DirRemover` for WebDAV. RFC 4918 says `DELETE` on a collection recursively deletes all contents, so there's no safe "remove only if empty" operation. A `ReadDir`\-then-`Remove` pattern has an unavoidable TOCTOU race that could cause data loss. Thanks to the optional interface pattern, WebDAV simply doesn't get sweep behavior, so no code change needed.

Edge Cases:
*   Unsharded blobs: `dirPath == RootPath`, so the sweep no-ops immediately
*   Trailing-slash `RootPath`: Handled by the `path.Clean()` normalization
*   Concurrent delete + put: `PutBlobInPath` recreates dirs via `MkSubdirAll` on `ENOENT`, so this is safe
*   Concurrent list + sweep: `walkDir` now gracefully handles `ENOENT`
*   `.shards` file: Lives in `RootPath`; sweep stops before reaching `RootPath`
*   Per-prefix shard overrides: Different depths share dirs safely (empty-dir semantics still hold)
*   Windows `os.Remove`: Calls `RemoveDirectory` for dirs, which only removes if empty

Tests

In `repo/blob/sharded/sharded_test.go`:
*   `TestDeleteBlobCleansUpEmptyShardDirs`, single blob, verify dirs get cleaned up
*   `TestDeleteBlobPreservesNonEmptyShardDirs`, shared prefix, delete one, shared dir stays
*   `TestDeleteBlobPartialChainCleanup`, same top shard, different sub-shards
*   `TestDeleteBlobCleanupPreservesShardsFile`, delete all blobs, `.shards` stays intact
*   `TestDeleteBlobCleanupWithFlatShards`, shards `[0]`, no dirs to clean

In `repo/blob/filesystem/filesystem_storage_test.go`:
*   `TestDeleteBlob_RemoveDirError_DoesNotFailDelete`, mock OS injects dir removal errors, blob delete still succeeds

Please let me know if you don't like anything about my changes, or you would like to go about resolving this Issue - but I'm hoping that this closes #2600 :)

### DIFF
--- a/repo/blob/sharded/sharded.go
+++ b/repo/blob/sharded/sharded.go
@@ -31,6 +31,12 @@ type Impl interface {
 	ReadDir(ctx context.Context, path string) ([]os.FileInfo, error)
 }
 
+// DirRemover is optionally implemented by Impl to enable
+// cleanup of empty shard directories after blob deletion.
+type DirRemover interface {
+	RemoveDirInPath(ctx context.Context, dirPath string) error
+}
+
 // Storage provides common implementation of sharded storage.
 type Storage struct {
 	Impl Impl
@@ -88,6 +94,10 @@ func (s *Storage) ListBlobs(ctx context.Context, prefix blob.ID, callback func(b
 
 		entries, err := s.Impl.ReadDir(ctx, directory)
 		if err != nil {
+			if os.IsNotExist(err) {
+				return nil // directory removed by concurrent sweep, skip
+			}
+
 			return errors.Wrap(err, "error reading directory")
 		}
 
@@ -196,8 +206,29 @@ func (s *Storage) DeleteBlob(ctx context.Context, blobID blob.ID) error {
 		return errors.Wrap(err, "error determining sharded path")
 	}
 
-	//nolint:wrapcheck
-	return s.Impl.DeleteBlobInPath(ctx, dirPath, filePath)
+	if err := s.Impl.DeleteBlobInPath(ctx, dirPath, filePath); err != nil {
+		return err //nolint:wrapcheck
+	}
+
+	if dr, ok := s.Impl.(DirRemover); ok {
+		s.sweepEmptyDirs(ctx, dr, dirPath)
+	}
+
+	return nil
+}
+
+func (s *Storage) sweepEmptyDirs(ctx context.Context, dr DirRemover, dir string) {
+	rootPath := path.Clean(s.RootPath)
+
+	for dir != rootPath && dir != "" && dir != "." && dir != "/" {
+		if err := dr.RemoveDirInPath(ctx, dir); err != nil {
+			return
+		}
+
+		log(ctx).Debugf("removed empty shard directory %v", dir)
+
+		dir = path.Dir(dir)
+	}
 }
 
 func (s *Storage) getParameters(ctx context.Context) (*Parameters, error) {
--- a/repo/blob/sharded/sharded.go
+++ b/repo/blob/sharded/sharded.go
@@ -206,29 +206,58 @@ func (s *Storage) DeleteBlob(ctx context.Context, blobID blob.ID) error {
 		return errors.Wrap(err, "error determining sharded path")
 	}
 
-	if err := s.Impl.DeleteBlobInPath(ctx, dirPath, filePath); err != nil {
-		return err //nolint:wrapcheck
+	//nolint:wrapcheck
+	return s.Impl.DeleteBlobInPath(ctx, dirPath, filePath)
+}
+
+// SweepEmptyDirectories walks the shard directory tree and removes empty
+// directories. It should be called after bulk blob deletions (e.g. during
+// maintenance) rather than after each individual DeleteBlob.
+func (s *Storage) SweepEmptyDirectories(ctx context.Context) error {
+	dr, ok := s.Impl.(DirRemover)
+	if !ok {
+		return nil
 	}
 
-	if dr, ok := s.Impl.(DirRemover); ok {
-		s.sweepEmptyDirs(ctx, dr, dirPath)
+	p, err := s.getParameters(ctx)
+	if err != nil {
+		return errors.Wrap(err, "error getting shard parameters")
+	}
+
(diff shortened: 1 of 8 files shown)


repo/blob/sharded/sharded.go
```diff
@@ -196,8 +206,29 @@ func (s *Storage) DeleteBlob(ctx context.Context, blobID blob.ID) error {
 		return errors.Wrap(err, "error determining sharded path")
 	}
 
-	//nolint:wrapcheck
-	return s.Impl.DeleteBlobInPath(ctx, dirPath, filePath)
+	if err := s.Impl.DeleteBlobInPath(ctx, dirPath, filePath); err != nil {
```
julio-lopez: This is the wrong place to perform this operation. It results in `DeleteBlob` doing a lot more than deleting a blob.

Also, is there a race here that would cause `PutBlob` in some cases?

julio-lopez (changes requested):

repo/blob/sharded/sharded.go
perfectra1n: You're right that `DeleteBlob` shouldn't own this...I'll move the sweep to maintenance, as a step after orphaned pack deletion. That's also more efficient (one pass vs per-blob) and eliminates any theoretical race since maintenance holds the exclusive lock.

Re the race...`os.Remove` fails on non-empty dirs, `PutBlobInPath` retries with `MkSubdirAll` on `IsNotExist`, and bulk deletes only run under the maintenance lock. But moving it to maintenance should make the concern moot regardless

perfectra1n:
If there's anything else you need from me, just let me know :)

onedr0p:
Bump @julio-lopez 😃

PR #5217, opened 2026-03-13
fix(snapshots): inaccessible entry causes parent directory to be skipped

### LINKED ISSUE #5045: Inaccessible mount point causes parent directory to be skipped

julio-lopez:
- Reverts #5044 (commit c8c461559568e5781471d96222864cac1c3f9d4f)
- Supersedes #5046
- Fixes #5045
- #5225
- #5228

Credit to @geoffreybennett for proposing the initial fix.

### DIFF
--- a/tests/end_to_end_test/snapshot_fail_test.go
+++ b/tests/end_to_end_test/snapshot_fail_test.go
@@ -422,9 +422,16 @@ func testPermissions(
 				e.RunAndExpectSuccess(t, "snapshot", "restore", parsed.manifestID, target)
 			}
 
-			require.Equal(t, expected.wantErrors, parsed.errorCount, "unexpected number of errors")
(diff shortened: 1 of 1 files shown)


geoffreybennett:
Tested; works for me now, thanks!

PR #5231, opened 2026-03-19
fix(general): do not included concatenated entries in log

### LINKED ISSUE #3093: Debug logs for concatenation are too long
When large files are split and then concatenated, a debug log is written with a JSON object of all of the concatenated data ([object_manager.go, line 118](https://github.com/kopia/kopia/blob/d2ddda7114bd4e3ff26ddb9c52fa2c778b806514/repo/object/object_manager.go#L118)). A recent example from a snapshot I'm running was over 1,000,000 characters long. When scrolling through the logs on the web UI, this is so long that it makes it very difficult to get useful data from the logs.

I would love to see a way to either:
1. Not emit these logs (not sure if they're adding any value)
2. Truncate these (is all of the concatenated data actually useful in debug logs?)
3. Filter these logs out in the web UI
4. Choose not to emit these via policy config (there is already a section for logging verbosity, so adding an option there would make sense)

majiayu000:
- Fixes #3093

## Changes

The debug log at `object_manager.go:132` was using `%v` to format the entire `concatenatedEntries` slice, which for large files with many indirect entries could produce 1M+ character log lines.

Changed `log(ctx).Debugf("concatenated: %v total: %v", concatenatedEntries, totalLength)` to `log(ctx).Debugf("concatenated %d entries, total: %v", len(concatenatedEntries), totalLength)`.

This logs just the entry count instead of dumping the full slice contents. `totalLength` is already a simple `int64` so it stays as-is.

## Test Plan

- `go build ./repo/object/...` — compiles cleanly
- `go vet ./repo/object/...` — no issues
- `go test ./repo/object/...` — all tests pass

Log-only change, no behavioral impact.

### DIFF
--- a/repo/object/object_manager.go
+++ b/repo/object/object_manager.go
@@ -129,7 +129,7 @@ func (om *Manager) Concatenate(ctx context.Context, objectIDs []ID, metadataComp
 		}
 	}
 
-	log(ctx).Debugf("concatenated %d entries, total: %v", len(concatenatedEntries), totalLength)
+	log(ctx).Debugf("concatenated %d entries, total object length: %d", len(concatenatedEntries), totalLength)
 
 	w := om.NewWriter(ctx, WriterOptions{
 		Prefix:             indirectContentPrefix,


repo/object/object_manager.go
```diff
@@ -129,7 +129,7 @@ func (om *Manager) Concatenate(ctx context.Context, objectIDs []ID, metadataComp
 		}
 	}
 
-	log(ctx).Debugf("concatenated: %v total: %v", concatenatedEntries, totalLength)
+	log(ctx).Debugf("concatenated %d entries, total: %v", len(concatenatedEntries), totalLength)
```
julio-lopez:
```suggestion
	log(ctx).Debugf("concatenated %d entries, total object length: %d", len(concatenatedEntries), totalLength)
```

majiayu000: Done, applied the suggested wording. Thanks!

PR #5234, opened 2026-03-19
fix(snapshots): ErrorEntry policy resolution to use child policy

### LINKED ISSUE #5232: Implement review suggestion on policy resolution for failed entries
- See https://github.com/kopia/kopia/pull/5217#discussion_r2957889891

Bingtagui404:
## Problem

In `processSingle()`, the `fs.ErrorEntry` case resolves error handling policy using `policyTree.EffectivePolicy()`, which returns the **parent directory's** policy. This means per-entry error handling rules (e.g., policies set on a specific problematic subdirectory) are ineffective for ErrorEntry items.

This is inconsistent with how directory processing works, where `childTree := policyTree.Child(entry.Name())` is used to get the entry-specific policy.

Fixes #5232 (from [PR #5217 review suggestion](https://github.com/kopia/kopia/pull/5217#discussion_r2957889891))

## Fix

Changed `policyTree.EffectivePolicy()` to `policyTree.Child(entry.Name()).EffectivePolicy()` in the `fs.ErrorEntry` branch of `processSingle()`, consistent with how `fs.Directory` derives `childTree`.

## Test Evidence

- Added `AddFileErrorEntry()` to `internal/mockfs/mockfs.go` to construct file-typed ErrorEntry (previously only directory-typed was possible).
- Added `TestUpload_ErrorEntryChildPolicy` with three sub-cases:
  - Child policy ignores dir errors only → dir-err ignored, file-err fatal
  - Child policy ignores file errors only → file-err ignored, dir-err fatal
  - Child policy disables unknown type ignore → unknown-err fatal, dir-err and file-err ignored

  This verifies `policyTree.Child()` resolution for all three ErrorEntry code paths (directory errors, file errors, and unknown types).

### DIFF
--- a/internal/mockfs/mockfs.go
+++ b/internal/mockfs/mockfs.go
@@ -204,7 +204,7 @@ func (imd *Directory) AddDir(name string, permissions os.FileMode) *Directory {
 	return subdir
 }
 
-// AddErrorEntry adds a fake directory with a given name and permissions.
+// AddErrorEntry adds a fake directory-typed error entry with a given name and permissions.
 func (imd *Directory) AddErrorEntry(name string, permissions os.FileMode, err error) *ErrorEntry {
 	imd, name = imd.resolveSubdir(name)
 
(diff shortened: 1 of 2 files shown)


snapshot/upload/upload.go
```diff
@@ -928,8 +928,13 @@ func (u *Uploader) processSingle(
 			prefix         string
 		)
 
+		// Use the child policy for the specific entry path, not the parent directory policy.
+		// This ensures per-entry error handling rules are respected, consistent with how
+		// directory processing derives childTree via policyTree.Child().
+		childPolicy := policyTree.Child(entry.Name()).EffectivePolicy()
```
julio-lopez, line 934: - Is `entry.Name()` guaranteed to return a non-empty string?
- What's the behavior for `.Child("")` ? What does it return?

julio-lopez (changes requested):
Please rebase and resolve the conflicts

snapshot/upload/upload.go
Bingtagui404: `entry.Name()` is not guaranteed to be non-empty by the `fs.ErrorEntry` interface itself, but in the current implementations here it comes from a basename for directory children (for example `os.DirEntry.Name()` in localfs).

Also, `policyTree.Child("")` returns the current tree, so even an empty name would safely fall back to the previous parent-policy behavior.

This matches the existing `Child(entry.Name())` pattern already used elsewhere in `processSingle()`.

I've also rebased onto current `master` and added a regression test covering child-scoped `IgnoreUnknownTypes`.

PR #5238, opened 2026-03-23
fix(notifications): add `Date` header in RFC 1123Z format

### LINKED ISSUE #5151: Notification mail: Date header field missing
When kopia sends notification mails, the header is missing the Date field.

In this case Thunderbird displays the current date when the mail was retrieved via IMAP, which can be confusing.

BugCoder762:
## Problem
Email messages sent by Kopia's notification system were missing the `Date` header, which can cause issues with email clients, servers, and spam filters that expect this standard header for proper message timestamping and delivery.

This violates RFC 5322 where the `Date` header is mandatory for Internet messages. Without it, clients like Thunderbird display the download time instead of the actual send time, breaking chronological order in inboxes.

## Solution
Added the `Date` header to outgoing email messages using RFC 1123Z format (e.g., `Mon, 23 Mar 2026 10:12:20 +0000`), which is the standard format for email timestamps as defined in RFC 5322.

## Changes
- **notification/sender/email/email_sender.go**: 
  - Added import for `internal/clock` package.
  - Modified the `Send` method to include `"Date: " + clock.Now().Format(time.RFC1123Z)` in the email headers.
  - Used `clock.Now()` instead of `time.Now()` to follow project conventions for time handling.

- **notification/sender/email/email_sender_test.go**:
  - Added `regexp` import for date validation.
  - Refactored tests to use `require.Contains` for individual header checks instead of exact string matching.
  - Added `require.Regexp` validation for the Date header format using RFC 1123Z regex pattern.
  - This makes tests more robust by validating semantic content rather than exact timestamps.

## Why RFC 1123Z?
- It's the standard format for email Date headers as defined in RFC 5322.
- Includes timezone information (+0000 for UTC).
- Widely supported by email clients and servers.

## Why clock.Now()?
- Follows Kopia's internal convention of using `internal/clock` for time operations outside dedicated timing packages.
- Improves testability and consistency across the codebase.

## Testing
- All existing tests pass.
- New tests validate the Date header presence and format.
- Verified with both HTML and plain text email formats.

## Checklist
- [x] Tests pass locally
- [x] Code follows project style guidelines
- [x] No breaking changes to existing functionality
- [x] Email format complies with RFC standards

Fixes #5151


notification/sender/email/email_sender_test.go
```diff
@@ -55,26 +56,29 @@ func TestEmailProvider(t *testing.T) {
...
+	// Verify headers are present
+	require.Contains(t, msgContent, "Subject: Test\r\n")
+	require.Contains(t, msgContent, "From: some-user@example.com\r\n")
+	require.Contains(t, msgContent, "To: another-user@example.com\r\n")
+	require.Contains(t, msgContent, "MIME-version: 1.0;\r\n")
+	require.Contains(t, msgContent, "Content-Type: text/html; charset=\"UTF-8\";\r\n")
+	require.Contains(t, msgContent, "X-ExtraHeader: value\r\n")
+
+	// Verify Date header is present and follows RFC 1123Z format
+	require.Regexp(t, regexp.MustCompile(`Date: [A-Za-z]{3}, \d{2} [A-Za-z]{3} \d{4} \d{2}:\d{2}:\d{2} [+-]\d{4}\r\n`), msgContent)
```
julio-lopez:
```suggestion
	require.Regexp(t, `Date: [A-Za-z]{3}, \d{2} [A-Za-z]{3} \d{4} \d{2}:\d{2}:\d{2} [+-]\d{4}\r\n`, msgContent)
```

notification/sender/email/email_sender_test.go
```diff
@@ -119,24 +123,27 @@ func TestEmailProvider_Text(t *testing.T) {
...
+	msgContent := msg.MsgRequest()
+
+	// Verify headers are present
+	require.Contains(t, msgContent, "Subject: Test\r\n")
+	require.Contains(t, msgContent, "From: some-user@example.com\r\n")
+	require.Contains(t, msgContent, "To: another-user@example.com\r\n")
+	require.Contains(t, msgContent, "X-ExtraHeader: value\r\n")
+
+	// Verify Date header is present and follows RFC 1123Z format
+	require.Regexp(t, regexp.MustCompile(`Date: [A-Za-z]{3}, \d{2} [A-Za-z]{3} \d{4} \d{2}:\d{2}:\d{2} [+-]\d{4}\r\n`), msgContent)
```
julio-lopez:
```suggestion
	require.Regexp(t, `Date: [A-Za-z]{3}, \d{2} [A-Za-z]{3} \d{4} \d{2}:\d{2}:\d{2} [+-]\d{4}\r\n`, msgContent)
```

notification/sender/email/email_sender_test.go
```diff
@@ -2,6 +2,7 @@ package email_test
 
 import (
 	"context"
+	"regexp"
```
julio-lopez:
```suggestion
```

notification/sender/email/email_sender_test.go
```diff
@@ -119,24 +122,27 @@ func TestEmailProvider_Text(t *testing.T) {
...
+	// Verify body is present
+	require.Contains(t, msgContent, "This is a test.\r\n")
+	require.Contains(t, msgContent, "* one\r\n")
+	require.Contains(t, msgContent, "* two\r\n")
+	require.Contains(t, msgContent, "* three\r\n")
+	require.Contains(t, msgContent, "# Header\r\n")
+	require.Contains(t, msgContent, "## Subheader\r\n")
+	require.Contains(t, msgContent, "- a\r\n")
+	require.Contains(t, msgContent, "- b\r\n")
+	require.Contains(t, msgContent, "- c\r\n")
```
julio-lopez, line 145: These checks are insufficient, as the body content should be verbatim.

notification/sender/email/email_sender_test.go
```diff
@@ -55,26 +55,29 @@ func TestEmailProvider(t *testing.T) {
...
+	require.Contains(t, msgContent, "From: some-user@example.com\r\n")
+	require.Contains(t, msgContent, "To: another-user@example.com\r\n")
+	require.Contains(t, msgContent, "MIME-version: 1.0;\r\n")
+	require.Contains(t, msgContent, "Content-Type: text/html; charset=\"UTF-8\";\r\n")
+	require.Contains(t, msgContent, "X-ExtraHeader: value\r\n")
+
+	// Verify Date header is present and follows RFC 1123Z format
+	require.Regexp(t, `Date: [A-Za-z]{3}, \d{2} [A-Za-z]{3} \d{4} \d{2}:\d{2}:\d{2} [+-]\d{4}\r\n`, msgContent)
+
+	// Verify body is present
```
julio-lopez, line 71: - The check should verify that the message body matches exactly.
- Also, this is removing an implicit check for "an empty line" between headers and the body.

julio-lopez (changes requested):
See inline comment

PR #5295, opened 2026-04-07
fix: prevent goroutine deadlock in workshare pool and parallelwork queue

davidcforbes:
## Summary

- Add panic recovery to `workshare.Pool` worker goroutines — ensures `wg.Done()`, semaphore release, and `activeWorkers` decrement always execute even if the process function panics, preventing pool deadlock
- Add panic recovery to `parallelwork.Queue` callbacks — converts panics to errors ensuring `completed()` is always called, preventing `dequeue()` from waiting forever
- Make `filesystemFile.Open()` and `filesystemDirectory.Iterate()` context-aware — wraps `os.Open()` with context cancellation and a 60-second default timeout, preventing indefinitely blocking file opens from stalling workshare pool workers
- **(follow-up commit)** Add optional `fs.Preflightable` interface; `localfs.filesystemFile` implements as `os.Open` + `Close`. `Uploader.processSingle` calls `Preflight` before workshare assignment so locked files surface as ignored errors immediately, complementing the timeout-based defense above by avoiding worker assignment for files that are already known-inaccessible.

Fixes #5294

## Details

Snapshot create hung for 17+ hours on Windows when `os.Open()` blocked indefinitely on a locked NTFS file inside a workshare pool worker. The worker never completed, so `AsyncGroup.Wait()` blocked forever, cascading to stall all pool workers.

The `Open()` and `Iterate()` methods in `fs/localfs/` already accepted a `context.Context` parameter but discarded it (using `_`). This change makes them actually respect context cancellation and deadlines.

The Preflight follow-up adds an optional interface check before workshare assignment. Entries that don't implement it (e.g. virtualfs, cachefs streams) fall through unchanged. Compile-time conformance is enforced with a `var _ =` assertion in localfs.

## Test plan

- [x] `TestWorkerRecoversPanic` — workshare pool worker panic doesn't deadlock `Wait()`
- [x] `TestWorkerRecoversPanic_PoolRemainsOperational` — pool stays usable after all workers panic
- [x] `TestProcessRecoversPanic` — parallelwork panic in callback becomes an error
- [x] `TestProcessRecoversPanicWithEnqueuedWork` — panic after enqueuing more work doesn't deadlock
- [x] `TestOpenWithContext_Success/FileNotFound/CancelledContext/Timeout/Directory` — context-aware file open
- [x] All existing tests pass (`go test ./internal/workshare/ ./internal/parallelwork/ ./fs/localfs/`)
- [ ] `make lint` — passes locally
- [ ] Manual: snapshot a Windows source with in-use Office temp files; locked-file errors appear as Ignored errors at the same logical position as today

### DIFF
--- a/fs/entry.go
+++ b/fs/entry.go
@@ -50,6 +50,15 @@ type File interface {
 	Open(ctx context.Context) (Reader, error)
 }
 
+// Preflightable is optionally implemented by File entries that support
+// a quick accessibility check before upload. This allows detecting locked
+// or inaccessible files early, before a workshare worker is assigned.
+type Preflightable interface {
+	// Preflight performs a quick accessibility check (e.g. open + close).
+	// Returns nil if the file is accessible, or an error if it cannot be read.
+	Preflight(ctx context.Context) error
+}
+
 // StreamingFile represents an entry that is a stream.
 type StreamingFile interface {
 	Entry
--- a/fs/localfs/local_fs.go
+++ b/fs/localfs/local_fs.go
@@ -114,6 +114,20 @@ func (fsf *filesystemFile) Open(ctx context.Context) (fs.Reader, error) {
 	return &fileWithMetadata{f}, nil
 }
 
+// Preflight performs a quick accessibility check by opening and immediately
+// closing the file. This catches locked/in-use files before a workshare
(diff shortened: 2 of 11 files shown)


davidcforbes:

PR #5300, opened 2026-04-09
chore(ci): generate govulncheck SARIF report

julio-lopez:
- Generate and uploads govulncheck SARIF report.
- Invoke govulncheck directly instead of using the govulncheck-action. This also has the benefit of using the same Go version from the go.mod file, without having to explicitly specify it and update it in the workflow file.


Rohit-BM18 (commented):
LGTM

PR #5319, opened 2026-04-17
feat(repository): Add a blob implementation for Storj with tests

ptandler:
This CL adds a blob storage implementation for the Storj cloud service.
It helps implement #3426.


ppenguin:
@ptandler Just stumbled on this PR, and took it as a motivation to finally publish my code base that has the same goal (and has been sitting in a private fork for more than a year in a different integration), see #5342.

I currently lack the time for a deeper comparison on how we might achieve synergies and whether this PR still has open issues to be solved, but of course feel free to make this comparison, so we might arrive at the optimal solution.

Cheers

PR #5333, opened 2026-04-25
fix(server): JWT v5, request limits, cookie flags, metrics auth, header injection

davidcforbes:
## Summary

Eight server-side security fixes:

- **JWT v4 → v5** with explicit algorithm validation and issuer/audience verification
- 20 MB request-body size limit on the HTTP handler matching the gRPC limit
- `HttpOnly` and `SameSite` flags added to auth and session cookies
- Require auth for `/metrics` endpoint (was anonymous on the main server)
- Use `mime.FormatMediaType` for `Content-Disposition` header (prevents header injection via filename)
- Add `kopia:sensitive` tag to RClone `EmbeddedConfig` so it's never logged
- Strip SSH config from SFTP dial error message (was leaking host/port/user)
- Deduplicate `reportSeverity` into the notification package

Includes a new `internal/server/security_test.go` (~200 LOC) covering each fix.

## Test plan

- [ ] `make test ./internal/server/...` — new tests + existing
- [ ] `make lint`
- [ ] Manual: hit `/metrics` without auth → 401; hit with auth → 200


davidcforbes:

PR #5334, opened 2026-04-25
fix(perf): allocation reduction, buffer reuse, fsync, cache scan

davidcforbes:
## Summary

Six allocation/hotpath fixes:

- Replace `fmt.Sprintf` with fast path in `contentCacheKeyForInfo`
- Use raw bytes instead of `String()` in index `shard()`
- Add fast path for single-char prefix in `comparePrefix`
- Reuse chunk buffer in object reader
- Add `fsync` before rename in disk index cache (durability)
- Store only `ModTime` in `expireUnused` map (less memory per entry)

Includes new benchmark/unit tests:
- `repo/content/content_cache_key_test.go`
- `repo/content/index/id_test.go`
- `repo/content/index/shard_bench_test.go`

## Test plan

- [ ] `make test`
- [ ] `go test -bench=. ./repo/content/index/...` — verify shard fast path improves
- [ ] `make lint`

### DIFF
--- a/repo/content/committed_content_index_disk_cache.go
+++ b/repo/content/committed_content_index_disk_cache.go
@@ -113,6 +113,10 @@ func writeTempFileAtomic(dirname string, data []byte) (string, error) {
 		return "", errors.Wrap(err, "can't write to temp file")
 	}
 
+	if err := tf.Sync(); err != nil {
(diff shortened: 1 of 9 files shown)


davidcforbes:

PR #5336, opened 2026-04-25
feat(mount): WinFsp/FUSE mount backend for Windows

davidcforbes:
## Summary

Adds WinFsp/FUSE mount support as an alternative to WebDAV+net use on Windows. Mounted snapshots now appear as local filesystems, enabling Explorer preview pane, thumbnails, and search.

- `internal/winfsp/winfspfs.go` — cgofuse filesystem with read-only snapshot access (`Getattr`, `Open`, `Read`, `Readdir`, `Readlink`, `Statfs`)
- `internal/mount/mount_winfsp.go` — WinFsp controller with `PreferWebDAV` fallback option
- `internal/mount/mount_net_use_helper.go` — extracted shared WebDAV+net-use logic for reuse by both backends
- Build with `-tags winfsp` (requires CGO + WinFsp SDK)
- Default Windows build (no tag) keeps WebDAV unchanged

Plus three follow-on fixes:

- Free-drive-letter assignment when caller passes `*` (cgofuse doesn't grok net-use's `*` convention)
- Set `FileSystemName=NTFS` so Explorer trusts the volume (preview pane works); set `volname=Kopia` for a friendly label
- **Always clean up mount state on `Unmount` failure** — previously left dead controllers in `s.mounts`, making the Mount button permanently broken until server restart. Includes a new `internal/server/api_mount_test.go` (~125 LOC).
- Open-file IPC: `file-viewer.js` + `open-file` IPC handler so the htmlui can open snapshot files in the system default viewer (15 Playwright tests). Companion htmlui PR adds the View button.

## Test plan

- [ ] `make test`
- [ ] Manual: `kopia mount` a snapshot on Windows; verify Explorer preview pane works
- [ ] Manual: trigger an unmount failure (e.g. file open from another process); verify a subsequent mount works (was permanently broken before this PR)

### DIFF
--- a/app/public/electron.js
+++ b/app/public/electron.js
@@ -33,6 +33,7 @@ import {
   isFirstRun,
   isPortableConfig,
 } from "./config.js";
+import { downloadObject, cleanupTempDir } from "./file-viewer.js";
 
 import Store from "electron-store";
 import log from "electron-log";
@@ -210,6 +211,7 @@ app.on("will-quit", function () {
   allConfigs().forEach((repositoryID) =>
     serverForRepo(repositoryID).stopServer(),
   );
+  cleanupTempDir();
 });
 
 app.on("login", (event, webContents, _request, _authInfo, callback) => {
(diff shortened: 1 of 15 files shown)


davidcforbes:

PR #5338, opened 2026-04-25
fix(tools): drop incompatible linter flags for golangci-lint v2 on Windows

davidcforbes:
## Summary

`golangci-lint` v2 reclassified `gofmt` and `goimports` as formatters, so the `-D` flag (which disables linters) no longer accepts them. Since `.golangci.yml` controls formatter configuration, these flags are unnecessary.

## Test plan

- [ ] `make lint` succeeds on Windows with golangci-lint v2.x


davidcforbes:

PR #5340, opened 2026-04-28
fix(repository): Delete superseded xs index blobs after range compaction

### LINKED ISSUE #4919: Delete compacted (single) epoch `xs` indexes once they are part of a compacted range (`xr`)
Objective: free up the storage space taken by `xs` index blobs when they are no longer needed.

Rationale:

The `xs<epoch>_` index blobs stop being read _some time_ after a covering `xr` index set has been created, and they can be deleted after _enough time_ has passed.

Context:

`xs` index blobs are produced by compacting the `xn<epoch>_...` index blobs from a _single epoch_. Epoch range compaction creates `xr<min_epoch>_<max_epoch>_...` index blobs that supersede all the blobs (`xn` and `xs`) for the epochs covered by the range. Then, kopia processes use the range-compacted `xr...` index set when they are available, instead of the `xs<epoch>_...` or `xn<epoch>_...` indexes for the superseded epochs.

Expected behavior:
`xs<epoch>_...` indexes should be deleted some time after they are superseded by a range compaction index set (`xr<low_epoch>_<high_epoch>_...` blobs).

Ref:
- #4901

kushshukla24:
**CleanupSupersededIndexes** only deleted _xn_ (uncompacted) index blobs once a covering _xs_ (single-epoch compaction) blob existed. _xs_ blobs themselves were never deleted, even after an _xr_ (epoch range checkpoint) blob fully covered their epoch and aged past **CleanupSafetyMargin**, so they accumulated indefinitely on object storage despite no reader ever consulting them.

Extend **CleanupSupersededIndexes** to also delete _xs<epoch>_..._ blobs whose epoch falls inside some **LongestRangeCheckpointSets** entry whose own blobs are older than **maxReplacementTime**. The new **rangeCoversEpochAndOldEnough** helper reuses **blobSetWrittenEarlyEnough** so the safety-margin discipline is identical to the existing _xn_ cleanup. **CleanupSupersededIndexesStats** now folds both passes into **DeletedBlobCount** and **DeletedTotalSize**.

Closes: https://github.com/kopia/kopia/issues/4919

@julio-lopez


kushshukla24:
@julio-lopez Please review. Let me know if anything is missing from my end.

lipuszb (approved):

PR #5342, opened 2026-04-28
Add storj backend

ppenguin:
Add a native storj storage backend because the performance of `rclone` has been found to be lacking.

I'm aware of #5319, but lack the time to study merits of its implementation compared to the code in this PR.

I've been using this implementation already for some time (most of the implementation was done more than a year ago) in a private integration, so it's somewhat battle tested.


moronimotta:
Hello! I enjoy browsing trending repositories to observe and learn more about how large-scale projects are organized. What is the advantage of using Go to develop a cross-platform backup tool compared to other programming languages, and what were the biggest challenges involved in writing tests specifically for different operating systems?

PR #5354, opened 2026-05-05
feat(cli): restrict insecure unauthenticated server to loopback binds

jkowalski:
Reject starting the server with --insecure and --without-password when --address would bind outside loopback (including empty host / all interfaces). Validate the actual listener after Listen so systemd socket activation cannot bypass the check.

Add hidden --allow-extremely-dangerous-unauthenticated-server-on-the-network to opt into the previous behavior for isolated environments.

Extract validation into internal/insecureserverbind with table-driven tests. Add end-to-end smoke tests for rejection and escape hatch.

### DIFF
--- a/internal/insecureserverbind/insecureserverbind.go
+++ b/internal/insecureserverbind/insecureserverbind.go
@@ -47,7 +47,13 @@ func stripProtocol(addr string) string {
 }
 
 // ParseListenHost extracts the host part of a server listen address flag value.
-// If isUnix is true, host is empty and the address refers to a unix domain socket.
+// If isUnix is true, host is empty and the address refers to a Unix domain socket.
+//
+// Unix detection runs after stripping a leading http:// or https:// (same as the server’s
+// stripProtocol). Any form that becomes unix:… is treated as a Unix socket, including:
+//   - unix:/path/to/socket
+//   - http://unix:/path/to/socket
+//   - https://unix:/path/to/socket
 func ParseListenHost(address string) (host string, isUnix bool, err error) {
 	stripped := stripProtocol(address)
 	if strings.HasPrefix(stripped, "unix:") {
--- a/internal/insecureserverbind/insecureserverbind.go
+++ b/internal/insecureserverbind/insecureserverbind.go
@@ -1,0 +1,123 @@
+// Package insecureserverbind validates listen addresses for insecure, unauthenticated Kopia servers.
+package insecureserverbind
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+)
+
+// AllowDangerousUnauthenticatedNetworkFlag is the CLI flag that disables bind restrictions.
+const AllowDangerousUnauthenticatedNetworkFlag = "allow-extremely-dangerous-unauthenticated-server-on-the-network"
+
+// AllowDangerousUnauthenticatedNetworkFlagHelp is the kingpin description for that flag.
+const AllowDangerousUnauthenticatedNetworkFlagHelp = "Allow unauthenticated server to listen on non-loopback addresses; " +
+	"exposes full repository and control API to the network without authentication (extremely dangerous)"
+
+// ErrDisallowedPublicBind is returned when the address would expose an unauthenticated server beyond loopback.
+var ErrDisallowedPublicBind = errors.New("refusing to expose unauthenticated server on non-loopback network bind")
+
+// RestrictionApplies reports whether insecure unauthenticated bind checks must run.
+func RestrictionApplies(insecure, withoutPassword, allowDangerousNetwork bool) bool {
+	return insecure && withoutPassword && !allowDangerousNetwork
+}
+
+// ValidateListenAddressIfRestricted runs [ValidateListenAddressFlag] only when [RestrictionApplies] is true.
+func ValidateListenAddressIfRestricted(insecure, withoutPassword, allowDangerousNetwork bool, address string) error {
+	if !RestrictionApplies(insecure, withoutPassword, allowDangerousNetwork) {
+		return nil
+	}
+
+	return ValidateListenAddressFlag(address)
+}
+
+// ValidateListenerAddrIfRestricted runs [ValidateListenerAddr] only when [RestrictionApplies] is true.
+func ValidateListenerAddrIfRestricted(insecure, withoutPassword, allowDangerousNetwork bool, addr net.Addr) error {
+	if !RestrictionApplies(insecure, withoutPassword, allowDangerousNetwork) {
+		return nil
+	}
+
+	return ValidateListenerAddr(addr)
+}
+
+func stripProtocol(addr string) string {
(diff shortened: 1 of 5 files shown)


internal/insecureserverbind/insecureserverbind.go
```diff
@@ -0,0 +1,123 @@
...
+	"net/url"
+	"strings"
+)
+
+// AllowDangerousUnauthenticatedNetworkFlag is the CLI flag that disables bind restrictions.
+const AllowDangerousUnauthenticatedNetworkFlag = "allow-extremely-dangerous-unauthenticated-server-on-the-network"
+
+// AllowDangerousUnauthenticatedNetworkFlagHelp is the kingpin description for that flag.
+const AllowDangerousUnauthenticatedNetworkFlagHelp = "Allow unauthenticated server to listen on non-loopback addresses; " +
+	"exposes full repository and control API to the network without authentication (extremely dangerous)"
```
julio-lopez: expand on why it is extremely dangerous ?
```suggestion
	"exposes full repository and control API to the network without authentication which allows any external attacker to take full control of the server host (extremely dangerous)"
```

internal/insecureserverbind/insecureserverbind.go
```diff
@@ -0,0 +1,123 @@
...
+
+func stripProtocol(addr string) string {
+	return strings.TrimPrefix(strings.TrimPrefix(addr, "https://"), "http://")
+}
+
+// ParseListenHost extracts the host part of a server listen address flag value.
+// If isUnix is true, host is empty and the address refers to a unix domain socket.
+func ParseListenHost(address string) (host string, isUnix bool, err error) {
+	stripped := stripProtocol(address)
+	if strings.HasPrefix(stripped, "unix:") {
```
julio-lopez, line 59: Not sure I follow this. IIUC, it would return true for `[http[s]://][unix:]anything`. That is, the following would return true:

- `http://unix:/somepath`
- `https://unix:/somepath/b`
- `unix:/somepath`

Is that the case?

internal/insecureserverbind/insecureserverbind_test.go
```diff
@@ -0,0 +1,210 @@
...
+		true, true, true, "http://0.0.0.0:0",
+	), "escape hatch skips validation")
+
+	err := ValidateListenAddressIfRestricted(true, true, false, "http://0.0.0.0:0")
+	require.Error(t, err)
+	require.ErrorIs(t, err, ErrDisallowedPublicBind)
+
+	require.NoError(t, ValidateListenAddressIfRestricted(
+		true, true, false, "http://127.0.0.1:0",
+	))
```
julio-lopez: nit: why aren't these either in separate tests (with corresponding names) or subtests? and optionally in a table-drive test?

internal/insecureserverbind/insecureserverbind_test.go
```diff
@@ -0,0 +1,210 @@
...
+	require.NoError(t, ValidateListenerAddrIfRestricted(true, true, true, pub))
+
+	err := ValidateListenerAddrIfRestricted(true, true, false, pub)
+	require.Error(t, err)
+	require.ErrorIs(t, err, ErrDisallowedPublicBind)
+
+	require.NoError(t, ValidateListenerAddrIfRestricted(
+		true, true, false,
+		&net.TCPAddr{IP: net.ParseIP("127.0.0.1"), Port: 1},
+	))
```
julio-lopez: ditto

internal/insecureserverbind/insecureserverbind_test.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+func TestValidateListenAddressFlag(t *testing.T) {
+	t.Parallel()
+
+	ok := []string{
+		"http://127.0.0.1:51515",
+		"http://localhost:0",
+		"http://LOCALHOST:9999",
```
julio-lopez, line 187: maybe add a mixed case case here for completeness?

julio-lopez (commented):
Overall 👍 LG: +1
🥇Thanks for doing this.

julio-lopez (dismissed):

julio-lopez (approved):

Blacks-Army:
This should have been clearly marked as a breaking change in the release notes. The current wording (“restrict insecure unauthenticated server to loopback binds”) underrepresents the actual impact, as it breaks existing deployments that rely on non-loopback interfaces and can silently prevent remote access.

jkowalski:
Running server with `--insecure --without-password` on a non-localhost is _extremely dangerous_. It is unfortunate that this was sometimes cited as recommended as a way of running Kopia in various discussion threads and LLM responses trained on those. 

Everyone running in such mode is urged to upgrade to 0.23.0.
See https://github.com/kopia/kopia/security/advisories/GHSA-2q4c-3mrw-63c3

Blacks-Army:
> Running server with --insecure --without-password on a non-localhost is extremely dangerous. It is unfortunate that this was sometimes cited as recommended as a way of running Kopia in various discussion threads and LLM responses trained on those.
> 
> Everyone running in such mode is urged to upgrade to 0.23.0.
> See GHSA-2q4c-3mrw-63c3

I understand the security concern, and I agree that exposing an unauthenticated insecure server directly to the internet is dangerous.

However, I think the release note still understates the impact of this change. Not everyone using _--insecure_ _--without-password_ on a non-loopback interface is doing so accidentally or because of bad advice from LLMs or old discussion threads.

In my case, Kopia is not publicly exposed directly. It is only reachable behind Traefik with OAuth2 Proxy authentication in front of it. That is a deliberate and fairly common reverse-proxy setup.

I am not arguing against the restriction itself, but this is still effectively a breaking behavioral change for existing deployments, and I think the release notes should communicate that more explicitly.

jkowalski:
Added `[BREAKING CHANGE]` tag to the release notes.

jkowalski:
> In my case, Kopia is not publicly exposed directly. It is only reachable behind Traefik with OAuth2 Proxy authentication in front of it. That is a deliberate and fairly common reverse-proxy setup.

This does improve the security posture, just make sure you're also disallowing connections from hosts other than the proxy itself (via iptables or similar), otherwise anybody on the network can attack your server and execute arbitrary code, read and write files through unauthenticated API.

Perhaps we should add something like `--reverse-proxy-host` which would allow connections only from that one host.

Blacks-Army:
> > In my case, Kopia is not publicly exposed directly. It is only reachable behind Traefik with OAuth2 Proxy authentication in front of it. That is a deliberate and fairly common reverse-proxy setup.
> 
> 
> 
> This does improve the security posture, just make sure you're also disallowing connections from hosts other than the proxy itself (via iptables or similar), otherwise anybody on the network can attack your server and execute arbitrary code, read and write files through unauthenticated API.
> 
> 
> 
> Perhaps we should add something like `--reverse-proxy-host` which would allow connections only from that one host.

Kopia and Traefik are connected through a dedicated traefik_kopia network, so Traefik is the only application that can access Kopia.

PR #5373, opened 2026-05-14
fix(repository): selection of next epoch to compact

### LINKED ISSUE #5371: Single epoch compaction may produce incomplete index sets in a specific scenario which may lead to data loss
The issue was discovered and diagnosed by @ashmrtn 

There is a scenario where single epoch compaction (the process that creates `xs` index sets) may use an incomplete (corresponding) `xr` index set.

The scenario is depicted in the figure below.

Image

Suppose:
- there are 7 settled epochs `[0..6]`;
- a range compaction for epochs 0 - 2 is happens before the single epoch compaction for `xs2` is performed;

Then:
1. Single compaction epoch 2 never occurs, this is not an issue as the compaction process is designed to handle this situation. However, this state is a precondition for the issue to occur, as it creates a gap in the sequence of epochs with single compactions (`xs` index sets).
2. the next elegible epoch for single-epoch compaction (`xs`) is epoch 3;
3. epoch 3 is compacted into a corresponding `xs3` index set.

Due to an incorrect assumption, the selection for the next epoch to compact, returns the next epoch after the highest epoch covered by a range compaction. In the figure below, the highest epoch covered by the range compaction is 2, so epoch 3 is the next epoch to be compacted into an `xs` index set. This only happens when there is a gap in the sequence of `xs` compacted epochs. In this case, epochs 0 and 1 have been compacted into `xs0` and `xs1`. However, there is a gap for epoch 2 because `xs2` is never created.

After this, epoch 3 continues to be compacted multiple times. The compaction process is idempotent when the same set of inputs are provided, so that in itself does not cause issues.

However, after some time, the `xn3...` index blobs for epoch 3 become eligible for deletion, after all, there is a an `xs3` index set that replaces all of the `xn3...` index blobs.

At this point, it is possible for epoch 3 to be compacted again (after some or all of the `xn3...` indexes have been removed), which produces an index set that does not have all the entries from epoch 3, it can be potentially empty.

In some cases, a kopia client may choose the incomplete index set for epoch 3. The selected index set depends on the "unique id" for the index set. This means that even in cases when an incomplete index set is created, it may be the case that that set is never used. Data loss happens when the incomplete index set is selected.
The same criteria applies for the selecting which `xs` index sets are used as inputs for range compactions. In this case, the incomplete index set for the epoch is included in the range compaction as well.

julio-lopez:
Thanks for @ashmrtn and @andrason for finding and diagnosing the issue, as well as suggesting an approach to fix it.

- Fixes #5371
- #5372


jkowalski (approved):
LGTM, would be great if you can also:

1. add an end-to-end test showing how the data corruption can happen (using some test hooks to make it deterministic)
2. show that the test is failing without the fix
3. show that it's passing with the fix

julio-lopez:
> LGTM, would be great if you can also:
> 
> 1. add an end-to-end test showing how the data corruption can happen (using some test hooks to make it deterministic)
> 2. show that the test is failing without the fix
> 3. show that it's passing with the fix

@andrason ^^^

PR #5386, opened 2026-05-24
Add Cloudflare R2 storage provider

agouny:
## Summary
- add a first-class Cloudflare R2 blob storage provider backed by the existing S3 implementation
- derive R2 endpoints from account ID and jurisdiction, with optional explicit endpoint override
- add CLI create/connect support and R2 environment variables
- reject unsupported S3 Object Lock retention options for R2 in CLI, server API, and provider code
- document R2 setup and limitations

Related UI PR: https://github.com/kopia/htmlui/pull/440

## Validation
- go test ./repo/blob/r2 ./repo/blob/s3 ./internal/server ./cli
- make -C site build
- live R2 repository validate-provider
- live R2 snapshot/restore with nested text files, binary file, empty file, empty directory, and symlink
- restored tree compared with diff -r
- htmlui validation in related PR: npm test -- SetupRepositoryR2, npm test -- SetupRepository, npm run lint, npm run build

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Pick the Cloud Storage Provider You Want
 Kopia supports saving your [encrypted](https://kopia.io/docs/features/#user-controlled-end-to-end-encryption) and [compressed](https://kopia.io/docs/features/#compression) snapshots to all of the following [storage locations](https://kopia.io/docs/features/#save-snapshots-to-cloud-network-or-local-storage):
 
 * **Amazon S3** and any **cloud storage that is compatible with S3**
+* **Cloudflare R2**
 * **Azure Blob Storage**
 * **Backblaze B2**
 * **Google Cloud Storage**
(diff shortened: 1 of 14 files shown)


basldfalksjdf:
Is this necessary? R2 already works with Kopia via the S3 repo.

julio-lopez:
> Is this necessary? R2 already works with Kopia via the S3 repo.

This ^^^

agouny:
Agreed, R2 can already be used through Kopia’s S3-compatible repository support.
This PR is not meant to add a completely new transport layer.

The motivation is to make R2 a first-class, safer preset:
- derive the R2 endpoint from account ID/jurisdiction
- force the S3 signing region to `auto`
- expose R2-specific env vars and UI fields
- avoid exposing S3-only options such as point-in-time/versioned behavior
- explicitly reject Object Lock/retention settings, since R2 does not support those S3 headers
- make R2 discoverable in CLI/UI/docs for users who do not know the exact S3-compatible settings

That said, totally understand if the project prefers not to add a new persisted storage type.
All the best!

PR #5432, opened 2026-06-17
refactor(testing): wait for routine to exit before test cleanup

julio-lopez:
Ensure background routine exits before cleaning up test environment.

- #5430

### DIFF
--- a/repo/content/content_manager_test.go
+++ b/repo/content/content_manager_test.go
@@ -1170,7 +1170,7 @@ func (s *contentManagerSuite) TestFlushWaitsForAllPendingWriters(t *testing.T) {
 	fs.AddFault(blobtesting.MethodPutBlob).SleepFor(2 * time.Second)
 
 	bm := s.newTestContentManagerWithTweaks(t, fs, nil)
-	t.Cleanup(func() { bm.CloseShared(ctx) })
+	t.Cleanup(func() { bm.CloseShared(testlogging.ContextForCleanup(t)) })
 
 	// write one content in another goroutine
 	// 'fs' is configured so that blob write takes several seconds to complete.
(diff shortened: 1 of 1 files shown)


repo/content/content_manager_test.go
```diff
@@ -1170,7 +1170,7 @@ func (s *contentManagerSuite) TestFlushWaitsForAllPendingWriters(t *testing.T) {
 	fs.AddFault(blobtesting.MethodPutBlob).SleepFor(2 * time.Second)
 
 	bm := s.newTestContentManagerWithTweaks(t, fs, nil)
-	defer bm.CloseShared(ctx)
+	t.Cleanup(func() { bm.CloseShared(ctx) })
```
julio-lopez:
```suggestion
	t.Cleanup(func() { bm.CloseShared(testlogging.ContextForCleanup(t)) })
```

repo/content/content_manager_test.go
```diff
@@ -1185,6 +1185,10 @@ func (s *contentManagerSuite) TestFlushWaitsForAllPendingWriters(t *testing.T) {
 		writeContentAndVerify(ctx, t, bm, seededRandomData(1, maxPackSize))
 	})
 
+	// wait for background writeContentAndVerify routine to finish before
+	// performing the rest of the test cleanup, such as tearing down 'bm'
+	t.Cleanup(wg.Wait)
```
julio-lopez, line 1190: @ashmrtn IIUC, this should be either:
- replaced with a defer, so it executes before the test `ctx` is cancelled; or
- more likely be moved back to the end of the function (and add a comment).

If the test fails before reaching `wg.Wait()`, that's OK, there's no need to prevent the context from being cancelled, since (1) the context is already cancelled anyway; (2) it does not matter if the background task experiences failures; and (3) we want the background task to exit as soon as possible.

Thoughts?

julio-lopez: - followup in #5435

PR #5433, opened 2026-06-17
add Date header in RFC 1123Z format. improve EmailProvider test assertions

BugCoder762:
This PR improves the robustness of the EmailProvider unit tests by moving from loose header verification (using require.Contains) to strict, exact-match verification of the entire email message content.

Key changes:

Normalized Verification: Introduced a regex-based normalization step for the dynamic Date header in both TestEmailProvider and TestEmailProvider_Text.

Strict Assertions: Replaced multiple individual require.Contains checks with a single require.Equal assertion for the full, normalized message body. This ensures that the message structure, header order, and line endings are exactly as expected, preventing regressions where headers might be accidentally reordered or missing.

Why it's needed:
The previous tests only verified that specific strings existed within the email content. This approach was brittle and allowed for structural issues (such as incorrect header placement or unexpected whitespace) to go unnoticed. By validating the exact string format, we ensure better consistency and accuracy in email generation.


BugCoder762:
Hi everyone,

This is my first time contributing an improvement to an open-source project, so I'm relatively new to this workflow.

Could someone please take a look at this PR when you have a moment? If I made any mistakes, if the format isn't right, or if you need me to adjust the code, please let me know. I'd be happy to learn and make any necessary changes.

Thank you for your time and for this great project!

BugCoder762:
"Hi team! Since I'm quite new to this, I'm having some trouble applying the suggestion directly in the interface and accepting the changes, and I'm not entirely sure how I should do it now. The changes are correct—they ensure the line starts with 'Date'—but I'm a bit lost on how to continue from here. Could someone guide me on what to do next?"

PR #5440, opened 2026-06-22
feat(repository): support inline and file-based web identity tokens for S3 storage

arthurdeschamps:
Hi! This PR adds support for web identity tokens for S3-based repositories, along with unit tests and an integration test (I ran it locally, works as expected).

Web identity token files are especially useful because they allow users to refresh tokens before they expire. Web identity tokens are usually short lived (they can't be longer than the underlying AWS policy's max duration, which has a default value of 1 hour), so without being able to refresh them, it's impossible to perform long-running ops such as backup uploads/downloads.

### DIFF
--- a/repo/blob/s3/s3_options.go
+++ b/repo/blob/s3/s3_options.go
@@ -24,6 +24,11 @@ type Options struct {
 	SecretAccessKey string `json:"secretAccessKey" kopia:"sensitive"`
 	SessionToken    string `json:"sessionToken" kopia:"sensitive"`
 
+	// Used for web-identity (OIDC) authentication. When set, the role in RoleARN is
+	// assumed using a refreshing credential provider.
+	WebIdentityToken     string `json:"webIdentityToken,omitempty" kopia:"sensitive"`
+	WebIdentityTokenFile string `json:"webIdentityTokenFile,omitempty"`
+
 	// Used for assume role authentication.
 	RoleARN      string                `json:"roleARN"`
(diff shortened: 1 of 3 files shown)


arthurdeschamps:
@julio-lopez hi! New to this project so I'm not sure who to tag, but you seem very active here so I thought I'd ask you if you could kindly take a look at my PR? Thanks in advance

jkowalski:
This is looking good, do you intend to add CLI support as well?

arthurdeschamps:
> This is looking good, do you intend to add CLI support as well?

I selfishly didn't intend to because I don't need it in my own project, but happy to add it to my PR if you think it'd be useful to other users

jkowalski:
> > This is looking good, do you intend to add CLI support as well?
> 
> I selfishly didn't intend to because I don't need it in my own project, but happy to add it to my PR if you think it'd be useful to other users

Please do, should be just <5 lines of code.

jkowalski:
> > > This is looking good, do you intend to add CLI support as well?
> > 
> > 
> > I selfishly didn't intend to because I don't need it in my own project, but happy to add it to my PR if you think it'd be useful to other users
> 
> Please do, should be just <5 lines of code.

Actually, perhaps let me take it back, expiring tokens don't like being stored long term in a config file, so it's good as-is, which is basically programmatic use only.

jkowalski (approved):

jkowalski:
Thanks for your contribution!

PR #5460, opened 2026-07-01
fix(repository): avoid nil-pointer crash when a cache directory is set but caching is disabled

### LINKED ISSUE #5406: env KOPIA_CACHE_DIRECTORY crashes if caching is deaktivated
hello,
because i can't set some parameters as default and if i connect to a repository my defaults are cleared, i call kopia from a shell-script, which sets the parameter i need and to run kopia with different repositorys without reconnect to them every time. for that, i swap between some configuration files.

but if i deactivate cache and set KOPIA_CACHE_DIRECTORY to a cache-dir, kopia crash:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc54849]

i deactivate caching because i don't like it, if some programs write much data to my drives. not all volumes have the needed space to do that and other don't like it, if much is written to it, like sd-cards and usb-sticks, this will damage them!
kopia writes on my tests so much data to the cache, the my root-partition gets full, which is a problem!

so i must move such dirs to a place, which don't make problems.

goodby

here my actual script with test-lines:
#!/usr/bin/ksh
set -x
CONFIG=test
CFN=/etc/kopia/$CONFIG/repository.config
#PARAM="--log-level=error --cache-directory=/hd/daten3/kopia --config-file=/etc/kopia/$CONFIG/repository.config"
#PARAM="--log-level=error --log-dir=/hd/daten3/kopia --disable-content-log --disbale-file-logging --parallel=1 --config-file=/etc/kopia/$CONFIG/repository.config"
#PARAM="--log-level=error --disable-content-log --disbale-file-logging --parallel=1 --config-file=/etc/kopia/$CONFIG/repository.config"
#PARAM="--log-level=error --disable-content-log --config-file=$CFN"
PARAM="--log-level=error --disable-content-log"
#export KOPIA_CACHE_DIRECTORY=/hd/daten3/kopia
export KOPIA_LOG_DIR=/hd/daten3/kopia
export KOPIA_LOG_DIR_MAX_FILES=10
export KOPIA_LOG_DIR_MAX_AGE=12h
export KOPIA_CONTENT_LOG_DIR_MAX_FILES=10
export KOPIA_CONTENT_LOG_DIR_MAX_AGE=12h
export KOPIA_CONFIG_PATH="$CFN"
grep '"caching": {}' "$CFN" >/dev/null || kopia.org cache set --content-cache-size-limit-mb=0 --metadata-cache-size-limit-mb=0

kopia.org $PARAM "$@"

maximilize:
Fixes #5406.

`snapshot create` (and any command that reads content) panics with a nil-pointer dereference when the repository has a cache directory configured but the cache size is zero (caching disabled) — e.g. after `cache set --content-cache-size-limit-mb=0 --metadata-cache-size-limit-mb=0` with `KOPIA_CACHE_DIRECTORY` set.

`NewContentCache` (`internal/cache/content_cache.go`) only falls back to a passthrough cache when *no* cache directory is configured. When a directory is set but the size is zero, `NewStorageOrNil` returns a nil storage and `NewPersistentCache` returns a nil cache, yet a `contentCacheImpl` is still built with that nil `pc`. The first blob fetch then dereferences it in `fetchBlobInternal` (`content_cache.go:105`, `c.pc.reportMissBytes`).

Fix: fall back to the passthrough cache when the persistent cache is nil, matching the existing no-directory path.

Added `TestContentCacheZeroSizeWithDirectory` — it panics without the fix and passes with it. The full `internal/cache` package stays green; gofmt and vet are clean.

### DIFF
--- a/internal/cache/content_cache.go
+++ b/internal/cache/content_cache.go
@@ -192,6 +192,13 @@ func NewContentCache(ctx context.Context, st blob.Storage, opt Options, mr *metr
 		return nil, errors.Wrap(err, "unable to create base cache")
 	}
 
+	if pc == nil {
+		// The cache is effectively disabled (zero size) even though a cache
+		// directory was configured, so there is no persistent cache to back the
+		// contentCacheImpl. Fall back to a passthrough to avoid a nil dereference.
+		return passthroughContentCache{st}, nil
+	}
+
 	return &contentCacheImpl{
 		st:             st,
 		pc:             pc,
--- a/internal/cache/content_cache_test.go
+++ b/internal/cache/content_cache_test.go
@@ -172,6 +172,30 @@ func TestDiskContentCache(t *testing.T) {
 	verifyContentCache(t, cc, cacheStorage)
 }
 
+// TestContentCacheZeroSizeWithDirectory is a regression test for #5406: a
+// configured cache directory combined with a zero cache size (cache disabled)
+// must fall back to a passthrough cache instead of building a nil-backed cache
+// that panics with a nil-pointer dereference when a blob is fetched.
+func TestContentCacheZeroSizeWithDirectory(t *testing.T) {
+	ctx := testlogging.Context(t)
+
+	cc, err := cache.NewContentCache(ctx, newUnderlyingStorageForContentCacheTesting(t), cache.Options{
+		BaseCacheDirectory: testutil.TempDirectory(t),
+		CacheSubDir:        "contents",
+		Sweep:              cache.SweepSettings{MaxSizeBytes: 0}, // cache disabled
+	}, nil)
+	require.NoError(t, err)
+
+	defer cc.Close(ctx)
+
+	var tmp gather.WriteBuffer
+	defer tmp.Close()
+
+	// Previously panicked here with a nil-pointer dereference on the nil persistent cache.
+	require.NoError(t, cc.GetContent(ctx, "content-1", "content-1", 0, -1, &tmp))
+	require.Equal(t, []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}, tmp.ToByteSlice())
+}
+
 func verifyContentCache(t *testing.T, cc cache.ContentCache, cacheStorage cache.Storage) {
 	t.Helper()
 


patvdv:
Using kopia 0.23.1 I ran into the same problem today I think. Moving a kopia server from a VM instance to a docker instance, I started with a bare-bones `repository.config` containing following for _caching_ settings:

```
  "caching": {
    "cacheDirectory": "/app/cache",
  },
```

That generated a panic:

```
panic: runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x74df74] goroutine 110 [running]: github.com/kopia/kopia/internal/cache.(*contentCacheImpl).fetchBlobInternal(0x40002591c0, {0x1d2a638, 0x40003971d0}, {0x4005a3c8c0, 0x36}, 0x4005064d40) github.com/kopia/kopia/internal/cache/content_cache.go:105 +0x124 github.com/kopia/kopia/internal/cache.(*contentCacheImpl).getContentFromFullBlob(0x40002591c0, {0x1d2a638, 0x40003971d0}, {0x4005a3c8c0, 0x36}, 0x2e1a, 0x1f9, 0x4005064d00)
...
```

at server start and when listing snapshots. Strangely enough running `kopia repository status` worked fine. The behaviour was consisted for both arm64 and amd64 images, the latter running in emulation mode.

I then manually re-added following to `repository.config`:

```
  "caching": {
    "cacheDirectory": "/app/cache",
    "maxCacheSize": 5242880000,
    "maxMetadataCacheSize": 5242880000,
    "maxListCacheDuration": 30
  },
```

And bingo: Kopia started correctly and worked without issues.

maximilize:
@patvdv That is exactly the crash this PR fixes, down to the frame.

With `cacheDirectory` set but no `maxCacheSize`, the size is 0, so `NewStorageOrNil` (`internal/cache/content_cache.go:184`) returns a nil storage and `NewPersistentCache` returns a nil `*PersistentCache` (`persistent_lru_cache.go:433-435`) — but `NewContentCache` still builds a `contentCacheImpl` around it. Nothing dereferences that nil until the first content read, at `fetchBlobInternal` (`content_cache.go:105`), which is your frame. That is also why `repository status` was fine: it reads metadata, not content blobs, so it never reaches that call. Server start and snapshot listing do.

Your workaround is the right one if you want caching. If you want it off instead, drop `cacheDirectory` — an empty base directory short-circuits to the passthrough cache at `content_cache.go:178` before the nil is built. This patch makes "directory set, size 0" take that same passthrough path instead of panicking.

Issue #5124, opened 2026-02-01
Backups to OneDrive through rclone hang randomly with "EOF" / "BLOB not found" on Periodic maintenance


michellabbe:
Backup using `rclone` to `OneDrive` work fine (Snapshots and Periodic maintenances) for several days, until suddently a `Periodic maintenance` fail after **0.0s** with this error:

```
Error: unable to determine if maintenance is required: error getting status: error reading schedule blob: error flushing dir cache: RC error: Post "https://127.0.0.1:50341/vfs/forget": EOF.
```

After this error occur, all Snapshots fail once and then nothing more seem to happen on this repo (no retry, no restart of rclone or anything), until `KopiaUI` is closed/relaunched.  Then a new `rclone` process is launched in parallel with the zombie one and everything works fine again for several days, until the same issue reappear again.

Reproduced on multiple different `Win10`/`Win11` computers (they are all setup the same way and all have the same issue), with at least these versions:
KopiaUI `v0.21.1`, `v0.22.1`, `v0.22.3`
rclone `v1.69.1`, `v1.70.2`, `v1.70.3`, `v1.72.1`
(I most likely always had these issues as I started using `KopiaUI` with `rclone` / `OneDrive` around february 2025, with Kopia `v0.19.0`).

Snapshots to S3 (`MinIO`) repositories work flawlessly on the same computers / KopiaUI.

I first thought it could be an issue with Bitdefender so I added `rclone.exe` and `KopiaUI.exe` in exclusions.  That didn't help.

`AI` suggests adding rate limits args in the rclone command like to avoid `429 Too Many Requests` throttling that could be returned by OneDrive API and not trapped by KopiaUI.

I'd like to experiment with that, but I don't see any way to add extra args for rclone in KopiaUI (also see #2846).
Is there some undocumented way through `repository.config` by any chance?

Checking `cli-logs`, the `BLOB not found` errors seem to start right after `maxConcurrency` hit **40** during the `full maintenance`:

```
[...]
2026-01-15T01:05:26.159017Z DEBUG kopia/repo [STORAGE] ListBlobs	{"prefix":"xn101_","resultCount":21,"error":null,"duration":"344.3025ms"}
2026-01-15T01:05:26.530672Z DEBUG kopia/repo [STORAGE] ListBlobs	{"prefix":"xn102_","resultCount":18,"error":null,"duration":"715.4524ms"}
[...]

Issue #4791, opened 2025-08-30
KopiaUI: scheduled tasks fail to run after several days: error flushing dir cache... EOF


yanghuidong:
When KopiaUI starts refresh, scheduled tasks run successfully at first, for a few days, and then they start to fail, all with similar error messages:

> Error: unable to determine if maintenance is required: error getting status: error reading schedule blob: error flushing dir cache: RC error: Post "https://127.0.0.1:53781/vfs/forget": EOF.

> Error: unable to create policy getter: unable to get policies: unable to get parent policies: unable to find manifest for source huidong@6-apple-macos:/Users/huidong/i: unable to load manifest contents: error refreshing indexes: mutable parameters: unable to read format blob: error getting kopia.repository blob: error flushing dir cache: RC error: Post "https://127.0.0.1:53781/vfs/forget": EOF.

We can see that the common underlying error is the following:

```
error flushing dir cache: RC error: Post "https://127.0.0.1:53781/vfs/forget": EOF.
```

Once we exit the application and start it again, the problem goes away, but again, it comes up in a few days.

Running KopiaUI `Version v0.21.1 0733cb4d2a731dbb92d927f66230694e014f4df2 kopia/htmlui f951773b9eeb5015c79bedacf49dee15f758bbcb built on Sat Jun 7 18:14:52 UTC 2025 fv-az1788-890`, on macOS 12.7.6, Intel chip.

Thank you for making Kopia.

---

Additional info:

The repositories in question are all using Rclone as the backend, and the underlying storage service is Cloudflare R2. I haven't tried other Kopia-native storage backends. Not sure if this is relevant to the error here.

rateschner:
+1

drakarah:
I'm having the same problem, I have 3 instances of kopia running in separate docker containers (3 separate remotes via rclone) and all of them failed at the same time with the exact message above and they are unable to recover, I need to restart them to get them working again.

Considering it's 3 separate processes independent of each other I'm guessing some outside factor is the cause but Kopia is unable to recover from it.

This is the log stack trace:
```
2025-12-29T15:43:15.896244989Z refresh error    {"err":"unable to refresh repository: error refreshing content index: mutable parameters: unable to read format
[...]

PR #5462, opened 2026-07-01
fix(rclone): make remote control calls resilient to transient failures

michaelmarcuccio:
## Problem

With the rclone storage backend, snapshots and maintenance intermittently
fail with errors like:

    error flushing dir cache: RC error: Post "https://127.0.0.1:PORT/vfs/forget": EOF

Reported in #4791 and #5124 across multiple OSes and rclone versions.

(The related state where *every* operation fails until KopiaUI is restarted
turned out to be a separate bug — rclone support files deleted from the OS
temp directory by temp cleaners — addressed in #5485.)

## Root cause

Before every `ListBlobs`, `GetMetadata`, and `GetBlob`, the rclone backend
issues a `vfs/forget` remote-control POST to flush rclone's directory cache
(added in #3284). Unlike all other blob operations — which are wrapped by
`retrying.NewWrapper` — this call had **no retry**, so a single transient
failure failed the whole operation. The most common trigger is an `EOF` when
Go's HTTP transport reuses a keep-alive connection that rclone has already
closed server-side (POSTs are not auto-retried by default).

Two secondary issues could leave rclone unresponsive rather than failing once:

- The RC HTTP client used a zero-value `http.Transport` (idle connections
  pooled indefinitely) with no request timeout, so a wedged rclone could hang
  operations instead of erroring.
- The stderr-scanner goroutine stopped permanently if rclone ever logged a
  line longer than `bufio.Scanner`'s 64 KiB limit; rclone would then block
  writing to stderr and stop responding to all requests.

## Changes

- Retry `vfs/forget` using the existing exponential-backoff helper, treating
  connection errors and 5xx as retriable and client errors, cancellations,
  client timeouts and process death as terminal.
- Mark the RC POST idempotent (`Idempotency-Key`) so the transport
  transparently replays it on a fresh connection when a pooled one was closed
  — fixing the common EOF at zero latency cost. (The body is a `*bytes.Buffer`,
  so `http.NewRequest` populates `GetBody`, which makes the POST replayable.)
- Give the RC HTTP client a request timeout and the shared certificate-pinned
  transport (standard dial/TLS/idle timeouts) used by the WebDAV path.
- Keep draining rclone's stderr after an over-long line so rclone never blocks
  on a full stderr pipe.
- Watch the rclone child process and fail fast with a clear
  `rclone process has exited` error instead of retrying against a dead process.

## Testing

- New unit test for the retry classifier.
- New end-to-end test that kills rclone mid-session and asserts operations
  fail fast with a clear error.
- `go test ./repo/blob/rclone/...` passes with `KOPIA_PROVIDER_TEST=1` against
  rclone v1.69.2, including provider validation.
- `golangci-lint` (repo config) and `gofmt` are clean.

Continues the rclone process-lifecycle hardening from #5040, whose
non-cancelling exec context the process watchdog here builds on.

Relates to #4791 (fixes the transient EOF failures; the persistent
restart-required variant is addressed by #5485)
Relates to #5124

### DIFF
--- a/repo/blob/rclone/rclone_storage.go
+++ b/repo/blob/rclone/rclone_storage.go
@@ -10,6 +10,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"io"
 	"net/http"
 	"os"
 	"os/exec"
@@ -22,6 +23,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/kopia/kopia/internal/osexec"
+	"github.com/kopia/kopia/internal/retry"
 	"github.com/kopia/kopia/internal/tlsutil"
 	"github.com/kopia/kopia/repo/blob"
 	"github.com/kopia/kopia/repo/blob/webdav"
@@ -35,9 +37,19 @@ const (
 
 	// defaultRcloneStartupTimeout is the time we wait for rclone to print the https address it's serving at.
 	defaultRcloneStartupTimeout = 15 * time.Second
+
+	// remoteControlRequestTimeout is the overall timeout for a single remote control request,
+	// so that a wedged rclone process does not hang storage operations indefinitely.
+	remoteControlRequestTimeout = 1 * time.Minute
+
+	// maxStderrLineLength is the maximum length of a single rclone stderr line the scanner can handle.
+	maxStderrLineLength = 1 << 20
 )
 
 var log = logging.Module("rclone")
+
+// errRCloneExited is returned when the rclone child process is no longer running.
+var errRCloneExited = errors.New("rclone process has exited")
 
 type rcloneStorage struct {
 	blob.Storage // the underlying WebDAV storage used to implement all methods.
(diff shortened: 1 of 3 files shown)


michaelmarcuccio:
Closing this in favor of #5485.

After further debugging on a live wedged instance, I root-caused the failures reported in #4791: kopia places the rclone support files (htpasswd, TLS key/cert, embedded config) in the OS temp directory, where temp cleaners delete them from under the running processes — after which rclone drops every authenticated connection until restart. Details and repro: https://github.com/kopia/kopia/issues/4791#issuecomment-4931275847

This PR was built on the hypothesis that the reported EOFs were transient stale-keep-alive failures. Going back through a month of my own logs and every report in #4791, I could not find a single confirmed instance of that mode — all documented cases match the support-file deletion. Rather than ask maintainers to review hardening for an unobserved failure mode, I'm withdrawing this PR. The individual pieces here (fail-fast on rclone process death, RC request timeouts, stderr-overflow drain) can be revisited later if there's still appetite once #5485 lands.

PR #5473, opened 2026-07-04
fix(cli): only run automatic maintenance after data-modifying operations

### LINKED ISSUE #3174: Repository maintenance should only run after specific write operations
Currently, repo maintenance runs when the `user@host` matches the repository owner (on a schedule) independent of the type of operation, be it a read-only data, read-write data, or control operation such as setting repository parameters.

While attempting to run maintenance in an opportunistic manner whenever kopia runs may be desirable,
most users, myself included, find it surprising that maintenance runs on read-only operations, such as `snapshot list`, or in control operations, such as `repo set-parameter`.

Repo maintenance should only run on a subset of data operations that modify the repo, for example, after `snapshot create`. And it should **NOT** run in the following scenarios:

- the operation encounters an error (for example, when `snapshot create` fails);
- read-only operations such as `snapshot list`, `snapshot restore`, `policy list`, etc.
- various "control only" operations such as connecting to the repo, setting repo parameters or updating policies, etc.

maximilize:
Fixes #3174.

Currently opportunistic automatic maintenance runs after *any* successful
repository-writer command, regardless of whether it modified snapshot data. As
noted in the issue, this is surprising for control-only operations such as
`policy set` or `repo set-parameter` that change configuration but write no
snapshot data.

Two of the three cases from the issue were already handled:

- **not on error** — `maybeRunMaintenance` is only called when the action
  returns no error (`cli/app.go`, the `err == nil` guard);
- **not on read-only operations** — reader actions (`repositoryReaderAction`,
  `directRepositoryReadAction`) never set `allowMaintenance`.

The remaining gap was **control-only write operations**: they go through
`repositoryWriterAction`, which unconditionally set `allowMaintenance: true`, so
`policy set`, `acl …`, `user …`, etc. all triggered maintenance.

## Change

`repositoryWriterAction` no longer triggers maintenance by default. A new
`repositoryWriterActionWithMaintenance` opts in, and only data-modifying
commands use it. Making "no maintenance" the default means a newly added
control command won't accidentally trigger maintenance.

Commands opted in to maintenance (data-modifying):
`snapshot create`, `snapshot delete`, `snapshot expire`, `snapshot migrate`,
`snapshot fix invalid-files`, `snapshot fix remove-files`,
`snapshot copy-history`/`move-history`.

Commands left on the default (control-only, no maintenance):
`policy set/edit/import/remove`, `acl add/delete/enable`,
`user add-set/delete/hash-password`,
`notification profile/template …`, `cache set`, `snapshot pin`,
`manifest delete`.

`snapshot pin` (writes a manifest flag but orphans no content) and
`manifest delete` (low-level, not meant for direct use) are treated as
control-only per review — neither triggers maintenance. Any content orphaned by
`manifest delete` is reclaimed by the next scheduled/opportunistic maintenance,
exactly as before (opportunistic maintenance was already throttled by schedule,
never guaranteed per operation).

## Test

`TestAutoMaintenanceOnlyAfterDataWrite` (cli): after `policy set` the
maintenance schedule has no runs; after `snapshot create` it does. Verified
RED→GREEN — without the change the test fails because `policy set` triggers a
full maintenance cycle.

### DIFF
--- a/cli/command_manifest_delete.go
+++ b/cli/command_manifest_delete.go
@@ -17,7 +17,7 @@ type commandManifestDelete struct {
 func (c *commandManifestDelete) setup(svc appServices, parent commandParent) {
 	cmd := parent.Command("delete", "Remove manifest items").Alias("remove").Alias("rm").Hidden()
 	cmd.Arg("item", "Items to remove").Required().StringsVar(&c.manifestRemoveItems)
-	cmd.Action(svc.repositoryWriterActionWithMaintenance(c.run))
+	cmd.Action(svc.repositoryWriterAction(c.run))
 
 	c.svc = svc
 }
--- a/cli/command_manifest_delete.go
+++ b/cli/command_manifest_delete.go
@@ -17,7 +17,7 @@ type commandManifestDelete struct {
 func (c *commandManifestDelete) setup(svc appServices, parent commandParent) {
 	cmd := parent.Command("delete", "Remove manifest items").Alias("remove").Alias("rm").Hidden()
 	cmd.Arg("item", "Items to remove").Required().StringsVar(&c.manifestRemoveItems)
-	cmd.Action(svc.repositoryWriterAction(c.run))
+	cmd.Action(svc.repositoryWriterActionWithMaintenance(c.run))
 
 	c.svc = svc
 }
--- a/cli/command_snapshot_pin.go
+++ b/cli/command_snapshot_pin.go
@@ -22,7 +22,7 @@ func (c *commandSnapshotPin) setup(svc appServices, parent commandParent) {
 	cmd.Flag("add", "Add pins").StringsVar(&c.addPins)
 	cmd.Flag("remove", "Remove pins").StringsVar(&c.removePins)
 	cmd.Arg("id", "Snapshot ID or root object ID").Required().StringsVar(&c.snapshotIDs)
-	cmd.Action(svc.repositoryWriterActionWithMaintenance(c.run))
+	cmd.Action(svc.repositoryWriterAction(c.run))
 }
 
 func (c *commandSnapshotPin) run(ctx context.Context, rep repo.RepositoryWriter) error {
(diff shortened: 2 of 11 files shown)


cli/command_manifest_delete.go
```diff
@@ -17,7 +17,7 @@ type commandManifestDelete struct {
 func (c *commandManifestDelete) setup(svc appServices, parent commandParent) {
 	cmd := parent.Command("delete", "Remove manifest items").Alias("remove").Alias("rm").Hidden()
 	cmd.Arg("item", "Items to remove").Required().StringsVar(&c.manifestRemoveItems)
-	cmd.Action(svc.repositoryWriterAction(c.run))
+	cmd.Action(svc.repositoryWriterActionWithMaintenance(c.run))
```
julio-lopez: Manifest deletion is a low(er)-level operation. In general, users should not modify manifests directly, and instead use the higher-level abstractions (policies, snapshots, etc).    

It is OK if this does not (auto) trigger maintenance.

cli/command_snapshot_pin.go
```diff
@@ -22,7 +22,7 @@ func (c *commandSnapshotPin) setup(svc appServices, parent commandParent) {
 	cmd.Flag("add", "Add pins").StringsVar(&c.addPins)
 	cmd.Flag("remove", "Remove pins").StringsVar(&c.removePins)
 	cmd.Arg("id", "Snapshot ID or root object ID").Required().StringsVar(&c.snapshotIDs)
-	cmd.Action(svc.repositoryWriterAction(c.run))
+	cmd.Action(svc.repositoryWriterActionWithMaintenance(c.run))
```
julio-lopez: As pointed out in the PR description, this is "borderline".

It makes sense to NOT (auto) trigger maintenance in this case.

What's a good way for communicating the difference in behavior to the user?

julio-lopez (changes requested):

maximilize:
Done — both now use the maintenance-free default (`repositoryWriterAction`):

- **`manifest delete`** — low-level, not meant for direct use; control-only now.
- **`snapshot pin`** — same.

`repositoryWriterActionWithMaintenance` is left to the data-modifying snapshot commands (create/delete/expire/migrate/fix-*/copy-move-history). `TestAutoMaintenanceOnlyAfterDataWrite` still passes; PR description updated.

On communicating the change: I don't think anything user-facing is warranted. Opportunistic maintenance was never guaranteed per-operation (it's throttled by the schedule), and correctness is unchanged — it still runs on schedule and after the next data op, `kopia maintenance run` stays available, and content orphaned by `manifest delete` is reclaimed at the next maintenance as before. For `pin` there is zero observable difference, since it orphans no content. #3174 was about this being surprising, so not running it is the less-surprising behavior. If you want something explicit, a `debug`-level log line when maintenance is skipped is the cheapest option — I'd lean toward nothing.

julio-lopez (approved):

PR #5484, opened 2026-07-09
fix(cli): reject negative values for --parallel flags

### LINKED ISSUE #2022: Parallel flag on many commands may have negative or zero values
Parallel flag on many commands may have negative or zero values.

For example with `snapshot migrate`:

- negative value like `-1` produces panic
- zero value produces stuck forever

In other commands results may be same when IntFlag used.

costajohnt:
Fixes #2022.

The `--parallel` flag accepted negative values on the commands that bind it with kingpin's `IntVar`, and a negative value (e.g. `snapshot migrate --parallel=-1`) panics.

This adds a small custom kingpin `Value` (`cli/flag_nonnegative_int.go`) that rejects negative input at parse time, so the CLI returns a clear error instead of panicking. It's wired into the 12 `IntVar` `--parallel` sites via `.SetValue(nonNegativeIntVar(&target))`, preserving each flag's existing default, placeholder, and help text. The 5 `UintVar` `--parallel` sites already reject negatives by type, so they're left as-is.

Zero is intentionally still allowed: some commands use `0` as a meaningful value (`snapshot create` defaults `--parallel` to `0` for auto-detect, `restore` documents `1=disable`), so the guard only rejects `< 0`.

## Out of scope (possible follow-ups)

- The issue also notes that `--parallel=0` "sticks forever" on some commands. That's per-command semantics (`0` legitimately means auto for `snapshot create`), so it isn't a value-level fix. It'd need a separate change deciding the right zero-handling per command.
- A few sibling int flags share the same negative-value shape (e.g. `--file-parallelism`, `--parallelism`, `--list-parallelism`). Kept out of this PR to stay scoped to `--parallel`, but they could get the same `nonNegativeIntVar` treatment in a follow-up.

## Tests

- Unit test on the value: negative rejected, zero/positive accepted, non-numeric rejected.
- CLI test: `snapshot migrate --parallel=-1` and `content verify --parallel=-1` now return an error instead of panicking.

`go build ./...`, `go vet ./cli/...`, and `go test ./cli/` all pass.


cli/command_benchmark_compression.go
```diff
@@ -41,7 +41,7 @@ func (c *commandBenchmarkCompression) setup(svc appServices, parent commandParen
 	cmd.Flag("data-file", "Use data from the given file").Required().ExistingFileVar(&c.dataFile)
 	cmd.Flag("by-size", "Sort results by size").BoolVar(&c.bySize)
 	cmd.Flag("by-alloc", "Sort results by allocated bytes").BoolVar(&c.byAllocated)
-	cmd.Flag("parallel", "Number of parallel goroutines").Default("1").IntVar(&c.parallel)
+	cmd.Flag("parallel", "Number of parallel goroutines").Default("1").SetValue(nonNegativeIntVar(&c.parallel))
```
julio-lopez: why not change the type to uint?

julio-lopez (changes requested):

cli/command_benchmark_compression.go
costajohnt: These all feed `int` fields downstream, some serialized (e.g. `restore.Options.Parallel`, `json:"parallel"`), so uint would mean casts everywhere or widening those core types. Custom value keeps it to one file and parses the same as the old int flag, just rejecting negatives. Happy to switch to uint if you'd rather.

costajohnt:
Hi @julio-lopez, just a gentle nudge on this one. I replied above to your uint question with the rationale for the custom nonnegative-int value (the flags feed downstream int fields, some serialized like restore.Options.Parallel), but I'm genuinely happy to go either way here. Do you have a preference? Otherwise I'll keep the current custom-value approach.

costajohnt:
I went ahead and tried the uint switch so this wasn't just me asserting it.

It builds fine and rejects `--parallel=-1` natively via `ParseUint`, so on that axis it works. It doesn't get past the linter though. These 12 flags feed `int` fields downstream, so each needs an `int()` cast, and 3 of those trip `G115: integer overflow conversion uint -> int (gosec)` under the pinned golangci-lint 2.6.1. G115 isn't in the gosec exclusions in `.golangci.yml`, and the same run is clean on the current branch.

I think the reason the two cases differ: the `--parallel` flags already on `UintVar` (benchmark encryption, hashing, splitters, ecc, crypto) are uint end to end, feeding `runInParallelNoInput(n uint, ...)`, so they never cast. The 12 here feed `int`, some serialized, like `restore.Options.Parallel` with `json:"parallel"`.

So it seems to come down to uint plus 3 `//nolint:gosec`, or the custom value in one file. I left it as the custom value for now since that keeps the suppressions out, but I'm happy to push the uint version if you'd rather have it that way, or widen the downstream types if you want it fixed at the root instead. Your call, just let me know.

costajohnt:
Switched to `uint` as you suggested, pushed in 7a14cf63. The custom `kingpin.Value` is gone.

One thing worth flagging, since it changed how I did the conversion. The flags feed APIs that take `int`, and a plain `int(c.parallel)` cast turns out to reintroduce the bug this PR is fixing: values above `math.MaxInt64` wrap to a negative int, so `--parallel=18446744073709551615` parses fine and then hands the callee a negative parallelism. That is also what gosec G115 was pointing at in three spots, so suppressing it with nolint would have buried a real case rather than a false positive.

Instead the conversion goes through a small helper that clamps at `math.MaxInt32`:

```go
func parallelismAsInt(v uint) int {
	if v > math.MaxInt32 {
		return math.MaxInt32
	}

	return int(v)
}
```

That keeps the conversion total, so gosec is clean without suppressions, and MaxInt32 is far past any useful degree of parallelism. Negative input is still rejected at parse time, now by kingpin itself (`strconv.ParseUint: parsing "-1": invalid syntax`).

I kept the CLI regression test for #2022 with its assertion updated to kingpin's message, and added a table test covering the conversion boundaries including `math.MaxUint`. Net change is smaller than what it replaces, 85 insertions against 125 deletions.

Verified locally with the repo-pinned golangci-lint 2.6.1: gosec reports no issues and a full run over `./cli/...` exits clean. `go test ./cli/...` passes.

Worth noting that no CI has ever run on this branch, the checks list is empty, so it may need a workflow approval from your side whenever you get a chance to look.

cli/flag_parallel.go
```diff
@@ -0,0 +1,20 @@
...
+
+// parallelismAsInt converts a --parallel-style flag value to the int that the
+// underlying APIs take.
+//
+// The flags are declared as uint so kingpin rejects negative input at parse
+// time, which is the fix for kopia/kopia#2022. A plain int(v) conversion would
+// undo that on 64-bit platforms: any value above math.MaxInt64 wraps to a
+// negative int, handing the callee exactly the negative parallelism that used
+// to panic. Clamping keeps the conversion total, and MaxInt32 is far beyond any
+// useful degree of parallelism.
```
julio-lopez: These comments (here and below) are verbose, tell the story about how this code came to be (fixed bug xyz) and noisy. Self documenting code with no comments is preferred, and when comments are needed, they should be sucint.


```suggestion
// parallelismAsInt converts --parallel-style flag values from 
// uint to int and clamps returned values at MaxInt32.
```

cli/flag_parallel_internal_test.go
```diff
@@ -0,0 +1,32 @@
+package cli
+
+import (
+	"math"
+	"testing"
+)
+
+// TestParallelismAsInt guards the uint->int conversion used by the --parallel
+// flags. A plain int(v) would wrap large values to a negative int, which is the
+// panic-on-negative-parallelism bug that kopia/kopia#2022 fixed.
```
julio-lopez:
```suggestion
// TestParallelismAsInt verifies the uint->int conversion used by the
// --parallel flags to prevent wrap large values to a negative int.
```

cli/flag_parallel.go
```diff
@@ -0,0 +1,20 @@
...
+func parallelismAsInt(v uint) int {
+	if v > math.MaxInt32 {
+		return math.MaxInt32
```
julio-lopez, line 9: It's slightly surprising to have a value changed, however it's an improvement over the existing behavior and works for (most) practical purposes. One could argue that it'd be good to check for a sane maximum allowed parallelism, it's not clear (to me) what that "sane" max value would be.

julio-lopez (changes requested):

cli/flag_parallel_test.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// TestNegativeParallelRejected verifies that a negative --parallel value is
+// rejected at flag-parse time (returning an error identifying the bad value)
+// rather than panicking once the command runs. See kopia/kopia#2022.
+//
+// The assertion checks the rejection message specifically, so the test fails if
+// the flag ever stops validating: a bare "command failed" check would pass even
+// without the fix, since these commands also fail later when no repository is
+// connected.
```
julio-lopez:
```suggestion
// TestNegativeParallelRejected verifies that a negative --parallel value is
// rejected at flag-parse time and returns an error identifying the bad value
// rather than panicking once the command runs.
```

cli/flag_parallel_test.go
```diff
@@ -0,0 +1,33 @@
...
+	t.Parallel()
+
+	env := testenv.NewCLITest(t, testenv.RepoFormatNotImportant, testenv.NewInProcRunner(t))
+
+	for _, args := range [][]string{
+		{"snapshot", "migrate", "--parallel=-1", "--all"},
+		{"content", "verify", "--parallel=-1"},
+	} {
+		_, _, err := env.Run(t, true, args...)
+		require.ErrorContains(t, err, "invalid syntax",
```
julio-lopez, line 30: What other tests do is they create and connect to the repository, and then issue commands with valid repo connections.

```suggestion
		// check rejection message so the test fails if
		// the flag ever stops validating, otherwise a
		// bare "command failed" check would pass
		// since these commands fail regardless given that
		// no repository is connected.
		require.ErrorContains(t, err, "invalid syntax",
```

julio-lopez (changes requested):

cli/flag_parallel.go
```diff
@@ -0,0 +1,13 @@
+package cli
+
+import "math"
+
+// parallelismAsInt converts --parallel-style flag values from
```
julio-lopez:
```suggestion
// parallelismAsInt converts --parallel-style flag values from
```

julio-lopez (approved):

julio-lopez:
Thanks @costajohnt

PR #5540, opened 2026-08-03
refactor(server): validate cert on re-connection

julio-lopez:


julio-lopez:
Will add tests and refactor after unblocking CI

julio-lopez:
FYI @jkowalski

PR #5546, opened 2026-08-05
publish container images to ghcr as well

aslilac:
I have a slight preference for using ghcr images when possible, because they're easier to discover without needing to leave github, and it's easier to see that the image is controlled by the same person/set of people as the codebase itself. I'm sure there are others feel the same. :)

so, I propose a small tweak to the scripts involved to publish future image tags to ghcr, in addition to the current docker hub image


aslilac (commented):
feedback

PR #5558, opened 2026-08-12
fix(providers): wait for ssh process to exit to avoid leaving zombie processes

### LINKED ISSUE #3717: kopia is leaving thousands of defunct processes making macos unusable due to limit of user processes being reached
MacOS Sonoma 14.4

````
 % kopia --version
0.15.0 build: dd1ebb11cfdb7902a2b87106dfeea038858bafb5 from:
````
(from homebrew)

I'm using sftp backend to a server to a directory which is on a disk that's mounted manually.

For example disk is mounted at /mnt/storage/ and user used for backup has its $HOME on that storage:

````
backup-user:x:1123:1000::/mnt/storage/kopia/backup:/bin/false
````

When I forget to mount that disk and snapshot task starts then kopia goes crazy. Starts so many processes that end up as defunct that it eats all resources up to process limits.

That means recovering from that situation (by trying to kill kopia) is hard as terminal cannot start due to limit of processes being reached etc.

````
% ulimit -u
6000
````

kopia defunct processes
````
% grep defunct kopia.txt|wc -l
    5480
````

pstree output for kopia:

[kopia.txt](https://github.com/kopia/kopia/files/14551381/kopia.txt)

### PR DESCRIPTION
**Pull request by AndiDog**
This is supposed to fix https://github.com/kopia/kopia/issues/3717. I'm still testing it but so far no more zombie processes from using the SFTP backend.

### CODE REVIEW

File: repo/blob/sftp/sftp_storage.go [hunk 1]
```diff
@@ -490,7 +486,8 @@ func getSFTPClientExternal(ctx context.Context, opt *Options) (*sftpConnection,
 	closeFunc := func() error {
 		p := cmd.Process
 		if p != nil {
-			p.Kill() //nolint:errcheck
+			p.Kill()   //nolint:errcheck
+			cmd.Wait() //nolint:errcheck
```
julio-lopez, line 490: What is the intent of this `Wait()`?

julio-lopez (changes requested):
Please add test(s).

repo/blob/sftp/sftp_storage.go
AndiDog: This was the main bug: [`Kill`](https://pkg.go.dev/os#Process.Kill) does not wait until the process exits, and you have to call `Wait` in any case, or else the operating system is forced to leave the process behind (a "zombie"/defunct process).

I've added a test that ensures no `ssh` processes get left behind when using the external SSH provider.

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -212,6 +260,55 @@ func TestSFTPStorageValid(t *testing.T) {
...
+				"-o", "IdentitiesOnly=yes",
+				"-o", "IdentityAgent=none",
+				"-o", "UserKnownHostsFile=" + knownHostsFile,
+				"-o", "StrictHostKeyChecking=yes",
+				"-o", "BatchMode=yes",
+			}, " "),
+		}, true)
+		if err != nil {
+			t.Fatalf("unable to connect to SSH: %v", err)
+		}
```
julio-lopez: `require.NoError`

repo/blob/sftp/sftp_storage.go
julio-lopez: Thanks.

IIUC, this only addresses the issue with (child) zombie processes. Other resource consumption issues could still remain due to the outer retry loop.

julio-lopez: small concern: `Wait()` could block for a while even though Kill cannot be ignored by the child process.

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -212,6 +260,55 @@ func TestSFTPStorageValid(t *testing.T) {
 		})
 	}
 
+	t.Run("ExternalSSH", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
+			t.Skip("ssh command may not be available on Windows")
+		}
```
julio-lopez: not needed / redundant given `testutil.SkipTestOnCIUnlessLinuxAMD64(t)` above.

```suggestion
```

julio-lopez (changes requested):

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -84,6 +85,52 @@ func mustRunCommand(t *testing.T, cmd string, args ...string) []byte {
...
+			listed[pid] = true
+		}
+	}
+
+	for _, pid := range pids {
+		require.False(t, listed[pid], "SSH process %v was not reaped", pid)
+	}
+}
+
+// sshChildProcesses returns the PIDs of the `ssh` processes started by the test process
```
julio-lopez:
```suggestion
// sshChildProcesses returns the PIDs of the `ssh` processes started by the test process.
```

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -159,7 +207,7 @@ func startDockerSFTPServerOrSkip(t *testing.T, idRSA string) (host string, port
 		return host, port, knownHostsFile
 	}
 
-	t.Skipf("SFTP server did not start!")
+	t.Error("SFTP server did not start!")
```
julio-lopez: Is the change from skipping to failing the tests needed?
```suggestion
	t.Fatal("SFTP server did not start!")
```

repo/blob/sftp/sftp_storage_test.go
julio-lopez:
```suggestion
		require.NoError(t, err, "unable to connect to SSH")
```

julio-lopez (changes requested):

repo/blob/sftp/sftp_storage_test.go
julio-lopez: Restore existing expected behavior: skip the test when the server fails to start (as the function name indicates). We can revisit this separately.


```suggestion
	t.Skip("SFTP server did not start!")
```

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
```
julio-lopez:
```suggestion
```

julio-lopez:
- Partial fix for #3717

repo/blob/sftp/sftp_storage_test.go
```diff
@@ -212,6 +259,50 @@ func TestSFTPStorageValid(t *testing.T) {
 		})
 	}
 
+	t.Run("ExternalSSH", func(t *testing.T) {
+
```
julio-lopez:
```suggestion
```

julio-lopez (approved):

julio-lopez:
Thanks @AndiDog

PR #5610, opened 2026-09-03
test(server): Add tests for TLS verification

ashmrtn:
Follow up to #5540 that adds tests for resuming TLS connections and fingerprint length checks


julio-lopez (approved):
🥇Thanks for doing this.

julio-lopez (approved):
✅ LGTM
🚀  🚢it

PR #5628, opened 2026-09-07
refactor(general): propagate `TLSConfigTrustingSingleCertificate` error

julio-lopez:
Improves failure mode by explicitly making returning an error when the provided cert fingerprint is invalid.  The previously existing behavior of creating a transport that fails all connection verification, is often hard to troubleshoot.

Ref:
- #5540
- #5610


julio-lopez:
@ashmrtn FYI. Thanks.

PR #5636, opened 2026-09-10
fix(testing): avoid clobbering storage used in `t.Cleanup`

julio-lopez:
- Followup #5635
- https://github.com/kopia/kopia/pull/5635#discussion_r3980311454


julio-lopez (commented):
👍 LG: +1

PR #5642, opened 2026-09-13
refactor(testing): server control tests

julio-lopez:
Avoid calling `RunAndProcessStderr` in a background goroutine in tests.
Simplifies tests by removing unnecessary async processing.
Improves test cleanup.

- Required for #5626


cli/command_server_control_test.go
```diff
@@ -152,8 +160,10 @@ func TestServerControl(t *testing.T) {
 	env.RunAndExpectSuccess(t, "server", "shutdown", "--address", sp.BaseURL, "--server-control-password", sp.ServerControlPassword)
```
julio-lopez, line 163: It seems that some tests can (spuriously) fail in this step with an `EOF` response when the server shuts down (too) quickly and closes the connection before sending the response.

Issue #3313, opened 2023-09-17
Bug:  The socket-activation test is not robust enough


PhracturedBlue:
The socket activation test seems to have a race condition that occasionally manifests as:
```
    command_server_control_linux_test.go:30: finished in 5: 'kopia --config-file /tmp/kopia-test3374049681/fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff/.kopia.config --no-use-keyring repo connect filesystem --path /tmp/kopia-test2469635346/fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff --override-username=test-user --override-hostname=test-host'
    command_server_control_linux_test.go:63: Activating socket on 35353, PID: 13505
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x131f02f]
```
I don't really have any idea how to fix it properly.  Here are some options:
  * We could remove the test, and leave the code
  * We could remove the test and the capability
  * We can pass in a mocked version of activation to the test (but this basically disables any testing of the activation module)
  * We can write our own activation module which is testable
  * We can fork the activation module and add testability code to it

PR #5643, opened 2026-09-13
fix(testing): TestServerControlSocketActivated flake

julio-lopez:
Fix `TestServerControlSocketActivated` flake.
Clear `runner.ExtraFiles` before calling running other commands. This appears to be the primary source of the spurious test failures.

Refactor `TestServerControlSocketActivated` to:
- start server from the test (go)routine instead of a background one;
- process server's output synchronously in the test goroutine;
- fix race: close the listener file descriptor synchronously in the test goroutine instead of async, which ensures the descriptor is closed before sending the first request to the (status) server, otherwise, shutdown hangs (when the listener close is delayed);
- check and assert server's exit status;
- ensure servers exits when tests fail.

Also, address potential race in `TestServerControlSocketActivatedTooManyFDs`

Ref:
- #3313
- #5626

### DIFF
--- a/tests/socketactivation_test/socketactivation_test.go
+++ b/tests/socketactivation_test/socketactivation_test.go
@@ -52,7 +52,7 @@ func TestServerControlSocketActivated(t *testing.T) {
 	l1File, err := testutil.EnsureType[*net.TCPListener](t, l1).File()
 	require.NoError(t, err, "failed to get filehandle for socket")
 
-	serverStopped := make(chan error)
+serverStopped := make(chan error, 1)
 
 	var sp testutil.ServerParameters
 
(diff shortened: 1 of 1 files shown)


tests/socketactivation_test/socketactivation_test.go
```diff
@@ -52,33 +53,22 @@ func TestServerControlSocketActivated(t *testing.T) {
...
-
-		l1File.Close()
-		close(serverStarted)
+	runner.ExtraFiles = append(runner.ExtraFiles, l1File)
+	wait, kill := env.RunAndProcessStderr(t, sp.ProcessOutput,
+		"server", "start", "--insecure", "--random-server-control-password", "--address=127.0.0.1:0")
+	runner.ExtraFiles = nil
 
-		wait()
+	t.Cleanup(func() { kill(); wait() })
```
julio-lopez:
```suggestion
	t.Cleanup(func() { kill() })
```

PR #5656, opened 2026-09-21
New Feature: Record Ignored (Excluded) Files in Snapshot Manifests

simon3z:
# Record Ignored (Excluded) Files in Snapshot Manifests

**Type:** feature, opt-in (default off). **Scope:** client + CLI only; no server changes.

## Summary

Kopia excludes files from snapshots via ignore rules (policy `ignore` rules and `.kopiaignore` files), but once a file is excluded there is no record of *what* was excluded and *why* — it simply does not appear in the snapshot. This PR adds an **opt-in** way to record the excluded files/directories in the snapshot manifest, plus a new command, `kopia snapshot ignored`, to list them. This makes exclusions auditable per snapshot.

## Why

Today the exclusion set is invisible:

- **Auditability gap.** When auditing or restoring a backup you cannot tell which missing files were deliberately ignored versus missing for another reason.
- **Silent over-exclusion.** Ignore patterns and `.kopiaignore` entries are often broad or hand-written (e.g. `build/`, `*.tmp`). If a pattern is broader than intended, the over-excluded files are silently absent with no way to discover it.
- **Compliance.** Some use cases require being able to enumerate exactly what a snapshot intentionally left out.

Example: a `.kopiaignore` containing `db/` excludes the whole `db` tree. Today you can only confirm `db/` is *not* in the snapshot; you cannot see that it was *excluded by the ignore rule* rather than genuinely absent.

The goal is to make the exclusion set auditable per snapshot, without changing default behavior, without a breaking manifest-format change, and without any per-file cost when it is not wanted.

## What this PR does

- New opt-in policy option **`FilesPolicy.IgnoreRecord`** (JSON `ignoreRecord`, default `false`), set via `kopia policy set ... --ignore-record=true`.
- New manifest field **`Manifest.IgnoredEntries`** (JSON `ignoredEntries`, `omitempty`) — the list of files/directories excluded by ignore rules during the snapshot.
- New command **`kopia snapshot ignored <ref>`** to print the recorded exclusions, where `<ref>` is the **snapshot ID shown by `kopia snapshot list`** or a source path.
- `kopia policy show` now displays the `Ignore record:` setting.
- Tests for the enabled and default-off paths.

Behavior:

- **Default off.** Unless a policy enables `IgnoreRecord`, uploads are unchanged and the manifest carries no ignored entries.
- **Directory semantics.** `ignorefs` never descends into an ignored directory, so a recorded directory entry implies its **entire subtree** was excluded. The command marks such entries with a trailing `/` and a note, which keeps the list small (mostly directory entries, not one line per file).

## How it works

- `fs/entry.go` adds `EntryWithIgnore` (`path` + `dir`), and `snapshot/manifest.go` stores `IgnoredEntries []*fs.EntryWithIgnore` with `json:"ignoredEntries,omitempty"`.
- `snapshot/policy/files_policy.go` adds the `IgnoreRecord *OptionalBool` field, merged like the other optional-bool fields, so it inherits/overrides through the normal policy tree (global → source → subdirectory) and is resolved per-directory via `policyTree.EffectivePolicy().FilesPolicy.IgnoreRecord.OrDefault(false)`.
- During upload, `snapshot/upload/upload.go` hooks into the existing `ignorefs` callback that is already invoked for every excluded entry and appends it to the manifest's list, but **only** when the effective policy at that location enables `IgnoreRecord` and only on the real upload pass (not the size-estimation pass, which re-walks the tree). Appends are mutex-protected and the list is reset at the start of each `Upload()`, so a reused uploader does not accumulate entries across snapshots.
- `cli/command_snapshot_ignored.go` (registered in `cli/command_snapshot.go`) resolves the argument as either the snapshot ID — via `snapshot.FindSnapshotsByRootObjectID`, the same lookup `kopia snapshot pin` uses — or a source path. So the ID you copy straight from `kopia snapshot list` just works. Only if the ID matches no snapshot does the command return a clear error.
- `cli/command_policy_set_files.go` adds the `--ignore-record` flag; `cli/command_policy_show.go` adds the `Ignore record:` row.

Output:

```
// snapshot <manifest-id> (user@host:/path/to/source): 1 ignored entry
  build/    (directory; entire subtree excluded)
```

or, when nothing was recorded:

```
// snapshot <manifest-id> (user@host:/path/to/source): no ignored entries recorded
```

## Usage

```
kopia policy set --global --ignore-record=true
kopia policy show --global            # shows:   Ignore record:   true
kopia snapshot create /path/to/source
kopia snapshot list /path/to/source             # shows the snapshot ID
kopia snapshot ignored <snapshot-id>            # a specific snapshot
kopia snapshot ignored /path/to/source          # all snapshots of that source
```

The option inherits/overrides through the policy tree like any other files-policy setting, so per-source and per-directory policies work too.

## Backward compatibility

- **`omitempty` manifest field.** Standard `encoding/json` ignores unknown fields on unmarshal, so an older client reading a newer manifest ignores `ignoredEntries`, and a newer client reading an older manifest sees the field absent. No manifest-format version bump is required.
- **Opt-in, default off.** Existing policies do not set `ignoreRecord`, so `IgnoreRecord.OrDefault(false)` is `false` and uploads behave exactly as before; no existing snapshot bytes change.
- **Server unchanged.** The manifest is opaque data persisted by the server; only the snapshotting client needs the new build.

## Testing

- Unit tests: `TestUploadRecordsIgnoredEntries` (enabled → entries recorded, including directory flags) and `TestUploadDoesNotRecordIgnoredByDefault` (default → list empty).
- End-to-end: with a real filesystem repository, `policy set --global --ignore-record=true` → `snapshot create` → `kopia snapshot ignored` prints the expected entries; verified for fresh and incremental snapshots, global and source-specific policies, and root plus subdirectory `.kopiaignore` files; by manifest ID, by source, and the object-ID error path.
- Backward compatibility: unmarshalling a manifest containing `ignoredEntries` into the pre-change struct via standard `json.Unmarshal` ignores the field.


simon3z:
Hi @jkowalski, first time contributing, just checking in to see if you have any feedback on this when you have a moment. Let me know if there's anything else needed to move this forward. Thanks!

